Issue #31, opened 2025-10-21
[BUG] HA Add-On not showing up


kingbear2:
## 🐛 Bug Description
After adding the custom repository to HA, the MCP server add-on doesn't show in the add-on store.

## 🔄 Steps to Reproduce
1. In HA > Settings > Add-ons > Add-on Store > 3 dots > Repositories > https://github.com/homeassistant-ai/ha-mcp > + Add
2. Refresh browser and search for MCP. Nothing shows up, and the repository doesn't even show in the list of "no results found"

## ✅ Expected Behavior
You should be able to see the "Home Assistant MCP Server" in the add-on store

## ❌ Actual Behavior
The "Home Assistant MCP Server" doesn't appear

## 📊 Environment
- **Home Assistant version**: 2025.10.3
- **MCP Server version**: 3.1.5
- **Operating System**: HAOS

## 📝 Error Logs
n/a

## 🔍 Additional Context
Restarted HA, rebooted entire HAOS, and updated from 2025.10.2 to 2025.10.3 and still can't get the add-on to show.

julienld:
should be fixed 

@kingbear2 can you confirm?

kingbear2:
I can confirm the add-on is now showing, but I have not installed it to see if it works as intended. But at least it’s showing which means it’s hopefully resolved.

PR #33, opened 2025-10-21
fix: align add-on schema with HA docs

julienld:
## Summary
- align schema syntax with Supervisor expectations using list()/str? strings (per https://developers.home-assistant.io/docs/add-ons/configuration#options--schema)
- keep default backup_hint while dropping empty secret_path default so option remains optional
- teach packaging to include new client/tools subpackages and add translations/en.yaml so UI retains field descriptions; confirmed pattern via `gh search code "list(" --filename config.yaml --repo home-assistant/addons`

## Testing
- uv run pytest tests/addon


julienld:
fix #31

PR #47, opened 2025-10-26
ci: streamline codex autofix actions

julienld:
## Summary
- replace the Codex auth helper with a reusable prepare action that validates secrets and installs the CLI
- simplify the PR context helper to emit a human-readable summary and feed the same prompt into the sync and async Codex runners
- update both Codex auto-fix workflows to consume the new helpers and shared prompt/context wiring

## Testing
- not run (workflow change only)

------
https://chatgpt.com/codex/tasks/task_e_68fd2e58c21c832ca0a9fc1d7343dc70

### DIFF
--- a/.github/actions/codex-run/action.yml
+++ b/.github/actions/codex-run/action.yml
@@ -2,7 +2,7 @@ name: Execute Codex automation
 description: Prepare authentication (if required) and invoke Codex in async or sync mode.
 inputs:
   mode:
-    description: "Execution mode: 'async' launches a cloud worker, 'sync' runs remediation locally."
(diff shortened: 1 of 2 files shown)


.github/actions/codex-run-async/action.yml
```diff
@@ -0,0 +1,56 @@
+name: Launch Codex async worker
+description: Start a Codex cloud worker to investigate the failing workflow asynchronously.
+inputs:
+  prompt:
```
julienld: @codex Combine all inputs into prompt. The logic should be in the caller

PR #120, opened 2025-11-26
fix: add pagination and limit to logbook tool to prevent oversized responses

### LINKED ISSUE #82: [BUG] Logbook tool return size is too large
## 🐛 Bug Description
When returning more than a few hours of data, the log book fills the context

## 🔄 Steps to Reproduce
"Get me the logbook of the last 48h"

## 🔍 Potential fix
The log book tool should be customizable. (max rows, search pattern, etc.)
Also a max size for some tool would be nice.
Looking forward forward for MCP specs update to fix the fact that the return of a big tool call is always passed to the context. Would want to pass it to a local file or similar that can be processed with scripts (grep, parse, ...). Could be part of the client implementation without being in the CMP specs too.

julienld:
## Summary
- Added default limit of 50 entries to logbook tool (configurable up to 500)
- Added `offset` parameter for pagination support
- Added `has_more` indicator and `pagination_hint` for multi-page navigation
- Response now includes `returned_entries`, `limit`, `offset`, and `has_more` metadata

## Test plan
- [x] Added comprehensive e2e tests in `tests/src/e2e/tools/test_logbook.py`
- [ ] CI tests pass
- [ ] Manual testing with large logbook queries

Closes #82

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/ha_mcp/tools/tools_utility.py
+++ b/src/ha_mcp/tools/tools_utility.py
@@ -61,6 +61,9 @@ def register_utility_tools(mcp: Any, client: Any, **kwargs: Any) -> None:
         effective_limit = limit if limit is not None else DEFAULT_LOGBOOK_LIMIT
         effective_limit = max(1, min(effective_limit, MAX_LOGBOOK_LIMIT))
 
+        # Ensure offset is non-negative
+        offset = max(0, offset)
+
         # Calculate start time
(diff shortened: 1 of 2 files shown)


julienld:
Recreating PR to reset review comments

PR #133, opened 2025-11-26
fix: add pagination and limit to logbook tool to prevent oversized responses

### LINKED ISSUE #82: [BUG] Logbook tool return size is too large

julienld:
## Summary
- Added default limit of 50 entries to logbook tool (configurable up to 500)
- Added `offset` parameter for pagination support
- Added `has_more` indicator and `pagination_hint` for multi-page navigation
- Response now includes `returned_entries`, `limit`, `offset`, and `has_more` metadata

## Test plan
- [x] Added comprehensive e2e tests in `tests/src/e2e/tools/test_logbook.py`
- [ ] CI tests pass
- [ ] Manual testing with large logbook queries

Closes #82

Generated with [Claude Code](https://claude.com/claude-code)


julienld:
## Pagination Stability Analysis

I've reviewed the pagination implementation and addressed the stability concern raised by @julienld.

### Current Behavior:
The pagination is performed **client-side** after retrieving the full result set from Home Assistant's logbook API. This means:
- We fetch ALL entries for the time period from HA
- We then slice the array based on `offset` and `limit` parameters
- This approach has a **stability issue**: if new logbook entries are created between page requests, the results can shift

### Example of the Issue:
```python
# Request 1 at time T: Get entries 0-49 (total: 100)
ha_get_logbook(hours_back=24, limit=50, offset=0)

# New entry is created in HA

# Request 2 at time T+1: Get entries 50-99 (total: 101 now)
ha_get_logbook(hours_back=24, limit=50, offset=50)
# Item #50 from request 1 might now be at position #51
```

### Mitigation:
I've added documentation in the docstring warning users about this limitation:

> **IMPORTANT - Pagination Stability:**
> Pagination is performed client-side on the full result set returned
> by Home Assistant. If new logbook entries are created between page
> requests, results may shift and items could be missed or duplicated
> across pages. For best results, use consistent time ranges (start/end)
> and retrieve pages in quick succession.

### Potential Future Improvements:
To achieve true stable pagination, we would need either:
1. Home Assistant API to support server-side pagination with cursors/tokens
2. Snapshot the full result set server-side and paginate from that
3. Use explicit `end_time` parameter to lock the time window

For now, the client-side approach is reasonable for most use cases where:
- Pages are retrieved quickly
- The time window is historical (not real-time)
- Perfect consistency isn't critical

### Fixes Applied:
All Copilot bot suggestions have been implemented:
- ✅ Added pagination metadata to error responses
- ✅ Enhanced pagination hint with all query parameters
- ✅ Improved test assertion messages
- ✅ Added negative offset test

Issue #279, opened 2025-12-05
[FEATURE] Partial dashboard update API


sergeykad:
# 🚀 Feature Request: Partial Dashboard Update API

### **Is your feature request related to a problem? Please describe.**

When using AI assistants (Claude, etc.) to modify Home Assistant dashboards via MCP, the current `ha_config_set_dashboard` tool requires submitting the complete dashboard configuration for any content change.

This creates several frustrations:

1. **Token inefficiency** — Small changes (e.g., fixing a typo in a markdown card) require sending/receiving the entire dashboard JSON, wasting tokens and context window
2. **Risk of data loss** — The AI must reconstruct the full config from memory, potentially losing content if it misremembers, truncates, or hallucinates parts of the config
3. **Slow iteration** — Each minor tweak requires a full round-trip of the complete dashboard structure
4. **Error-prone** — Complex dashboards with multiple views/sections increase the chance of reconstruction errors

### **Describe the solution you'd like**

Add card-level CRUD operations that allow modifying individual cards without touching the rest of the dashboard:

```python
ha_dashboard_update_card(
    url_path: str,        # Dashboard to modify
    view_index: int,      # Which view (0-based)
    section_index: int,   # Which section within view  
    card_index: int,      # Which card within section
    card_config: dict     # New card configuration (replaces entire card)
)

ha_dashboard_add_card(
    url_path: str,
    view_index: int,
    section_index: int,
    card_config: dict,
    position: int = -1    # Optional: insert position, -1 for end
)

ha_dashboard_remove_card(
    url_path: str,
    view_index: int,
    section_index: int,
    card_index: int
)
```

**Example improvement:**

Updating a markdown card's Jinja template (e.g., changing unit conversion logic):
- **Current:** ~5KB full dashboard JSON required
- **Proposed:** ~200 bytes for just the card config

### **Describe alternatives you've considered**

1. **JSON path-based patching** — More flexible but harder to validate:
   ```python
   ha_config_patch_dashboard(
       url_path: str,
       json_path: str,    # e.g., "views[0].sections[1].cards[0].content"
       value: Any         # New value at that path
   )
   ```
   Downside: Requires JSON path parsing, more room for invalid paths.

2. **Section-level operations** — Middle ground between full dashboard and card-level:
   ```python
   ha_dashboard_update_section(url_path, view_index, section_index, section_config)
   ```
   Downside: Still requires full section config, less granular than card-level.

[...]

PR #295, opened 2025-12-07
Implements issue #279: card-level CRUD operations for dashboards

sergeykad:
- Add ha_dashboard_add_card: insert card at position or append
   - Add ha_dashboard_update_card: replace card config entirely
   - Add ha_dashboard_remove_card: remove card by index

   Features:
   - Supports both flat views and sections-based dashboards
   - Rejects strategy dashboards with clear error messages
   - Optimistic locking: verifies config unchanged before save
   - Returns previous_card on update for audit/undo capability
   - Token-optimized error messages using Strunk's principles

   Tests: 88 unit tests, 15 E2E tests (4 new card operation tests)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,6 +79,7 @@ archive/
 docs/review/
 docs/development/
 .local/
+/.claude/settings.local.json
 fastmcp/
 
 # HACS frontend - large (~51MB) JavaScript assets, downloaded during test setup
--- a/src/ha_mcp/tools/tools_config_dashboards.py
+++ b/src/ha_mcp/tools/tools_config_dashboards.py
@@ -327,11 +327,16 @@ def register_config_dashboard_tools(mcp: Any, client: Any, **kwargs: Any) -> Non
 
             # Extract config from WebSocket response
             config = response.get("result") if isinstance(response, dict) else response
+
+            # Compute hash for optimistic locking in subsequent operations
+            config_hash = _compute_config_hash(config) if isinstance(config, dict) else None
+
             return {
                 "success": True,
                 "action": "get",
                 "url_path": url_path,
                 "config": config,
+                "config_hash": config_hash,
             }
         except Exception as e:
             logger.error(f"Error getting dashboard config: {e}")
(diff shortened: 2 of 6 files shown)


julienld:
## Suggestion: Simplify Optimistic Locking with Hash Parameter

The current implementation does 3 API calls per operation (get → verify-get → save). We can reduce this to 2 calls while keeping the stateless design.

### Proposed Approach

1. **Add `config_hash` to `ha_config_get_dashboard` response:**
```python
return {
    "success": True,
    "config": {...},
    "config_hash": "abc123..."  # NEW
}
```

2. **Add mandatory `config_hash` parameter to card operations:**
```python
async def ha_dashboard_update_card(
    url_path: str | None = None,
    view_index: int = 0,
    card_index: int = 0,
    section_index: int | None = None,
    card_config: dict | str = None,
    config_hash: str,  # REQUIRED for partial updates
) -> dict[str, Any]:
```

3. **Return new hash after successful modification:**
```python
return {
    "success": True,
    "action": "update_card",
    "previous_card": {...},
    "updated_card": {...},
    "config_hash": "def456..."  # NEW: hash after modification
}
```

4. **Validation logic:**
   - Fetch current config, compute hash, compare to provided `config_hash`
   - If mismatch → return error asking agent to re-read dashboard
   - If match → apply change, save, return new hash

### Benefits

| Aspect | Current | Proposed |
|--------|---------|----------|
| API calls per operation | 3 | 2 |
| Server stateless | ✅ | ✅ |
| Conflict detection | ✅ | ✅ |
| Contract | Implicit (hidden verify) | Explicit (agent must provide hash) |
| Chain operations | Must re-read each time | Use returned hash for next op |

### Why Mandatory?

Making `config_hash` required ensures agents always read the dashboard first before modifying. This:
- Guarantees agents have fresh indices (view_index, section_index, card_index)
- Makes the read-then-modify contract explicit
- Prevents blind modifications with stale data

### Flow Example

```
# Agent reads dashboard once
result = ha_config_get_dashboard(url_path="my-dash")
hash = result["config_hash"]  # "abc123"

# Agent chains operations using returned hashes
r1 = ha_dashboard_update_card(..., config_hash=hash)
r2 = ha_dashboard_add_card(..., config_hash=r1["config_hash"])
r3 = ha_dashboard_remove_card(..., config_hash=r2["config_hash"])

# If any operation returns "config changed" error → re-read dashboard
```

This makes the contract explicit, reduces API overhead by ~33%, and allows chaining operations without re-reading.

PR #355, opened 2025-12-23
feat(tools): add enable/disable tools for integrations and entities

cj-elevate:
## Summary

Add 4 new tools for managing integration and entity enabled states via MCP:

### Integration tools (`tools_integrations.py`)
- **`ha_enable_integration`**: Re-enable disabled integrations
- **`ha_disable_integration`**: Disable integrations without removing config  
- **`ha_delete_config_entry`**: Permanently remove orphaned config entries (with confirmation safety)

### Entity tool (`tools_registry.py`)
- **`ha_set_entity_enabled`**: Enable/disable individual entities

## Motivation

These tools fill a gap in the current toolset where users couldn't programmatically manage the enabled state of integrations and entities through MCP.

### Common use cases:
- Re-enabling sensors that were auto-disabled by integrations
- Disabling unused entities to reduce clutter and resource usage
- Cleaning up orphaned config entries from dead/replaced hardware
- Troubleshooting integrations by toggling enabled state

## Implementation Details

- Uses WebSocket API (`config_entries/disable`, `config/entity_registry/update`)
- Follows existing patterns in the codebase
- Includes proper error handling and user-friendly messages
- `ha_delete_config_entry` requires explicit `confirm=True` for safety

## Test plan

- [ ] Tested enable/disable integration cycle
- [ ] Tested entity enable/disable
- [ ] Tested delete config entry with orphaned entry
- [ ] Verified error handling for non-existent entries

### DIFF
--- a/src/ha_mcp/tools/tools_integrations.py
+++ b/src/ha_mcp/tools/tools_integrations.py
@@ -1,14 +1,14 @@
 """
 Integration management tools for Home Assistant MCP server.
 
-This module provides tools to list and query Home Assistant integrations
-(config entries) via the REST API.
+This module provides tools to list, enable, and disable Home Assistant integrations
+(config entries) via the REST and WebSocket APIs.
 """
 
 import logging
 from typing import Any
 
-from .helpers import log_tool_usage
+from .helpers import get_connected_ws_client, log_tool_usage
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


julienld:
@cj-elevate : thanks for the PR. I will review it.

julienld:
Here is my review.

On the content, it is good. Great additions. I thought about adding a hint to use the backup tool before deleting an entity, but it is not required in my opinion since this is for orphaned stuff. Integrations usually re-create entities when you restart them. You can lookup also BACKUP_HINT in the code if you want to know more. People should already have daily backups.

On the form:
- if possible, make tool description as minimal as you can. Descriptions are injected into context and take space. We already have a lot of tools. Might need to "fold" the tools in the future.
- Please don't use tool_registry.py. It's for registration, not implementation. I still have some chore to do there already so it might be confusing. Consider existing tool files or a new one.
- consider the use of coerce_bool_param.
- see also comments from Gemini, but I would just merge enable+disable into ha_set_integration_enabled or ha_set_integration_status... something consistent with ha_set_entity_enabled.


I consider the above optional. The only non-optionnal would be : 
- needs some tests. e2e or similar.

Issue #525, opened 2026-01-31
Release DCR support out of beta


julienld:
see OAuth 2.1 with DCR support in beta https://github.com/homeassistant-ai/ha-mcp/blob/master/docs/OAUTH.md // https://github.com/homeassistant-ai/ha-mcp/issues/245

julienld:
need to fix https://github.com/homeassistant-ai/ha-mcp/issues/522

JB09:
#532 

`AssertionError
2026-02-02 17:34:49,597 ha_mcp.__main__ INFO: OAuth mode logging configured at INFO level
2026-02-02 17:34:49,598 ha_mcp.auth.provider INFO: Using OAUTH_ENCRYPTION_KEY from environment
2026-02-02 17:34:49,598 ha_mcp.auth.provider INFO: HomeAssistantOAuthProvider initialized (base_url will be auto-detected)
2026-02-02 17:34:49,838 ha_mcp.tools.registry INFO: Auto-discovery registered tools from 33 modules
2026-02-02 17:34:49,838 ha_mcp.__main__ INFO: Server created with OAuthProxyClient
2026-02-02 17:34:49,838 ha_mcp.__main__ INFO: Starting OAuth-enabled MCP server with 93 tools on None/mcp
2026-02-02 17:34:49,866 asyncio ERROR: Task exception was never retrieved
future: <Task finished name='Task-2' coro=<FastMCP.run_async() done, defined at /usr/local/lib/python3.13/site-packages/fastmcp/server/server.py:594> exception=AssertionError()>
Traceback (most recent call last):
  File "/usr/local/lib/python3.13/site-packages/fastmcp/server/server.py", line 616, in run_async
    await self.run_http_async(
    ...<3 lines>...
    )
  File "/usr/local/lib/python3.13/site-packages/fastmcp/server/server.py", line 2552, in run_http_async
    app = self.http_app(
        path=path,
    ...<3 lines>...
        stateless_http=stateless_http,
    )
  File "/usr/local/lib/python3.13/site-packages/fastmcp/server/server.py", line 2617, in http_app
    return create_streamable_http_app(
        server=self,
    ...<16 lines>...
        middleware=middleware,
    )
  File "/usr/local/lib/python3.13/site-packages/fastmcp/server/http.py", line 308, in create_streamable_http_app
    auth_routes = auth.get_routes(mcp_path=streamable_http_path)
[...]

PR #533, opened 2026-02-02
fix: Add workaround for ChatGPT's non-standard /token/.well-known/openid-configuration request

julienld:
## Summary

Fixes ChatGPT MCP connector compatibility by adding a workaround for its non-standard OAuth request to `/token/.well-known/openid-configuration`.

## Problem

After completing the OAuth flow, ChatGPT makes a request to `/token/.well-known/openid-configuration` which returns 404. This prevents users from connecting their ChatGPT MCP connector to ha-mcp OAuth servers.

**User report:**
> Appears to be working! There was a 404 error for /token endpoint at the very end, but I was able to create the connector on ChatGPT. edit: looks like I can't actually connect to my instance without the token endpoint being mapped to the openid-configuration

**Logs showed:**
```
GET /token/.well-known/openid-configuration HTTP/1.1" 404 Not Found
```

## Root Cause

ChatGPT incorrectly combines the token endpoint path (`/token`) with the discovery endpoint path (`/.well-known/openid-configuration`), creating a non-standard request path.

This is a bug in ChatGPT's OAuth implementation, but we need to support it for compatibility.

## Solution

Added a new route at `/token/.well-known/openid-configuration` that serves the same OpenID Configuration metadata as the standard `/.well-known/openid-configuration` endpoint.

**Changes:**
- `src/ha_mcp/auth/provider.py`: Added route with explanatory comment about the ChatGPT bug
- `tests/src/unit/test_oauth.py`: Added test verifying the route exists

## Testing

- ✅ All 44 OAuth unit tests pass
- ✅ New test verifies route existence and configuration
- ✅ Route serves same metadata as standard endpoints for consistency

## Related

- Complements PR #531 (OpenID Configuration endpoint)
- Part of ongoing ChatGPT OAuth compatibility work

### DIFF
--- a/tests/src/unit/test_oauth.py
+++ b/tests/src/unit/test_oauth.py
@@ -617,36 +617,28 @@ class TestOAuthRoutes:
         # Note: Full handler testing requires ASGI app context, which is tested in E2E tests
 
     @pytest.mark.asyncio
-    async def test_openid_configuration_endpoint(self, provider):
-        """Test OpenID Configuration endpoint exists for ChatGPT compatibility."""
+    @pytest.mark.parametrize("discovery_path,description", [
+        ("/.well-known/openid-configuration", "standard OpenID Configuration endpoint"),
+        ("/token/.well-known/openid-configuration", "ChatGPT bug workaround endpoint"),
+    ])
+    async def test_openid_configuration_endpoints(self, provider, discovery_path, description):
+        """Test OpenID Configuration endpoints exist for ChatGPT compatibility.
+
+        Covers:
+        - Standard /.well-known/openid-configuration (required by ChatGPT)
+        - Non-standard /token/.well-known/openid-configuration (ChatGPT bug workaround)
+
+        Both should serve the same metadata as /.well-known/oauth-authorization-server.
+        """
         routes = provider.get_routes()
-        openid_route = next(
-            (r for r in routes if r.path == "/.well-known/openid-configuration"),
+        route = next(
+            (r for r in routes if r.path == discovery_path),
             None
         )
 
-        # Verify the route exists (required by ChatGPT MCP connector)
-        assert openid_route is not None
-        assert openid_route.path == "/.well-known/openid-configuration"
-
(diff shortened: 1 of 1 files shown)


julienld:
## Implementation Summary

**Choices Made:**
- Added `/token/.well-known/openid-configuration` route serving the same metadata as standard OpenID Configuration endpoint
- Refactored tests using `pytest.mark.parametrize` to reduce code duplication (per Gemini suggestion)
- Added explanatory comments documenting this as a ChatGPT bug workaround

**Problems Encountered:**
- None - straightforward implementation addressing a non-standard ChatGPT OAuth request

**Analysis:**
- Confirmed this is NOT a trailing slash bug on our side - our code properly strips trailing slashes via `rstrip('/')`
- ChatGPT incorrectly requests OpenID configuration at the token endpoint path, violating RFC 8414 and OpenID Connect Discovery specs
- According to specs, discovery should be at issuer root (`/.well-known/openid-configuration`), NOT at token endpoint (`/token/.well-known/openid-configuration`)
- This workaround is harmless - just serves the same public metadata at an additional path for compatibility

**Testing:**
- ✅ All 44 OAuth unit tests pass
- ✅ Parametrized test covers both standard and workaround endpoints
- ✅ CI checks passed before merge

**Related Issues:**
- Complements PR #531 (standard OpenID Configuration endpoint)
- Part of ongoing ChatGPT OAuth compatibility work (scope auto-granting, discovery endpoints)

julienld:
Alright. This will be release Tuesday on the stable channel. OAuth is still considerer in beta. Would appreciate your feedback on issue #525 for things that work or don't.

I did a PR to remove the need for MCP_BASE_URL and OAUTH_ENCRYPTION_KEY... so if you could make sure you are testing that way it would be appreciated. (should test restarting the container)... Thinking of it... The OAUTH_ENCRYPTION_KEY auto-generation won't work on a container. There need to be a persistent volume somewhere. I might tag you when I get the solution to test it out.

Thanks for testing!

PS: I'm not actively monitoring closed PRs. Issues are the place to report bugs.

PR #586, opened 2026-02-09
fix: remove internal info leaks from error responses (#517)

### LINKED ISSUE #517: [BUG] ha_deep_search leaks Python tracebacks to MCP clients in error responses
### What Happened?

While reviewing the codebase against the MCP specification, I found that the `ha_deep_search` tool exposes full Python tracebacks to clients when an error occurs. If the search fails for any reason — a connection timeout, a bad query, an internal exception — the response includes `traceback.format_exc()` output with file paths, line numbers, and class names from the server.

This is in `src/ha_mcp/tools/tools_search.py` around line 531. The error dict is built manually and includes `"traceback"` and `"error_type"` keys that no other tool uses. The rest of the codebase handles errors through `exception_to_structured_error()`, which keeps internal details server-side.

The MCP spec's security considerations section says servers must sanitize tool outputs ([spec reference](https://modelcontextprotocol.io/specification/2025-11-25/server/tools)). Tracebacks reveal internal structure and should only appear in server logs, not in client-facing responses.

### Copy-Paste from Conversation

The relevant code in `tools_search.py`:

```python
except Exception as e:
    import traceback
    return {
        "success": False,
        "error": str(e),
        "error_type": type(e).__name__,
        "traceback": traceback.format_exc(),
        "query": query,
        "search_types": parsed_search_types,
        "limit": limit,
        "suggestions": [
            "Check Home Assistant connection",
            "Try simpler search terms",
            "Check search_types are valid: 'automation', 'script', 'helper'",
        ],
    }
```

**Suggested fix:** Remove the `traceback` and `error_type` keys, log the traceback server-side with `logger.error`, and replace the manual dict with `exception_to_structured_error()` to match every other tool in the project. The `suggestions` list is already good and should stay.

sergeykad:
## What does this PR do?                                                                                                                                                                                                                                                                                          
                                                                                   
  Closes #517                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                    
  - **`ha_deep_search`**: replaced `traceback.format_exc()` with `exception_to_structured_error()` — was leaking full Python stack traces to MCP clients. Also added `add_timezone_metadata()` on both success and error paths for consistency with sibling tools.
  - **`ha_dashboard_find_card`**: replaced ad-hoc error dict with `exception_to_structured_error()` — was leaking `error_type` (Python exception class names) to clients.

  ## Type of change
  - [x] 🐛 Bug fix

  ## Testing
  - Added `test_deep_search_error_handling.py` — verifies no `traceback`/`error_type` in error response, verifies search-specific suggestions preserved
  - Added `test_dashboard_find_card_error.py` — verifies no `traceback`/`error_type` in error response, verifies dashboard-specific suggestions preserved

  ## Checklist

### DIFF
--- a/src/ha_mcp/tools/tools_search.py
+++ b/src/ha_mcp/tools/tools_search.py
@@ -645,7 +645,7 @@ def register_search_tools(mcp: Any, client: Any, **kwargs: Any) -> None:
                     f"Unexpected error response structure, could not add suggestions: "
                     f"{type(error_response.get('error'))}"
                 )
-            return await add_timezone_metadata(client, error_response)
+            return error_response
 
     @mcp.tool(
         annotations={
--- a/tests/src/unit/test_deep_search_error_handling.py
+++ b/tests/src/unit/test_deep_search_error_handling.py
@@ -65,13 +65,12 @@ class TestDeepSearchErrorHandling:
 
         result = await deep_search_tool(query="test_query")
 
-        data = result["data"]
-        assert data["success"] is False
-        assert isinstance(data["error"], dict), "error must be structured dict, not raw string"
-        assert "code" in data["error"]
-        assert "message" in data["error"]
-        assert "traceback" not in data
-        assert "error_type" not in data
+        assert result["success"] is False
+        assert isinstance(result["error"], dict), "error must be structured dict, not raw string"
+        assert "code" in result["error"]
+        assert "message" in result["error"]
+        assert "traceback" not in result
+        assert "error_type" not in result
 
     @pytest.mark.asyncio
     async def test_error_includes_search_specific_suggestions(
@@ -84,24 +83,9 @@ class TestDeepSearchErrorHandling:
 
         result = await deep_search_tool(query="test_query")
 
-        data = result["data"]
-        suggestions = data["error"]["suggestions"]
+        suggestions = result["error"]["suggestions"]
         assert "Check Home Assistant connection" in suggestions
         assert "Try simpler search terms" in suggestions
-
-    @pytest.mark.asyncio
-    async def test_error_response_includes_timezone_metadata(
-        self, mock_mcp, mock_client, mock_smart_tools, deep_search_tool
-    ):
-        """Error responses should be wrapped with timezone metadata."""
-        mock_smart_tools.deep_search = AsyncMock(
-            side_effect=RuntimeError("fail")
-        )
-
-        result = await deep_search_tool(query="test_query")
-
-        assert "metadata" in result, "error response must include timezone metadata"
-        assert "home_assistant_timezone" in result["metadata"]
 
     @pytest.mark.asyncio
     @pytest.mark.parametrize(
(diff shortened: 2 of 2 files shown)


kingpanther13 (dismissed):
Good work addressing the info leak in #517 — the move to `exception_to_structured_error()` and removal of `error_type`/`traceback` fields is the right call. The new unit tests are solid and the E2E test hardening (replacing `.get()` with key assertions) is a nice improvement.

However, I found a few issues that should be addressed before merging:

---

### 1. Response shape inconsistency in `ha_deep_search` (bug)

The **success path** returns a raw dict:
```python
return cast(dict[str, Any], result)
```

But the **error path** wraps the response with `add_timezone_metadata()`:
```python
return await add_timezone_metadata(client, error_response)
```

This means clients get two structurally different response shapes from the same tool:
- Success: `{"success": True, "automations": [...], ...}`
- Error: `{"data": {"success": False, "error": {...}}, "metadata": {"home_assistant_timezone": ...}}`

This is particularly notable because commit `3930f5b` explicitly removed `add_timezone_metadata` from the success path with the rationale:
> "Deep search returns entity IDs, names, and scores with no timestamps, so wrapping the response in add_timezone_metadata was unnecessary and changed the API contract."

That reasoning applies equally to the error path. The error path should match the success path — remove `add_timezone_metadata` from the error return as well, and just `return error_response` directly. The unit test (`test_deep_search_error_handling.py`) would also need to be updated accordingly (it currently expects `result["data"]`).

---

### 2. Missing server-side logging for `ha_deep_search` errors

The old code leaked the traceback to clients (bad), but the new code doesn't log it server-side either. When `ha_deep_search` fails, the exception is silently converted to a structured error with no `logger.error()` or `logger.exception()` call. This means the full traceback is lost entirely rather than being redirected to server logs where it would be useful for debugging.

The issue bot analysis for #517 specifically recommended:
> "Use `logger.exception()` within the except block to ensure the full traceback is still available in the server logs for debugging without being exposed to the client."

Note that the `ha_dashboard_find_card` handler already has proper logging (`logger.error(..., exc_info=True)`) before calling `exception_to_structured_error()`. The deep search handler should follow the same pattern. Something like:

```python
except Exception as e:
    logger.error("Deep search failed: query=%s, error=%s", query, e, exc_info=True)
    error_response = exception_to_structured_error(...)
```

---

### 3. Malformed line breaks in function calls

Two call sites have an unusual formatting artifact where the closing argument starts on a new line with a leading comma:

```python
result = await _exact_match_search(
    client, query, domain_filter, limit
, offset)
```

and:

```python
result = await _partial_results_search(
    client, query, domain_filter, limit
, offset)
```

While syntactically valid, this is non-standard Python style. These should be:
```python
result = await _exact_match_search(
    client, query, domain_filter, limit, offset
)
```

---

Issues 1 and 2 are the important ones — the response shape inconsistency is a functional bug and the missing logging is a debugging regression. The formatting nit (#3) is minor but easy to fix while you're in there.

PR #587, opened 2026-02-09
fix: bound OAuth client cache to prevent memory leak on token rotation

### LINKED ISSUE #522: [BUG] OAuth client cache grows without bound on token rotation
### What Happened?

In OAuth mode, `OAuthProxyClient` in `__main__.py` (around line 51) caches Home Assistant clients in a plain dict keyed by `f"{ha_url}:{ha_token}"`. Each time a user's token rotates or a new user connects, a new client is created and cached, but old entries are never removed. Over time this leaks memory and holds open stale HTTP connections.

```python
client_key = f"{ha_url}:{ha_token}"
if client_key not in self._oauth_clients:
    self._oauth_clients[client_key] = HomeAssistantClient(...)
```

The MCP spec's security best practices cover session and credential lifecycle management ([reference](https://modelcontextprotocol.io/specification/2025-11-25/basic/security_best_practices)). Unbounded caches that hold credential-derived state risk resource exhaustion.

### Copy-Paste from Conversation

The cache is a plain `dict` with no eviction, no TTL, and no size limit. Evicted clients also need `await client.close()` called for proper httpx connection cleanup.

**Suggested fix:** Replace the plain dict with a bounded, TTL-based cache (e.g., `cachetools.TTLCache` with max 100 entries and a 1-hour TTL matching token expiry). Add an eviction callback that closes the HTTP client.

sergeykad:
Replace the unbounded dict in OAuthProxyClient with cachetools.TTLCache (maxsize=100, ttl=1h) to prevent memory growth and connection leaks from OAuth token rotation.                                                                                                                                            
                                                                                                                                                                                                                                                                                                                    
  - Detect cache evictions (both TTL and maxsize) via a secondary tracking TTLCache and schedule async client.close() to release httpx connections                                                                                                                                                                  
  - Use .get() instead of in + [] to avoid a TOCTOU race with TTLCache's lazy expiry                                                                                                                                                                                                                                
  - Add cachetools>=5.0.0 as an explicit dependency (previously only transitive via fastmcp)                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                    
  Closes #522                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                    
  Type of change                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                    
  - 🐛 Bug fix                                                                                                                                                                                                                                                                                                      
  - ✨ New feature                                                                                                                                                                                                                                                                                                   
  - 📚 Documentation                                                                                                                                                                                                                                                                                                
  - 🔧 Maintenance/refactor
  - 💥 Breaking change

  Testing

  - I have tested these changes with a LLM agent
  - All automated tests pass (uv run pytest)
  - Code follows style guidelines (uv run ruff check)

  Checklist

  - I have updated documentation if needed

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -791,6 +791,25 @@ await mcp.call_tool("ha_config_get_script", {"script_id": "nonexistent"})
 
 **HA API uses singular field names:** `trigger` not `triggers`, `action` not `actions`.
 
+**E2E tests: poll after creating entities.** After creating an entity (automation, script, helper, etc.), HA needs time to register it. Never search/query immediately — use polling helpers from `tests/src/e2e/utilities/wait_helpers.py`:
+```python
+from ..utilities.wait_helpers import wait_for_tool_result
+
+# BAD: entity may not be registered yet
+create_result = await mcp_client.call_tool("ha_config_set_automation", {"config": config})
+result = await mcp_client.call_tool("ha_deep_search", {"query": "my_sensor"})  # may return empty
+
+# GOOD: poll until the entity appears in results
+data = await wait_for_tool_result(
+    mcp_client,
+    tool_name="ha_deep_search",
+    arguments={"query": "my_sensor", "search_types": ["automation"], "limit": 10},
+    predicate=lambda d: len(d.get("automations", [])) > 0,
+    description="deep search finds new automation",
+)
+```
+Other available helpers: `wait_for_entity_state()`, `wait_for_entity_attribute()`, `wait_for_condition()`. See `wait_helpers.py` for the full set.
+
 ## Release Process
 
 Uses [semantic-release](https://python-semantic-release.readthedocs.io/) with conventional commits.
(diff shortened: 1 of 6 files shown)


julienld:
Quick clarification on my previous comment: when I said "Even though Fernet encryption is non-deterministic" - I meant **before the base64 refactor in #534**.

**Timeline:**
- **Before Feb 2**: Tokens used Fernet encryption (non-deterministic, but cache key still used decrypted values)
- **After Feb 2 (#534)**: Tokens use base64 encoding (deterministic - same credentials = identical token string)

So in the **current codebase** (base64), the same credentials produce the **exact same token string** every time. The cache key would definitely be identical for the same user.

This makes the leak scenario even less clear to us - with deterministic base64 encoding and stateless tokens, we don't see how the cache grows beyond the number of unique credential pairs.

PR #590, opened 2026-02-09
feat: Add config entry options flow tools

saphid:
## Summary
Add 3 tools for configuring existing integrations via their options flow (the "Configure" button in the HA UI):

- `ha_start_config_entry_options_flow`: Start options flow for an entry (use `ha_get_integration()` to find entry IDs)
- `ha_submit_config_entry_options_step`: Submit form data or navigate menu
- `ha_abort_config_entry_options_flow`: Abort the flow without saving changes

## Motivation
Configuring integrations with complex settings (many dropdowns, multi-step forms) is slow via browser automation. The REST API allows:
- Batch submission of all field values in one request
- Programmatic configuration of sensors, entity mappings, etc.
- Much faster than click-wait-scroll-click per dropdown

## API Reference
These tools expose undocumented but stable HA REST endpoints:
- `POST /api/config/config_entries/options/flow` - start flow
- `POST /api/config/config_entries/options/flow/{flow_id}` - submit step
- `DELETE /api/config/config_entries/options/flow/{flow_id}` - abort flow

## Example Usage
```python
# Find integrations that support options (uses existing tool)
integrations = await ha_get_integration()
# Look for entries with supports_options=true

# Start options flow
flow = await ha_start_config_entry_options_flow(entry_id='abc123')
# flow.type='menu', flow.menu_options=['sensors', 'general', ...]

await ha_submit_config_entry_options_step(flow_id, {'next_step_id': 'sensors'})
# Now at sensor selection form

await ha_submit_config_entry_options_step(flow_id, {'rpm': True, 'speed': True})
# Flow completes automatically on final form submit

# Only use abort if you need to cancel without saving:
await ha_abort_config_entry_options_flow(flow_id)
```

## Changes
- `src/ha_mcp/client/rest_client.py`: Add client methods for options flow (start, submit, abort)
- `src/ha_mcp/tools/tools_config_entry_options.py`: New tools module (auto-discovered)
- `tests/src/e2e/tools/test_config_entry_options.py`: E2E tests for all 3 tools + error cases

## Review Feedback Addressed
- **Removed `ha_list_config_entries`** — duplicates existing `ha_get_integration()` (per @julienld)
- **Renamed `finish` → `abort`** — DELETE aborts the flow; successful flows complete on final submit (per @julienld)
- **Updated descriptions** to reference `ha_get_integration()` for discovering entry IDs
- **Added E2E tests** covering start/submit/abort flows and invalid ID error cases
- **Error handling** uses structured errors via `create_validation_error` and `exception_to_structured_error`

### DIFF
--- a/.claude/skills/bat-adhoc/SKILL.md
+++ b/.claude/skills/bat-adhoc/SKILL.md
@@ -1,5 +1,5 @@
 ---
-name: bat
+name: bat-adhoc
 description: Run bot acceptance tests to validate MCP tools work correctly from a real AI agent's perspective. Use when testing PRs, detecting regressions, or verifying tool changes end-to-end with Claude/Gemini CLIs.
 disable-model-invocation: true
 argument-hint: [scenario-description or --help]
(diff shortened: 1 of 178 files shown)


julienld:
Thanks for this contribution @saphid - this fills a real gap. Today we can create config-entry-based helpers (template, group, utility_meter, etc.) via `ha_create_config_entry_helper()`, but there's no way to modify their configuration after creation. The options flow tools solve that, and they work for any integration with `supports_options: true`.

Before we move forward, I want to flag some overlap with existing tools:

**`ha_list_config_entries` duplicates `ha_get_integration()`**

The existing `ha_get_integration()` in `tools_integrations.py` already lists all config entries with the same fields (entry_id, domain, title, state, supports_options) plus fuzzy search, domain filtering, and options inspection. The `get_config_entries()` client method is also redundant since `get_config_entry()` already calls the same `GET /config/config_entries/entry` endpoint internally. Please drop both and update the options flow tool descriptions to reference `ha_get_integration()` for discovering entry IDs (e.g., "Use `ha_get_integration()` to find entries with `supports_options: true`").

**Rename "finish" to "abort":** `ha_finish_config_entry_options_flow` sends HTTP DELETE, which aborts the flow. Successful flows complete automatically when the final form step is submitted. Renaming to `ha_abort_config_entry_options_flow` avoids misleading AI agents into thinking they must call this to "complete" a configuration change.

Once these are addressed and CI is green, this should be good to merge.

PR #639, opened 2026-02-17
docs: classify BAT metrics as primary vs secondary

julienld:
## Summary
- Classifies metrics across all BAT skill files into **primary** (decide pass/fail) vs **secondary** (report but don't decide on)
- Aligns `bat-adhoc` with `bat-story-eval` — adhoc previously treated duration as a primary comparison metric
- Explicitly lists "not used for evaluation" category (cached tokens, wall time) in all relevant files
- Changes `cost regression` label to `cost investigation` to reflect that token increases need root-cause analysis before being called regressions

## Files changed
- `.claude/skills/bat-adhoc/SKILL.md` — split comparison metrics into primary/secondary tiers
- `.claude/skills/bat-story-eval/SKILL.md` — added Metrics section with tiers, updated trend labels
- `.claude/skills/bat-story-eval/references/evaluation-protocol.md` — renamed Efficiency Metrics to Secondary, added exclusion list
- `.claude/skills/bat-story-eval/references/regression-protocol.md` — clarified token regression as secondary, added false positive guidance

## Test plan
- [ ] Docs-only change, no code affected
- [ ] Verify metric classification is consistent across all 4 files

### DIFF
--- a/.claude/skills/bat-adhoc/SKILL.md
+++ b/.claude/skills/bat-adhoc/SKILL.md
@@ -98,7 +98,8 @@ EOF
 *Secondary (report but don't decide on these alone):*
 - **Tool call count**: Compare `aggregate.total_tool_calls`, `aggregate.total_turns` — directional signal, not conclusive (agent exploration varies between runs)
 - **Duration**: Compare `aggregate.total_duration_ms` — noisy due to network, cache misses, server load. Only flag large (>2x) regressions.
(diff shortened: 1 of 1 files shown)


julienld (commented):
Good catch — moved variation testing out of the metrics list into its own tip. It's a testing technique, not a metric.

sergeykad (approved):

Issue #651, opened 2026-02-19
[BUG] Server crashes on initialize with Claude Desktop 1.1.3541 (protocolVersion 2025-11-25)


mpierce80:
### 💬 What Happened?

After Claude Desktop auto-updated this morning (Feb 19, 2026), ha-mcp stopped 
working entirely. The server starts, receives the initialize handshake, then 
immediately crashes. This worked perfectly for over a week with no config changes.

The root cause appears to be Claude Desktop now sending protocolVersion 
"2025-11-25" in the initialize message, which ha-mcp does not support, causing 
it to exit immediately.

Running ha-mcp manually from terminal with correct env vars works fine with no 
errors. The crash is specific to the Claude Desktop MCP handshake.

### 📋 Copy-Paste from Conversation

```markdown
Environment:
- Claude Desktop: 1.1.3541 (1e65e4) 2026-02-18T19:23:37.000Z
- ha-mcp: 6.6.1 (also tested @latest and git main)
- OS: macOS
- Install method: uvx / install-macos.sh script
- HA Version: 2025.10.3

MCP Log showing the crash:
---
[info] Initializing server...
[info] Using MCP server command: /Users/mattpierce/.local/bin/uvx 
  args: [ '--refresh', 'ha-mcp@latest' ]
[info] Server started and connected successfully
[info] Message from client: {
  "method":"initialize",
  "params":{
    "protocolVersion":"2025-11-25",
    "capabilities":{
      "extensions":{
        "io.modelcontextprotocol/ui":{
          "mimeTypes":["text/html;profile=mcp-app"]
        }
      }
    },
    "clientInfo":{"name":"claude-ai","version":"0.1.0"}
  },
  "jsonrpc":"2.0",
  "id":0
}
[info] Server transport closed
[info] Client transport closed
[error] Server disconnected. Server transport closed unexpectedly, 
this is likely due to the process exiting early.
---

Troubleshooting already attempted:
- Cleared uvx cache (rm -rf ~/.cache/uv)
- Tested with ha-mcp==6.6.1 pinned version
- Tested with --refresh ha-mcp@latest
- Tested with git+https://github.com/homeassistant-ai/ha-mcp
- Ran official install-macos.sh reinstall script
- Verified HA URL and token work correctly via curl
- Verified ha-mcp launches correctly in terminal with env vars
- Fully quit and relaunched Claude Desktop multiple times
- All attempts produce identical crash on initialize handshake

[...]

PR #662, opened 2026-02-22
deps: pin and upgrade all runtime dependencies

julienld:
## Summary
- Pin all 8 runtime dependencies to exact versions (`==`) to prevent upstream breakage
- Remove `mcp` as direct dependency (transitive via `fastmcp`, no need to pin separately)
- Upgrade `fastmcp` 3.0.0→3.0.2, `python-dotenv` 1.2.1→1.2.2, `mcp` 1.24.0→1.26.0 (transitive)
- Future upgrades managed by Renovate/Dependabot via PRs

## Test plan
- [ ] CI passes
- [ ] `uv lock --check` passes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,15 +24,14 @@ classifiers = [
 ]
 
 dependencies = [
-    "fastmcp>=3.0.0,<4.0.0",
-    "mcp>=1.24.0",  # Required for protocol version 2025-11-25 support (see #651)
-    "httpx[socks]>=0.27.0,<1.0",
-    'jq>=1.8.0; sys_platform != "win32"',
-    "pydantic>=2.5.0",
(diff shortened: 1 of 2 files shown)


julienld (commented):
Addressed both Gemini comments in 7a0e0bb:

1. **`pyproject.toml` - mcp dependency**: Restored `mcp==1.24.0` as a direct pinned dependency with the explanatory comment about protocol version support (#651). It was incorrectly removed since `mcp` is directly imported in `server.py` and `auth/provider.py`.

2. **`uv.lock` - revision change**: The revision field is a content-derived counter that changes when the dependency list changes. Since we modified the dependencies, the revision updating is expected and correct behavior. `uv lock --check` passes.

PR #679, opened 2026-02-27
feat: serve bundled HA skills as MCP resources

### LINKED ISSUE #670: feat: serve bundled HA skills as MCP resources via FastMCP SkillsDirectoryProvider
## Problem

The [homeassistant-ai/skills](https://github.com/homeassistant-ai/skills) repo provides HA best-practice skills (automation patterns, helper selection, device control, etc.), but users must manually install them into their agent's skill directory (`~/.claude/skills/`, `~/.cursor/skills/`, etc.). This means:

- Users who don't know about the skills repo never benefit from them
- Each client platform requires separate manual installation
- Skills aren't discoverable through the MCP server itself

## Proposal

FastMCP 3.0 introduced [SkillsDirectoryProvider](https://gofastmcp.com/servers/providers/skills) — a provider that serves skill directories (SKILL.md + supporting files) as standard MCP resources via `skill://` URIs. ha-mcp already depends on `fastmcp>=3.0.0`.

The idea: **bundle the skills from `homeassistant-ai/skills` into ha-mcp and serve them via `SkillsDirectoryProvider`**. Any MCP client connecting to ha-mcp would automatically discover them via `list_resources()` and can read them on demand.

### Implementation sketch

```python
# server.py — during init
from fastmcp.server.providers.skills import SkillsDirectoryProvider

skills_dir = Path(__file__).parent / "resources" / "skills"
self.mcp.add_provider(SkillsDirectoryProvider(roots=skills_dir))
```

Bundle the skill files in `src/ha_mcp/resources/skills/home-assistant-best-practices/` (SKILL.md + 6 reference files), add a `ENABLE_SKILLS` config toggle, and include them in package data.

### What this gets us

- **Zero-install discoverability**: Any MCP client can call `list_resources()` and see the skills without manual setup
- **Minimal idle context cost**: Resources are NOT auto-injected into the LLM context. Only metadata (URIs, names, descriptions — ~200 tokens) appears in `list_resources()`. The actual ~70KB of skill content is only loaded when explicitly read via `resources/read`
- **Client-agnostic**: Works at the MCP protocol level, so any client that supports MCP resources can access them

### Limitations to be aware of

- **Resources are not auto-injected** — MCP resources are "application-driven" per [the spec](https://modelcontextprotocol.io/docs/concepts/resources). The client decides how to present them. In practice:
  - Claude Code: user types `@` and selects the skill resource
  - VS Code/Copilot: user selects via "Add Context > MCP Resources"
  - Claude Desktop: resource picker UI
  - Other clients: support varies
- **No client currently auto-loads resources** based on `annotations.priority` or `annotations.audience`, though the MCP spec defines these fields for that purpose
- **Skill content gets stale** if the upstream skills repo updates and ha-mcp hasn't re-bundled. Could be mitigated with CI that syncs from the skills repo on release

kingpanther13:
## Summary

- Bundle skills from [homeassistant-ai/skills](https://github.com/homeassistant-ai/skills) (~68KB) and serve them as MCP resources via FastMCP's `SkillsDirectoryProvider` using `skill://` URIs
- Add optional `ResourcesAsTools` transform for MCP clients that don't support resources natively (exposes `list_resources`/`read_resource` tools)
- Add `ENABLE_SKILLS` (default: true) and `ENABLE_SKILLS_AS_TOOLS` (default: false) config toggles, exposed in the HA add-on config panel
- Add automated weekly sync workflow to keep bundled skills in sync with the [skills repo](https://github.com/homeassistant-ai/skills)

### How it works

- **MCP resources** (default): Any client that supports `list_resources()` discovers the skills automatically. Resources are NOT auto-injected into LLM context — clients must explicitly request them. Clients that don't support resources silently ignore them (MCP capability negotiation).
- **Skills as tools** (opt-in): For tool-only clients (Gemini CLI, etc.), the `ResourcesAsTools` transform creates two synthetic tools. Enabled via `ENABLE_SKILLS_AS_TOOLS=true`.

### Auto-sync with skills repo

Skill files are bundled from [homeassistant-ai/skills](https://github.com/homeassistant-ai/skills). A `sync-skills.yml` GitHub Actions workflow runs weekly (Mondays 08:00 UTC) and on manual trigger — it fetches the latest skill files from the skills repo and opens a PR automatically if anything changed. This ensures bundled skills stay up to date without manual intervention.

### Files changed

| Area | Files |
|------|-------|
| Bundled skills | `src/ha_mcp/resources/skills/home-assistant-best-practices/` (SKILL.md + 6 reference files from [homeassistant-ai/skills](https://github.com/homeassistant-ai/skills)) |
| Core | `config.py` (2 new settings), `server.py` (`_register_skills()` method) |
| Packaging | `pyproject.toml` (package-data glob for skills) |
| Add-on | `config.yaml`, `start.py`, `translations/en.yaml` (both production and dev) |
| CI | `.github/workflows/sync-skills.yml` (weekly auto-sync from skills repo) |
| Docs | `README.md` (bundled skills section) |
| Tests | `tests/addon/test_skills_config.py`, `tests/src/e2e/tools/test_skills_resources.py` |

Closes #670

## Test plan

- [ ] Addon tests pass: `pytest tests/addon/test_skills_config.py`
- [ ] E2E tests pass: `pytest tests/src/e2e/tools/test_skills_resources.py`
- [ ] CI lint/type checks pass
- [ ] Verify skills appear in `list_resources()` with a real MCP client
- [ ] Verify `ENABLE_SKILLS=false` hides resources
- [ ] Verify `ENABLE_SKILLS_AS_TOOLS=true` adds synthetic tools

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -52,6 +52,9 @@ jobs:
     timeout-minutes: 5
 
     steps:
+    - name: Install git for submodule checkout
+      run: apt-get update -qq && apt-get install -y -qq git >/dev/null 2>&1
+
     - uses: actions/checkout@v6
       with:
         submodules: true


kingpanther13:
Yes it works, and yes it's loaded dynamically/incrementally at least on the resources side. Still trying to test the resources as tools side but it should work the same way from what I can tell. However, when testing with Sonnet it doesn't seem to automatically call the resources tool for the skills, it'll only call when I ask it to, so I'm going to modify the PR a bit to see if I can get like a bootstrap prompt on all relevant tools to tell the LLM to call the skills.md but only if the toggle is enabled in settings. This may actually be a nice way to cut down on our tool descriptions, from what I can tell for clients that don't support the resources thing we can just do the resources as tools and point it to them.

kingpanther13:
Ok I improved it, I put in like a bootstrap prompt that will prompt the llm to review the resources or resources tools depending on what's enabled. Sonnet and opus do fantastic with it but Haiku willfully skipped reading the deeper stuff. Makes me wonder if this could be a different alternative in the future to the other stuff I've been working on, essentially just putting tool descriptions into skills instead and having more than just one skills.md.

kingpanther13:
## Alternative approach: flattening the skill reference chain

During testing, we noticed that weaker models (e.g., Haiku) will read SKILL.md when prompted by the server `instructions`, but then skip reading the specific reference files that SKILL.md directs them to. They complete hop 1 (instructions → SKILL.md) but not hop 2 (SKILL.md → references/automation-patterns.md).

**Proposed alternative:** Move the decision tree, anti-patterns, and reference file mappings into the SKILL.md YAML frontmatter. The server would parse these at startup and include them directly in the `instructions` field, along with the specific `skill://` URIs for each reference file. This gives the LLM everything it needs in one shot — no intermediate hop through SKILL.md required.

**Trade-off:** Slightly more idle tokens in the system prompt (~500-600 tokens at session start) in exchange for fewer tool calls and better compliance from lesser models. The reference files themselves are still only loaded on-demand when the LLM reads the specific one matching its task.

**Impact on auto-updates:** None — frontmatter is structured YAML, which is easier to programmatically generate than markdown tables. The server parsing logic just reads more fields from the same frontmatter it already parses.

Will consider implementing if others think it's worth the change.

sergeykad (dismissed):
`_parse_skill_instructions` and` _extract_section` apply regex to extract structured data from the description field, which is designed for LLM consumption, not code parsing. This works against the intended design of the format and produces fragile logic that will break as the skill content evolves.

kingpanther13:
### Point 3 — Deeper analysis: pros, cons, and why this took deliberation

For transparency, here's the full trade-off analysis we went through before implementing this:

**Why we were initially hesitant:**

1. **Nested directory structure**: The skills repo has a `skills/` subdirectory inside it (plus repo-level files like README, LICENSE, .git). Placing the submodule directly at `src/ha_mcp/resources/skills` (as originally suggested) would mean `_get_skills_dir()` points to `resources/skills/skills/` — a confusing double-nested path. We worked around this by placing it at `resources/skills-vendor/` instead, so the path is `resources/skills-vendor/skills/`.

2. **Extra files in the wheel**: The submodule root includes repo metadata (~8KB of README, LICENSE, .gitignore etc.) that aren't skill content. With the `package_data` glob `resources/skills-vendor/**/*`, these get bundled into the wheel. With the old committed-copies approach, we only copied the actual skill files. In practice ~8KB is negligible, and dotfiles/dot-directories are excluded by setuptools glob behavior, so this turned out to be a non-issue.

3. **CI checkout complexity**: Every workflow that builds, tests, or packages the code now needs `submodules: true` on its checkout step. Missing it in any workflow means a silent degradation (skills directory doesn't exist, `_get_skills_dir()` returns None, skills gracefully disabled). We had to audit all 6+ workflow files to add it. The old approach with committed copies "just worked" with a bare checkout.

4. **Local dev experience**: `git clone` without `--recurse-submodules` produces a repo where skills don't work. The old approach with committed copies worked out of the box. This needs a note in CONTRIBUTING.md (as you mentioned).

**Why we ultimately agreed with your approach:**

1. **Single source of truth** — The biggest win. No sync script, no sync workflow, no risk of committed copies drifting from the submodule. This was the core problem you identified and it's completely eliminated.

2. **Less machinery** — Deleted `scripts/sync-skills.sh` (50+ lines) and `.github/workflows/sync-skills.yml` entirely. That's fewer moving parts to maintain, fewer CI minutes, and no more automated PRs when skills change.

3. **Force-push concern vanishes** — Your point 10 about `--force-with-lease` not protecting in fresh CI checkouts becomes completely moot. No sync workflow = no force-push = no risk.

4. **Industry standard pattern** — Submodule-as-package-resource is how many projects handle vendored content. The `submodules: true` checkout step is well-understood.

Net: the pros clearly outweigh the cons. The hesitation was about the nested path awkwardness and CI audit surface, but both turned out manageable.

sergeykad (approved):

PR #689, opened 2026-02-28
feat: add generic integration options inspection

transportrefer:
## Summary
Add generic support for inspecting config-entry options flows.

## Changes
- add `start_options_flow(entry_id)` and `submit_options_flow_step(flow_id, user_input)` to the REST client
- add `ha_get_integration_options`
- add E2E coverage for the new inspection path

## Notes
- this PR is read/inspect-focused
- it exposes persisted options plus the initial options-flow metadata for any config entry that supports Options Flow
- follow-up write helpers can build on this if the direction makes sense

### DIFF
--- a/.github/workflows/_update-addon-config.yml
+++ b/.github/workflows/_update-addon-config.yml
@@ -1,0 +1,84 @@
+name: Update Add-on Config Version
+
+# Reusable workflow to update homeassistant-addon/config.yaml after Docker images are published.
+# Called from semver-release.yml and hotfix-release.yml to avoid duplication.
+on:
+  workflow_call:
+    inputs:
+      version:
+        description: 'Version to set in config.yaml (without v prefix)'
+        required: true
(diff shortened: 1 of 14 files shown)


julienld:
Hi @transportrefer, this has been superseded by #712 which takes a different design approach for integration options inspection.

**Design decision:** Rather than a dedicated `ha_get_integration_options` tool (which would start a flow internally and either leak it or add hidden side effects), the inspection use case is covered by composing two existing tools:
1. `ha_get_integration(entry_id)` → returns persisted `options` field
2. `ha_start_options_flow(entry_id)` → inspect schema, then `ha_abort_options_flow(flow_id)` to clean up

This also avoids the resource leak (confirmed from HA Core source: flows have no TTL and can block new option flows for the same entry on integrations that check `_async_in_progress()`).

Thank you for the contribution and for identifying the options flow inspection use case!

Issue #726, opened 2026-03-07
feat: CodeMode sandbox tool for custom one-off tools, and multi-step tool chaining


kingpanther13:
## Summary

Add an optional sandboxed code execution tool that lets LLMs **create custom one-off tools on the fly** when no existing tool covers the user's request, using [FastMCP 3.1's CodeMode](https://gofastmcp.com/servers/transforms/code-mode) sandbox (`pydantic-monty`).

## Problem

The ha-mcp tool catalog — no matter how large — will never cover every possible Home Assistant operation. Today, when a user asks for something there's no dedicated tool for (e.g., "edit entity categories", "pull debug logs", "interact with a third-party add-on"), the LLM fails in one of several ways:

- **Gives up entirely**: "Sorry, I don't have a tool for that" — even though the operation is technically possible through existing lower-level APIs.
- **Confabulates**: Claims the operation is impossible or unsupported by Home Assistant, when it's really just unsupported by the current tool catalog.
- **Thrashes through similar tools**: Tries a series of related-but-wrong tools, failing repeatedly, wasting tokens and round-trips before ultimately giving up anyway.

But the LLM *could* accomplish many of these tasks by writing custom code that calls the existing lower-level tools or HA APIs in the right combination. The gap isn't always capability — it's that the LLM has no way to run custom code to bridge the gaps.

## Proposal

Add an `ha_execute_code` tool that accepts Python code with access to `call_tool(name, args)`. The code runs in [Monty](https://github.com/pydantic/monty) — a sandboxed Python interpreter (Rust-based, no filesystem/network access, configurable resource limits).

### Primary use case: On-the-fly custom tools

The core value is giving the LLM a **general-purpose fallback**. When no existing tool matches the user's request, the LLM writes custom Python code on the spot to accomplish it:

```
User: "Move my kitchen sensors to the 'Climate' category"
LLM:  No dedicated tool exists for editing entity categories.
      → Writes Python that calls the HA REST API via available tools to modify categories
      → Returns the result

User: "Show me the debug logs for the Zigbee integration"
LLM:  No tool exposes debug logs directly.
      → Writes Python to access the debug log endpoint and filter for the relevant integration
      → Returns the result

User: "Check the status of my AdGuard add-on"
LLM:  No tool covers third-party add-on interaction.
      → Writes Python to query the Supervisor add-on API
      → Returns the result
```

[...]

PR #637, opened 2026-02-17
feat: Consolidate dashboard tools (category gateway proxy for tool consolidation Part 1)

kingpanther13:
## Problem

With 80+ MCP tools, idle context cost is significant — every tool's name, description, and parameter schema is loaded into the LLM context on every request, whether or not those tools are needed. This wastes tokens and dilutes attention on the tools that matter.

PR #616 tackles this by deferring tool loading. PR #627 attempted server-side proxying with meta-tools (`ha_find_tools` / `ha_get_tool_details` / `ha_execute_tool`), but the 3-step pattern creates a discoverability black hole — LLMs don't proactively check generic meta-tools.

## Solution

**Domain-named category gateways** — group lesser-used tools behind MCP gateway tools with clear, descriptive names. Each gateway:

- Shows tool names with **typed required parameter signatures** in the MCP description (always visible to the LLM)
- Allows **direct execution** — LLMs call `ha_manage_dashboards(tool="ha_config_set_dashboard", args={"url_path": "..."})` without any discovery step
- Accepts `args` as a **native JSON object** (no double-serialization required) or as a JSON string for backwards compatibility
- Uses **Literal enum** on the `tool` parameter so valid tool names appear in the MCP `inputSchema`, enabling constrained decoding

### Key design principles

1. **Discovery is optional, not mandatory.** The gateway description includes enough information (tool names + required params with types) for LLMs to call tools directly. This avoids the token-compounding problem where a discovery response gets injected into conversation history and inflates every subsequent API call.

2. **Show, don't tell.** The gateway description includes a concrete example call rather than abstract templates, which helps smaller models understand the pattern immediately.

3. **Minimize per-call friction.** `args` accepts `dict | str` so LLMs can pass arguments as native JSON objects instead of serializing JSON inside a JSON string.

4. **Read/write split preserves MCP annotations.** Separate gateways for read-only vs write operations so clients like Claude Code can auto-approve read-only calls without confirmation prompts.

## Phase 1 (this PR): Dashboard gateways

Proof-of-concept with the biggest context savings. 9 tools from `tools_config_dashboards` + `tools_resources` (post-#660 consolidation) split into 2 gateways:

### ha_dashboard_info (read-only, `readOnlyHint: true`)

| Tool | Description |
|------|-------------|
| `ha_config_get_dashboard` | Get/list dashboard configurations |
| `ha_dashboard_find_card` | Find cards by entity, type, or heading |
| `ha_get_dashboard_guide` | Get dashboard design guide |
[...]

PR #727, opened 2026-03-07
feat: Search-based tool discovery with categorized call proxies

kingpanther13:
## What does this PR do?

Adds an opt-in `CategorizedSearchTransform` (`ENABLE_TOOL_SEARCH=true`) that replaces the full 89+ tool catalog with a unified BM25 search tool and three categorized call proxies (read/write/delete). Reduces idle context from ~46K tokens to ~2-3K tokens while maintaining full functionality and client compatibility.

**How it works:**
- `search_tools` — unified BM25 search across all tools (results include full annotations so the LLM knows read vs write vs delete)
- `call_read_tool` — `readOnlyHint: true`, clients can auto-approve
- `call_write_tool` — `destructiveHint: true`, requires confirmation
- `call_delete_tool` — `destructiveHint: true`, requires confirmation

**Pinned tools** (always visible for individual permission gating):
`ha_restart`, `ha_reload_core`, `ha_backup_create`, `ha_backup_restore`, `ha_get_overview`, `ha_report_issue`, and `list_resources`/`read_resource` (when `ENABLE_SKILLS_AS_TOOLS=true`)

**Bootstrap for LLM awareness:**
- Server `instructions` updated with tool discovery workflow when enabled
- `ha_get_overview` response includes `tool_discovery` section with hints and pinned tool list
- Search tool description lists all categories and instructs the LLM to search first

**BM25 search quality tuning (`SearchKeywordsTransform`):**
- A lightweight FastMCP `Transform` subclass that appends extra keywords to specific tool descriptions so BM25 ranks them correctly for common queries
- Only active when `ENABLE_TOOL_SEARCH=true` — original tool docstrings are unchanged
- Fixes three ranking issues found during search quality testing:
  - "find entities" → `ha_search_entities` ranks above `ha_deep_search`
  - "get/read automation" → `ha_config_get_automation` ranks above `ha_config_set_automation`
  - "create helper" → `ha_config_set_helper` ranks above `ha_config_remove_helper`
- To tune rankings for additional tools, add an entry to the `_SEARCH_KEYWORDS` dict in `server.py` — no other files need changing

Subclasses FastMCP 3.1's `BM25SearchTransform` (~200 lines). Tools auto-categorized by existing MCP annotations. Works with ANY MCP client. Pure opt-in — disable the flag and everything works exactly as before.

Related: #637 (category gateways approach), #726 (CodeMode sandbox)

## Type of change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 📚 Documentation
- [x] 🔧 Maintenance/refactor
- [x] 💥 Breaking change

## Testing

## Checklist

### DIFF
--- a/.github/ISSUE_TEMPLATE/agent_behavior.yml
+++ b/.github/ISSUE_TEMPLATE/agent_behavior.yml
@@ -2,9 +2,6 @@ name: "🤖 AI Agent Behavior Feedback"
 description: "Report AI agent inefficiency, wrong tool usage, or suggest workflow improvements"
 title: "[AGENT] "
 labels: ["agent-behavior"]
-assignees:
-  - julienld
-
 body:
   - type: markdown
     attributes:
--- a/.github/ISSUE_TEMPLATE/runtime_bug.yml
+++ b/.github/ISSUE_TEMPLATE/runtime_bug.yml
@@ -2,9 +2,6 @@ name: "🐛 Runtime Bug"
 description: "Report a bug that occurs while ha-mcp is running (errors, crashes, incorrect behavior)"
 title: "[BUG] "
 labels: ["bug"]
-assignees:
-  - julienld
-
 body:
   - type: markdown
     attributes:
--- a/.github/ISSUE_TEMPLATE/startup_bug.yml
+++ b/.github/ISSUE_TEMPLATE/startup_bug.yml
@@ -2,9 +2,6 @@ name: "🚀 Startup/Installation Bug"
 description: "Report a bug preventing ha-mcp from starting, installing, or connecting"
 title: "[BUG] "
 labels: ["bug", "startup"]
-assignees:
-  - julienld
-
 body:
   - type: markdown
     attributes:
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -14,7 +14,7 @@ This repository uses a worktree-based development workflow.
 
 **Directory Structure:**
 ```
-/home/julien/github/ha-mcp/           # Main repository (checkout master here)
+<repo-root>/                           # Main repository (checkout master here)
 ├── AGENTS.md                          # This file (canonical source)
 ├── CLAUDE.md -> AGENTS.md             # Symlink for convenience
 ├── worktree/                          # Git worktrees (gitignored)
@@ -40,7 +40,7 @@ This repository uses a worktree-based development workflow.
 
 ```bash
 # Correct - worktrees go in worktree/ subdirectory
-cd /home/julien/github/ha-mcp
+cd <repo-root>
 git worktree add worktree/issue-42 -b issue-42
 git worktree add worktree/feat-new-feature -b feat/new-feature
 
--- a/src/ha_mcp/__main__.py
+++ b/src/ha_mcp/__main__.py
@@ -573,7 +573,7 @@ async def _run_http_with_graceful_shutdown(
 _registered_landing_paths: set[str] = set()
 
 
-def register_browser_landing(mcp_instance: Any, path: str) -> None:
+def register_browser_landing(mcp_instance: "FastMCP | _DeferredMCP", path: str) -> None:
     """Register a GET handler that returns 405 with a helpful message.
 
     Browsers and misconfigured clients that send GET instead of POST will see


src/ha_mcp/transforms/categorized_search.py
```diff
@@ -0,0 +1,207 @@
...
+
+def _categorize_tool(tool: Tool) -> str:
+    """Categorize a tool as read, write, or delete based on annotations and name."""
+    annotations = tool.annotations
+    if annotations and annotations.readOnlyHint:
+        return "read"
+    # Check for delete/remove pattern in tool name
+    if any(pattern in tool.name for pattern in _DELETE_PATTERNS):
+        return "delete"
+    return "write"
```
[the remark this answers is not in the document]
kingpanther13: Fixed: \ now requires both \ AND a name pattern (\/\) for delete categorization. A tool with \ in the name but without \ correctly falls through to write.

src/ha_mcp/transforms/categorized_search.py
```diff
@@ -0,0 +1,207 @@
...
+    def __init__(
+        self,
+        *,
+        max_results: int = 10,
+        always_visible: list[str] | None = None,
+        search_tool_name: str = "search_tools",
+        search_tool_description: str | None = None,
+        call_read_name: str = "call_read_tool",
+        call_write_name: str = "call_write_tool",
+        call_delete_name: str = "call_delete_tool",
```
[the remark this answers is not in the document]
kingpanther13: Fixed: All tool defaults renamed to \, \, \, \ — matching the \ convention. All references in server.py, tools_search.py, and descriptions updated accordingly.

src/ha_mcp/transforms/categorized_search.py
```diff
@@ -0,0 +1,207 @@
...
+
+        async def categorized_call(
+            name: Annotated[str, "The name of the tool to call"],
+            arguments: Annotated[
+                dict[str, Any] | None, "Arguments to pass to the tool"
+            ] = None,
+            ctx: Context = None,  # type: ignore[assignment]
+        ) -> Any:
+            # Rebuild cache on each call to pick up catalog changes
+            await transform._rebuild_category_cache(ctx)
```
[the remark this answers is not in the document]
kingpanther13: Fixed: Added \ flag. Cache is now built lazily on first proxy call and reused for subsequent calls, instead of rebuilding on every single call.

src/ha_mcp/tools/tools_search.py
```diff
@@ -544,6 +544,33 @@ async def ha_get_overview(
...
+                    "ha_report_issue",
+                    "ha_restart",
+                    "ha_reload_core",
+                    "ha_backup_create",
+                    "ha_backup_restore",
+                    "search_tools",
+                    "call_read_tool",
+                    "call_write_tool",
+                    "call_delete_tool",
+                ],
```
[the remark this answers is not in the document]
kingpanther13: Fixed: Added \ tuple constant in \, exported via \. \ now imports and uses it instead of a hardcoded list.

src/ha_mcp/tools/tools_search.py
```diff
@@ -544,6 +544,30 @@ async def ha_get_overview(
...
+                    "ha_call_delete_tool. Do NOT assume a capability is "
+                    "unavailable without searching first."
+                ),
+                "pinned_tools": sorted([
+                    *DEFAULT_PINNED_TOOLS,
+                    "ha_search_tools",
+                    "ha_call_read_tool",
+                    "ha_call_write_tool",
+                    "ha_call_delete_tool",
+                ]),
```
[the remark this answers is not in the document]
kingpanther13: Not fixing: the names are defaults and the server never customizes them. Over-engineering for a hypothetical.

src/ha_mcp/transforms/categorized_search.py
```diff
@@ -0,0 +1,255 @@
...
+                    correct_proxy = transform._call_delete_name
+                else:
+                    raise ValueError(
+                        f"Tool '{name}' not found. Use ha_search_tools to "
+                        f"discover available tools."
+                    )
+                raise ValueError(
+                    f"Tool '{name}' is a {actual_category} tool. "
+                    f"Use {correct_proxy} instead of {proxy_name}."
+                )
```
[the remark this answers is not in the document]
kingpanther13: Fixed: replaced ValueError with ToolError + create_error_response per style guide.

src/ha_mcp/server.py
```diff
@@ -248,6 +276,74 @@ def _build_skill_block(
...
+    # Tools pinned outside the search transform for individual permission gating.
+    # These are always visible in list_tools() regardless of search transform.
+    _PINNED_TOOLS: ClassVar[list[str]] = [
+        "ha_restart",
+        "ha_reload_core",
+        "ha_backup_create",
+        "ha_backup_restore",
+        "ha_get_overview",
+        "ha_report_issue",
+    ]
```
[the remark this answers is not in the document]
kingpanther13: Fixed: _PINNED_TOOLS now uses list(DEFAULT_PINNED_TOOLS) as single source of truth.

src/ha_mcp/transforms/categorized_search.py
```diff
@@ -0,0 +1,255 @@
...
+                "Execute a write tool discovered via ha_search_tools. "
+                "Creates or updates data.",
+            )
+        if name == self._call_delete_name:
+            return self._make_categorized_proxy(
+                self._call_delete_name, "delete",
+                ToolAnnotations(destructiveHint=True),
+                "Execute a delete/remove tool discovered via ha_search_tools. "
+                "Permanently removes data.",
+            )
```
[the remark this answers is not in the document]
kingpanther13: Fixed: extracted descriptions to module constants (_READ_PROXY_DESC, _WRITE_PROXY_DESC, _DELETE_PROXY_DESC) shared between transform_tools() and get_tool().

kingpanther13:
I've done a lot of BAT testing with this in various modes with various tools pinned, and per the BAT tests it actually takes longer and uses more tokens in stories where the context is cleared between stories. If it's a longer test when the llm has a better idea of what's going on, it does seem to save tokens and time eventually, but the BAT tests tend to clear context between tests. I also tested an iteration of codemode with similar results. The problem is that I really only am capable of reliably testing with Claude at the moment and Claude now has its own automatic deferred toolsearch function. Haiku shows improvement because it doesn't use Toolsearch. 

Therefore I believe that this is still beneficial for people that use LLMs that don't have deferred tools built in, but it won't be useful for those of use who use Claude. I'm willing to tinker more with suggestions if people want to try pinning different tools and stuff, but I'm pretty exhausted from testing. I think it's pretty much ready how it is but I'll modify with any suggestions that are given. If there's any interest in me adding codemode as a toggleable mode as well I can copy over my work for that too (https://github.com/kingpanther13/ha-mcp-fork/pull/11 is what I've been using for testing but it will obviously need some work before it's made into a full PR here).

src/ha_mcp/server.py
```diff
@@ -208,7 +219,37 @@ def _build_skills_instructions(self) -> str | None:
...
+                "   a) DIRECT CALL (preferred): Call the tool directly by "
+                "name. All discovered tools are callable without a proxy.\n"
+                "   b) VIA PROXY: For permission-gated execution, use the "
+                "matching proxy:\n"
+                "      - ha_call_read_tool \u2014 safe, read-only operations\n"
+                "      - ha_call_write_tool \u2014 creates or modifies data\n"
+                "      - ha_call_delete_tool \u2014 removes data permanently\n\n"
+                "Once you know a tool\u2019s name, you do NOT need to search "
+                "again \u2014 call it directly.\n\n"
+                "A few critical tools are listed directly (ha_restart, "
```
sergeykad: The instruction text hardcodes specific tool names (`ha_restart`, `ha_backup_create`, etc.), but `_PINNED_TOOLS` (line 310) derives from `DEFAULT_PINNED_TOOLS` which includes `ha_search_entities`, `ha_config_get_automation`, and `ha_config_set_automation` — none of which appear in this instruction. These two lists will drift silently as tools are added or renamed. Consider generating this sentence from `DEFAULT_PINNED_TOOLS` rather than hardcoding names.

src/ha_mcp/server.py
```diff
@@ -264,7 +305,163 @@ def _build_skill_block(
 
         return f"\n### Skill: {skill_name} ({uri})\n{description.strip()}"
 
+    # Tools pinned outside the search transform for individual permission gating.
+    # These are always visible in list_tools() regardless of search transform.
+    _PINNED_TOOLS: ClassVar[list[str]] = list(DEFAULT_PINNED_TOOLS)
```
sergeykad, line 309: `_PINNED_TOOLS` is derived from `DEFAULT_PINNED_TOOLS` at class definition time. If `DEFAULT_PINNED_TOOLS` ever changes, the instruction text at line 244 won't update automatically. Both the instructions and this list should come from a single source of truth.

src/ha_mcp/server.py
```diff
@@ -264,7 +305,163 @@ def _build_skill_block(
...
+        remain directly visible in list_tools() for individual permission
+        gating. ResourcesAsTools (list_resources/read_resource) are also
+        pinned when enabled.
+        """
+        if not self.settings.enable_tool_search:
+            return
+
+        try:
+            from .transforms import CategorizedSearchTransform
+        except ImportError:
```
sergeykad, line 412: If `CategorizedSearchTransform` fails to import (missing dependency, broken module), the server silently continues with `enable_tool_search=True` in settings but the transform not applied — the full tool catalog is exposed, which contradicts the user's configuration. This is a broken state that's invisible until the user notices their LLM has too many tokens. Consider raising instead of returning, or at minimum logging at `ERROR` level rather than `WARNING`.

src/ha_mcp/tools/tools_search.py
```diff
@@ -580,6 +580,32 @@ async def ha_get_overview(
             except Exception as e:
                 logger.debug(f"Failed to fetch notifications for overview: {e}")
```
sergeykad: `logger.debug` is invisible at the default `INFO` log level in production. If notifications fail to load, the overview silently returns without them — the user has no way to know. This should be `logger.warning` so it surfaces in normal operation.

src/ha_mcp/tools/tools_search.py
```diff
@@ -580,6 +580,32 @@ async def ha_get_overview(
...
 
+        # Include tool discovery hint when search transform is active
+        from ..config import get_global_settings
+
+        settings = get_global_settings()
+        if settings.enable_tool_search:
```
sergeykad, line 587: This import is inside the function body and runs on every `ha_get_overview` call. If the import fails (e.g., circular import or missing module), it raises `ImportError` at runtime, mid-response. Move this import to the top of the file with a try/except guard so failures are caught at startup, not during tool execution.

src/ha_mcp/tools/tools_search.py
```diff
@@ -580,6 +580,32 @@ async def ha_get_overview(
...
+            from ..transforms.categorized_search import DEFAULT_PINNED_TOOLS
+
+            result["tool_discovery"] = {
```
sergeykad, line 588: The hint says to execute via `ha_call_read_tool`, `ha_call_write_tool`, or `ha_call_delete_tool`, but the server instructions (server.py:240) document direct calls as the **preferred** path — "DIRECT CALL (preferred): Call the tool directly by name."

This hint is inconsistent: it omits the direct-call option, which may cause the LLM to always route through proxies even when unnecessary. Add a note that discovered tools can also be called directly.

src/ha_mcp/transforms/categorized_search.py
```diff
@@ -0,0 +1,367 @@
...
+
+    async def get_tool(
+        self, name: str, call_next: GetToolNext, *, version: VersionSpec | None = None
+    ) -> Tool | None:
+        tool = await call_next(name, version=version)
+        return self._enrich(tool) if tool else None
+
+# Proxy tool descriptions (shared between transform_tools and get_tool)
+_READ_PROXY_DESC = (
+    "Execute a read-only tool discovered via ha_search_tools. "
```
sergeykad: These proxy descriptions hardcode `ha_search_tools` by name, but `CategorizedSearchTransform.__init__` accepts a configurable `search_tool_name` parameter (default `"ha_search_tools"`). If the caller passes a custom name, the proxy descriptions will refer to the wrong tool. Build these strings inside `_make_categorized_proxy` using `self._search_tool_name` so they stay correct.

src/ha_mcp/transforms/categorized_search.py
```diff
@@ -0,0 +1,367 @@
...
+        call_delete_name: str = "ha_call_delete_tool",
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(
+            max_results=max_results,
+            always_visible=always_visible,
+            search_tool_name=search_tool_name,
+            # Use a placeholder call_tool_name — we override transform_tools
+            # so the default call_tool is never surfaced.
+            call_tool_name="_unused_call_tool",
```
sergeykad: `_unused_call_tool` is not unused — it's passed to `BM25SearchTransform` as the `call_tool_name` placeholder so the parent doesn't surface its default proxy. Naming it `_unused_` implies dead code and will mislead future readers. A name like `_hidden_call_tool` or `_base_call_proxy` would be clearer.

src/ha_mcp/transforms/categorized_search.py
```diff
@@ -0,0 +1,367 @@
...
+        self._call_read_name = call_read_name
+        self._call_write_name = call_write_name
+        self._call_delete_name = call_delete_name
+        self._search_tool_description = search_tool_description
+
+        # Category caches built lazily from the tool catalog
+        self._read_tools: set[str] = set()
+        self._write_tools: set[str] = set()
+        self._delete_tools: set[str] = set()
+        self._cache_built = False
```
sergeykad: Once `_cache_built` is set to `True` it is never reset. If HA entities or tools change at runtime (new integration loaded, tool registered dynamically), the category cache drifts silently with no way to recover until server restart. Consider invalidating the cache on a TTL or when `list_tools` is called more than N seconds after the last rebuild.

src/ha_mcp/transforms/categorized_search.py
```diff
@@ -0,0 +1,367 @@
...
+
+    async def _rebuild_category_cache(self, ctx: Any) -> None:
+        """Rebuild the read/write/delete category sets from the catalog."""
+        catalog = await self.get_tool_catalog(ctx)
+        self._read_tools.clear()
```
sergeykad: The three `set.clear()` calls happen **before** `get_tool_catalog(ctx)` awaits. If the catalog fetch raises, the sets are left empty and `_cache_built` remains `False` — the next call retries, which is correct. However if `_cache_built` were set to `True` before the error (it isn't here, but easy to introduce), all proxy calls would silently route to an empty set and raise `ToolError` for every tool name.

Safer pattern: collect into local variables, then atomically update the instance sets only on success:
```python
read, write, delete = set(), set(), set()
for tool in catalog:
    ...
self._read_tools, self._write_tools, self._delete_tools = read, write, delete
self._cache_built = True
```

src/ha_mcp/transforms/categorized_search.py
```diff
@@ -0,0 +1,367 @@
...
+                f'client.{proxy}(name="{tool.name}", arguments={{...}}) '
+                f'or {proxy}(name="{tool.name}", arguments={{...}})'
+            )
+            results.append(data)
+        return results
+
+    def _make_categorized_proxy(
+        self,
+        proxy_name: str,
+        category: str,
```
sergeykad: `category: str` accepts any string. Since only `"read"`, `"write"`, and `"delete"` are valid, annotate this as `Literal["read", "write", "delete"]` to catch typos at type-check time:
```python
from typing import Literal
def _make_categorized_proxy(
    self,
    proxy_name: str,
    category: Literal["read", "write", "delete"],
    ...
```

src/ha_mcp/transforms/categorized_search.py
```diff
@@ -0,0 +1,367 @@
...
+    ) -> Tool:
+        """Create a call proxy that validates tool category before execution."""
+        transform = self
+
+        async def categorized_call(
+            name: Annotated[str, "The name of the tool to call"],
+            arguments: Annotated[
+                dict[str, Any] | None, "Arguments to pass to the tool"
+            ] = None,
+            ctx: Context = None,  # type: ignore[assignment]
```
sergeykad, line 258: `ctx: Context = None` is an incorrect annotation — `None` is not a valid `Context`. This causes type checker errors and misleads readers about whether `ctx` can be `None`. The correct annotation is `ctx: Context | None = None` (or `Optional[Context] = None` for older style).

src/ha_mcp/transforms/categorized_search.py
```diff
@@ -0,0 +1,367 @@
...
+        ) -> Any:
+            # Lazily build category cache if not already populated
+            if not transform._cache_built:
```
sergeykad: There is no `asyncio.Lock` protecting `_rebuild_category_cache`. If two concurrent `ha_call_*` calls arrive with `_cache_built=False`, both will call `_rebuild_category_cache` simultaneously — the second rebuild starts before the first finishes, resulting in interleaved set writes and a corrupted cache. Add an `asyncio.Lock` initialized in `__init__` and acquire it in `_rebuild_category_cache`:
```python
self._cache_lock = asyncio.Lock()

async def _rebuild_category_cache(self, ctx):
    async with self._cache_lock:
        if self._cache_built:  # re-check under lock
            return
        ...
```

src/ha_mcp/transforms/categorized_search.py
```diff
@@ -0,0 +1,367 @@
...
+                    transform._call_delete_name,
+                )
+                and arguments["name"] in all_known
+            ):
+                logger.warning(
+                    "Detected double-wrapped proxy call for '%s' via %s — unwrapping",
+                    arguments["name"],
+                    name,
+                )
+                name = arguments["name"]
```
sergeykad, line 294: After unwrapping, `name` is updated to `arguments["name"]` but the category check on the next line still uses `allowed` — the set for the **original** proxy that was called. If the inner tool is in a different category (e.g., the LLM called `ha_call_read_tool` but the inner tool is actually `write`), the unwrapped call will succeed when it should be rejected, or be rejected when it should succeed. After unwrapping, re-evaluate which `allowed` set applies:
```python
name = arguments["name"]
arguments = arguments.get("arguments") or {}
# Re-evaluate category for the unwrapped tool
if name in transform._read_tools:
    allowed = transform._read_tools
elif name in transform._delete_tools:
    allowed = transform._delete_tools
else:
    allowed = transform._write_tools
```

tests/src/unit/test_categorized_search.py
```diff
@@ -0,0 +1,295 @@
...
+    def test_contains_critical_tools(self):
+        assert "ha_restart" in DEFAULT_PINNED_TOOLS
+        assert "ha_get_overview" in DEFAULT_PINNED_TOOLS
+        assert "ha_backup_create" in DEFAULT_PINNED_TOOLS
+        assert "ha_backup_restore" in DEFAULT_PINNED_TOOLS
+        assert "ha_report_issue" in DEFAULT_PINNED_TOOLS
+        assert "ha_reload_core" in DEFAULT_PINNED_TOOLS
+
+    def test_is_immutable_tuple(self):
+        assert isinstance(DEFAULT_PINNED_TOOLS, tuple)
```
sergeykad, line 299: The `categorized_call` closure — the core dispatch function created by `_make_categorized_proxy` — has no tests. This is the most critical path: it validates categories, handles double-unwrap, and calls the actual tool. The following are also untested:

- `categorized_call` happy path: correct tool in correct category succeeds
- `categorized_call` wrong-category error: correct proxy name appears in error message
- `categorized_call` unknown tool error: `RESOURCE_NOT_FOUND` is returned
- Double-unwrap: double-wrapped call is detected and unwrapped correctly
- Double-unwrap category re-evaluation: unwrapped tool ends up using the right `allowed` set
- `_rebuild_category_cache`: populates read/write/delete sets correctly from tool catalog
- `_rebuild_category_cache`: fetch failure leaves sets empty and `_cache_built=False`
- `SearchKeywordsTransform`: keywords appended, overrides replace, neither applied when absent

These gaps mean a regression in the dispatch or categorization logic would not be caught by the test suite.

src/ha_mcp/server.py
kingpanther13: Fixed: instruction text now generated from `DEFAULT_PINNED_TOOLS` via `join()` so it stays in sync automatically.

src/ha_mcp/server.py
kingpanther13: Fixed: same change — the instruction text at line 244 now reads from `DEFAULT_PINNED_TOOLS` directly, so both the instruction text and `_PINNED_TOOLS` share a single source of truth.

src/ha_mcp/server.py
kingpanther13: Fixed: changed to `logger.error` with an explicit message about the consequence (full catalog exposed). Kept `return` rather than raising — the server is still functional, just not in the mode the user configured.

src/ha_mcp/tools/tools_search.py
kingpanther13: Fixed: changed to `logger.warning` for consistency with the system_info failure handler on line 558.

src/ha_mcp/tools/tools_search.py
kingpanther13: Fixed: hint now mentions direct calling as the preferred path, consistent with server instructions.

src/ha_mcp/transforms/categorized_search.py
kingpanther13: Fixed: renamed to `_base_call_proxy` with updated comment explaining it's a placeholder for the base class's single call proxy that we replace with categorized proxies.

src/ha_mcp/transforms/categorized_search.py
kingpanther13: Fixed: `category` parameter now typed as `Literal['read', 'write', 'delete']`.

src/ha_mcp/transforms/categorized_search.py
kingpanther13: Fixed: added `asyncio.Lock` in `__init__` and `async with self._cache_lock` in `_rebuild_category_cache` with an early return if `_cache_built` is already True inside the lock.

src/ha_mcp/tools/tools_search.py
kingpanther13: Fixed: moved both imports to the top of the file. Traced the import chains — `config` and `transforms.categorized_search` have no dependencies on `ha_mcp.tools`, so there's no circular import risk. `server.py` already imports both at module level.

src/ha_mcp/transforms/categorized_search.py
kingpanther13: Fixed: proxy descriptions are now built via `_build_proxy_descriptions(search_tool_name)` in `__init__`, so they correctly reference whatever name is configured. Stored as `self._proxy_descs` dict keyed by category.

src/ha_mcp/transforms/categorized_search.py
kingpanther13: Fixed: replaced the boolean `_cache_built` flag with hash-based staleness detection matching the BM25 base class pattern. `_rebuild_category_cache` now computes a SHA256 hash of tool names + categories on every call and only rebuilds when the hash changes. Also uses local-variable accumulation + atomic swap for the sets, with double-checked locking via `asyncio.Lock`.

src/ha_mcp/transforms/categorized_search.py
kingpanther13: Addressed via the hash-based rebuild in the previous comment — the cache now uses local-variable accumulation (`read`, `write`, `delete` local sets) and atomically swaps them into the instance attributes only after the loop completes. Combined with `asyncio.Lock`, this fully prevents any caller from observing partially-built sets.

src/ha_mcp/transforms/categorized_search.py
kingpanther13: Not changing: `ctx: Context = None` with `# type: ignore[assignment]` follows FastMCP's own convention — `BaseSearchTransform._make_call_tool()` (base.py:232) uses this exact pattern. While `Context | None = None` would technically work with FastMCP's DI (it handles Union types), diverging from the upstream pattern this code directly inherits from would create an inconsistency for anyone reading both. The `# type: ignore` comment makes the intent explicit.

src/ha_mcp/transforms/categorized_search.py
kingpanther13: Not changing — this is intentional security behavior. If the LLM double-wraps a write tool inside `ha_call_read_tool`, rejecting it and pointing to `ha_call_write_tool` is the correct response. Re-evaluating the category after unwrapping would allow bypassing permission gating — a read proxy should only execute read tools regardless of how the arguments got wrapped. The error message at line 296 is actionable: it tells the LLM exactly which proxy to use instead.

kingpanther13: To clarify the security guarantee: there is no path where an improper tool call succeeds. The `allowed` set is bound to the proxy's category *before* unwrapping and is never reassigned after. So if a write tool gets double-wrapped inside `ha_call_read_tool`, after unwrapping the inner tool name is still checked against `_read_tools` — it will always be rejected with an actionable error pointing to `ha_call_write_tool`. The reviewer's concern that 'the unwrapped call will succeed when it should be rejected' does not apply to this code.

tests/src/unit/test_categorized_search.py
kingpanther13: Fixed: added comprehensive tests covering all the gaps you identified. New test classes:

**`TestCategorizedCallDispatch`** — proxy execution:
- Happy path for read/write/delete proxies
- Wrong-category rejection (write via read, read via write, delete via read) — verifies error code and correct proxy suggestion in message
- Unknown tool returns `RESOURCE_NOT_FOUND`
- `None` arguments handled correctly

**`TestDoubleUnwrap`** — double-wrapped call detection:
- Correct same-category unwrap succeeds
- Wrong-category after unwrap still rejected (security guarantee: `allowed` set is never reassigned)
- Non-proxy names with `name`/`arguments` keys are NOT unwrapped

**`TestRebuildCategoryCache`** — cache lifecycle:
- Populates all three sets correctly from catalog
- Rebuilds when catalog hash changes (new tool added)
- No-op when hash unchanged

**`TestSearchKeywordsTransform`**:
- Keywords appended to description
- Overrides replace description entirely
- Override takes priority when both exist
- No-match leaves description unchanged
- `get_tool` also applies enrichment
- Returns `None` for missing tools

PR #761, opened 2026-03-13
fix: clarify ha_search_entities vs ha_deep_search descriptions to prevent tool misuse

sergeykad:
## What does this PR do?

LLM agents were defaulting to `ha_deep_search` for simple entity lookups because its description sounded more powerful/comprehensive. This updates both docstrings to make the intended use case explicit and redirect agents to the right tool.

**`ha_search_entities`** — now leads with:
> PRIMARY tool for finding entities (lights, sensors, switches, etc.) by name, area, or domain. Use this first when looking up any entity ID.

**`ha_deep_search`** — now opens with:
> Search inside automation, script, and helper *configurations* — not for finding entity IDs.

Also adds UAT story `s13` for the "find automations that reference a specific entity" workflow, which exercises `ha_deep_search` and was missing from the catalog.

## Type of change
- [x] 🐛 Bug fix

## Testing

Verified with two UAT runs against a Docker HA container:
- "Find the kitchen light entity_id" → agent called `ha_search_entities` only ✓
- "Which automations reference light.bed_light?" → agent called `ha_deep_search` only ✓

## Checklist

### DIFF
--- a/tests/uat/stories/catalog/s13_find_automations_by_entity.yaml
+++ b/tests/uat/stories/catalog/s13_find_automations_by_entity.yaml
@@ -38,6 +38,10 @@ prompt: >
 teardown: []
 
 verify:
+  questions:
+    - "Which automation references light.bed_light?"
+    - "What does the automation do with light.bed_light?"
+    - "Is this an automation, script, or both?"
   ha_checks:
     - type: automation_exists
       alias: "Bed Light Evening"


kingpanther13:
@julienld I have the same issue, except I don't have the issue when I use categorized search tools (#727) because I managed to change the weighting of it with the bm25 searches. However I agree that combining the tools would be a good idea and having deep search as a parameter, that would be less confusing. I think the llm just sees the word "search" and assumes it's the only search tool.

PR #777, opened 2026-03-17
Feature: Show friendly landing page when browsers visit MCP endpoint directly

kingpanther13:
## What does this PR do?

We have had multiple issues pop up with people unable to connect to their MCP server for a variety of reasons. I used to be able to tell people to connect directly to their url via a browser and that if they saw the line of json then their URL was correct, but FastMCP recently changed it so now it just says "method not allowed." Instead of showing that message, I feel that we can put our own message there telling people that this is the correct URL, they just have to put it in the LLM. It also contains a hint to disable the "ai bot prevention" feature on Cloudflare...something that I've had at least two people have issues with.

This registers a custom GET route (via FastMCP's `custom_route()` API) at the MCP endpoint path that returns a plain-text message confirming the server is up and providing setup guidance. POST/DELETE requests from LLMs are completely unaffected — Starlette routing matches the MCP handler first for those methods.

Covers all HTTP modes: web, SSE, and OAuth.

## Type of change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 💥 Breaking change

## Testing

## Checklist

### DIFF
--- a/site/package-lock.json
+++ b/site/package-lock.json
@@ -2776,9 +2776,9 @@
       }
     },
     "node_modules/h3": {
-      "version": "1.15.8",
-      "resolved": "https://registry.npmjs.org/h3/-/h3-1.15.8.tgz",
-      "integrity": "sha512-iOH6Vl8mGd9nNfu9C0IZ+GuOAfJHcyf3VriQxWaSWIB76Fg4BnFuk4cxBxjmQSSxJS664+pgjP6e7VBnUzFfcg==",
+      "version": "1.15.9",
+      "resolved": "https://registry.npmjs.org/h3/-/h3-1.15.9.tgz",
+      "integrity": "sha512-H7UPnyIupUOYUQu7f2x7ABVeMyF/IbJjqn20WSXpMdnQB260luADUkSgJU7QTWLutq8h3tUayMQ1DdbSYX5LkA==",
       "license": "MIT",
       "dependencies": {
         "cookie-es": "^1.2.2",
--- a/src/ha_mcp/tools/tools_config_helpers.py
+++ b/src/ha_mcp/tools/tools_config_helpers.py
@@ -25,6 +25,48 @@ from .util_helpers import (
 logger = logging.getLogger(__name__)
 
 
+def _format_schedule_days(
+    monday: list | None,
+    tuesday: list | None,
+    wednesday: list | None,
+    thursday: list | None,
+    friday: list | None,
+    saturday: list | None,
+    sunday: list | None,
+) -> dict[str, list[dict[str, Any]]]:
+    """Format schedule day data, ensuring time strings include seconds.
+
+    Returns a dict of day_name -> formatted time ranges, only for days
+    where data was provided (not None).
+    """
+    day_params = {
+        "monday": monday,
+        "tuesday": tuesday,
+        "wednesday": wednesday,
+        "thursday": thursday,
+        "friday": friday,
+        "saturday": saturday,
+        "sunday": sunday,
+    }
+    formatted_days: dict[str, list[dict[str, Any]]] = {}
+    for day_name, day_schedule in day_params.items():
+        if day_schedule is not None:
+            formatted_ranges = []
+            for time_range in day_schedule:
+                formatted_range: dict[str, Any] = {}
+                for key in ["from", "to"]:
+                    if key in time_range:
+                        time_val = time_range[key]
+                        if isinstance(time_val, str) and time_val.count(":") == 1:
+                            time_val = f"{time_val}:00"
+                        formatted_range[key] = time_val
+                if "data" in time_range:
+                    formatted_range["data"] = time_range["data"]
+                formatted_ranges.append(formatted_range)
+            formatted_days[day_name] = formatted_ranges
+    return formatted_days
+
+
 def register_config_helper_tools(mcp: Any, client: Any, **kwargs: Any) -> None:
     """Register Home Assistant helper configuration tools."""
 
@@ -552,34 +594,10 @@ def register_config_helper_tools(mcp: Any, client: Any, **kwargs: Any) -> None:
                     # Schedule parameters: monday-sunday with time ranges
                     # Each day is a list of {"from": "HH:MM:SS", "to": "HH:MM:SS"}
                     # with optional "data" dict for additional attributes
-                    day_params = {
-                        "monday": monday,
-                        "tuesday": tuesday,
-                        "wednesday": wednesday,
-                        "thursday": thursday,
-                        "friday": friday,
-                        "saturday": saturday,
-                        "sunday": sunday,
-                    }
-                    for day_name, day_schedule in day_params.items():
-                        if day_schedule is not None:
-                            # Ensure time format has seconds
-                            formatted_ranges = []
-                            for time_range in day_schedule:
-                                formatted_range = {}
-                                for key in ["from", "to"]:
-                                    if key in time_range:
-                                        time_val = time_range[key]
-                                        # Add seconds if not present
-                                        if time_val.count(":") == 1:
-                                            time_val = f"{time_val}:00"
-                                        formatted_range[key] = time_val
-                                # Pass through the optional 'data' dict
-                                # for additional attributes (e.g. mode, brightness)
-                                if "data" in time_range:
-                                    formatted_range["data"] = time_range["data"]
-                                formatted_ranges.append(formatted_range)
-                            message[day_name] = formatted_ranges
+                    message.update(_format_schedule_days(
+                        monday, tuesday, wednesday, thursday,
+                        friday, saturday, sunday,
+                    ))
 
                 elif helper_type == "zone":
                     # Zone parameters - HA validates required fields (latitude, longitude)
@@ -677,7 +695,7 @@ def register_config_helper_tools(mcp: Any, client: Any, **kwargs: Any) -> None:
                 # Person and zone have entity registry entries with unique_id
                 # used as the config store identifier. Tags use their own tag
                 # registry and don't have entity registry entries.
-                config_store_types = {"person", "zone"}
+                config_store_types = {"person", "zone", "schedule"}
 
                 updated_data: dict[str, Any] = {}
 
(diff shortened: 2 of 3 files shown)


src/ha_mcp/__main__.py
```diff
@@ -567,6 +567,24 @@ async def _run_http_with_graceful_shutdown(
...
+    @mcp.custom_route(path, methods=["GET"])
+    async def _browser_landing(request):  # noqa: ARG001
+        return PlainTextResponse(
+            "HA-MCP server is up and running. To connect, please follow the "
+            "setup instructions (https://github.com/homeassistant-ai/ha-mcp#-installation), "
+            "and paste the URL for this page into your LLM. If using Cloudflare "
+            "and you're unable to connect via your LLM, make sure the "
+            '"Block AI training bots" setting is set to '
+            '"do not block (allow crawlers)".'
+        )
```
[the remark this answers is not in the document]
kingpanther13: Added unit test in `tests/src/unit/test_browser_landing.py` — creates a FastMCP instance, registers the GET route, and verifies 200 + expected text via `httpx.ASGITransport`.

src/ha_mcp/__main__.py
```diff
@@ -567,6 +567,24 @@ async def _run_http_with_graceful_shutdown(
...
 
 
+def _register_browser_landing(path: str) -> None:
+    """Register a GET handler so browsers see a friendly page instead of 405."""
+    from starlette.responses import PlainTextResponse
+
+    mcp = _get_mcp()
+
+    @mcp.custom_route(path, methods=["GET"])
+    async def _browser_landing(request):  # noqa: ARG001
```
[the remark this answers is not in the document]
kingpanther13: Done — added `Request` type hint, `PlainTextResponse` return type, renamed param to `_`, removed unused noqa directive.

src/ha_mcp/__main__.py
```diff
@@ -567,6 +567,35 @@ async def _run_http_with_graceful_shutdown(
...
+        return PlainTextResponse(
+            "HA-MCP server is up and running. To connect, please follow the "
+            "setup instructions (https://homeassistant-ai.github.io/ha-mcp/), "
+            "and paste the URL for this page into your LLM. If using Cloudflare "
+            "and you're unable to connect via your LLM, make sure the "
+            '"Block AI training bots" setting is set to '
+            '"do not block (allow crawlers)".',
+            status_code=405,
+            headers={"Allow": "POST, DELETE"},
+        )
```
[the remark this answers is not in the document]
kingpanther13: It's not that much text, I don't see any real reason to change this

src/ha_mcp/__main__.py
```diff
@@ -567,6 +567,35 @@ async def _run_http_with_graceful_shutdown(
     )
 
 
+def register_browser_landing(mcp_instance: "FastMCP", path: str) -> None:
```
sergeykad, line 576: `custom_route` unconditionally appends to `_additional_http_routes` (a plain list). Starlette iterates routes in order with no duplicate detection — calling this function twice on the same path silently adds a dead second route. No log message, no exception.

Add a guard before registering:
```python
existing = [getattr(r, 'path', None) for r in mcp_instance._additional_http_routes]
if path in existing:
    logger.warning('register_browser_landing: %r already registered, skipping', path)
    return
```

src/ha_mcp/__main__.py
```diff
@@ -567,6 +567,35 @@ async def _run_http_with_graceful_shutdown(
     )
 
 
+def register_browser_landing(mcp_instance: "FastMCP", path: str) -> None:
+    """Register a GET handler that returns 405 with a helpful message.
+
+    Browsers and misconfigured clients that send GET instead of POST will see
+    a human-readable explanation instead of a bare "Method Not Allowed" error.
+    The 405 status and Allow header are preserved so automated clients still
```
sergeykad: "The 405 status and Allow header are **preserved**" is misleading — "preserved" implies these values originate upstream and are being passed through. They are constructed and returned fresh by this handler. Change to "are set explicitly by this handler" or "are returned by this handler".

src/ha_mcp/__main__.py
```diff
@@ -567,6 +567,35 @@ async def _run_http_with_graceful_shutdown(
...
+    Browsers and misconfigured clients that send GET instead of POST will see
+    a human-readable explanation instead of a bare "Method Not Allowed" error.
+    The 405 status and Allow header are preserved so automated clients still
+    get the correct HTTP semantics.
+
+    Args:
+        mcp_instance: The FastMCP server to register the route on.
+        path: The MCP endpoint path (e.g. "/mcp" or a secret path).
+    """
+    from starlette.requests import Request
```
sergeykad: Deferred imports inside the function body mean `ImportError` surfaces at call time with no actionable context. In `start.py` this gets caught by the `BaseException` handler and logged as `"MCP server crashed: No module named 'starlette'"` — no indication of which feature failed or what to install.

fastmcp already imports starlette at module level (`transport.py` line 16-17), so there is no startup-cost reason to defer. Move these to module scope, or if keeping them deferred, wrap with a `RuntimeError` that identifies the dependency:
```python
try:
    from starlette.requests import Request
    from starlette.responses import PlainTextResponse
except ImportError as e:
    raise RuntimeError(
        'register_browser_landing requires starlette (a fastmcp dependency)'
    ) from e
```

src/ha_mcp/__main__.py
```diff
@@ -567,6 +567,35 @@ async def _run_http_with_graceful_shutdown(
...
+    from starlette.responses import PlainTextResponse
+
+    @mcp_instance.custom_route(path, methods=["GET"])
```
sergeykad, line 596: This works because FastMCP registers custom routes with lowest precedence (after the MCP transport route), and the MCP transport only claims `POST` and `DELETE` — so `GET` falls through to this handler. But nothing documents this invariant.

If the MCP transport ever adds `GET` support, this landing page silently shadows legitimate MCP traffic with no error. Add a comment:
```python
# Safe because the MCP streamable-http transport claims only POST and DELETE.
# FastMCP registers custom routes at lowest precedence (after the MCP route),
# so GET requests fall through here without intercepting MCP traffic.
@mcp_instance.custom_route(path, methods=["GET"])
```

src/ha_mcp/__main__.py
```diff
@@ -567,6 +567,35 @@ async def _run_http_with_graceful_shutdown(
...
+    async def _browser_landing(_: Request) -> PlainTextResponse:
+        return PlainTextResponse(
+            "HA-MCP server is up and running. To connect, please follow the "
+            "setup instructions (https://homeassistant-ai.github.io/ha-mcp/), "
+            "and paste the URL for this page into your LLM. If using Cloudflare "
+            "and you're unable to connect via your LLM, make sure the "
+            '"Block AI training bots" setting is set to '
+            '"do not block (allow crawlers)".',
+            status_code=405,
+            headers={"Allow": "POST, DELETE"},
```
sergeykad, line 608: This codebase uses `stateless_http=True` everywhere, which means the MCP transport does not handle `DELETE` (session termination only exists in stateful mode). A maintainer seeing `DELETE` here has no way to tell if it belongs or is a copy-paste from the MCP spec.

Add a clarifying comment:
```python
# DELETE is included per the MCP Streamable HTTP spec (used for session
# termination), even though this deployment uses stateless mode.
headers={"Allow": "POST, DELETE"},
```

homeassistant-addon/start.py
```diff
@@ -147,8 +147,13 @@ def main() -> int:
...
     try:
         log_info("Importing ha_mcp module...")
-        from ha_mcp.__main__ import _get_timestamped_uvicorn_log_config, mcp
+        from ha_mcp.__main__ import (
+            _get_timestamped_uvicorn_log_config,
+            mcp,
+            register_browser_landing,
+        )
 
+        register_browser_landing(mcp, secret_path)
```
sergeykad: `register_browser_landing` is called inside the `try` block that catches `BaseException` and logs `"MCP server crashed: {e}"`. If route registration raises — e.g. because FastMCP adds duplicate detection in a future version — the message implies the server started and then crashed, when it never started.

Move this call outside the try block (or tighten the try to wrap only `mcp.run()`) so registration failures are clearly distinguished from runtime crashes:
```python
register_browser_landing(mcp, secret_path)
log_info("Browser landing route registered")
try:
    log_info("Starting MCP server...")
    mcp.run(...)
```

tests/src/unit/test_browser_landing.py
```diff
@@ -0,0 +1,46 @@
...
+    ) as client:
+        resp = await client.post(
+            "/mcp",
+            json={"jsonrpc": "2.0", "method": "initialize", "id": 1, "params": {}},
+            headers={"Content-Type": "application/json"},
+        )
+
+    # The MCP handler errors (no lifespan in test), but the key assertion is
+    # that POST was NOT intercepted by the landing page route.
+    assert "HA-MCP server is up and running" not in resp.text
```
sergeykad, line 67: `assert "HA-MCP server is up and running" not in resp.text` is a negative string check — it passes for any response that doesn't contain this string, including a 405 from a different handler. The primary assertion should be that POST was **not** swallowed by the landing page:
```python
assert resp.status_code != 405  # POST must not be intercepted by the landing route
assert "HA-MCP server is up and running" not in resp.text
```

Two additional cases worth covering:

**Custom path** — the production use always involves a `MCP_SECRET_PATH`. No test verifies that `register_browser_landing(server, "/secret-abc")` mounts at `/secret-abc` and leaves `/mcp` returning 404 (not the landing page).

**HEAD method** — Starlette automatically extends GET routes to handle HEAD. No test confirms HEAD on `/mcp` returns 405 with the correct `Allow` header.

src/ha_mcp/__main__.py
kingpanther13: Fixed. Added a module-level `_registered_landing_paths` set that guards against duplicate registration. Used this instead of accessing `mcp_instance._additional_http_routes` to avoid coupling to FastMCP's private internals.

src/ha_mcp/__main__.py
kingpanther13: Fixed. Changed "preserved" to "set explicitly by this handler".

src/ha_mcp/__main__.py
kingpanther13: Fixed. Moved both imports to module scope — starlette is a required fastmcp dependency so there's no startup-cost reason to defer. Skipped the `try/except ImportError` wrapper since if starlette is missing, fastmcp itself won't import and the server won't start at all.

src/ha_mcp/__main__.py
kingpanther13: Fixed. Added the comment documenting the routing invariant.

src/ha_mcp/__main__.py
kingpanther13: Fixed. Added inline comment explaining DELETE is per MCP Streamable HTTP spec even though stateless mode doesn't use it.

homeassistant-addon/start.py
kingpanther13: Fixed. Moved both the import and `register_browser_landing()` call above the try block so registration failures are clearly distinguished from runtime crashes.

tests/src/unit/test_browser_landing.py
kingpanther13: Fixed. All three additions:

1. Added `assert resp.status_code != 405` before the negative string check on POST
2. Added `test_custom_path_mounts_at_correct_path` — registers at `/secret-abc`, verifies it serves there and `/mcp` returns 404
3. Added `test_head_returns_405_with_allow_header` — confirms Starlette's auto-HEAD returns 405 with correct Allow header

PR #789, opened 2026-03-20
Fix schedule helper update ignoring day data and 'data' field

airlabno:
## Summary

Fixes schedule helper updates silently ignoring day data (monday-sunday time blocks) and their `data` attributes.

**The bug:** When updating a schedule via `ha_config_set_helper`, the update path fell through to the generic `config/entity_registry/update` branch, which only handles metadata (name, icon, area, labels). The schedule day configuration was completely ignored — the API returned success but nothing changed.

**The fix:**
- Add `"schedule"` to `config_store_types` so it routes through the config store update path
- Implement `schedule/update` websocket message handler that mirrors the create path, including day schedule formatting and `data` field passthrough

This is a companion to PR #578 which added `data` field support for schedule **create** and **list** — this PR completes the CRUD cycle by fixing **update**.

## Test plan

- [x] Added E2E test `test_schedule_update_with_data_field` that creates a schedule with data, updates data values, and verifies persistence via list
- [ ] Existing `test_schedule_with_data_field` (create) and `test_schedule_full_lifecycle` should continue to pass
- [ ] Manual verification: update a schedule's time block mode and confirm it persists

### DIFF
--- a/src/ha_mcp/tools/tools_config_helpers.py
+++ b/src/ha_mcp/tools/tools_config_helpers.py
@@ -25,6 +25,48 @@ from .util_helpers import (
 logger = logging.getLogger(__name__)
 
 
+def _format_schedule_days(
+    monday: list | None,
+    tuesday: list | None,
+    wednesday: list | None,
+    thursday: list | None,
+    friday: list | None,
+    saturday: list | None,
+    sunday: list | None,
+) -> dict[str, list[dict[str, Any]]]:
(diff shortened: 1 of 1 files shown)


airlabno:
Addressed the review feedback — extracted the duplicated day-formatting logic into a shared `_format_schedule_days()` module-level helper function, used by both the create and update paths. See 4bbe215.

PR #368, opened 2025-12-26
feat: OAuth 2.1 Authentication with DCR and Consent Form

julienld:
Relates to #245

## Summary

Implements OAuth 2.1 authentication with Dynamic Client Registration (DCR) and consent form for ha-mcp server, enabling Claude.ai and other MCP clients to connect without pre-shared credentials.

## Key Features

✅ **OAuth 2.1 with PKCE** - Public client authentication without client secrets
✅ **Dynamic Client Registration (RFC 7591)** - Clients self-register with the server
✅ **Custom Consent Form** - Users authorize with their Home Assistant URL and Long-Lived Access Token
✅ **Stateless Encrypted Tokens** - Credentials encrypted in token with Fernet, survive server restarts
✅ **Full Toolset** - All 97 tools exposed via OAuth with dynamic credential injection
✅ **Production Ready** - 24/24 unit tests passing, clean code

### Stateless Token Design
- Access tokens contain Fernet-encrypted Home Assistant credentials (URL + token)
- No server-side session storage required
- Tokens persist across server restarts when `OAUTH_ENCRYPTION_KEY` is set
- Credentials decrypted on-demand from token string

### OAuthProxyClient Pattern
- Intercepts tool calls via `__getattr__`
- Extracts credentials from OAuth token claims
- Dynamically creates `HomeAssistantClient` instances per request
- Enables full toolset without hardcoded credentials

### Consent Flow
1. Client initiates OAuth with authorization request
2. Server redirects to custom consent form (`/consent`)
3. User enters HA URL and Long-Lived Access Token
4. Server validates credentials against HA `/api/config`
5. Issues encrypted stateless access token + refresh token

### Environment Variables
```bash
# Required for OAuth mode
OAUTH_ENCRYPTION_KEY=<base64-key>  # For token persistence across restarts

# Optional - auto-generates if not set (tokens invalid after restart)
```

### Running OAuth Server
```bash
# Generate encryption key (do once, save securely)
python -c "from cryptography.fernet import Fernet; print(Fernet.generate_key().decode())"

# Start server
export OAUTH_ENCRYPTION_KEY="your-key-here"
ha-mcp-web --oauth-base-url https://your-tunnel.trycloudflare.com
```

### New Files
- `src/ha_mcp/auth/provider.py` - OAuth provider with encrypted tokens
- `src/ha_mcp/auth/consent_form.py` - Custom consent HTML
- `tests/src/unit/test_oauth.py` - OAuth unit tests (24 tests)

### Modified Files
[...]

PR #790, opened 2026-03-20
fix: OAuth token refresh broken and state lost on container restart

### LINKED ISSUE #787: [BUG] OAuth session does not persist in Docker OAuth mode and requires re-entering Home Assistant long-lived token repeatedly
### 💬 What Happened?

Currently on 7.0.0

I set up ha-mcp in OAuth mode using the Docker container and exposed it through Cloudflare. The initial authentication flow works correctly. Claude registers the client, I get the consent screen, I enter my Home Assistant long-lived access token, and the MCP server connects successfully. After that, Claude can list tools and use the server normally.

The problem is that the authentication does not stay valid. After some time, usually by the next day, the MCP server is disconnected again and Claude makes me go back through the full consent flow and re-enter the Home Assistant long-lived token. This keeps happening repeatedly, so the OAuth session does not seem to persist across time/restarts the way I expected.

I originally thought it might be a Docker volume issue, so I mounted /home/mcpuser/.ha-mcp to a persistent host folder and confirmed that the container can read and write there. I can see a logs folder being written into that mounted directory, so persistence is definitely working at the filesystem level. Even with that fixed, the OAuth authentication still drops and requires full re-authentication.

From the logs, it looks like Claude first tries POST /mcp and gets 401 invalid_token, then it goes back through OAuth discovery, registration, consent, token exchange, and reconnects successfully. So the server is reachable and auth works, but the issued auth state is not staying valid. My guess is that either the OAuth state is not actually being persisted, or Claude-compatible OAuth state is being lost/invalidated between sessions even though the mounted directory is writable.

Here is my current docker compose file. Note that I did need to add in the "HOMEASSISTANT_TOKEN" variable otherwise I was getting immediate errors in the docker container logs.

Compose file:

```
services:
  ha-mcp-oauth:
    image: ghcr.io/homeassistant-ai/ha-mcp:latest
    container_name: ha-mcp-oauth
    command: ["ha-mcp-oauth"]
    ports:
      - "8086:8086"
    environment:
      HOMEASSISTANT_URL: "http://192.168.1.60:8123"
      MCP_BASE_URL: "https://ha-mcp.my.url"
      HOMEASSISTANT_TOKEN: "MY_TOKEN"
    volumes:
      - /volume2/dockerssd/ha-mcp:/home/mcpuser/.ha-mcp
    restart: unless-stopped
```

Error shown in logs without "HOMEASSISTANT_TOKEN":

HOMEASSISTANT_TOKEN
  Value error, Home Assistant token must be provided [type=value_error, input_value='', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/value_error

```markdown
None
```

kingpanther13:
## What does this PR do?

Fixes #787 — OAuth token refresh is broken (produces invalid access tokens) and all OAuth state is lost on container restart, forcing users to re-authenticate repeatedly.

**Root cause:** Three bugs in `exchange_refresh_token`, all present since the original OAuth implementation (PR #368):

1. **Token refresh generates invalid tokens** — `exchange_refresh_token` created random `ha_access_*` strings instead of stateless base64-encoded tokens. `load_access_token` can't decode them → 401 Unauthorized after the first hour.

2. **Credential mapping is wrong** — `_refresh_to_access_map` stored `client_id` instead of the actual access token string, so HA credentials could never be recovered during refresh.

3. **No disk persistence** — All OAuth state (clients, refresh tokens, mappings) was stored in in-memory Python dicts. Container restart = complete state loss.

**Fix:**

- `exchange_refresh_token` now decodes the HA token from the old stateless access token via `_decode_credentials` and encodes a new one via `_encode_credentials`
- `exchange_authorization_code` correctly maps refresh_token → access_token (not client_id)
- Added disk persistence via `~/.ha-mcp/oauth_state.json` (atomic writes via tmp+rename, loaded on init)
- State file uses restrictive permissions (0o600 file, 0o700 directory)
- Orphaned `.tmp` files cleaned up on save failure
- Expired refresh tokens pruned during load
- Removed dead `_access_to_refresh_map` (never populated in stateless flow)
- Added `state_dir` constructor parameter for configurability

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 💥 Breaking change

## Testing
- [x] Docker integration test: full OAuth flow via curl (register → authorize → consent → exchange → refresh → chained refresh → restart → refresh after restart)

## Checklist

### DIFF
--- a/src/ha_mcp/auth/provider.py
+++ b/src/ha_mcp/auth/provider.py
@@ -242,7 +242,7 @@ class HomeAssistantOAuthProvider(OAuthProvider):
                 try:
                     os.close(tmp_fd)
                 except OSError:
-                    pass
+                    pass  # Best-effort cleanup; fd may already be closed
             if tmp_path is not None:
                 try:
                     tmp_path.unlink()
--- a/tests/src/unit/test_oauth.py
+++ b/tests/src/unit/test_oauth.py
@@ -895,7 +895,6 @@ class TestOAuthStatePersistence:
     @pytest.mark.asyncio
     async def test_expired_tokens_pruned_on_load(self, tmp_path):
         """Test that expired refresh tokens are not loaded from disk."""
-        import json
 
         state = {
             "clients": {},


src/ha_mcp/auth/provider.py
```diff
@@ -177,6 +185,71 @@ def _decode_credentials(self, token: str) -> str | None:
...
+            os.chmod(tmp_file, 0o600)
+            tmp_file.rename(self._state_file)
+            logger.debug(f"OAuth state saved to {self._state_file}")
+        except Exception as e:
+            # Clean up tmp file if it was written but not renamed
+            try:
+                tmp_file = self._state_file.with_suffix('.tmp')
+                if tmp_file.exists():
+                    tmp_file.unlink()
+            except OSError:
```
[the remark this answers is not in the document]
kingpanther13: Fixed in eed8d8d — added `logger.debug(f"Failed to clean up tmp file: {cleanup_err}")` to the inner `except OSError` block.

src/ha_mcp/auth/provider.py
```diff
@@ -177,6 +185,71 @@ def _decode_credentials(self, token: str) -> str | None:
...
+            }
+
+            self._state_file.parent.mkdir(parents=True, exist_ok=True)
+            os.chmod(self._state_file.parent, 0o700)
+            tmp_file = self._state_file.with_suffix('.tmp')
+            tmp_file.write_text(json.dumps(state, indent=2))
+            os.chmod(tmp_file, 0o600)
+            tmp_file.rename(self._state_file)
+            logger.debug(f"OAuth state saved to {self._state_file}")
+        except Exception as e:
```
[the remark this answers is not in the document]
kingpanther13: Fixed in eed8d8d — narrowed to `except (OSError, TypeError)`. Dropped `IOError` since it's an alias for `OSError` in Python 3.

src/ha_mcp/auth/provider.py
```diff
@@ -177,6 +185,71 @@ def _decode_credentials(self, token: str) -> str | None:
...
+                    expires_at=rt_data.get("expires_at"),
+                )
+
+            self._refresh_to_access_map = state.get("refresh_to_access_map", {})
+
+            logger.info(
+                f"OAuth state loaded: {len(self.clients)} clients, "
+                f"{len(self.refresh_tokens)} refresh tokens"
+            )
+        except Exception as e:
```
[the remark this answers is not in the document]
kingpanther13: Fixed in eed8d8d — narrowed to `except (OSError, json.JSONDecodeError, KeyError, TypeError)`. Dropped `IOError` since it's an alias for `OSError` in Python 3.

src/ha_mcp/auth/provider.py
```diff
@@ -98,6 +101,7 @@ def __init__(
             client_registration_options: Options for client registration
             revocation_options: Options for token revocation
             required_scopes: Scopes required for all requests
+            state_dir: Directory for persisting OAuth state (default: ~/.ha-mcp)
```
sergeykad: The docstring says "default: `~/.ha-mcp`" but the actual default is the file `~/.ha-mcp/oauth_state.json`. Change to: "Directory for persisting OAuth state; writes `oauth_state.json` inside it. Defaults to `~/.ha-mcp`."

More importantly: the state file will contain `_refresh_to_access_map` values that are base64-encoded HA long-lived access tokens — trivially decodable. The 0o600 file permission is the only protection. This is an appropriate tradeoff for a local Docker deployment, but it should be stated explicitly here so operators understand what they are protecting:

> Note: the state file contains HA access tokens in base64-encoded form. The 0o600 permissions are the security boundary. Treat this file like a credential file.

src/ha_mcp/auth/provider.py
```diff
@@ -177,6 +184,79 @@ def _decode_credentials(self, token: str) -> str | None:
             logger.debug(f"Failed to decode token: {e}")
             return None
 
+    def _save_state(self) -> None:
```
sergeykad, line 193: Two things:

**1. `ValueError` from `json.dumps` is not caught.**
`json.dumps` raises `ValueError` for circular references or encoder failures. The outer `except (OSError, TypeError)` misses it — the exception escapes, the cleanup block is skipped, and the `.tmp` file is left on disk permanently. Add `ValueError` to the exception tuple.

**2. `_save_state` docstring should note that `ha_credentials` is intentionally excluded.**
A maintainer adding a new credential field may not realise it also needs explicit handling. Add: "Note: `ha_credentials` is not persisted — HA tokens are recoverable from the stateless access tokens stored in `_refresh_to_access_map`."

src/ha_mcp/auth/provider.py
```diff
@@ -177,6 +184,79 @@ def _decode_credentials(self, token: str) -> str | None:
...
+            logger.debug(f"OAuth state saved to {self._state_file}")
+        except (OSError, TypeError) as e:
+            # Clean up tmp file if it was written but not renamed
+            try:
+                tmp_file = self._state_file.with_suffix('.tmp')
+                if tmp_file.exists():
+                    tmp_file.unlink()
+            except OSError as cleanup_err:
+                logger.debug(f"Failed to clean up tmp file: {cleanup_err}")
+            logger.warning(f"Failed to save OAuth state: {e}")
```
sergeykad: A failed save means every active session is lost on the next restart. `WARNING` conveys the same urgency as a deprecation notice. Use `logger.error` and include the state file path and the consequence:
```python
logger.error(
    "Failed to save OAuth state to %s — active sessions will not survive restart: %s",
    self._state_file, e,
)
```

src/ha_mcp/auth/provider.py
```diff
@@ -177,6 +184,79 @@ def _decode_credentials(self, token: str) -> str | None:
...
+                        "scopes": rt.scopes,
+                        "expires_at": rt.expires_at,
+                    }
+                    for tid, rt in self.refresh_tokens.items()
+                },
+                "refresh_to_access_map": dict(self._refresh_to_access_map),
+            }
+
+            self._state_file.parent.mkdir(parents=True, exist_ok=True)
+            os.chmod(self._state_file.parent, 0o700)
```
sergeykad: `os.chmod(self._state_file.parent, 0o700)` runs on every save. If `state_dir` points to a shared directory this modifies its permissions on every token operation. Only set permissions when the directory is first created:
```python
created = not self._state_file.parent.exists()
self._state_file.parent.mkdir(parents=True, exist_ok=True)
if created:
    os.chmod(self._state_file.parent, 0o700)
```

src/ha_mcp/auth/provider.py
```diff
@@ -177,6 +184,79 @@ def _decode_credentials(self, token: str) -> str | None:
...
+            tmp_file = self._state_file.with_suffix('.tmp')
```
sergeykad: The `.tmp` filename is fixed (`oauth_state.tmp`). If two processes start simultaneously (old container still shutting down, new container starting), both write to the same path. `Path.rename()` is atomic on Linux for same-filesystem moves, so the last writer wins — but the first writer's data is silently discarded with no log.

Use `tempfile.mkstemp(dir=self._state_file.parent, suffix='.tmp')` to get a unique name per process, then rename to the final path. Also worth a comment that source and destination must be on the same filesystem for the rename to be atomic.

src/ha_mcp/auth/provider.py
```diff
@@ -177,6 +184,79 @@ def _decode_credentials(self, token: str) -> str | None:
...
+                tmp_file = self._state_file.with_suffix('.tmp')
+                if tmp_file.exists():
+                    tmp_file.unlink()
+            except OSError as cleanup_err:
+                logger.debug(f"Failed to clean up tmp file: {cleanup_err}")
+            logger.warning(f"Failed to save OAuth state: {e}")
+
+    def _load_state(self) -> None:
+        """Load persisted OAuth state from disk."""
+        if not self._state_file.exists():
```
sergeykad: TOCTOU: if the file is deleted between `exists()` returning `True` and `read_text()`, the `FileNotFoundError` is caught and logged as an error — a normal first-start condition appearing as a failure in the logs.

Remove the guard and handle `ENOENT` explicitly:
```python
import errno
try:
    text = self._state_file.read_text()
except OSError as e:
    if e.errno == errno.ENOENT:
        return  # Normal first-start, no prior state
    logger.error("Cannot read OAuth state file %s: %s", self._state_file, e)
    return
```

src/ha_mcp/auth/provider.py
```diff
@@ -177,6 +184,79 @@ def _decode_credentials(self, token: str) -> str | None:
...
+            return
+
+        try:
+            state = json.loads(self._state_file.read_text())
+            now = time.time()
+
+            for cid, client_data in state.get("clients", {}).items():
+                self.clients[cid] = OAuthClientInformationFull.model_validate(client_data)
```
sergeykad: If construction of the 8th client out of 10 raises (e.g. schema mismatch), the first 7 are already written into `self.clients` while their paired refresh tokens may already be in `self.refresh_tokens`. The exception is caught and logged, but the provider continues with inconsistent in-memory state — clients with orphaned refresh tokens, or vice versa.

Populate into local dicts first, then assign atomically only if everything succeeds:
```python
new_clients = {cid: OAuthClientInformationFull.model_validate(d) for cid, d in state.get('clients', {}).items()}
new_tokens = {tid: RefreshToken(**d) for tid, d in ...}  # with expiry check
new_map = {k: v for k, v in stored_map.items() if k in new_tokens}
# Commit atomically:
self.clients, self.refresh_tokens, self._refresh_to_access_map = new_clients, new_tokens, new_map
```

src/ha_mcp/auth/provider.py
```diff
@@ -691,31 +773,19 @@ def _revoke_internal(
...
         if access_token_str:
-            if access_token_str in self.access_tokens:
-                del self.access_tokens[access_token_str]
-
-            associated_refresh = self._access_to_refresh_map.pop(access_token_str, None)
-            if associated_refresh:
-                if associated_refresh in self.refresh_tokens:
-                    del self.refresh_tokens[associated_refresh]
-                self._refresh_to_access_map.pop(associated_refresh, None)
+            self.access_tokens.pop(access_token_str, None)
```
sergeykad, line 829: The old `access_token_str` path cascaded to also revoke the associated refresh token via `_access_to_refresh_map`. That map is removed in this PR, so the cascade is gone.

Revoking an access token no longer invalidates the session — the paired refresh token remains valid and can generate new access tokens indefinitely (RFC 7009 §2.1 recommends cascading revocation).

Note: in the old code the cascade was actually broken for the stateless flow (the map was never populated by `exchange_authorization_code`), so this is removing dead code rather than working functionality. But that should be stated explicitly — without a comment, every future reviewer will question whether this is a security gap:
```python
# Stateless access tokens are not stored in self.access_tokens, so there is
# no server-side cascade to revoke. The paired refresh token is only revoked
# when revoke_token() is called with a RefreshToken argument directly.
self.access_tokens.pop(access_token_str, None)
```

tests/src/unit/test_oauth.py
```diff
@@ -765,6 +775,221 @@ async def test_complete_oauth_flow(self, provider):
...
+            expires_at=int(time.time() + 86400),
+        )
+        provider.refresh_tokens["corrupt_refresh"] = refresh_token
+        # Map to a non-decodable string
+        provider._refresh_to_access_map["corrupt_refresh"] = "not_valid_base64_!!!"
+
+        with pytest.raises(TokenError, match="Cannot recover credentials"):
+            await provider.exchange_refresh_token(
+                client_info, refresh_token, ["homeassistant"]
+            )
```
sergeykad, line 991: Four gaps worth adding to `TestOAuthStatePersistence`:

**1. Chained refresh across restart** — `test_state_persists_across_restart` tests one refresh after restart. A second refresh (refresh2 → refresh3) across a restart is the most fragile path: `_refresh_to_access_map` entries for tokens issued post-load must also be saved and restored correctly.

**2. Access-token revocation does not cascade to refresh token** — call `revoke_token` with the stateless `AccessToken` returned from the initial code exchange, then assert the paired refresh token is still present in `provider.refresh_tokens`. This documents the current behaviour and will catch any future attempt to restore the cascade.

**3. `_save_state` failure on a read-only filesystem** — patch `Path.write_text` to raise `OSError`, call `exchange_authorization_code`, and assert no exception propagates and the returned token is still valid. This verifies that persistence failures are non-fatal.

**4. `ha_credentials` does not appear in the saved state file** — populate `provider.ha_credentials` before calling `_save_state`, then read the written JSON and assert `'ha_credentials'` is not a key. Protects against accidentally serializing raw HA tokens.

src/ha_mcp/auth/provider.py
kingpanther13: Fixed in 954c76d — updated docstring to clarify `oauth_state.json` is written inside `state_dir`, and added the security note about base64-encoded tokens and the 0o600 permission boundary.

src/ha_mcp/auth/provider.py
kingpanther13: Fixed in 954c76d — added `ValueError` to the exception tuple, added docstring note that `ha_credentials` is intentionally excluded (HA tokens are recoverable from stateless access tokens in `_refresh_to_access_map`).

src/ha_mcp/auth/provider.py
kingpanther13: Fixed in 954c76d — upgraded to `logger.error` with the state file path and consequence message:
```python
logger.error(
    "Failed to save OAuth state to %s — active sessions will "
    "not survive restart: %s",
    self._state_file, e,
)
```

src/ha_mcp/auth/provider.py
kingpanther13: Fixed in 954c76d — chmod now only runs when the directory is first created:
```python
created = not self._state_file.parent.exists()
self._state_file.parent.mkdir(parents=True, exist_ok=True)
if created:
    os.chmod(self._state_file.parent, 0o700)
```

src/ha_mcp/auth/provider.py
kingpanther13: Fixed in 954c76d — switched to `tempfile.mkstemp(dir=..., suffix='.tmp')` for unique filenames per process. Also added a comment noting source and destination must be on the same filesystem for atomic rename. Used `Path.replace()` instead of `Path.rename()` for cross-platform atomicity (rename doesn't overwrite on Windows).

src/ha_mcp/auth/provider.py
kingpanther13: Fixed in 954c76d — removed the `exists()` guard entirely. Now handles `ENOENT` explicitly:
```python
try:
    text = self._state_file.read_text()
except OSError as e:
    if e.errno == errno.ENOENT:
        return  # Normal first-start, no prior state
    logger.error("Cannot read OAuth state file %s: %s", self._state_file, e)
    return
```
Also split the exception handling — file I/O errors are caught separately from JSON/validation errors, so first-start doesn't show up as an error in logs.

src/ha_mcp/auth/provider.py
kingpanther13: Fixed in 954c76d — `_load_state` now populates local dicts first, then assigns atomically:
```python
new_clients = {cid: OAuthClientInformationFull.model_validate(d) for cid, d in ...}
new_tokens = {tid: RefreshToken(**d) for tid, d in ... if not expired}
new_map = {k: v for k, v in stored_map.items() if k in new_tokens}

# Commit atomically
self.clients, self.refresh_tokens, self._refresh_to_access_map = (
    new_clients, new_tokens, new_map
)
```
If validation fails partway through, in-memory state stays empty rather than partially populated.

src/ha_mcp/auth/provider.py
kingpanther13: Fixed in 954c76d — added the explanatory comment to `_revoke_internal`:
```python
if access_token_str:
    # Stateless access tokens are not stored in self.access_tokens,
    # so there is no server-side cascade to revoke. The paired
    # refresh token is only revoked when revoke_token() is called
    # with a RefreshToken argument directly.
    self.access_tokens.pop(access_token_str, None)
```
Also added a test (`test_access_token_revocation_does_not_cascade_to_refresh`) that documents this behavior — revoking an access token leaves the paired refresh token intact.

tests/src/unit/test_oauth.py
kingpanther13: All four added in 954c76d:

1. **Chained refresh across restart** (`test_chained_refresh_across_restart`) — full flow: code exchange → refresh1 → restart → refresh2 → verify HA token preserved. Exercises `_refresh_to_access_map` entries created post-load.

2. **Access-token revocation doesn't cascade** (`test_access_token_revocation_does_not_cascade_to_refresh`) — revoke stateless access token, assert paired refresh token still in `provider.refresh_tokens`.

3. **`_save_state` failure is non-fatal** (`test_save_state_failure_is_nonfatal`) — patches `Path.write_text` to raise `OSError`, verifies `exchange_authorization_code` succeeds and returned token is valid.

4. **`ha_credentials` excluded from state file** (`test_ha_credentials_not_in_saved_state`) — populates `ha_credentials`, calls `_save_state` via code exchange, reads JSON from disk, asserts `'ha_credentials'` is not a key.

julienld:
### Post-merge note: disk persistence doesn't survive container restarts

The `state_dir` defaults to `Path.home() / ".ha-mcp"` (`/root/.ha-mcp` in Docker), which is inside the ephemeral container filesystem. Without a `VOLUME` declaration in the Dockerfile or a user-configured volume mount, all OAuth state is still lost on container restart — the same symptom as the original #787.

The HA add-on has the same issue unless `/root/.ha-mcp` falls on a path persisted by Supervisor.

Opened #837 with analysis and possible fixes (stateless refresh tokens, volume mount, or long-lived tokens).

PR #796, opened 2026-03-21
fix: block registry-disable on automation/script entities

### LINKED ISSUE #794: [AGENT]  `ha_config_get_automation` returns RESOURCE_NOT_FOUND for `automation.*` identifier after disabling automation entity via `ha_set_entity(enabled=false)` (transient until Core restart)

When disabling an automation, if the agent uses `ha_set_entity(..., enabled=false)` instead of `automation.turn_off`, this unexpectedly makes the automation disappear entirely from the HA UI until both the entity is reenabled _AND_ HA Core is restarted. It also makes the automation fail to appear in queries to the `ha_config_get_automation` tool.

This is arguably a HA Core bug, but AFAIK `ha-mcp` is one of the few ways to get into this scenario (the UI doesn't disable the automation entities), so a first step is to add MCP documentation to ensure that agents **never call `ha_set_entity(automation.*, enabled=false)`** and instead call `automation.turn_off`.

```markdown
**Note: I used `Composer 2 fast` to debug and create this bug report, and it's a bit special, so it didn't mention the worst part, which is that the automation disappears from the UI until you restart HA core. I also edited the environment section to fill in a few things it didn't bother to add.**

# Draft for GitHub: homeassistant-ai/ha-mcp

**Open:** https://github.com/homeassistant-ai/ha-mcp/issues/new?template=runtime_bug.yml

**Suggested title:** `ha_config_get_automation` returns RESOURCE_NOT_FOUND for `automation.*` identifier after disabling automation entity via `ha_set_entity(enabled=false)` (transient until Core restart)

---

## Bug description

After disabling an automation using **`ha_set_entity`** with **`enabled: false`** (entity registry disable, `disabled_by: user`), **`ha_config_get_automation`** with an **`automation.<slug>`** identifier returns **`RESOURCE_NOT_FOUND`**, even though the automation still exists. **`ha_get_entity`** for the same `entity_id` still succeeds, and **`ha_config_get_automation`** with the automation’s **numeric `unique_id`** still succeeds.

Disabling the same automation with **`automation.turn_off`** (service call) does **not** trigger this behavior.

In at least one reproduction, **`ha_set_entity(enabled=true)`** did **not** restore **`automation.*`** lookups; a **Home Assistant Core restart** cleared the problem.

## Steps to reproduce

1. Create a minimal automation via `ha_config_set_automation` (e.g. alias `test automation`, time trigger, trivial action).
2. Call `ha_config_get_automation` with identifier `automation.test_automation` → **success**.
3. Call `ha_set_entity` with `entity_id: automation.test_automation`, `enabled: false`.
4. Call `ha_config_get_automation` with identifier `automation.test_automation` → **RESOURCE_NOT_FOUND**.
5. (Optional) Call `ha_set_entity` with `enabled: true` and retry step 4 — may still fail until Core restart.
6. Confirm `ha_config_get_automation` with numeric `unique_id` still works; confirm `ha_get_entity` still returns the entity.

## Expected vs actual

- **Expected:** Config API should resolve automation config by `entity_id` whenever the entity exists in the registry (or ha-mcp should document that registry-disabled automations must be queried by `unique_id` only).
- **Actual:** `automation.*` lookup fails after registry disable; behavior is inconsistent with `automation.turn_off` and is only fully cleared by Core restart in some cases.

## Workaround

- Restart Home Assistant Core, **or**
- Use **`ha_config_get_automation(<numeric unique_id>)`** instead of `automation.*`, **or**
- Prefer **`automation.turn_off`** instead of **`ha_set_entity(enabled=false)`** when the goal is only to stop the automation from running.

## Environment (fill in when submitting)

- ha-mcp: 7.0.0 (from `ha_report_issue` output)
- Home Assistant Core: *2026.3.2*
- Installation: *HA OS*

## Additional context

- `ha_report_issue` was run from Cursor; its snapshot may show “connection error” to HA while local logs still contain successful tool calls — treat server-side log buffer as the source of truth for repro sequence.
```

kingpanther13:
## What does this PR do?

Prevents `ha_set_entity(enabled=False)` from being used on `automation.*` and `script.*` entities, addressing #794.

Registry-disabling (via `enabled=False`) removes entities from the HA state machine entirely, hiding them from the UI and making them unqueryable via state APIs until re-enabled **AND** the integration is reloaded. This is a destructive side effect that agents are very unlikely to intend when asked to "disable an automation."

**Changes:**
- **Guardrail**: When `enabled=False` is used on an automation or script entity, the tool now returns a clear error directing the agent to use `automation.turn_off` / `script.turn_off` via `ha_call_service()` instead
- **Description update**: The `enabled` parameter description now warns about the destructive nature of registry-level disable (entity removed from state machine, hidden from UI, reload required to restore)
- **Docstring update**: Added an `ENABLED/DISABLED WARNING` section to the tool docstring explaining the difference and the automation/script restriction
- **Unit tests**: 5 new tests covering the guardrail (automation blocked, script blocked, string "false" blocked, enable still allowed, other domains still allowed)

**Reproduction of the issue** (performed on a real HA instance before writing this PR):
1. `automation.turn_off` → state becomes "off", entity stays visible and queryable ✅
2. `ha_set_entity(enabled=False)` → entity disappears from state machine (404), invisible in UI ❌
3. `ha_set_entity(enabled=True)` → registry updated but entity still 404 until `automation.reload`

Closes #794

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 💥 Breaking change

## Testing

## Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -23,10 +23,14 @@ ENABLE_WEBSOCKET=true
 DEBUG=false
 LOG_LEVEL=INFO
 
-# Tool Search: replace the full tool catalog (~80 tools) with search-based
+# Tool Search: replace the full tool catalog (~92 tools) with search-based
 # discovery (~4 proxy tools). Reduces idle context from ~46K to ~5K tokens.
 # Recommended for models without deferred tool support or with limited context.
 # ENABLE_TOOL_SEARCH=false
+
+# Managed YAML config editing: allows ha_config_set_yaml to add/replace/remove
+# top-level keys in configuration.yaml and packages/*.yaml.
+# ENABLE_YAML_CONFIG_EDITING=false
 
 # Optional: MCP Server Configuration
 MCP_SERVER_NAME=ha-mcp
--- a/.gemini/styleguide.md
+++ b/.gemini/styleguide.md
@@ -2,7 +2,7 @@
 
 ## Project Context
 
-**ha-mcp** is a Model Context Protocol (MCP) server that enables AI assistants to control Home Assistant smart homes. It provides 80+ tools for entity control, automations, device management, and configuration via Home Assistant's REST and WebSocket APIs.
+**ha-mcp** is a Model Context Protocol (MCP) server that enables AI assistants to control Home Assistant smart homes. It provides 92+ tools for entity control, automations, device management, and configuration via Home Assistant's REST and WebSocket APIs.
 
 **Key Technologies:**
 - Python 3.13, FastMCP framework
@@ -11,7 +11,7 @@
 - Architecture: Tool registry with lazy loading, service layer pattern, WebSocket state verification
 
 **Code Organization:**
-- `src/ha_mcp/tools/` - 80+ MCP tools (tool modules auto-discovered)
+- `src/ha_mcp/tools/` - 92+ MCP tools (tool modules auto-discovered)
 - `src/ha_mcp/client/` - REST and WebSocket clients
 - `tests/src/e2e/` - End-to-end tests with real Home Assistant instance
 - `tests/src/unit/` - Unit tests for utilities
--- a/.github/workflows/gemini-triage.yml
+++ b/.github/workflows/gemini-triage.yml
@@ -253,7 +253,7 @@ jobs:
             Base your evaluation ONLY on the issue text provided below.
 
             **Project**: Home Assistant MCP Server (ha-mcp)
-            - MCP server with 95+ tools for AI assistants to control Home Assistant
+            - MCP server with 92+ tools for AI assistants to control Home Assistant
             - Supports: Claude Desktop, Claude Code, Cursor, Gemini CLI, Continue, etc.
             - Installation: uvx, pip, docker, Home Assistant add-on
             - Users can run \`ha_report_issue\` tool to auto-collect diagnostic info
@@ -359,7 +359,7 @@ jobs:
 
             **Project Context:**
             - ha-mcp is an MCP (Model Context Protocol) server that connects AI assistants to Home Assistant
-            - It provides 95+ tools for controlling smart home devices, automations, and scenes
+            - It provides 92+ tools for controlling smart home devices, automations, and scenes
             - Installation methods: uvx, pip, Docker, or Home Assistant add-on
             - Codebase structure: src/ha_mcp/ (tools, client, server code)
             - Common issues: connection timeouts, authentication errors, tool failures
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -63,7 +63,7 @@ Custom agent workflows are located in `.claude/agents/`:
 
 ## Project Overview
 
-**Home Assistant MCP Server** - A production MCP server enabling AI assistants to control Home Assistant smart homes. Provides 80+ tools for entity control, automations, device management, and more.
+**Home Assistant MCP Server** - A production MCP server enabling AI assistants to control Home Assistant smart homes. Provides 92+ tools for entity control, automations, device management, and more.
 
 - **Repo**: `homeassistant-ai/ha-mcp`
 - **Package**: `ha-mcp` on PyPI
@@ -421,7 +421,7 @@ Balance improvement against regression risk. Consider:
 ### Setup
 ```bash
 uv sync --group dev        # Install with dev dependencies
-uv run ha-mcp              # Run MCP server (80+ tools)
+uv run ha-mcp              # Run MCP server (92+ tools)
 cp .env.example .env       # Configure HA connection
 ```
 
@@ -484,7 +484,7 @@ src/ha_mcp/
 │   ├── rest_client.py       # HTTP REST API client
 │   ├── websocket_client.py  # Real-time state monitoring
 │   └── websocket_listener.py
-├── tools/             # 28 modules, 80+ tools
+├── tools/             # 28 modules, 92+ tools
 │   ├── registry.py          # Lazy auto-discovery
 │   ├── smart_search.py      # Fuzzy entity search
 │   ├── device_control.py    # WebSocket-verified control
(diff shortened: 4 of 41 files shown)


src/ha_mcp/tools/tools_entities.py
```diff
@@ -449,6 +466,38 @@ async def ha_set_entity(
...
 
+            # Block registry-disable on automation and script entities.
+            # Registry-disabling (enabled=False) removes the entity from the HA
+            # state machine entirely, making it invisible in the UI and
+            # unqueryable via state APIs until re-enabled AND the integration is
+            # reloaded.  For automations and scripts the correct way to
+            # "disable" them is via their domain services (automation.turn_off /
+            # script.turn_off) which simply prevent them from running while
+            # keeping them visible and manageable.
+            if enabled is not None and not is_bulk:
```
sergeykad: **1. `not is_bulk` makes the guardrail fragile**

This condition is redundant — bulk already rejects `enabled` at lines 441-450. But it creates a fragile implicit dependency: if bulk ever supports `enabled`, the guardrail silently stops protecting automation/script entities.

Consider removing `not is_bulk` and iterating all `entity_ids` to make the guardrail self-contained:

```python
if enabled is not None:
    try:
        _enabled_check = coerce_bool_param(enabled, "enabled")
    except ValueError:
        _enabled_check = None

    if _enabled_check is False:
        blocked = [eid for eid in entity_ids if eid.split(".")[0] in ("automation", "script")]
        if blocked:
            _domain = blocked[0].split(".")[0]
            # ... raise error ...
```

src/ha_mcp/tools/tools_entities.py
```diff
@@ -449,6 +466,38 @@ async def ha_set_entity(
...
+                try:
+                    _enabled_check = coerce_bool_param(enabled, "enabled")
+                except ValueError:
+                    _enabled_check = None  # will be caught later by existing validation
```
sergeykad: **2. Duplicate `coerce_bool_param` with silent swallow relying on buggy downstream**

`coerce_bool_param` is called here and again at line 112 in `_update_single_entity`. The `except ValueError` silently sets `None` trusting downstream validation — but that downstream validation at line 114 itself has a bug: it does `return create_error_response(...)` instead of `raise_tool_error(create_error_response(...))`. Per CLAUDE.md: "All tool-level failures must raise `ToolError` (sets `isError=true` per MCP spec)." Same issue at line 125 for `hidden`.

So this guardrail trusts a downstream handler that itself violates the error handling contract. Hoist the coercion once before the guardrail and pass the parsed value to `_update_single_entity`. Fix lines 114 and 125 to use `raise_tool_error()` (Boy Scout Rule — same function).

tests/src/unit/test_tools_entities.py
```diff
@@ -1087,3 +1087,116 @@ async def test_bulk_label_add_operation(self, mock_mcp, mock_client):
...
+
+        error_text = str(exc_info.value)
+        assert "script" in error_text.lower()
+        assert "turn_off" in error_text
+        mock_client.send_websocket_message.assert_not_called()
+
+    @pytest.mark.asyncio
+    async def test_disable_automation_string_false_blocked(self, set_entity_tool, mock_client):
+        """enabled='false' (string) on automation entity should also be blocked."""
+        with pytest.raises(ToolError):
```
sergeykad: **3. Missing error message assertions and test variants**

Unlike `test_disable_automation_blocked` and `test_disable_script_blocked`, this test doesn't assert on the error message content. Without that, it could pass due to a different validation error rather than the guardrail.

Add:
```python
with pytest.raises(ToolError) as exc_info:
    await set_entity_tool(entity_id="automation.morning", enabled="false")
error_text = str(exc_info.value)
assert "automation" in error_text.lower()
assert "turn_off" in error_text
```

Also missing:
- Test for `"False"` (capital F) — common from Python-based agents
- Test for single-element list `entity_id=["automation.test"]` — distinct code path through `is_bulk` logic

src/ha_mcp/tools/tools_entities.py
kingpanther13: Fixed in bd3000d — removed `not is_bulk`, guardrail now iterates all `entity_ids`:

```python
blocked = [
    eid for eid in entity_ids
    if eid.split(".")[0] in ("automation", "script")
]
if blocked:
    ...
```

Self-contained regardless of future bulk `enabled` support. Also added a test for single-element list `["automation.test"]` to cover the code path that goes through `is_bulk=False` with a list input.

src/ha_mcp/tools/tools_entities.py
kingpanther13: Good catch on both issues. Fixed in bd3000d + 6c6a0e7:

1. **`return` → `raise_tool_error`** at lines 114 and 125 in `_update_single_entity` — both `enabled` and `hidden` validation now properly raise `ToolError` per the error handling contract.

2. **Updated tests** — `test_enabled_invalid_value_returns_error` and `test_hidden_invalid_value_returns_error` now use `pytest.raises(ToolError)` instead of checking a dict return.

The guardrail's `except ValueError: _enabled_check = None` comment updated to note it's caught by the now-fixed downstream validation.

tests/src/unit/test_tools_entities.py
kingpanther13: All addressed in bd3000d:

1. **Error message assertions** added to `test_disable_automation_string_false_blocked` — now asserts `"automation"` and `"turn_off"` in error text.

2. **`"False"` (capital F) test** added — `test_disable_automation_string_capital_false_blocked` covers the common Python-agent case.

3. **Single-element list test** added — `test_disable_automation_single_element_list_blocked` with `entity_id=["automation.test"]` covers the distinct code path through `is_bulk=False` with list input.

PR #805, opened 2026-03-22
Clarify Claude.ai connection error and stateless session logs

### LINKED ISSUE #783: [BUG] Couldn't reach MCP server
### 💬 What Happened?

I cant connect to Claude.
I get:

INFO     Starting MCP server 'ha-mcp' with  transport.py:273
                             transport 'http' (stateless) on                    
                             http://0.0.0.0:9583/private_5qqqqqqqqqqqqqqqqqqqqq       
...
INFO:     000.00.000.00:0 - "POST /private_5qqqqqqqqqqqqqqqqqqqqq HTTP/1.1" 200 OK
INFO:mcp.server.streamable_http:Terminating session: None

I have haos, with the mcp server and proxy running.
Cloudflared is on a seperate lxc, and I assume connected correctly via 8123 port.
So basically: Cloudflare tunnel → HA port 8123 → webhook proxy → ha-mcp on 9583

It didnt work on 7.0.0. and it doenst on 7.0.1.

```markdown
Title: MCP server reports 200 OK but session terminates immediately — Claude.ai never connects (7.0.0, 7.0.1)
Environment:

ha-mcp addon version: 7.0.1 (also tested 7.0.0)
HAOS as KVM VM
Cloudflared on a separate LXC container
Architecture: Cloudflare tunnel → HA :8123 → webhook proxy → ha-mcp :9583

What happens:
INFO     Starting MCP server 'ha-mcp' with transport 'http' (stateless) on http://0.0.0.0:9583/private_5qqq...
INFO:     x.x.x.x:0 - "POST /private_5qqq... HTTP/1.1" 200 OK
INFO:mcp.server.streamable_http:Terminating session: None

The request reaches the addon (200 OK is returned), but the session is immediately terminated with session: None. Claude.ai never establishes a working connection.
What's expected:
Claude.ai connects successfully and HA tools become available.
What I've tried:

Upgraded from 7.0.0 to 7.0.1 — same behavior
Cloudflared tunnel confirmed working on port 8123 for regular HA access
Webhook proxy path appears to be routing correctly (POST reaches ha-mcp)

Suspicion:
Terminating session: None suggests the addon isn't creating/tracking a session object, possibly a bug in how stateless HTTP transport handles the initial handshake, or a missing/malformed header that Claude.ai's MCP client expects.
```

kingpanther13:
## What does this PR do?

Improves the setup experience by addressing the most common sources of user confusion when connecting to ha-mcp — especially via Claude.ai and Cloudflare.

### Claude.ai connection messaging
- **"Couldn't reach the MCP server"** — Claude.ai shows this during its initial handshake but connects successfully afterward. Added notes to the Claude.ai setup guide, FAQ, and setup wizard explaining this is expected, with steps to verify the connection is actually working (check Configure button, or ask Claude in a new conversation).

### Stateless session log noise
- **"Terminating session: None"** — The MCP SDK logs this on every request in stateless HTTP mode, which looks alarming but is routine cleanup. Added a log filter that annotates these messages with "(Normal — stateless mode)" and a FAQ entry explaining the behavior.

### Cloudflare "Block AI training bots" (most common connection issue)
Cloudflare's AI bot blocking silently breaks LLM connections. Added detailed step-by-step instructions for disabling it, with a screenshot, to **every place a Cloudflare user might look**:
- **Browser landing page** (#777) — expanded with full walkthrough and screenshot link
- **Setup wizard** (`setup.astro`) — amber warning box in the Cloudflare Tunnel section
- **FAQ** — new troubleshooting section
- **Cloudflared deployment guide** (`cloudflared.md`)
- **Add-on DOCS.md** — under the Cloudflared section

### Setup wizard improvements
- **UI-based clients** (Claude.ai, ChatGPT, Raycast, Open WebUI) now render setup steps as formatted instruction blocks instead of inside a monospace code box
- **"Verify Your Server"** step added to all HTTP-based setups — tells users to visit the MCP URL in their browser to confirm the server is running
- Config code box + copy button hidden for UI clients (not relevant when there's no config file)

Also filed upstream: [modelcontextprotocol/python-sdk#2329](https://github.com/modelcontextprotocol/python-sdk/issues/2329)

Closes #783

## Type of change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [x] 📚 Documentation
- [x] 🔧 Maintenance/refactor
- [ ] 💥 Breaking change

## Testing
- [x] Setup wizard tested locally via `npm run dev`

## Checklist

### DIFF
--- a/tests/src/unit/test_browser_landing.py
+++ b/tests/src/unit/test_browser_landing.py
@@ -34,6 +34,7 @@ async def test_get_returns_405_with_helpful_message(mcp_app):
     assert resp.status_code == 405
     assert "HA-MCP server is up and running" in resp.text
     assert "Block AI training bots" in resp.text
+    assert '"do not block (allow crawlers)"' in resp.text
     assert "dash.cloudflare.com" in resp.text
     assert resp.headers["allow"] == "POST, DELETE"
 


src/ha_mcp/__main__.py
```diff
@@ -325,6 +325,22 @@ def __getattr__(self, name: str) -> Any:
...
+    session, producing an INFO log that looks alarming but is routine.
+    """
+
+    def filter(self, record: logging.LogRecord) -> bool:
+        if (
+            record.name == "mcp.server.streamable_http"
+            and "Terminating session: None" in record.getMessage()
+        ):
+            record.msg = f"{record.msg} (Normal — stateless mode)"
+        return True
```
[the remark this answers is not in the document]
kingpanther13: Already addressed — class is named `StatelessSessionLogFilter` (no underscore) and tests exist in `tests/src/unit/test_stateless_session_log_filter.py`.

docs/FAQ.md
```diff
@@ -46,6 +46,38 @@ The demo environment resets weekly. Your changes won't persist.
...
+To disable it:
+
+1. Log in to [Cloudflare](https://dash.cloudflare.com)
+2. In the left sidebar, click **Domains**, then click **Overview**
+3. Click on the domain you use for connecting to Home Assistant
+4. On the right side of the page, find **"Control AI Crawlers"**
+5. Under **"Block AI training bots"**, open the dropdown
+6. Select **"do not block (allow crawlers)"**
+
+![Cloudflare AI Crawlers Setting](https://github.com/user-attachments/assets/3b949eff-951f-40d7-a460-17dfbfea045f)
```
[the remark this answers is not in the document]
kingpanther13: Fixed. Moved screenshot to `site/public/images/cloudflare-ai-crawlers-setting.jpg` and updated all references to use the permanent GitHub Pages URL.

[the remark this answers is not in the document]
kingpanther13: done

homeassistant-addon/start.py
```diff
@@ -169,12 +169,16 @@ def main() -> int:
...
     from ha_mcp.__main__ import (
+        StatelessSessionLogFilter,
         _get_timestamped_uvicorn_log_config,
         mcp,
         register_browser_landing,
     )
 
     register_browser_landing(mcp, secret_path)
+    logging.getLogger("mcp.server.streamable_http").addFilter(
+        StatelessSessionLogFilter()
```
sergeykad, line 180: `_setup_logging()` in `__main__.py` already adds a `StatelessSessionLogFilter` to this same logger. If `_setup_logging()` is ever called before this point in the addon path, the logger ends up with two filter instances. The second filter sees the already-annotated message — `"Terminating session: None"` is still a substring of `"Terminating session: None (Normal — stateless mode)"` — and appends the suffix again:

```
Terminating session: None (Normal — stateless mode) (Normal — stateless mode)
```

Either guard the filter against double-annotation (e.g. skip if `"(Normal" in record.getMessage()`), or remove this explicit `addFilter` call and rely solely on `_setup_logging()` in the addon flow.

docs/FAQ.md
```diff
@@ -46,6 +46,38 @@ The demo environment resets weekly. Your changes won't persist.
...
 ## Troubleshooting
 
+### Claude.ai says "Couldn't reach the MCP server"
+
+**This is normal.** Claude.ai shows this error during its initial connection handshake, but the server connects successfully afterward. To verify you're actually connected:
+
+1. Look for a **"Configure"** button on the connector — click it
+2. If you see tools listed, you're connected and ready to go
+3. You can also start a new conversation and ask Claude if it can see your Home Assistant via the MCP connection — this is the easiest way to confirm it's truly connected
+4. You can also check your server logs — successful requests (HTTP 200) after the initial error confirm the connection is working
```
sergeykad: Items 3 and 4 are alternatives ("you can also…"), not sequential steps — but they sit inside a numbered list that implies sequence. Consider making them unnumbered bullets, or folding them into item 2 as a single note.

docs/FAQ.md
```diff
@@ -46,6 +46,38 @@ The demo environment resets weekly. Your changes won't persist.
...
+
+This is a known Claude.ai behavior that affects all MCP servers, not just ha-mcp.
+
+### "Terminating session: None" in server logs
+
+**This is normal.** ha-mcp runs in stateless HTTP mode, which means each request creates a temporary session that is cleaned up immediately after. The `Terminating session: None` log message is the MCP SDK reporting this routine cleanup — it does not mean your connection was dropped.
```
sergeykad: "a temporary session that is cleaned up immediately after" — passive voice, and "immediately after" is left dangling. Suggested rewrite:

> ha-mcp runs in stateless HTTP mode, which means each request creates and discards a temporary session. The `Terminating session: None` log message is the MCP SDK reporting this routine cleanup — the connection stays active.

src/ha_mcp/__main__.py
```diff
@@ -590,18 +609,36 @@ def register_browser_landing(mcp_instance: "FastMCP | _DeferredMCP", path: str)
...
+        "3. Click on the domain you use for connecting to Home Assistant\n"
+        '4. On the right side, find "Control AI Crawlers"\n'
+        '5. Under "Block AI training bots", open the dropdown\n'
+        '6. Select "do not block (allow crawlers)"\n'
+        "\n"
+        "Screenshot of the setting:\n"
+        "https://homeassistant-ai.github.io/ha-mcp/images/cloudflare-ai-crawlers-setting.jpg\n"
+    )
+
     # Safe because the MCP streamable-http transport claims only POST and DELETE.
```
sergeykad: The first line is inaccurate. The MCP transport itself supports GET, POST, and DELETE — GET is used for SSE streams in stateful mode. What restricts the MCP endpoint to POST and DELETE here is FastMCP's route registration: when `stateless_http=True`, `create_streamable_http_app` registers the Starlette Route with `methods=["POST", "DELETE"]` explicitly, so Starlette 405s any GET before the MCP handler runs. The transport layer is not the safety mechanism.

Suggested replacement for the first line:
```
# Safe because FastMCP registers the MCP route with methods=["POST", "DELETE"]
# in stateless mode, so Starlette rejects GET requests before the MCP handler runs.
```

src/ha_mcp/__main__.py
```diff
@@ -325,6 +325,22 @@ def __getattr__(self, name: str) -> Any:
...
+    In stateless HTTP mode every request creates and tears down a temporary
+    session, producing an INFO log that looks alarming but is routine.
+    """
+
+    def filter(self, record: logging.LogRecord) -> bool:
+        if (
+            record.name == "mcp.server.streamable_http"
+            and "Terminating session: None" in record.getMessage()
+        ):
+            record.msg = f"{record.msg} (Normal — stateless mode)"
```
sergeykad: The mutation is correct today because the MCP SDK uses an f-string (`logger.info(f"Terminating session: {self.mcp_session_id}")`), so `record.args` is always `()`. But if the SDK ever switches to `%`-style formatting (`logger.info("Terminating session: %s", session_id)`), `record.msg` becomes `"Terminating session: %s"` and the mutation leaves `record.args = (None,)` intact — which still renders correctly via `getMessage()`, but leaves the record internally inconsistent for any structured-logging consumer that reads `record.msg` and `record.args` separately (e.g., Sentry, JSON log formatters).

The idiomatic fix is to normalize before mutating:
```python
record.msg = record.getMessage() + " (Normal — stateless mode)"
record.args = ()
```
This is safe regardless of how the SDK formats the message.

tests/src/unit/test_stateless_session_log_filter.py
```diff
@@ -0,0 +1,55 @@
...
+        self.log_filter = StatelessSessionLogFilter()
+
+    def _make_record(self, name: str, msg: str) -> logging.LogRecord:
+        return logging.LogRecord(
+            name=name,
+            level=logging.INFO,
+            pathname="",
+            lineno=0,
+            msg=msg,
+            args=(),
```
sergeykad, line 21: All four test cases construct records with `args=()`, which is not how the SDK would emit this log if it ever switched from an f-string to `%`-style formatting. A test covering the realistic `args=(None,)` case would catch any regression if the mutation logic is changed:

```python
def test_annotates_stateless_termination_printf_style(self):
    record = self._make_record("mcp.server.streamable_http", "Terminating session: %s")
    record.args = (None,)
    result = self.log_filter.filter(record)
    assert result is True
    assert "Normal — stateless mode" in record.getMessage()
    assert record.args == ()  # args must be cleared after normalization
```

tests/src/unit/test_browser_landing.py
```diff
@@ -33,6 +33,8 @@ async def test_get_returns_405_with_helpful_message(mcp_app):
 
     assert resp.status_code == 405
     assert "HA-MCP server is up and running" in resp.text
+    assert "Block AI training bots" in resp.text
+    assert "dash.cloudflare.com" in resp.text
```
sergeykad, line 38: The most actionable instruction in the Cloudflare section — `"do not block (allow crawlers)"` — is not asserted. If the message were ever shortened and that line dropped, no test would fail. Worth adding:
```python
assert '"do not block (allow crawlers)"' in resp.text
```

homeassistant-addon/start.py
```diff
@@ -169,12 +169,16 @@ def main() -> int:
     # Import and register browser landing before server start
     log_info("Importing ha_mcp module...")
     from ha_mcp.__main__ import (
+        StatelessSessionLogFilter,
```
sergeykad, line 172: `StatelessSessionLogFilter` is a cosmetic concern — it annotates a log message but has no effect on server operation. Adding it to this import block means a rename or removal of the class (e.g., during a `__main__.py` refactor) causes the entire add-on to fail at startup with a bare `ImportError`, before logging is even initialized, with no actionable message for the user.

Consider importing it inside a try/except and skipping `addFilter` on failure, or moving the filter registration into a dedicated setup helper that `start.py` can call without importing the class directly.

homeassistant-addon/start.py
sergeykad: Correction: this is not a bug. The code reviewer confirmed that `start.py` uses its own `logging.basicConfig()` and never calls `_setup_logging()` from `__main__.py`. The two `addFilter` calls are in mutually exclusive code paths — the add-on path and the standard entrypoints never both execute in the same process. The explicit registration in `start.py` is necessary and correct.

src/ha_mcp/__main__.py
```diff
@@ -325,6 +325,22 @@ def __getattr__(self, name: str) -> Any:
 _LOG_DATE_FORMAT = "%Y-%m-%d %H:%M:%S"
 
 
+class StatelessSessionLogFilter(logging.Filter):
```
sergeykad, line 328: An alternative worth considering: downgrade the log level to DEBUG instead of rewriting the message.

```python
def filter(self, record: logging.LogRecord) -> bool:
    if (
        record.name == "mcp.server.streamable_http"
        and "Terminating session: None" in record.getMessage()
    ):
        record.levelno = logging.DEBUG
        record.levelname = "DEBUG"
    return True
```

The message disappears from normal INFO output but stays visible under `--debug`. That matches what the SDK should do — stateless teardown is a DEBUG-level event. Changing the level is a cleaner mutation than rewriting the message text, and it sidesteps the `record.args` concern raised on the `filter()` method.

A `# TODO: remove when modelcontextprotocol/python-sdk#2329 is resolved` comment would make the intent clear and give a path to deleting this workaround once the SDK logs at DEBUG natively.

kingpanther13: Implemented — switched to downgrading to DEBUG instead of annotating the message. Added the TODO referencing python-sdk#2329. This also makes the `record.args` concern moot since we're not touching the message text anymore.

src/ha_mcp/__main__.py
kingpanther13: Resolved — switched to downgrading the log level to DEBUG (per your other suggestion) instead of mutating the message, so record.args is never touched.

docs/FAQ.md
kingpanther13: Good catch — moved the alternatives out of the numbered list into a plain paragraph.

docs/FAQ.md
kingpanther13: Adopted your suggested wording — much cleaner.

tests/src/unit/test_stateless_session_log_filter.py
kingpanther13: Added — test now covers the `args=(None,)` case. Assertion checks levelno/levelname since we switched to the DEBUG downgrade approach.

src/ha_mcp/__main__.py
kingpanther13: Good catch — fixed. Updated the comment to correctly attribute the method restriction to FastMCP's route registration in stateless mode, not the MCP transport itself.

tests/src/unit/test_browser_landing.py
kingpanther13: Added — `assert '\"do not block (allow crawlers)\"' in resp.text` now verifies the actionable instruction is present.

homeassistant-addon/start.py
kingpanther13: Pushing back on this one. All four imports in this block come from the same module (`ha_mcp.__main__`). If `StatelessSessionLogFilter` fails to import, the others (`mcp`, `register_browser_landing`, `_get_timestamped_uvicorn_log_config`) will too — they're defined in the same file. And `mcp` is the FastMCP server instance: wrapping it in try/except would let the addon start as a zombie process that serves nothing, which is much harder to debug than a clear ImportError at startup. Since these imports can only fail together, wrapping any of them individually doesn't add real resilience.

Issue #709, opened 2026-03-06
[FEATURE] Systematic verification for config tool operations (prevent false success)


kingpanther13:
## What would you like to see?

Add post-operation verification to all config/state-modifying tools, following the pattern already established by automations, scripts, and helpers. Currently 9 out of 13 config tool modules report success based solely on the HA API acknowledgment without confirming the operation actually completed — leading to "false success" responses.

This was identified while investigating three related issues:
- #610 — `ha_config_set_automation` reported success with a fabricated entity_id when the automation never loaded (fix in #708)
- #685 — `ha_import_blueprint` reported success but never called `blueprint/save`
- #698 — `ha_config_set_automation` silently created duplicates (fix in #708)

All three share the same root cause: **trusting the HA API acknowledgment as proof of completion**.

## Why do you need this?

When a tool reports `"success": true`, agents and users trust that the operation completed. False successes cause:
- **Silent data loss**: user thinks automation/blueprint/zone was created, moves on, never realizes it wasn't saved
- **Cascading failures**: agent builds on a resource that doesn't exist (e.g. creates automation referencing a non-existent blueprint)
- **Difficult debugging**: no error to investigate — everything looked fine

The tools that already have verification (automations, scripts, helpers, entities) use `wait_for_entity_registered()` / `wait_for_entity_removed()` — the infrastructure exists, it just isn't applied consistently.

### Audit results

**Tools WITH proper verification (4/13)** — these are the model to follow:
| Tool | Verification method |
|---|---|
| `tools_config_automations.py` | `wait_for_entity_registered()` after create, `wait_for_entity_removed()` after delete |
| `tools_config_scripts.py` | Same pattern |
| `tools_config_helpers.py` | Same pattern, plus robust fallback for delete |
| `tools_entities.py` | Fetches entity state after update to confirm changes |

**Tools WITHOUT verification (9/13):**

#### High Risk — user-visible false success
| Tool | File | Operations | Issue |
|---|---|---|---|
| `ha_import_blueprint` | `tools_blueprints.py` | import | Only calls `blueprint/import`, skips `blueprint/save` (separate bug, #685) |
| `ha_config_set_group` | `tools_groups.py` | create/update | Predicts `group.{object_id}` without verifying entity exists |
| `ha_set_integration_enabled` | `tools_integrations.py` | enable/disable | Trusts WS response, no state check |
[...]

PR #812, opened 2026-03-23
fix: surface connection errors in ha_get_overview instead of returning empty data

### LINKED ISSUE #811: [BUG] Not able to view any entities within HA
### 💬 What Went Wrong?

Installing went perfectly, i setup, added my (previous working env vars from my M1 mac) and then quit Claude and re-opened. When i reopened and asked if Claude could see my Home Assitant it said it could, but it was an empty install (which is not right, it should show 500+ entities like before). Upon copying the EXACT config back to my M1 and creating a new token just to be sure it was working on the M1 again.

### 📋 AI-Generated Content / Logs / Config

```markdown
2026-03-23T19:25:36.708Z [Home Assistant] [info] Message from client: {"method":"tools/call","params":{"name":"ha_get_overview","arguments":{"detail_level":"minimal"}},"jsonrpc":"2.0","id":4} { metadata: undefined }
2026-03-23T19:25:36.737Z [Home Assistant] [info] Message from server: {"jsonrpc":"2.0","id":4,"result":{"content":[{"type":"text","text":"{\"success\":true,\"system_summary\":{\"total_entities\":0,\"total_domains\":0,\"total_services\":0,\"total_areas\":0},\"domain_stats\":{},\"area_analysis\":{},\"ai_insights\":{\"most_common_domains\":[],\"controllable_devices\":[],\"monitoring_sensors\":[],\"automation_ready\":false},\"notification_count\":0}"}],"structuredContent":{"success":true,"system_summary":{"total_entities":0,"total_domains":0,"total_services":0,"total_areas":0},"domain_stats":{},"area_analysis":{},"ai_insights":{"most_common_domains":[],"controllable_devices":[],"monitoring_sensors":[],"automation_ready":false},"notification_count":0},"isError":false}} { metadata: undefined }

  "mcpServers": {
    "Home Assistant": {
      "command": "uvx",
      "args": [
        "--refresh",
        "ha-mcp@latest"
      ],
      "env": {
        "HOMEASSISTANT_URL": "https://hostname",
        "HOMEASSISTANT_TOKEN": "ey..."
      }
    }
  }
```

### 🖥️ Client Application

Claude Desktop

### 📦 Installation Method

uvx ha-mcp

### 💻 Operating System

macOS

### OS Version

15.7.4

### 📖 Setup Guide Followed

https://homeassistant-ai.github.io/ha-mcp/guide-macos/

### 🔄 Exact Steps You Followed

Steps listed above, I followed the above setup, 
copied over env vars
quit claude
asked if it could see
got response with fresh install

### 🐍 Python Version

Python 3.9.6

### 📦 ha-mcp Version

latest

### 🏠 Home Assistant Version

Latest

### 💡 Additional Context

I run Caddy locally to resolve domain names to internal ips so using the direct hostname resolves properly on both macs in the browser and ping does show the internal IP instead of external proxy.

Works on my M1 mac, but not Intel

sergeykad:
## What does this PR do?

Fixes #811 — `ha_get_overview` was returning `success: true` with `total_entities: 0` when the HA connection was broken, making it look like an empty install rather than a connectivity problem.

**Root cause:** `asyncio.gather(..., return_exceptions=True)` captured the exception from `get_states()` silently. The tool then returned an empty but "successful" response.

**Changes:**

- **Entities fetch failure** → raises immediately via the existing `exception_to_structured_error` handler, producing a proper `ToolError` with connection troubleshooting suggestions
- **Services fetch failure** → logs at warning and sets `partial: true` + `warnings[]` in the response so callers can distinguish a real zero-service HA from a fetch failure
- **`get_config` swallow** → pre-existing `except Exception` already logged at warning; no change needed
- **`import random`** → moved from inside function body to module level

**Tests added:**
- `test_entities_fetch_failure_raises_error` — regression test for #811 core bug
- `test_services_fetch_failure_continues_with_empty_services` — verifies `partial: true` + `warnings[]` on services failure
- `assert "partial" not in result` added to happy-path test to catch regressions

## Type of change
- [x] 🐛 Bug fix

## Testing

## Checklist

### DIFF
--- a/src/ha_mcp/tools/smart_search.py
+++ b/src/ha_mcp/tools/smart_search.py
@@ -4,6 +4,7 @@ Smart search tools for Home Assistant MCP server.
 
 import asyncio
 import logging
+import random
 import time
 from typing import Any
 
@@ -387,9 +388,20 @@ class SmartSearchTools:
                 return_exceptions=True,
             )
 
-            # Process results, handling any exceptions gracefully
-            entities = results[0] if not isinstance(results[0], Exception) else []
-            services = results[1] if not isinstance(results[1], Exception) else []
+            # Entities are mandatory — surface connection/auth errors immediately.
+            # Services failure is logged at warning (affects total count and service catalog).
+            # Registry failures are logged at debug (area enrichment only).
+            if isinstance(results[0], Exception):
+                raise results[0]
+
+            entities = results[0]
+            partial_warnings: list[str] = []
+            if isinstance(results[1], Exception):
+                logger.warning(f"Could not fetch services: {results[1]}")
+                partial_warnings.append(f"Services unavailable: {results[1]}")
+                services = []
+            else:
+                services = results[1]
 
             # Handle area registry result
             area_registry: list[dict[str, Any]] = []
@@ -537,8 +549,6 @@ class SmartSearchTools:
             }
 
             # Prepare domain stats with entity filtering and truncation info
-            import random
-
             formatted_domain_stats = {}
             for domain, stats in sorted_domains:
                 all_entities = stats["all_entities"]
@@ -569,7 +579,7 @@ class SmartSearchTools:
                 }
 
             # Build base response
-            base_response = {
+            base_response: dict[str, Any] = {
                 "success": True,
                 "system_summary": {
                     "total_entities": len(entities),
@@ -581,6 +591,10 @@ class SmartSearchTools:
                 "area_analysis": area_stats,  # Now included in all detail levels
                 "ai_insights": ai_insights,
             }
+
+            if partial_warnings:
+                base_response["partial"] = True
+                base_response["warnings"] = partial_warnings
 
             # Add level-specific fields
             if detail_level == "full":
--- a/tests/src/unit/test_performance_parallelization.py
+++ b/tests/src/unit/test_performance_parallelization.py
@@ -8,9 +8,10 @@ Tests observable behavior of performance optimizations:
 """
 
 import asyncio
-from unittest.mock import patch
+from unittest.mock import AsyncMock, patch
 
 import pytest
+from fastmcp.exceptions import ToolError
 
 from ha_mcp.tools.smart_search import SmartSearchTools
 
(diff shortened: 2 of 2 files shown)


kingpanther13 (approved):
## Review Summary

**Verdict: Approve** — This is a clean, well-targeted bug fix with good test coverage.

### What the PR does well

1. **Correct root cause analysis.** The `asyncio.gather(..., return_exceptions=True)` pattern silently swallowed `get_states()` failures, making a broken connection indistinguishable from an empty HA install. Re-raising the entities exception so the outer handler converts it to a `ToolError` is the right fix.

2. **Good severity differentiation.** Entities are mandatory (connection/auth errors should surface immediately), while services are supplementary (degraded response via `partial: true` + `warnings[]` is appropriate). This matches the existing pattern where registry failures (areas, entities, devices) are already logged at debug and silently degraded.

3. **Clean `partial` flag design.** Returning `success: true` with `partial: true` and `warnings[]` lets consumers distinguish "HA genuinely has zero services" from "couldn't fetch services." This is a useful API contract addition.

4. **Solid test coverage.** Three additions:
   - `test_entities_fetch_failure_raises_error` — regression test for the core #811 bug
   - `test_services_fetch_failure_continues_with_empty_services` — verifies partial response
   - `assert "partial" not in result` on happy path — catches regressions

5. **Minor housekeeping is welcome** — `import random` moved to module level, type annotation added to `base_response`.

### Minor observations (non-blocking)

- The implicit behavior when **both** entities and services fail is correct (entities check comes first → raises before services are even inspected), but a brief comment or an additional test documenting this priority could help future maintainers. Not blocking.

### Relationship to #709

This PR is **related to but does not close** [#709](https://github.com/homeassistant-ai/ha-mcp/issues/709). Both deal with "false success" responses, but the mechanisms are different:

- **#709** catalogs **write operations** where the HA API acknowledges a mutation that never took effect (e.g., automation create returns success but entity never loads). The fix pattern is post-operation verification via `wait_for_entity_registered()`.
- **#811/#812** is a **read operation** where `return_exceptions=True` swallowed a connection error, making a broken connection look like an empty install. The fix pattern is exception inspection in gather results.

They share the same symptom (user/agent sees `success: true` when something went wrong) but different root causes and fix strategies. Worth cross-referencing in #709 as a related instance of the false-success pattern, but this PR correctly does **not** claim to close it. `ha_get_overview` wasn't in #709's audit because #709 scoped itself to config/state-modifying tools — this is a read-only overview tool that happened to have the same symptom class.

PR #813, opened 2026-03-24
refactor: consolidate 2 redundant tools (merge after #806)

### LINKED ISSUE #774: [FEATURE] Consolidate 5 redundant tools: areas+floors, rename entity, operation status

Three tool consolidations to reduce tool count by 5 (93 → 88). All merge tools that share internal helpers or are near-identical code in the same file.

### 1. Merge area + floor tools (6 → 3, saves 3 tools)

`ha_config_list_areas` + `ha_config_list_floors`, `ha_config_set_area` + `ha_config_set_floor`, `ha_config_remove_area` + `ha_config_remove_floor`

All 6 live in `tools_areas.py` and are structurally identical — same WebSocket pattern (`config/{area,floor}_registry/{list,create,update,delete}`), same error handling, nearly identical parameters. The only differences are floor-specific params (`level`) vs area-specific params (`floor_id`, `picture`).

**Approach:** Add a `type: Literal["area", "floor"]` parameter (default `"area"`) to each combined tool. Type-specific params are optional and only apply to their respective type. For the list tool, could also just return both areas and floors in a single call since both datasets are small.

### 2. Merge `ha_rename_entity` + `ha_rename_entity_and_device` (2 → 1, saves 1 tool)

Both already share `_rename_entity_internal()` in `tools_registry.py`. The combo tool is a thin wrapper that additionally looks up the `device_id` from the entity registry and calls `_update_device_internal()`.

**Approach:** Add optional `new_device_name` param to `ha_rename_entity`. If provided, do the device lookup + rename. If omitted, behavior is identical to today. The `new_entity_name` param from the combo tool maps to the existing `name` param on `ha_rename_entity`.

### 3. Merge `ha_get_operation_status` + `ha_get_bulk_status` (2 → 1, saves 1 tool)

Both are read-only tools in `tools_service.py` checking async device operation status. Both delegate to `device_tools`. The bulk version loops over individual operations and aggregates results.

**Approach:** Make `ha_get_operation_status` accept `operation_id: str | list[str]`. Single string returns single status (current behavior). List returns aggregated results (current bulk behavior).

## Why do you need this?

These are duplicates or near-duplicates that inflate the tool count without adding distinct functionality. Each consolidation follows patterns already established in the codebase — the `str | list` pattern is used by `ha_get_entity`/`ha_set_entity`, and the area+floor tools are already in the same file sharing the same structure.

## Additional context

Per AGENTS.md: "When a tool's functionality is fully covered by another tool, remove the redundant tool rather than deprecating it. Fewer tools reduces cognitive load for AI agents and improves decision-making."

kingpanther13:
## What does this PR do?

Consolidates 2 redundant tools to reduce tool count from 93 → 91, following the directive in AGENTS.md: *"When a tool's functionality is fully covered by another tool, remove the redundant tool rather than deprecating it."*

### 1. Rename entity tools (2 → 1, saves 1 tool)

`ha_rename_entity` now accepts an optional `new_device_name` parameter. If provided, the device is looked up from the entity registry and renamed alongside the entity. `ha_rename_entity_and_device` is removed.

- The `new_entity_name` param from the old combo tool maps to the existing `name` param
- Partial success handling preserved (entity renamed but device rename failed)
- Empty string `new_device_name` normalized to `None` (entity-only rename)
- Registry lookup failure now distinguishable from "entity has no device" via warning field
- Extracted `extract_tool_error_message()` helper replacing brittle `json.loads(str(te))`

### 2. Operation status tools (2 → 1, saves 1 tool)

`ha_get_operation_status` now accepts `operation_id: str | list[str]`. A single string returns single status (existing behavior). A list returns aggregated results (bulk behavior). `ha_get_bulk_status` is removed.

- Added `parse_json_param` coercion for JSON-stringified lists from MCP clients

### Area/floor consolidation — reverted

The original PR consolidated area+floor tools (6 → 3). This was reverted per reviewer feedback — areas and floors remain separate tools.

### Additional changes

- Merged upstream master (post-#806 and #839)
- Regenerated `tools.json` and README tool table via `extract_tools.py`
- Corrected stale tool counts across all docs (now 92+)
- Added assertions to operation status E2E tests

Closes #774

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [x] 🔧 Maintenance/refactor
- [ ] 💥 Breaking change

## Testing

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
   <!-- mcp-name: io.github.homeassistant-ai/ha-mcp -->
 
   <p align="center">
-    <img src="https://img.shields.io/badge/tools-92-blue" alt="90+ Tools">
+    <img src="https://img.shields.io/badge/tools-92-blue" alt="95+ Tools">
     <a href="https://github.com/homeassistant-ai/ha-mcp/releases"><img src="https://img.shields.io/github/v/release/homeassistant-ai/ha-mcp" alt="Release"></a>
     <a href="https://github.com/homeassistant-ai/ha-mcp/actions/workflows/e2e-tests.yml"><img src="https://img.shields.io/github/actions/workflow/status/homeassistant-ai/ha-mcp/e2e-tests.yml?branch=master&label=E2E%20Tests" alt="E2E Tests"></a>
     <a href="LICENSE.md"><img src="https://img.shields.io/github/license/homeassistant-ai/ha-mcp.svg" alt="License"></a>
@@ -239,7 +239,7 @@ This project is licensed under the MIT License - see the [LICENSE](LICENSE) file
 - **[@ryphez](https://github.com/ryphez)** — Codex Desktop UI MCP quick setup guide.
 - **[@Danm72](https://github.com/Danm72)** — Entity registry tools (`ha_set_entity`, `ha_get_entity`) for managing entity properties.
 - **[@Raygooo](https://github.com/Raygooo)** — SOCKS proxy support.
-- **[@cj-elevate](https://github.com/cj-elevate)** — Integration & entity management tools (enable/disable/delete).
+- **[@cj-elevate](https://github.com/cj-elevate)** — Integration & entity management tools (enable/disable/delete); person/zone/tag config store routing.
 - **[@maxperron](https://github.com/maxperron)** — Beta testing.
 - **[@kingbear2](https://github.com/kingbear2)** — Windows UV setup guide.
 - **[@konradwalsh](https://github.com/konradwalsh)** — Financial support via [GitHub Sponsors](https://github.com/sponsors/julienld). Thank you! ☕
@@ -249,6 +249,9 @@ This project is licensed under the MIT License - see the [LICENSE](LICENSE) file
 - **[@saphid](https://github.com/saphid)** — Config entry options flow tools (initial design, #590).
 - **[@adraguidev](https://github.com/adraguidev)** — Fix menu-based config entry flows for group helpers (#647).
 - **[@transportrefer](https://github.com/transportrefer)** — Integration options inspection (`ha_get_integration` schema support, #689).
+- **[@teh-hippo](https://github.com/teh-hippo)** — Fix blueprint import missing save step.
+- **[@smenzer](https://github.com/smenzer)** — Documentation fix.
+- **[@The-Greg-O](https://github.com/The-Greg-O)** — REST API for config entry deletion.
 - **[@restriction](https://github.com/restriction)** — Responsible disclosure: python_transform sandbox missing call target validation.
 
 ---
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,7 @@ classifiers = [
 ]
 
 dependencies = [
-    "fastmcp==3.1.1",
+    "fastmcp==3.2.0",
     "httpx[socks]==0.28.1",
     "pydantic==2.12.5",
     "python-dotenv==1.2.2",
--- a/uv.lock
+++ b/uv.lock
@@ -362,7 +362,7 @@ wheels = [
 
 [[package]]
 name = "fastmcp"
-version = "3.1.1"
+version = "3.2.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "authlib" },
@@ -387,9 +387,9 @@ dependencies = [
     { name = "watchfiles" },
     { name = "websockets" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/25/83/c95d3bf717698a693eccb43e137a32939d2549876e884e246028bff6ecce/fastmcp-3.1.1.tar.gz", hash = "sha256:db184b5391a31199323766a3abf3a8bfbb8010479f77eca84c0e554f18655c48", size = 17347644, upload-time = "2026-03-14T19:12:20.235Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/d0/32/4f1b2cfd7b50db89114949f90158b1dcc2c92a1917b9f57c0ff24e47a2f4/fastmcp-3.2.0.tar.gz", hash = "sha256:d4830b8ffc3592d3d9c76dc0f398904cf41f04910e41a0de38cc1004e0903bef", size = 26318581, upload-time = "2026-03-30T20:25:37.692Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/70/ea/570122de7e24f72138d006f799768e14cc1ccf7fcb22b7750b2bd276c711/fastmcp-3.1.1-py3-none-any.whl", hash = "sha256:8132ba069d89f14566b3266919d6d72e2ec23dd45d8944622dca407e9beda7eb", size = 633754, upload-time = "2026-03-14T19:12:22.736Z" },
+    { url = "https://files.pythonhosted.org/packages/4f/67/684fa2d2de1e7504549d4ca457b4f854ccec3cd3be03bd86b33b599fbf58/fastmcp-3.2.0-py3-none-any.whl", hash = "sha256:e71aba3df16f86f546a4a9e513261d3233bcc92bef0dfa647bac3fa33623f681", size = 705550, upload-time = "2026-03-30T20:25:35.499Z" },
 ]
 
 [[package]]
@@ -436,7 +436,7 @@ dev = [
 [package.metadata]
 requires-dist = [
     { name = "cryptography", specifier = "==46.0.6" },
-    { name = "fastmcp", specifier = "==3.1.1" },
+    { name = "fastmcp", specifier = "==3.2.0" },
     { name = "httpx", extras = ["socks"], specifier = "==0.28.1" },
     { name = "pydantic", specifier = "==2.12.5" },
     { name = "python-dotenv", specifier = "==1.2.2" },


src/ha_mcp/tools/tools_areas.py
```diff
@@ -104,373 +147,354 @@ async def ha_config_set_area(
...
+        Create or update a Home Assistant area (room) or floor.
 
-        Provide name only to create a new area. Provide area_id to update existing.
-        Areas organize entities by physical location for room-based control.
+        For areas: provide name to create, area_id to update.
+        For floors: provide name to create, floor_id to update.
+        Areas organize entities by physical location; floors organize areas into vertical levels.
         """
+        operation = "create"
+        registry = "area" if type == "area" else "floor"
```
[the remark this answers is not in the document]
kingpanther13: Fixed in 3dfab6b — `operation` is now computed before the `try` block based on whether an update ID is provided.

src/ha_mcp/tools/tools_registry.py
```diff
@@ -380,25 +409,138 @@ async def ha_rename_entity(
...
+                    import json as _json
+
+                    try:
+                        error_data = _json.loads(str(te))
+                        device_error = error_data.get("error", {}).get(
+                            "message", str(te)
+                        )
+                    except Exception:
+                        device_error = str(te)
+                    results["device_rename"] = {"success": False, "error": device_error}
```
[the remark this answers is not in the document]
kingpanther13: Fixed in 3dfab6b — moved `import json` to module level and narrowed the except clause to `(json.JSONDecodeError, TypeError)`.

src/ha_mcp/tools/tools_areas.py
```diff
@@ -21,76 +24,116 @@
...
             str | None,
             Field(
-                description="Floor ID to assign this area to (use ha_config_list_floors to find IDs, empty string to remove)",
+                description=(
+                    "When type='area': Floor ID to assign this area to (empty string to remove). "
+                    "When type='floor': Floor ID to update (omit to create new floor)."
+                ),
                 default=None,
             ),
         ] = None,
```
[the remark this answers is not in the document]
kingpanther13: This dual purpose is intentional per the issue #774 specification. The parameter description already documents both uses. Adding a separate `update_id` parameter would increase parameter count rather than simplify the interface. The `ha_config_remove_area` tool uses separate params because both `area_id` and `floor_id` can be required simultaneously (to delete an area vs a floor), whereas in the set tool the `type` parameter disambiguates which registry `floor_id` refers to.

src/ha_mcp/tools/tools_areas.py
```diff
@@ -104,373 +147,355 @@ async def ha_config_set_area(
...
-        Create or update a Home Assistant area (room).
+        Create or update a Home Assistant area (room) or floor.
 
-        Provide name only to create a new area. Provide area_id to update existing.
-        Areas organize entities by physical location for room-based control.
+        For areas: provide name to create, area_id to update.
+        For floors: provide name to create, floor_id to update.
+        Areas organize entities by physical location; floors organize areas into vertical levels.
         """
+        registry = "area" if type == "area" else "floor"
```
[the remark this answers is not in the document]
kingpanther13: Duplicate of earlier comment — fixed in 3dfab6b.

src/ha_mcp/tools/tools_registry.py
```diff
@@ -380,25 +410,136 @@ async def ha_rename_entity(
...
+                except ToolError as te:
+                    try:
+                        error_data = json.loads(str(te))
+                        device_error = error_data.get("error", {}).get(
+                            "message", str(te)
+                        )
+                    except (json.JSONDecodeError, TypeError):
+                        device_error = str(te)
+                    results["device_rename"] = {"success": False, "error": device_error}
+                    return {
```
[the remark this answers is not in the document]
kingpanther13: Import and except clause fixed in 3dfab6b. The `str(te)` JSON parsing pattern is inherited from the original `ha_rename_entity_and_device` implementation — refactoring how `ToolError` payloads are accessed would be a separate concern beyond this consolidation PR.

src/ha_mcp/tools/tools_areas.py
```diff
@@ -21,76 +24,122 @@
...
                 default=None,
             ),
         ] = None,
         floor_id: Annotated[
             str | None,
             Field(
-                description="Floor ID to assign this area to (use ha_config_list_floors to find IDs, empty string to remove)",
+                description=(
+                    "When type='area': Floor ID to assign this area to (empty string to remove). "
+                    "When type='floor': Floor ID to update (omit to create new floor)."
```
sergeykad: `floor_id` means "assign area to this floor" when `type="area"`, but "floor ID to update" when `type="floor"`. A single parameter with two completely different semantic meanings depending on a discriminator is genuinely confusing for an LLM caller. This is the core design issue with the area/floor consolidation.

src/ha_mcp/tools/tools_service.py
```diff
@@ -243,13 +293,51 @@ async def ha_call_service(
...
+        to check multiple operations at once (bulk status).
+
+        The timeout_seconds parameter applies to single-operation checks only.
+        Bulk checks poll each operation individually with a short internal timeout.
+
+        Use this to track operations initiated by ha_bulk_control or ha_call_service.
+        For current entity states, use ha_get_state instead.
+        """
+        try:
+            if isinstance(operation_id, list):
```
sergeykad: MCP clients sometimes send `'["op1", "op2"]'` as a JSON string rather than a native list. Other tools in this codebase (e.g., `ha_bulk_control`) use `parse_json_param` / `parse_string_list_param` to handle this. This tool does not -- a list sent as a JSON string will be treated as a single operation ID and fail. Add `parse_json_param` handling here.

src/ha_mcp/tools/tools_registry.py
```diff
@@ -380,25 +410,141 @@ async def ha_rename_entity(
...
+                )
+                if entity_registry_result.get("success"):
+                    entities = entity_registry_result.get("result", [])
+                    for ent in entities:
+                        if ent.get("entity_id") == entity_id:
+                            device_id = ent.get("device_id")
+                            break
+                else:
+                    logger.warning(
+                        f"Could not query entity registry to find device for {entity_id}. "
```
sergeykad, line 455: When `new_device_name` is provided but the registry query fails, this logs a warning and skips device rename. The response later says "No associated device found" -- indistinguishable from "entity truly has no device". The caller has no way to know the lookup failed vs the entity having no device. Should include a warning field in the response when the registry lookup fails so the LLM can decide whether to retry.

src/ha_mcp/tools/tools_registry.py
```diff
@@ -380,25 +410,141 @@ async def ha_rename_entity(
...
+            # Step 2: If new_device_name was requested, rename the device
+            if new_device_name is None:
+                return entity_rename_result
+
+            results: dict[str, Any] = {
+                "entity_rename": entity_rename_result,
+                "device_rename": None,
+            }
+
+            if device_id and new_device_name:
```
sergeykad, line 477: `new_device_name=""` is falsy, so this condition is False even though `new_device_name is not None` (which caused us to enter the combined response path on line 459). Result: `device_rename` stays `None` and the response says "Device name was not changed" -- ambiguous behavior. Should either treat `""` as `None` (entity-only rename) or validate early with a clear error.

src/ha_mcp/tools/tools_registry.py
```diff
@@ -380,25 +410,141 @@ async def ha_rename_entity(
...
+                try:
+                    device_rename_result = await _update_device_internal(
+                        device_id=device_id,
+                        name=new_device_name,
+                    )
+                except ToolError as te:
+                    try:
+                        error_data = json.loads(str(te))
```
sergeykad: Parsing `ToolError`'s string representation as JSON to extract error data is brittle. If `ToolError.__str__` format ever changes, this breaks silently. This was copied from the old `ha_rename_entity_and_device` (which used `import json as _json` inline) -- worth improving rather than carrying forward. Consider accessing `ToolError` attributes directly or using a shared helper.

tests/src/e2e/workflows/device_control/test_operation_status.py
```diff
@@ -0,0 +1,107 @@
...
+from ...utilities.assertions import safe_call_tool
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.mark.device
+class TestOperationStatusConsolidation:
+    """Test consolidated ha_get_operation_status tool."""
+
+    async def test_single_operation_id_invalid(self, mcp_client):
```
sergeykad, line 21: All four tests in this file (`test_single_operation_id_invalid`, `test_list_operation_ids_empty`, `test_list_operation_ids_invalid`, `test_single_vs_list_different_dispatch`) call `safe_call_tool` and log results, but have zero `assert` statements. They verify non-crashing but nothing about correctness -- effectively dead tests. At minimum, assert on the response structure (`success`, `error` fields) and the dispatch behavior (single vs list returns different shapes).

src/ha_mcp/tools/tools_areas.py
kingpanther13: Fixed in fc683ba — renamed tools to `_or_floor` suffix for discoverability, kept `area_id` and `floor_id` as separate params (matching the remove tool's pattern), and clarified the `floor_id` description.

src/ha_mcp/tools/tools_service.py
kingpanther13: Fixed in fc683ba — added `parse_json_param` coercion with `ValueError` fallback for plain string IDs.

src/ha_mcp/tools/tools_registry.py
kingpanther13: Fixed in fc683ba — added `registry_lookup_failed` flag. Response now includes a warning field and distinct message when the registry query itself fails vs entity genuinely having no device.

src/ha_mcp/tools/tools_registry.py
kingpanther13: Fixed in fc683ba — empty/whitespace `new_device_name` is now normalized to `None` early, so it takes the entity-only rename path cleanly.

src/ha_mcp/tools/tools_registry.py
kingpanther13: Fixed in fc683ba — extracted `extract_tool_error_message()` helper in `helpers.py` that pairs with `raise_tool_error()`. Removed inline `json` import from `tools_registry.py`.

tests/src/e2e/workflows/device_control/test_operation_status.py
kingpanther13: Fixed in fc683ba — added assertions on response structure (`success`/`error` fields) to all 4 tests, plus a new `test_json_string_list_coercion` test.

kingpanther13:
@sergeykad — I disagree with most of these points but I'll make the change anyway since it's not worth going back and forth on.

For the record:

> Forces the LLM to understand a meta-parameter before reasoning about the actual fields

The `type` parameter defaults to `"area"`, so for the common case (area ops) the LLM doesn't need to think about it. Only floor operations require setting `type="floor"`. Discriminator patterns are standard API design.

> Makes `floor_id` mean two unrelated things depending on `type`

They're not unrelated — `floor_id` consistently means "a floor identifier." When `type="area"` it's "which floor to assign to," when `type="floor"` it's "which floor to update." The concept is the same (a floor), the operation context differs. This was already addressed with the updated description.

> Doesn't save tokens — the combined description has to explain both flows anyway

This is incorrect. Each separate tool carries its own full parameter schema, annotations, title, and description in the MCP tools/list response. One combined tool with a few conditional params is smaller than two full tool schemas. Fewer tools in the listing also means less processing overhead per request.

> Creates a tool that doesn't map to any real HA concept

Areas and floors are in the same spatial-organization domain and HA's own UI manages them on the same page. They share the same WebSocket registry API pattern (`config/{area,floor}_registry/{list,create,update,delete}`), same error handling, and most parameters (name, icon, aliases). The only differences are `level` vs `floor_id`/`picture`.

> The other consolidations make sense because they share domain objects and API shapes. Areas and floors don't

Areas and floors share more than the rename or operation-status tools do — same file, same API structure, same shared params. But this isn't a hill worth dying on.

That said — reverting the area/floor consolidation now to keep things moving. The rename and operation-status consolidations stay (2 tools saved).

PR #814, opened 2026-03-24
fix: add exact_match to all search tools, badge search, and dashboard deep search

### LINKED ISSUE #801: ha_deep_search: fuzzy matching false positives + dashboards not searched → unreliable for post-rename verification

After renaming several entities (Zigbee2MQTT devices from IEEE address-based IDs to human-readable names), the agent used `ha_deep_search` to verify that no automations or scripts still referenced the old entity IDs. The search returned 5-8 matches for every query, but when the agent fetched the full configs with `include_config: true`, none of the returned automations actually contained the queried string — they were all fuzzy/partial matches. The agent had to manually scan all configs to determine there were no real references, which defeated the purpose of the search. Since there is no way to disable fuzzy matching, the tool cannot be reliably used for exact reference checking.

A second, more severe issue compounds this: ha_deep_search does NOT search dashboard configurations at all. After renaming entities, the agent ran ha_deep_search and got a clean result, creating false confidence that all references were updated. In reality, 6 dashboard cards (across the bedroom and mobile dashboards) still referenced the old entity IDs and were silently broken — only discovered by the user noticing the dashboard was no longer showing temperature. The combination of false positives from fuzzy matching AND false negatives from missing dashboard coverage makes ha_deep_search unreliable as a post-rename verification tool. The tool description should either (a) add dashboard search support, or (b) prominently warn that dashboards are not searched.

A third gap: ha_dashboard_find_card does NOT search view badges. The mobile dashboard had a badge (the chip row at the top of the view) still referencing the old entity ID. ha_dashboard_find_card returned no matches for that entity, yet the badge was broken and visible to the user. Badges are part of the view config under views[n].badges and are a distinct location from cards — they need to be included in the search scope.

Root cause analysis — this is a three-way failure:

MCP tool gaps (fixable in ha-mcp):
- ha_deep_search has no exact match mode, causing false positive noise
- - ha_deep_search does not cover dashboards or badges at all
- - - ha_dashboard_find_card misses view badges (views[n].badges)
Agent failure (fixable in skill/system prompt):
- The agent should know that entity renames affect dashboards and proactively check them after every rename, regardless of what ha_deep_search reports. This is not a tool problem — it is the agent not owning the full impact of a rename operation.
Their dangerous interaction:
- ha_deep_search returning a clean result created false confidence that actively discouraged further checking. If the tool had warned "note: dashboards not searched", the agent would have continued. The tool's silence looked like completeness, which is worse than no verification tool at all.
Both fixes are needed independently: the MCP fix (add dashboard/badge coverage + exact match) would prevent false confidence even if the agent misbehaves. The skill fix (agent always checks dashboards post-rename) would catch breakage even with the current incomplete tools.

```markdown
a=== ha-mcp Bug Report Info ===

ha-mcp Version: 7.1.0
Installation Method: pypi
Platform: Windows 11 (AMD64)
Python Version: 3.14.3
Home Assistant Version: 2026.3.2
Connection Status: Connected
Entity Count: 245

=== Recent Tool Calls ===
  2026-03-21T15:18:01 | ha_deep_search | OK | 927ms
  2026-03-21T15:17:53 | ha_deep_search | OK | 361ms
  2026-03-21T15:13:41 | ha_rename_entity | OK | 37ms
  2026-03-21T15:12:44 | ha_deep_search | OK | 274ms  (include_config: true)
  2026-03-21T15:12:20 | ha_deep_search | OK | 295ms

=== Scenario ===
User asked agent to rename Zigbee2MQTT sensor entities from IEEE address-based IDs
(e.g. sensor.0x187a3efffe304339_temperature) to human-readable names
(e.g. sensor.bedroom_temperature).

After renaming, agent called ha_deep_search twice to verify no references remained.
Both calls returned 5-8 matches but none actually contained the queried string
when configs were fetched with include_config: true — all false positives from fuzzy matching.

ha_dashboard_find_card was then used to check dashboards — found 6 broken cards.
Fixed those. However ha_dashboard_find_card returned 0 matches for
sensor.0x187a3efffe304339_temperature on the mobile dashboard, yet a badge
(views[0].badges[1]) was still broken — badges are not searched by ha_dashboard_find_card.

=== Issues Found ===
1. ha_deep_search: no exact match mode — fuzzy matching causes false positives,
   making it unreliable for post-rename reference verification.
2. ha_deep_search: does not search dashboard configurations at all —
   false negatives for broken dashboard cards.
3. ha_dashboard_find_card: does not search view badges (views[n].badges) —
   badge references to renamed entities are silently missed.
4. Agent behavior: the agent did NOT proactively check dashboards after renaming entities.
   The broken dashboard cards and badge were only discovered because the user explicitly
   reported them. The correct post-rename workflow should always include scanning all
   dashboards with ha_dashboard_find_card for each renamed entity — without waiting for
   the user to notice something is broken. The agent should own the full impact of a rename,
   not just the automation/script side.
```

kingpanther13:
## What does this PR do?

Addresses all three tool gaps and the broader fuzzy-matching discussion from #801, plus community feedback:

### 1. `ha_dashboard_find_card` now searches view badges and header cards
- **Badges** (`views[n].badges`): the chip row at the top of a view. Previously, entity references in badges were silently missed during rename operations.
- **Header cards** (`views[n].header.card`): sections-view header cards (typically Markdown with Jinja2 templates) that can contain entity references. Reported by @Patch76.

Both are now included in search results with appropriate `jq_path` for targeted updates.

### 2. `exact_match` parameter added to all search tools (default: `True`)
Per maintainer discussion — fuzzy matching caused more harm than good for the common case of searching for known values. The following tools now have `exact_match=True` by default:

- **`ha_deep_search`** — uses exact substring matching in `_search_in_dict` and name scoring, eliminating the false positives reported in the issue
- **`ha_search_entities`** — routes directly to substring matching, skipping the fuzzy search path entirely
- **`ha_get_integration`** — skips fuzzy scoring when filtering by query

All three tools accept `exact_match=False` to re-enable fuzzy matching when searching with approximate terms or user typos.

### 3. `ha_deep_search` gains `dashboard` search type
Pass `search_types=["dashboard"]` (or include alongside other types) to search all storage-mode dashboard configurations. This closes the gap where `ha_deep_search` returned a clean result while dashboards still had broken entity references.

### Design decisions

- **`exact_match` defaults to `True`** per maintainer consensus in #801 — LLMs can opt into fuzzy with `exact_match=False` when queries may contain typos
- **Dashboard search is opt-in** (`search_types` default unchanged) — adding dashboards to the default would increase latency and context usage for every `ha_deep_search` call
- **Badge results use `card_type: "badge"`** with `badge_index` (and `card_index: None`) so callers can distinguish them from card results
- **Scoring logic deduplicated** into `_score_deep_match` helper per Gemini Code Assist review

Closes #801

## Type of change
- [x] 🐛 Bug fix
- [x] ✨ New feature

## Testing

## Checklist


src/ha_mcp/tools/smart_search.py
```diff
@@ -767,28 +787,47 @@ async def deep_search(
...
+                            if query_lower in entity_id.lower()
+                            or query_lower in friendly_name.lower()
+                            else 0
+                        )
+                        total_score = max(name_exact, config_match_score)
+                        threshold = 100
+                    else:
+                        total_score = max(name_score, config_match_score)
+                        threshold = self.settings.fuzzy_threshold
+
```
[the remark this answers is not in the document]
kingpanther13: Addressed in 87ef420 — extracted `_score_deep_match` helper to deduplicate the scoring logic across automations, scripts, and helpers.

src/ha_mcp/tools/smart_search.py
```diff
@@ -996,6 +1044,97 @@ async def fetch_helper_list(helper_type: str) -> list[dict[str, Any]]:
...
+                        return_exceptions=True,
+                    )
+                    for dash_result in dash_results:
+                        if isinstance(dash_result, list):
+                            results["dashboards"].extend(dash_result)
+                        elif isinstance(dash_result, Exception):
+                            logger.debug(f"Dashboard search failed: {dash_result}")
+
+                except Exception as e:
+                    logger.debug(f"Dashboard search error: {e}")
```
sergeykad: This silently swallows the error and returns `success=True` with an empty `dashboards` list — the same false-confidence problem issue #801 was filed to fix. If the dashboard listing fails, the LLM has no way to know the search was incomplete and will treat zero results as "no broken references".

This should raise `ToolError`, not recover silently. Follow the same pattern as the rest of `deep_search`: let `exception_to_structured_error` raise so the caller knows the operation failed.

src/ha_mcp/tools/tools_search.py
```diff
@@ -183,6 +197,9 @@ async def ha_search_entities(
             coerce_bool_param(group_by_domain, "group_by_domain", default=False)
             or False
         )
+        exact_match_bool = (
+            coerce_bool_param(exact_match, "exact_match", default=True) or False
```
sergeykad: `coerce_bool_param(..., default=True) or False` undermines the `True` default. If `coerce_bool_param` returns `None` for any reason, `None or False` evaluates to `False` — silently enabling fuzzy mode when exact-match was expected. The `or False` suffix is safe when `default=False` (used that way elsewhere), but here it contradicts the documented default.

Drop the `or False`. The same issue applies to the `ha_deep_search` call below and in `tools_integrations.py`.

src/ha_mcp/tools/tools_integrations.py
```diff
@@ -96,8 +114,15 @@ async def ha_get_integration(
...
-            include_opts = coerce_bool_param(include_options, "include_options", default=False)
-            include_schema_bool = coerce_bool_param(include_schema, "include_schema", default=False)
+            include_opts = coerce_bool_param(
+                include_options, "include_options", default=False
+            )
+            include_schema_bool = coerce_bool_param(
+                include_schema, "include_schema", default=False
+            )
+            exact_match_bool = (
+                coerce_bool_param(exact_match, "exact_match", default=True) or False
```
sergeykad: Same `coerce_bool_param(..., default=True) or False` problem as in `tools_search.py`: the `or False` suffix inverts the `True` default when `coerce_bool_param` returns `None`. Drop the `or False`.

src/ha_mcp/tools/tools_integrations.py
```diff
@@ -202,24 +239,22 @@ async def ha_get_integration(
...
-
-                # Perform fuzzy search with both exact and fuzzy matching
                 matches = []
                 query_lower = query.strip().lower()
 
                 for entry in formatted_entries:
-                    domain_lower = entry['domain'].lower()
-                    title_lower = entry['title'].lower()
+                    domain_lower = entry["domain"].lower()
+                    title_lower = entry["title"].lower()
```
sergeykad: `formatted_entries` was built with `entry.get("domain")` and `entry.get("title")`, so either field can be `None`. Calling `.lower()` on `None` raises `AttributeError`, crashing the tool for any config entry with a missing domain or title (custom integrations, broken entries, or HA version differences).

Fix: `(entry.get("domain") or "").lower()` and `(entry.get("title") or "").lower()`.

src/ha_mcp/tools/tools_config_dashboards.py
```diff
@@ -128,6 +157,46 @@ def _find_cards_in_config(
         if not isinstance(view, dict):
             continue
 
+        # Search view-level badges (only when filtering by entity_id,
+        # since badges don't have card_type or heading)
+        if entity_id is not None and card_type is None and heading is None:
```
sergeykad: Badge search is skipped whenever `card_type` or `heading` is set. If a caller passes `card_type="badge"` — which the updated docstring shows as a valid example — the badge path never runs and zero results are returned with no explanation. The implementation and the docstring contradict each other.

Either support `card_type="badge"` as a first-class trigger for the badge path, or remove `"badge"` from the docstring examples.

tests/src/e2e/tools/test_search_entities.py
```diff
@@ -353,24 +370,36 @@ async def test_search_entities_pagination_metadata(mcp_client):
...
+            f"No pagination needed: {results_count} of {total_matches} shown, has_more=False"
+        )
 
     # total_matches should always be >= results_count
-    assert total_matches >= results_count, \
+    assert total_matches >= results_count, (
         f"total_matches ({total_matches}) should be >= results count ({results_count})"
+    )
 
     logger.info("Pagination metadata test passed")
```
sergeykad, line 405: All 51 additions in this file are formatting changes and adding `exact_match=False` to two existing tests to preserve their pre-PR semantics. None of the three issue #801 gaps have new E2E tests:

- `exact_match=True` path in `ha_search_entities` and `ha_deep_search`: not tested
- `search_types=["dashboard"]` in `ha_deep_search`: not tested
- Badge search in `ha_dashboard_find_card`: not tested
- Header card search: not tested

The PR description marks "All automated tests pass" unchecked, but given these are the core features being added, tests should be present before merge.

src/ha_mcp/tools/smart_search.py
kingpanther13: Fixed in 2547b4b — dashboard search now does `logger.error` + `raise` instead of silently logging at debug level. The outer `except Exception` in `deep_search` will catch it and raise `ToolError` via `exception_to_structured_error`.

src/ha_mcp/tools/tools_search.py
kingpanther13: Fixed in 2547b4b — dropped `or False` from all three `coerce_bool_param(..., default=True)` calls (tools_search.py x2, tools_integrations.py x1).

src/ha_mcp/tools/tools_integrations.py
kingpanther13: Fixed in 2547b4b — same fix as above, dropped the `or False`.

src/ha_mcp/tools/tools_integrations.py
kingpanther13: Fixed in 2547b4b — changed to `(entry.get("domain") or "").lower()` and `(entry.get("title") or "").lower()`. Pre-existing bug but good catch.

src/ha_mcp/tools/tools_config_dashboards.py
kingpanther13: Fixed in 2547b4b — badge search now triggers when `card_type is None or card_type == "badge"`, so `card_type="badge"` works as a first-class filter.

tests/src/e2e/tools/test_search_entities.py
kingpanther13: Added in 2547b4b:

**Unit tests** (`test_find_cards_badges_headers.py`):
- String/dict/typed badge search, wildcard matching, card_type="badge" filter
- Header card search by entity and by type, empty/missing header handling
- Strategy dashboard returns empty

**E2E tests** (`test_deep_search_exact_match.py`):
- exact_match=True default with precise filtering
- exact_match=False fuzzy opt-in
- search_types=["dashboard"] finds dashboard content
- Dashboard search absent from default results
- ha_search_entities exact_match default

PR #827, opened 2026-03-27
feat: add managed YAML config editing tool (ha_config_set_yaml)

### LINKED ISSUE #581: [FEATURE] Add managed YAML config editing for features without API equivalents

A new tool (e.g. `ha_edit_yaml_config`) that inserts, updates, or removes configuration blocks in `configuration.yaml` and package files — for Home Assistant features that exist only in YAML and have no REST/WebSocket API.

The MCP cannot create any of the following:

- Template sensors and binary sensors (`template:`)
- Package files (`packages/*.yaml`)
- Platform-based sensors (`sensor:`, `binary_sensor:` with `platform:` entries)
- YAML-defined input helpers (`input_boolean`, `input_number`, etc.)

The tool should perform structured, validated edits — not raw file writes:

```python
ha_edit_yaml_config(
    file: str = "configuration.yaml",       # or a package file path
    action: str = "add" | "replace" | "remove",
    yaml_path: str = "template",            # top-level key or dotted path
    content: str = "...",                    # YAML block to insert/replace
    backup: bool = True                     # auto-backup before edit
)
```

Safeguards:
- Back up the file before every edit (e.g. to `www/backups/`)
- Validate YAML syntax before writing
- Call `homeassistant.check_config` after writing to verify HA can parse the result
- Restrict edits to a whitelist of top-level keys (block `homeassistant:`, `http:`, `recorder:`, etc.)
- Reject path traversal (already enforced by the filesystem tools)

### Why do you need this?

When an LLM needs to create a template sensor or any other YAML-only config, it must ask the user to copy-paste YAML into config files manually. This breaks the conversational workflow.

Example: a user asks "create a template sensor that shows my vacuum's last clean time in a human-readable format." The LLM generates the YAML but then says "please add this to your `configuration.yaml` and restart HA." With this tool, the LLM would insert the block directly and call `homeassistant.reload_all` — no manual steps required.

Automations and scripts already work end-to-end through their API-based tools (`ha_config_set_automation`, `ha_config_set_script`). YAML-only features lack equivalent coverage.

### Additional context

- Issue #194 designed the filesystem tools with `configuration.yaml` as read-only — a sound decision for raw file writes. A structured, validated edit tool would bridge the gap without sacrificing safety.
- Discussion #338 proposed `old_str`/`new_str` partial edits for dashboards; the same pattern could apply here.
- The `ha_mcp_tools` custom component already runs inside HA with config directory access and could implement the validation layer.
- Home Assistant's `config/core` API exposes `check_config`, enabling post-edit verification.

kingpanther13:
## What does this PR do?

Adds `ha_config_set_yaml` — a managed YAML config editing tool for Home Assistant features that exist only in YAML and have no REST/WebSocket API equivalent (e.g., `command_line` sensors, platform-based `mqtt` sensors, `rest` sensors in packages).

The tool performs structured, validated edits to `configuration.yaml` and `packages/*.yaml` files with strong safeguards:

- **Key allowlist**: Only YAML-only keys can be edited (`template`, `sensor`, `binary_sensor`, `command_line`, `rest`, `mqtt`, `shell_command`, `switch`, `light`, `fan`, `cover`, `climate`, `notify`, `group`, `utility_meter`). Keys manageable via `ha_config_set_helper` (input_*, counter, timer, schedule) are excluded.
- **Backup before every edit**: Written from in-memory content to `www/yaml_backups/` with timestamps
- **YAML validation**: Content is validated with `yaml.safe_load` before writing; null/empty content is rejected
- **Atomic writes**: Writes to temp file then `os.replace()` to prevent corruption on disk-full/IO errors
- **Path traversal blocking**: Reuses existing filesystem security patterns
- **Config check after write**: Calls `homeassistant.check_config` to verify HA can load the edited file
- **Type mismatch protection**: `add` action errors on incompatible types instead of silently overwriting
- **Restrictive tool description**: Explicitly steers agents toward UI/API alternatives first (Template Helpers, `ha_config_set_automation`, etc.)
- **Single feature gate**: Requires `ENABLE_YAML_CONFIG_EDITING=true` (no filesystem flag dependency)
- **Pinned tool**: Always visible in `list_tools()` when categorized search transform is active

### Tool signature
```python
ha_config_set_yaml(
    yaml_path: str,          # top-level key (e.g., "template", "sensor")
    action: str,             # "add" | "replace" | "remove"
    content: str | None,     # YAML block (required for add/replace)
    file: str = "configuration.yaml",  # or packages/*.yaml
    backup: bool = True,
)
```

### Changes
- **New**: `src/ha_mcp/tools/tools_yaml_config.py` — MCP tool module
- **New**: `tests/src/e2e/workflows/filesystem/test_yaml_config.py` — E2E tests
- **Edit**: `src/ha_mcp/config.py` — `ENABLE_YAML_CONFIG_EDITING` toggle
- **Edit**: `custom_components/ha_mcp_tools/__init__.py` — `edit_yaml_config` service handler
- **Edit**: `custom_components/ha_mcp_tools/const.py` — allowed key constants (trimmed to YAML-only)
- **Edit**: `custom_components/ha_mcp_tools/services.yaml` — service schema
- **Edit**: `custom_components/ha_mcp_tools/manifest.json` — version bump (0.3.0 → 0.4.0)
- **Edit**: `src/ha_mcp/transforms/categorized_search.py` — pinned in `DEFAULT_PINNED_TOOLS`
- **Edit**: `src/ha_mcp/server.py` — BM25 search keywords
- **Edit**: `.env.example` — documented new env var

### Known limitations
- YAML comments are not preserved after editing (PyYAML round-trip). Backup retains the original.
- Files with `!include` or `!secret` directives will fail to parse (standard YAML only). Package files are the better target for managed edits.

## Type of change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 💥 Breaking change

## Testing
- [x] E2E tests added for security boundaries, validation, CRUD, backup, and config check

## Checklist

Closes #581

### DIFF
--- a/custom_components/ha_mcp_tools/__init__.py
+++ b/custom_components/ha_mcp_tools/__init__.py
@@ -31,6 +31,8 @@ from .const import (
     ALLOWED_YAML_CONFIG_FILES,
     ALLOWED_YAML_KEYS,
     DOMAIN,
+    YAML_KEY_DEFAULT_POST_ACTION,
+    YAML_KEY_POST_ACTIONS,
 )
 
 _LOGGER = logging.getLogger(__name__)
@@ -690,6 +692,12 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
             }
             if backup_path_str:
                 result["backup_path"] = backup_path_str
+
+            # Surface the post-edit action required to activate the change
+            post_info = YAML_KEY_POST_ACTIONS.get(
+                yaml_path, YAML_KEY_DEFAULT_POST_ACTION
+            )
+            result.update(post_info)
 
             # Run HA config check to verify the file is loadable
             try:
--- a/custom_components/ha_mcp_tools/const.py
+++ b/custom_components/ha_mcp_tools/const.py
@@ -33,3 +33,24 @@ ALLOWED_YAML_KEYS = frozenset(
         "utility_meter",
     }
 )
+
+# Post-edit action required for each YAML key.
+# Only template, mqtt, and group have YAML reload services in HA core
+# (verified against homeassistant/components/homeassistant/__init__.py,
+# async_handle_reload_all). All others require a full HA restart.
+YAML_KEY_POST_ACTIONS: dict[str, dict[str, str]] = {
+    "template": {
+        "post_action": "reload_available",
+        "reload_service": "homeassistant.reload_custom_templates",
+    },
+    "mqtt": {
+        "post_action": "reload_available",
+        "reload_service": "mqtt.reload",
+    },
+    "group": {
+        "post_action": "reload_available",
+        "reload_service": "group.reload",
+    },
+}
+# Default for keys not in YAML_KEY_POST_ACTIONS:
+YAML_KEY_DEFAULT_POST_ACTION = {"post_action": "restart_required"}
--- a/src/ha_mcp/tools/tools_yaml_config.py
+++ b/src/ha_mcp/tools/tools_yaml_config.py
@@ -119,6 +119,10 @@ def register_yaml_config_tools(mcp: Any, client: Any, **kwargs: Any) -> None:
         Safeguards: file backup, YAML validation, top-level key whitelist,
         path traversal blocking, post-edit config check.
 
+        IMPORTANT: Check 'post_action' in the response. Most keys require
+        a full HA restart ('restart_required'). Only template, mqtt, and
+        group support reload ('reload_available' with 'reload_service').
+
         Note: YAML comments are not preserved after editing. The backup
         retains the original file with comments intact.
         """
--- a/tests/src/e2e/workflows/filesystem/test_yaml_config.py
+++ b/tests/src/e2e/workflows/filesystem/test_yaml_config.py
@@ -5,7 +5,7 @@ This test suite validates:
 - Security boundaries: path traversal, file allowlist, key allowlist
 - CRUD operations: add, replace, remove actions
 - Validation: null content rejection, type mismatch errors
-- Safeguards: backup creation, config check integration
+- Safeguards: backup creation, config check integration, post-edit action hints
 - Feature flag behavior (disabled by default)
 
 These tests require:
@@ -554,3 +554,60 @@ class TestYamlConfigSafeguards:
                 f"Config check result should be in response: {data}"
             )
             logger.info(f"Config check result: {inner.get('config_check')}")
+
+    async def test_post_action_reload_for_template(self, mcp_client_with_yaml_config):
+        """Template key should return post_action=reload_available."""
+        service_check = await _check_service_available(mcp_client_with_yaml_config)
+        _skip_if_unavailable(service_check, "Post action reload")
+
+        content = "- sensor:\n    - name: Post Action Test\n      state: 'ok'"
+
+        async with MCPAssertions(mcp_client_with_yaml_config) as mcp:
+            data = await mcp.call_tool_success(
+                TOOL_NAME,
+                {
+                    "yaml_path": "template",
+                    "action": "add",
+                    "content": content,
+                    "file": "packages/_e2e_test_post_action_reload.yaml",
+                    "backup": False,
+                },
+            )
+            inner = data.get("data", data)
+            assert inner.get("success") is True, f"Add should succeed: {data}"
+            assert inner.get("post_action") == "reload_available", (
+                f"template should have post_action=reload_available: {data}"
+            )
+            assert "reload_service" in inner, (
+                f"reload_service should be present for reloadable keys: {data}"
+            )
+            logger.info(
+                f"post_action={inner.get('post_action')}, "
+                f"reload_service={inner.get('reload_service')}"
+            )
+
+    async def test_post_action_restart_for_shell_command(self, mcp_client_with_yaml_config):
+        """shell_command key should return post_action=restart_required."""
+        service_check = await _check_service_available(mcp_client_with_yaml_config)
+        _skip_if_unavailable(service_check, "Post action restart")
+
+        async with MCPAssertions(mcp_client_with_yaml_config) as mcp:
+            data = await mcp.call_tool_success(
+                TOOL_NAME,
+                {
+                    "yaml_path": "shell_command",
+                    "action": "add",
+                    "content": "test_cmd: echo hello",
+                    "file": "packages/_e2e_test_post_action_restart.yaml",
+                    "backup": False,
+                },
+            )
+            inner = data.get("data", data)
+            assert inner.get("success") is True, f"Add should succeed: {data}"
+            assert inner.get("post_action") == "restart_required", (
+                f"shell_command should have post_action=restart_required: {data}"
+            )
+            assert "reload_service" not in inner, (
+                f"reload_service should NOT be present for restart-only keys: {data}"
+            )
+            logger.info(f"post_action={inner.get('post_action')}")


custom_components/ha_mcp_tools/const.py
```diff
@@ -5,3 +5,59 @@
...
+        "recorder",
+        "logger",
+        "history",
+        "logbook",
+        "system_log",
+        "automation",
+        "script",
+        "scene",
+    }
+)
```
sergeykad, line 35: **1. Redundant blocklist.** `BLOCKED_YAML_KEYS` is unnecessary — anything not in `ALLOWED_YAML_KEYS` is already rejected by the subsequent check in the service handler. The blocklist only changes the error message text. A single allowlist with a clear error message is cleaner.

Also, `ALLOWED_YAML_KEYS` includes keys that have full UI/API paths (`input_boolean`, `input_number`, `input_text`, `input_select`, `input_datetime`, `input_button`, `counter`, `timer`, `schedule`). These are all manageable via `ha_config_set_helper`. Including them here creates a bypass around the safer API-based tools, contradicting the tool's own description ("Only use when NO UI or API alternative exists").

src/ha_mcp/tools/tools_yaml_config.py
```diff
@@ -0,0 +1,210 @@
...
+
+
+def register_yaml_config_tools(mcp: Any, client: Any, **kwargs: Any) -> None:
+    """Register YAML config editing tools with the MCP server.
+
+    Requires both HAMCP_ENABLE_FILESYSTEM_TOOLS=true (filesystem access)
+    and ENABLE_YAML_CONFIG_EDITING=true (dedicated toggle).
+    """
+    if not is_filesystem_tools_enabled():
+        logger.debug(
```
sergeykad: **2. Dual feature flag is counterproductive.** Requiring `HAMCP_ENABLE_FILESYSTEM_TOOLS=true` forces users to also enable raw `read_file`/`write_file`/`delete_file` just to use this managed, validated tool. That defeats the purpose — a user wanting only validated YAML editing gets exposed to unrestricted filesystem access.

More fundamentally — if the safeguards (allowlist, validation, backups, path restrictions) are sufficient, this tool needs no feature flag at all. It's no more dangerous than `ha_config_set_automation`, which writes automations without any flag. If the safeguards aren't sufficient, fix them rather than pushing the safety decision to the user.

src/ha_mcp/tools/tools_yaml_config.py
```diff
@@ -0,0 +1,210 @@
...
+                service_data["content"] = content
+
+            # Call the custom component service
+            result = await client.call_service(
+                MCP_TOOLS_DOMAIN,
+                "edit_yaml_config",
+                service_data,
+                return_response=True,
+            )
+
```
sergeykad, line 180: **3. Service errors returned as successful MCP responses.** When `call_service` returns `{"success": False, ...}`, the tool returns it as a normal response without raising `ToolError`. Per MCP spec and CLAUDE.md, tool-level failures must set `isError=true`. LLM clients will interpret this as a successful call.

Also, `add_timezone_metadata` is called on both success and error responses. CLAUDE.md: "Never add `add_timezone_metadata` to errors."

Suggested fix:
```python
if isinstance(result, dict):
    if not result.get("success", True):
        raise_tool_error(result)
    return await add_timezone_metadata(client, result)
    ErrorCode.SERVICE_CALL_FAILED,
    "Unexpected response format from YAML config service",
    context={"file": file},
))
```

custom_components/ha_mcp_tools/__init__.py
```diff
@@ -486,7 +500,213 @@ async def handle_delete_file(call: ServiceCall) -> ServiceResponse:
...
+                    "success": False,
+                    "error": f"'content' is required for action '{action}'.",
+                }
+            try:
+                parsed_content = yaml.safe_load(content)
+            except yaml.YAMLError as err:
+                return {
+                    "success": False,
+                    "error": f"Invalid YAML content: {err}",
+                }
```
sergeykad, line 556: **4. Two silent data corruption paths:**

(a) `yaml.safe_load(content)` returns `None` for empty/null content (`""`, `"---"`, `"~"`, `"null"`). The tool proceeds to write `key: null` to configuration.yaml with no error. HA will fail to load that key on restart. Add a `None` check after `safe_load`.

(b) The `add` action's `else` branch (line 575) silently replaces on type mismatch (e.g., existing list replaced with dict). User loses config data with no warning. Should either error or return a warning — the user can use `replace` explicitly if that's their intent.

custom_components/ha_mcp_tools/__init__.py
```diff
@@ -486,7 +500,213 @@ async def handle_delete_file(call: ServiceCall) -> ServiceResponse:
...
+                    backup_dir.mkdir, parents=True, exist_ok=True
+                )
+                timestamp = datetime.now().strftime("%Y%m%d_%H%M%S")
+                safe_name = normalized.replace(os.sep, "_")
+                backup_file = backup_dir / f"{safe_name}.{timestamp}.bak"
+                await hass.async_add_executor_job(
+                    shutil.copy2, str(target_file), str(backup_file)
+                )
+                backup_path_str = str(backup_file.relative_to(config_dir))
+                _LOGGER.info("Backup created: %s", backup_path_str)
```
sergeykad, line 609: **5. Non-atomic write risks corrupting configuration.yaml.** `Path.write_text` truncates then writes. If it fails midway (disk full, I/O error), the file is left truncated/corrupt. Should write to a temp file first, then `os.replace` for atomicity.

Also, the backup uses `shutil.copy2` from disk (line 558) instead of the already-read `raw_content`. If the file changes between read and backup, the backup is stale. Write `raw_content` directly to the backup file instead.

custom_components/ha_mcp_tools/__init__.py
```diff
@@ -486,7 +500,213 @@ async def handle_delete_file(call: ServiceCall) -> ServiceResponse:
...
+
+            # Perform the action
+            if action == "add":
+                if yaml_path in data:
+                    existing = data[yaml_path]
+                    # Merge: list extends list, dict merges dict, else replace
+                    if isinstance(existing, list) and isinstance(parsed_content, list):
+                        data[yaml_path] = existing + parsed_content
+                    elif isinstance(existing, dict) and isinstance(
+                        parsed_content, dict
```
sergeykad, line 619: **6. Missing `check_config` call.** Issue #581 explicitly requests calling `homeassistant.check_config` after writing to verify HA can parse the result. This is especially important because `yaml.safe_load` can't validate HA-specific features (custom component configs, integration schemas). A bad edit could break HA on next restart with no immediate feedback.

custom_components/ha_mcp_tools/const.py
kingpanther13: Fixed in aeb8ebc. Removed `BLOCKED_YAML_KEYS` entirely — the allowlist alone is sufficient. Also removed `input_boolean`, `input_number`, `input_text`, `input_select`, `input_datetime`, `input_button`, `counter`, `timer`, and `schedule` from `ALLOWED_YAML_KEYS` since they're all manageable via `ha_config_set_helper`.

src/ha_mcp/tools/tools_yaml_config.py
kingpanther13: Fixed in aeb8ebc. Removed the `is_filesystem_tools_enabled()` gate entirely. The tool now only requires `ENABLE_YAML_CONFIG_EDITING=true`, consistent with `ha_config_set_automation` having no feature flag. Updated docstrings and `.env.example` accordingly.

src/ha_mcp/tools/tools_yaml_config.py
kingpanther13: Fixed in aeb8ebc. Now checks `result.get("success", True)` — if False, calls `raise_tool_error(result)`. Only calls `add_timezone_metadata` on success. Unexpected response formats also raise `ToolError` with `ErrorCode.SERVICE_CALL_FAILED`.

custom_components/ha_mcp_tools/__init__.py
kingpanther13: Fixed in aeb8ebc.

(a) Added explicit `None` check after `yaml.safe_load(content)` — rejects with "Content parsed as null/empty. Provide non-empty YAML."

(b) The `add` action's type-mismatch branch now returns an error with the existing/new types and suggests using `action='replace'` explicitly instead of silently overwriting.

custom_components/ha_mcp_tools/__init__.py
kingpanther13: Fixed in aeb8ebc.

- Write is now atomic: writes to `target_file.with_suffix(".tmp")` then `os.replace()` into place.
- Backup now writes `raw_content` directly to the backup file via `backup_file.write_text(raw_content)` instead of `shutil.copy2` from disk, eliminating the TOCTOU race.

custom_components/ha_mcp_tools/__init__.py
kingpanther13: Fixed in aeb8ebc. The service handler now calls \`homeassistant.check_config\` after writing and includes the result in the response as \`config_check\` (values: "ok", "errors", or "unavailable"). If errors are found, they're included in \`config_check_errors\`. Uses the existing HA service API — the REST client's \`check_config()\` method was already in place.

kingpanther13:
All 6 review comments addressed in aeb8ebc + E2E tests added:

1. Removed redundant `BLOCKED_YAML_KEYS` + trimmed allowlist (removed helper-manageable keys)
2. Removed `HAMCP_ENABLE_FILESYSTEM_TOOLS` dependency — single flag only
3. Fixed error handling: `raise_tool_error` on failure, no `add_timezone_metadata` on errors
4. Null content rejection + type mismatch error instead of silent replace
5. Atomic write (temp + `os.replace`) + backup from `raw_content`
6. Added `homeassistant.check_config` call after write
7. Added comprehensive E2E test suite (`test_yaml_config.py`) covering security boundaries, validation, CRUD, backup, and config check

@sergeykad ready for re-review.

Issue #804, opened 2026-03-22
[FEATURE] Support for 'ENABLED_TOOL_MODULES' in ha-add-on


snok3r-krk:
### What would you like to see?

Currently, I do not believe that 'ENABLED_TOOL_MODULES' is supported in ha-add-on. I think it makes sense to add, as maintaining a deny list per connection may be be finicky and error-prone in some cases.

### Why do you need this?

An option to define which tool_modules are allowed

PR #857, opened 2026-04-03
feat: user-configurable tool disable/pin in add-on settings

### LINKED ISSUE #798: [FEATURE] User configurable tools and pinned tools on config screen in HAOS addon UI (for after #727 is merged)

After working on #727 I realized that some users might want an easy way to enable and disable tools within the HAOS addon settings. We could implement a simple menu to allow users to select what tools they would like loaded; in addition to this, we could potentially allow users that are using the toolsearch function (introduced in #727) to configure what tools are pinned. We would have some tools that are mandatory to be pinned at all times, most obviously the basic search tools, but others can be optional. This would provide a lot of configurability and freedom for users to choose what setup works best for them. 

I initially was going to make the selection of pinned tools a user option in #727 but I felt that I could expand that out to allowing selection of all tools whether using toolsearch/727, or not. 

### Why do you need this?

Will allow users to create their own custom variations. When using toolsearch: This will allow them to decide what tools should be pinned or deferred in the toolsearch function with #727, allowing a lot of configurability. When not using toolsearch: will allow users to decide what tools are and are not available in the first place, allowing users to customize how much context their ha-mcp setup uses without needing to use toolsearch if they choose not to use it. 

### Additional context

This likely will require two menus or a multi step menu: one for disabling/enabling tools in the first place, and another for deciding what tools will and won't be pinned or using toolsearch. They don't need to be mutually exclusive, the pinned options menu simply won't do anything if toolsearch is disabled, and the disable/enable menu can still work whether or whether not toolsearch is used; if toolsearch is being used then disabling tools by that menu will just simply remove them from the list and the proxied list as well. 

Changing the list will require restarting the addon, and likely will require reconnection of the AI client as well in order to refresh the tool list on the client side when not in toolsearch mode (#727). When in toolsearch mode restarting the addon will still be required, but reconnecting to AI won't be if the tool being enabled/disabled is one of the proxied tools. 

Also: I want to add functionality for the user to change the default amount of tools returned by our toolsearch tool. This should be within the scope of the PR that will close this issue as the idea is adding more things for the user to configure in the settings. Currently the default is 5. Any more than that will use way too much context, but reducing to 2-3 may actually be beneficial for users as the search often finds the proper tool within the first 2-3 results. 

### Addendum 

When this is created it will make the feature flag/toggle in config settings for ha_config_set_yaml unnecessary, so the PR that will close this issue should remove that toggle as well. Same for the codemode tool if it gets merged first. I will have small tool descriptions on each tool to explain if it's read/write/destructive/etc, and with warnings for particularly dangerous tools. I'm unsure if I should keep the toggled for enabling skills as resources and enabling skills as tools the same or if they should be part of this new system since they already enable/disable tools, and I also need to consider how toggling toolsearch on and off will affect things.... Perhaps it shouldn't make a difference  for the main enable/disable since they're the same tools either way, but I need a good interface that shows the option to pin/unpin the tools too. Furthermore, need to set this up in a way that won't break things and will work properly when tools are added or taken away.

kingpanther13:
## What does this PR do?

Adds a full per-tool and per-group configuration UI to the HAOS add-on settings page, using the Grocy-style nested collapsible sections pattern. Each of the 92 tools can be individually set to **enabled-unpinned**, **enabled-pinned** (always visible in tool search), or **disabled** via a dropdown.

**UI structure:**
- Single collapsible **"Advanced tool configuration"** parent section
- 23 tool groups inside (Add-ons, Automations, HACS, System, etc.), each collapsible
- Group-level `enabled` master toggle — disables all tools in the group when off
- Per-tool dropdown: `enabled-unpinned` | `enabled-pinned` | `disabled`
- Description includes link to the [tools page](https://homeassistant-ai.github.io/ha-mcp/tools) for full tool details

**New/restored config options:**
- `enable_yaml_config_editing` — standalone bool toggle restored (off by default), gates `ha_config_set_yaml`
- `tools` — nested object with per-group, per-tool dropdowns (auto-generated)
- `tool_search_max_results` — configurable 2-10 (default 5)

**Key behaviors:**
- Group `enabled: false` overrides all individual tool states in that group
- Mandatory tools (`ha_search_entities`, `ha_get_state`, `ha_get_overview`, `ha_report_issue`) are re-enabled by the server regardless of user config
- Default-pinned tools match the existing `DEFAULT_PINNED_TOOLS` (ha_restart, ha_reload_core, ha_backup_create/restore, ha_search_entities, ha_get_overview, ha_report_issue, ha_config_get/set_automation, ha_config_set_yaml)
- `enable_yaml_config_editing: false` forces `ha_config_set_yaml` to disabled regardless of the tools dropdown
- Uses native FastMCP `mcp.disable()`/`mcp.enable()` visibility transforms

**Precedence rules (applied in order by `_apply_tool_visibility()`):**
1. Group toggles — `enabled: false` disables all tools in the group
2. Individual tool dropdowns — `disabled` removes the tool
3. `enable_yaml_config_editing` toggle — overrides `ha_config_set_yaml` dropdown (off = always disabled, on = enabled-pinned)
4. Mandatory tool re-enable — `ha_search_entities`, `ha_get_overview`, `ha_get_state`, `ha_report_issue` are force-enabled last, overriding everything above

Note: HAOS Supervisor doesn't support conditionally disabling config options, so mandatory tools still show dropdowns in the UI even though the server ignores them. The add-on config descriptions now document this behavior directly.

**Auto-generation:**
- `extract_tools.py` (from PR #839) extended to auto-generate config.yaml tool sections and translations from tool source code via AST parsing
- New tools added to the codebase are automatically included when the script runs post-merge (via `sync-tool-docs.yml` from PR #849)
- Config sections are between `GENERATED ... START/END` markers — never edit by hand
- `homeassistant-addon-dev/translations/en.yaml` is a symlink to the addon version

Closes #798

## Type of change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 💥 Breaking change

## Testing

## Checklist


kingpanther13:
The point of this PR is to make the addon more congruent with the docker/command line versions. Currently users can configure what tools are enabled and disabled via command line, but that isn't possible when running ha-mcp as an HAOS addon. 

"Per-tool pinning duplicates toolsearch. Toolsearch (#727) already handles tool prioritization automatically. Exposing pin configuration in the UI duplicates what the algorithm does." The  per-tool pinning does not duplicate toolsearch. The pinned tools were decided by me via what I thought was best per BAT testing, it is NOT chosen automatically or by an algorithm. Opening up per tool pinning to users allows users to decide what tools are best to be pinned for their specific use case. 

As for "Per-tool enable/disable is premature. No user has reported that tool count causes problems. If it does, the right fix is improving toolsearch, not adding 92 dropdowns." There have been multiple issues regarding this, #727 was meant to address this, but #727 was made before I realized that Claude and other LLMs have their own built in toolsearch now which would render toolsearch useless. This allows users to granularly control things at their own desire. 

As for "enable_yaml_config_editing is redundant. YAML editing requires a HACS addon to be installed. That dependency gates the feature already." I don't know why Claude mentioned that in my PR description, it really had no need to. What that means to say is that I'm leaving that separate toggle intact right now instead of removing it.... That tool should be disabled by default, but I realized if I removed its individual toggle it would be hidden in a long list of tools so I left it separate.

kingpanther13:
As for the  maintenance concerns, once merged this will automatically update itself using the same mechanism behind #839/#849 so it shouldn't require any extra maintenance. Per Claude once #857 is merged, the sync-tool-docs.yml workflow (from PR
  #849) runs extract_tools.py on every push to master that touches tool
  source files. Since we extended extract_tools.py to also generate
  config.yaml and translations, new tools will be auto-committed. Zero
  maintenance.

kingpanther13:
Good questions — dug into the code to confirm:

**Precedence (mandatory tools in disabled groups):** Yes, this is handled. `_apply_tool_visibility()` in `server.py` runs disables first (groups by tag, then individual tools by name), then re-enables mandatory tools last via `mcp.enable()`. Last call wins, so mandatory tools always survive regardless of group toggle. There's also a unit test (`test_mandatory_tools_cannot_be_disabled`) that explicitly disables the "Search & Discovery" group and asserts `ha_search_entities`, `ha_get_state`, and `ha_get_overview` are still visible. DOCS.md documents this as: "Mandatory tools that cannot be disabled (silently re-enabled)."

The gap is purely UI — the config dropdowns still show for mandatory tools even though the server ignores the setting. HAOS Supervisor doesn't support conditionally disabling individual config options, so there's no clean way to grey them out.

**`ha_config_set_yaml` default state:** The tool IS disabled at runtime (both `start.py` and `server.py` force it into the disabled list when `enable_yaml_config_editing` is false). The issue is the config.yaml default showing `enabled-pinned` — gives the impression it's active when it's actually overridden. @kingpanther13 said he'd address this.

kingpanther13:
Pushed a commit to clarify the config descriptions in `translations/en.yaml`:

- **Enable YAML config editing**: Now explicitly says it controls `ha_config_set_yaml` and overrides the tool dropdown when off
- **Advanced tool configuration**: Now documents group toggle override behavior, mandatory tool re-enable, and the `ha_config_set_yaml` / toggle relationship

This way users see the precedence rules right in the HAOS settings page.

julienld:
After digging in, I don't think this approach is ready to merge.

**Config drift between merge and release.** `sync-tool-docs.yml` already runs `extract_tools.py` on every tool-touching push to master, but its `git add` only stages `site/src/data/tools.json README.md homeassistant-addon/DOCS.md` — `homeassistant-addon/config.yaml` is missing, so the regenerated tool list is discarded. The release pipeline (`_update-addon-config.yml`) only bumps the version field. Even if the missing `git add` were fixed, regenerating on every master push would still be wrong — `config.yaml` needs to match the tools in the published Docker image, so generation belongs in the release pipeline, not on merge.

**Contradicts our non-breaking rename/consolidation policy.** `AGENTS.md` explicitly states that tool renames and consolidations are non-breaking as long as the same outcome is achievable. With this PR, a user who has disabled a tool in their config will silently lose that setting if the tool is renamed or consolidated — the server ignores unknown names without warning. This concern applies equally to `DISABLED_TOOLS`.

**Simpler path to close the gap.** HAOS users currently have no way to access `DISABLED_TOOLS`/`PINNED_TOOLS` without SSH. A plain text field in the add-on config mirroring the env vars directly closes the gap in ~30 lines with no generation pipeline and no drift risk. For a proper tool selection UI, I'd rather wait until we migrate the entire MCP server to a custom component — a config screen could read available tools at startup and working at category level would sidestep the rename/consolidation problem entirely. If you don't want to wait that long, the right place for this UI is `ha_mcp_tools`, not the add-on config schema.

kingpanther13:
Appreciate the thorough review. I'd like to push back on a few points before reworking anything, since getting the UI right took a lot of iteration.

**On the plain text field approach:** It has the exact same rename/consolidation problem as point 2 — a user who has `ha_old_name` in a text field loses their setting just as silently when it gets renamed. On top of that, a text field is strictly worse UX: users have to locate the tool list elsewhere to know what to type, and typos produce no feedback at all. The whole reason I moved away from text fields during development was that they're error-prone with no way for users to know they made a mistake. The nested dropdown approach is gated behind an "Advanced tool configuration" label specifically because it's for users who understand what they're doing.

**On point 1 (config drift):** Fixable. I can either add `homeassistant-addon/config.yaml`, `homeassistant-addon-dev/config.yaml`, and `homeassistant-addon/translations/en.yaml` to the `git add` in `sync-tool-docs.yml`, OR — which I think matches your intent better — move the addon config/translations generation into `_update-addon-config.yml` so it regenerates as part of the release pipeline and stays in lockstep with the published Docker image. The second option is cleaner; happy to go that direction.

**On point 2 (rename/consolidation):** This is a real concern, but I don't think there's a clean fix for *any* per-tool config mechanism — `DISABLED_TOOLS` env var has the exact same behavior today. The best I can offer is:
1. Add a rename map (old_name → new_name) maintained alongside the tool code, applied when parsing user config
2. Log a warning when a user-configured tool name doesn't match any registered tool

That doesn't eliminate the risk but it surfaces it to the user and gives us a migration path when we do rename things. It's a maintenance burden, but it's tractable.

**Why I think this is worth keeping:** This PR is really an extension of #727. Once tool search is enabled, a user who clicks "allow all" on `ha_call_write_tool` is implicitly trusting *every* destructive tool the server exposes — including ones they may not know about or want. Without per-tool disable, the only safety net is reviewing each proxy call individually. The ability to pre-disable dangerous tools (or entire groups) before they ever enter the proxy's tool space is a genuine safety feature for users running #727.

It's also still useful for users who choose *not* to use #727 — they can trim their tool list to reduce idle context usage, or hide tools they simply don't want the assistant to touch (e.g. disable file writes, disable HACS operations, etc.). Both use cases matter.

I opened #798 myself, but #804 is an independent user request for the same thing, so the demand is real beyond just my own use case.

I understand the long-term vision is a custom component where this UI can be dynamic, but there's no concrete timeline on that migration. In the meantime HAOS users have no way to access these controls without SSH.

**My ask:** Before I spend more time reworking anything, could you tell me which of these you're open to?
1. Keep the current approach if I fix point 1 (move generation to release pipeline) and address point 2 (rename map + warning log)?
2. Rework to something different you'd prefer (and if so, what)?
3. Not willing to consider this at all until the custom component migration?

I want to make sure I don't burn more cycles on the wrong direction.

kingpanther13:
@julienld please let me know how you would like me to handle this... I could either make some or all of the changes I suggested in my previous comment, or I can rework it, or just close it altogether. 

 I was thinking there's another option. The tools list is already generated in the docs list on the addon. I could do a text field type thing and just put a note to get the tools from the docs tab.

 I think that would be worse UX than what I have currently, and would still run into the issue of tool name changes being a breaking change... It actually could be worse, because if a name changes and that tool is in the disabled list it might stop the addon from starting completely until the tool name is manually updated in the config page, unless I don't include typo detection. At least the version I have now won't have that issue, it would just silently reenable the tool. That all could probably be worked out but I just wanted to lay out the options,  I'll go whatever direction you want. If you haven't done it already I'd request for you to at least load my PR up on an haos supervised installation so you can see how it looks, I personally think it's pretty smooth, and it's all nested under an "advanced settings" thing anyways.

julienld:
@kingpanther13 :  My main concern is that we’re writing 1300 lines of code specifically for the add-on. It would be much better to have a configuration UI that works across all installation methods. Perhaps we could use the custom component to handle this via an Integration setup screen? This would make sense if we transition to a custom component in the future.

Alternatively, we could use the built-in FastMCP HTTP server to expose a custom webpage. This could function as an add-on UI, a custom page for the HTTPS/Docker versions, or even a command-line tool like `uvx --from=ha-mcp websettings` for the stdio/uv version (which would write settings locally).

Or perhaps... we could simply create a tool that enables or disables other tools. This would allow for requests like: "Reduce the number of tools to only what’s relevant for my current Home Assistant setup to help me build automations and dashboards."

kingpanther13:
@julienld ah I see what you're saying now. I had considered doing those methods but decided not to because they would be about the same amount of lines of code, but if it works across all install methods then maybe it would be worth it. To be clear most of the 1300 lines here are auto generated by the extract tools python file and are automatically regenerated whenever tools are modified, but you're right that it's only useful for the addon. 

As for a custom tool and controlling via the llm itself, I was actually proposing issues #863 and #913 for after something like this was made so that settings could be controlled from within the llm.

I'm leaning toward doing a  custom fastmcp page because that would probably work better with #863 down the road. I'll do a bit more digging.

julienld:
Interesting, MCP Apps would also work with a custom component. It would permit to surface any Home Assistant integration config screen using the config flow (including our custom component settings).

kingpanther13:
I just made #960, it is very early draft. Need to find the time to actually load it up and test it today, but I went with the http route because I've always had a lot of trouble with custom component/integration stuff with HA.

kingpanther13:
Closing in favor of #960.

PR #859, opened 2026-04-03
fix: init submodules and use portable path in /wt skill

sergeykad:
## What does this PR do?

Fixes two issues in `.claude/skills/wt/SKILL.md`:

1. **Hardcoded path** — `cd /home/julien/github/ha-mcp` replaced with `cd "$(git rev-parse --show-toplevel)"` so the skill works for any contributor
2. **Missing submodule init** — adds `git submodule update --init --recursive` after worktree creation so the `skills-vendor` submodule is available

Without the submodule init, lefthook pre-commit unit tests fail (`test_resources.py` checks for `skills-vendor/skills/` directory).

## Type of change
- [x] 🐛 Bug fix

## Testing

## Checklist

### DIFF
--- a/.claude/skills/wt/SKILL.md
+++ b/.claude/skills/wt/SKILL.md
@@ -20,13 +20,13 @@ git checkout master
 git pull origin master
 
 # Create worktree
-git worktree add worktree/$ARGUMENTS -b $ARGUMENTS
+git worktree add worktree/"$ARGUMENTS" -b "$ARGUMENTS"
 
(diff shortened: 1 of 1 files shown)


kingpanther13 (approved):
Clean fix. Portable path, proper quoting, submodule init — all correct.

Re: Gemini's security flag on `$ARGUMENTS` quoting — false positive. The PR already quotes it (`worktree/"$ARGUMENTS"` is functionally identical to `"worktree/$ARGUMENTS"`).

PR #865, opened 2026-04-05
fix: fetch addon stats from /addons/{slug}/stats endpoint

teancom:
## What does this PR do?

The include_stats parameter on ha_get_addon was reading cpu_percent, memory_percent, etc. from the /addons list response, but that endpoint doesn't include stats fields — they were always null.

The Supervisor API exposes per-addon stats at /addons/{slug}/stats. This fix calls that endpoint for each running addon when include_stats=True.

Tested against a live HAOS instance via the supervisor/api WebSocket command — the endpoint returns cpu_percent, memory_usage, memory_limit, memory_percent, network_rx/tx, and blk_read/write.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 💥 Breaking change

## Testing

## Checklist

### DIFF
--- a/src/ha_mcp/tools/tools_addons.py
+++ b/src/ha_mcp/tools/tools_addons.py
@@ -175,7 +175,7 @@ async def list_addons(
                         "memory_usage": s.get("memory_usage"),
                         "memory_limit": s.get("memory_limit"),
                     }
-            except (ToolError, Exception) as exc:
+            except Exception as exc:
                 logger.warning("Failed to fetch stats for addon %s: %s", slug, exc)
             return slug, None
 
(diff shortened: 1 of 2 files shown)


kingpanther13:
No worries! That was actually from Claude not me, I'm sorry I didn't catch that and edit it out it really isn't that big of a deal. I'll re-review once you push your changes

teancom:
Oh, I can't blame Claude, I'm just an idiot. And now I'm not at home. Will push to the right branch when I get back from the Easter egg hunt.

teancom:
Alright, this time with feeling. I managed to git push to the correct branch this time. 😅

teancom:
Hmm. That wasn't a force push, and that third commit *is* the one I was referencing before - it has the tests and fixes the redundant ToolError.

PR #870, opened 2026-04-05
feat: add pagination and detail_level to ha_list_services, ha_get_device, ha_get_integration

### LINKED ISSUE #852: [BUG] ha_list_services returns 260KB unfiltered response that exceeds MCP client token limits
## 💬 What Happened?

Calling `ha_list_services()` with no arguments returns all 469 services across 94 domains in a single 260KB JSON response. The MCP client (Claude Code) truncates it with:

> Error: result (239,065 characters) exceeds maximum allowed tokens.

The tool executes successfully (50ms), but the response is so large the LLM never sees it. An LLM exploring "what services are available?" will call it bare and hit this wall every time. With `domain` filter the response is manageable, but the tool's defaults don't prevent the unfiltered firehose.

The tool has no pagination (`limit`/`offset`), no summary mode, and no result size guardrails — unlike `ha_search_entities`, `ha_deep_search`, and `ha_hacs_search` which all have pagination.

**Potential solutions:**
- Add `limit`/`offset` pagination (consistent with other search tools)
- Add a `detail_level` parameter (similar to `ha_get_overview`) — a `summary` mode returning just service key + name + description (no full field schemas) would drastically reduce response size while remaining useful
- Keep the full unfiltered response accessible via explicit `detail_level="full"` for cases where it's needed

## 📋 Copy-Paste from Conversation

**Tool call:**
```
ha_list_services() with no arguments
```

**MCP client response:**
```
Error: result (239,065 characters) exceeds maximum allowed tokens. Output has been saved to 
/home/.claude/projects/.../tool-results/mcp-...-ha_call_read_tool-1775063774202.txt.
Format: JSON with schema: {success: boolean, domains: [string], services: {persistent_notification.create: {...}, 
persistent_notification.dismiss: {...}, ...}, total_count: number, filters_applied: {domain: null, query: null}}
```

**`ha_report_issue` diagnostic output:**

```
=== ha-mcp Bug Report Info ===

ha-mcp Version: 7.1.0
Installation Method: addon
Platform: Linux 6.12.67-haos (x86_64)
Python Version: 3.13.12
Home Assistant Version: 2026.4.0
Connection Status: Connected
Entity Count: 3308
Location Name: Home
Time Zone: America/New_York

=== Recent Tool Calls ===
  2026-04-01T17:16:11 | ha_list_services | OK | 50ms | 259,911 bytes response
```

**Related:** #851 (BM25 search improvements — includes `ha_list_services` query scoring)

---

## Addendum: Broader audit of tools with excessive response sizes

A systematic audit was run against **every ha-mcp tool that returns bulk/list data** on a live HA instance (3,308 entities, 158 integrations, ~469 services) to identify which other tools have the same "firehose" problem as `ha_list_services`. Tools already optimized (`ha_search_entities`, `ha_deep_search`) were excluded.

### Critical — overflow the LLM context

| Tool | Call pattern | Response size | Pagination? | Summary mode? |
|------|-------------|---------------|:-----------:|:------------:|
| **`ha_get_device()`** | No args (list all devices) | **599 KB** | ❌ | ❌ |
| **`ha_get_overview("full")`** | Non-default `detail_level` | **434 KB** | N/A | Has levels, but… |
| **`ha_list_services()`** | No args | **239 KB** | ❌ | ❌ |
| **`ha_get_overview("standard")`** | Non-default `detail_level` | **198 KB** | N/A | …even "standard" overflows |
| **`ha_get_logbook(limit=500)`** | Max allowed limit, 24 h | **91 KB** | ✅ (default 50) | Has compact mode |

#### `ha_get_device()` — worst offender (599 KB)

When called with no arguments it dumps **every device** with full details (device_id, name, manufacturer, model, area_id, connections, identifiers, all associated entities). An LLM's natural first step to discover devices is a bare `ha_get_device()` call — instant overflow.

It has filter parameters (`area_id`, `manufacturer`, `integration`) but **no `limit`/`offset` pagination** and **no `detail_level`** parameter.

**Needs:** pagination + a summary mode (e.g. just name/manufacturer/area vs full fields+entities).

#### `ha_get_overview("standard"/"full")` — 198 KB / 434 KB

The default "minimal" level is safe (samples 10 entities/domain). But "standard" lists ALL entities per domain with friendly_name only and still hits 198 KB. "Full" adds entity_id + state + system info for 434 KB. The jump from "minimal" to "standard" is too steep.

> **Note:** PR #728 is supposed to address `ha_get_overview`.

#### `ha_get_logbook(limit=500)` — 91 KB

Already has pagination (default limit=50 works fine), but the max allowed limit of 500 can still produce 91 KB. Even 1 hour of logbook had **5,258 entries** and 24 hours had **52,370 entries** on this test instance. The default is safe; the max should probably be capped lower or a warning/cap applied.

### Borderline — could overflow on larger systems

| Tool | Call pattern | Response size | Risk |
|------|-------------|---------------|------|
| **`ha_get_integration()`** | No args | ~40 KB (158 entries) | No pagination. Systems with 300+ integrations will overflow |
| **`ha_config_get_dashboard(url_path=...)`** | Specific dashboard config | ~10 KB (simple) | Complex dashboards with many views/cards: 50–500 KB+ |

`ha_get_integration()` has no `limit`/`offset` at all. With 158 entries it fit inline (~40 KB), but on larger HA installations this will overflow. Worth adding pagination proactively.

### Safe — tested and within bounds

| Tool | Size | Notes |
|------|------|-------|
| `ha_config_list_areas()` | ~5 KB | 26 areas |
| `ha_config_list_floors()` | <1 KB | 0 floors |
| `ha_config_list_groups()` | ~15 KB | 2 groups (one had ~200 entity_ids) |
| `ha_config_list_dashboard_resources()` | ~2 KB | 11 resources |
| `ha_hacs_list_installed()` | ~10 KB | 32 repos |
| `ha_get_updates()` | ~2 KB | 3 updates |
| `ha_get_logbook()` default | ~14 KB | 50 entries, pagination works |
| `ha_get_overview("minimal")` default | Small | Sampling works |
| `ha_get_history()` | Has limits | Default 100, max 1000, per-entity |
| `ha_get_statistics()` | Small | Pre-aggregated by period |
| `ha_config_get_dashboard(list_only=True)` | ~2 KB | Metadata only |

### Suggested priority

1. **`ha_get_device()`** — most urgent, 599 KB, zero guardrails
2. **`ha_list_services()`** — this issue, 239 KB
3. **`ha_get_overview()`** — 198 KB at "standard" (PR #728 may address this)
4. **`ha_get_integration()`** — proactive fix before it becomes a problem on larger systems

### Audit test environment

- ha-mcp 7.2 (addon), HA 2026.4.0, 3,308 entities, 158 integrations, ~469 services
- MCP client: Claude Code (Opus 4.6)
- All sizes measured via Claude Code's overflow error messages (`result (N characters) exceeds maximum allowed tokens`)

kingpanther13:
## What does this PR do?

Adds `limit`/`offset` pagination and `detail_level` controls to the three tools identified in #852 that return oversized responses on large HA instances:

- **`ha_list_services()`** (primary issue, 260KB): Paginated (default 50/page), `detail_level` parameter (`summary`/`full`). Summary mode omits field schemas, reducing per-service size ~80%.
- **`ha_get_device()`** (worst offender, 599KB): Paginated (default 50/page), `detail_level` (`summary`/`full`). Summary mode omits per-device entity lists. Single device lookups remain full detail.
- **`ha_get_integration()`** (proactive fix, ~40KB): Paginated (default 50/page). Prevents overflow on larger installations.

Shared `build_pagination_metadata()` helper extracted to `util_helpers.py` for consistent pagination fields (`total_count`, `offset`, `limit`, `count`, `has_more`, `next_offset`). Existing `_build_pagination_metadata` in `tools_search.py` refactored as a thin wrapper.

All new parameters have sensible defaults — existing callers with no args get paginated summary results (50/page) instead of the full firehose. Single-item lookups are unaffected.

## Type of change
- [x] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 💥 Breaking change

## Testing

**Tests added/updated:**
- `test_list_services.py`: Updated for summary mode defaults; added `test_pagination_limit_and_offset`, `test_detail_level_full_includes_fields`, `test_detail_level_summary_omits_fields`
- `test_device_registry.py`: Updated `test_list_all_devices` for pagination metadata and summary mode; added `test_list_devices_pagination`, `test_list_devices_full_detail`
- `test_list_integrations.py`: Updated all `data["total"]` → `data["total_count"]` references; added `test_pagination_limit_and_offset`

## Checklist

Closes #852

### DIFF
--- a/tests/src/e2e/tools/test_logbook.py
+++ b/tests/src/e2e/tools/test_logbook.py
@@ -87,30 +87,26 @@ async def test_logbook_limit_capped_at_maximum(mcp_client):
     data = get_logbook_data(raw_data)
 
     # Verify limit is capped at 500
-    assert data["limit"] == 500, (
-        f"Limit should be capped at 500, got {data['limit']}"
-    )
+    assert data["limit"] == 500, f"Limit should be capped at 500, got {data['limit']}"
 
     logger.info(f"Limit correctly capped at {data['limit']}")
 
 
 @pytest.mark.asyncio
 async def test_logbook_minimum_limit(mcp_client):
-    """Test that logbook limit has a minimum of 1."""
+    """Test that logbook limit of 0 is rejected (must be at least 1)."""
     logger.info("Testing logbook minimum limit")
 
-    result = await mcp_client.call_tool(
+    data = await safe_call_tool(
+        mcp_client,
         "ha_get_logs",
-        {"hours_back": 1, "limit": 0},  # Request zero
+        {"hours_back": 1, "limit": 0},
     )
 
-    raw_data = assert_mcp_success(result, "Logbook with zero limit")
-    data = get_logbook_data(raw_data)
+    # limit=0 should be rejected by coerce_int_param (min_value=1)
+    assert not data.get("success"), "limit=0 should be rejected"
 
-    # Verify limit is at least 1
-    assert data["limit"] >= 1, f"Limit should be at least 1, got {data['limit']}"
-
-    logger.info(f"Minimum limit enforced: {data['limit']}")
+    logger.info("Zero limit correctly rejected")
 
 
 @pytest.mark.asyncio
@@ -164,22 +160,19 @@ async def test_logbook_pagination_with_offset(mcp_client):
 
 @pytest.mark.asyncio
 async def test_logbook_negative_offset(mcp_client):
-    """Test that negative offset is clamped to 0."""
+    """Test that negative offset is rejected."""
     logger.info("Testing logbook with negative offset")
 
-    # Use safe_call_tool — offset clamping is applied before the HA API call,
-    # so the clamped value appears in both success and error responses
-    raw_data = await safe_call_tool(
+    data = await safe_call_tool(
         mcp_client,
         "ha_get_logs",
         {"hours_back": 1, "limit": 10, "offset": -5},
     )
-    data = get_logbook_data(raw_data)
 
-    # Verify offset is clamped to 0 (present in both success and error context)
-    assert data["offset"] == 0, f"Negative offset should be clamped to 0, got {data['offset']}"
+    # Negative offset should be rejected by coerce_int_param (min_value=0)
+    assert not data.get("success"), "Negative offset should be rejected"
 
-    logger.info("Negative offset correctly clamped to 0")
+    logger.info("Negative offset correctly rejected")
 
 
 @pytest.mark.asyncio
@@ -211,9 +204,7 @@ async def test_logbook_has_more_indicator(mcp_client):
         )
         logger.info(f"Pagination hint: {data['pagination_hint']}")
 
-    logger.info(
-        f"has_more={has_more} (total={total}, limit=2, offset=0)"
-    )
+    logger.info(f"has_more={has_more} (total={total}, limit=2, offset=0)")
 
 
 @pytest.mark.asyncio
@@ -388,7 +379,9 @@ async def test_logs_error_log_source(mcp_client):
     assert "returned_lines" in data, "Response should contain returned_lines"
     assert data["limit"] == 20, f"Limit should be 20, got {data['limit']}"
 
-    logger.info(f"Retrieved {data['returned_lines']} of {data['total_lines']} log lines")
+    logger.info(
+        f"Retrieved {data['returned_lines']} of {data['total_lines']} log lines"
+    )
 
 
 @pytest.mark.asyncio
@@ -416,7 +409,10 @@ async def test_logs_invalid_level(mcp_client):
             {"source": "system", "level": "INVALID"},
         )
 
-    assert "invalid" in str(exc_info.value).lower() or "level" in str(exc_info.value).lower()
+    assert (
+        "invalid" in str(exc_info.value).lower()
+        or "level" in str(exc_info.value).lower()
+    )
     logger.info("Invalid level correctly raises ToolError")
 
 
--- a/tests/src/e2e/workflows/services/test_list_services.py
+++ b/tests/src/e2e/workflows/services/test_list_services.py
@@ -97,8 +97,8 @@ class TestServiceDiscovery:
         if "light.turn_on" in services:
             turn_on = services["light.turn_on"]
             assert "name" in turn_on, "Service should have name"
-            # Domain filter auto-enables full detail — fields should be present
-            assert "fields" in turn_on, "Domain filter should auto-enable full detail"
+            # Default is summary mode — fields are omitted even with domain filter
+            assert "fields" not in turn_on, "Summary mode should omit fields"
 
         logger.info("Domain filter test passed")
 


src/ha_mcp/tools/tools_integrations.py
```diff
@@ -266,16 +271,23 @@ async def ha_get_integration(
...
+            paginated_entries = formatted_entries[offset_int : offset_int + limit_int]
+
             result_data: dict[str, Any] = {
                 "success": True,
-                "total": len(formatted_entries),
-                "entries": formatted_entries,
+                **build_pagination_metadata(
+                    total_entries, offset_int, limit_int, len(paginated_entries)
+                ),
+                "total": total_entries,
```
sergeykad: **1. Duplicate total field.** Response contains both `total_count` (from `build_pagination_metadata`) and `total` with the same value. The other two tools (`ha_get_device`, `ha_list_services`) only have `total_count`. Remove `"total"` — tests already use `total_count`, and keeping both creates ambiguity for consumers.

src/ha_mcp/tools/tools_registry.py
```diff
@@ -608,42 +620,60 @@ async def ha_get_device(
...
-        - Full device details including integration_type, ieee_address/node_id, radio_metrics/node_status, entities
+        **Zigbee:** integration="zha" or "zigbee2mqtt". Returns ieee_address, radio metrics.
+        **Z-Wave:** integration="zwave_js". Returns node_id, node_status.
         """
         try:
+            limit_int = coerce_int_param(
+                limit, "limit", default=50, min_value=1, max_value=200
+            )
+            offset_int = coerce_int_param(offset, "offset", default=0, min_value=0)
+            effective_detail = (
```
sergeykad: **2. Silent `detail_level` override.** When a user explicitly passes `detail_level="summary"` alongside a filter (e.g., `ha_get_device(area_id="kitchen", detail_level="summary")`), their choice is silently overridden to `"full"`. This contradicts the PR's goal of compact responses.

Consider using `None` as the default for `detail_level` so you can distinguish "user didn't specify" (auto-promote) from "user explicitly chose summary" (respect it). Same issue in `ha_list_services` at line 108.

src/ha_mcp/tools/tools_integrations.py
```diff
@@ -123,6 +124,10 @@ async def ha_get_integration(
             exact_match_bool = coerce_bool_param(
                 exact_match, "exact_match", default=True
             )
+            limit_int = coerce_int_param(
```
sergeykad, line 127: **3. `coerce_int_param` silently clamps out-of-range values.** Negative values for `limit`/`offset` indicate a caller bug, but they're silently clamped to `min_value` with no feedback. This is inconsistent with `coerce_bool_param` which raises `ValueError` for unrecognized inputs.

Consider raising `ValueError` for negative values (genuinely nonsensical for pagination) and at least logging at debug level when clamping values above `max_value`.

src/ha_mcp/tools/tools_registry.py
```diff
@@ -911,21 +951,31 @@ def get_device_info(device: dict[str, Any]) -> dict[str, Any]:
...
-                        not in device_info.get("integration_sources", [])
+                    elif integration_lower not in device_info.get(
+                        "integration_sources", []
                     ):
                         continue
 
-                device_info["entities"] = device_to_entities.get(device.get("id"), [])
+                # In summary mode, omit entity lists to reduce response size
+                if effective_detail == "full":
+                    device_info["entities"] = device_to_entities.get(
```
sergeykad, line 962: **4. Unnecessary entity registry work in summary mode.** In summary mode, entities are never included in the response, but the tool still fetches the full entity registry and builds `device_to_entities`. On large HA instances this is wasted overhead.

The `entity_to_device` reverse map is still needed for `entity_id` parameter resolution, but `device_to_entities` construction and entity formatting can be gated on `effective_detail == "full"` (and no single-device lookup).

src/ha_mcp/tools/util_helpers.py
```diff
@@ -232,6 +228,28 @@ def parse_string_list_param(
     raise ValueError(f"{param_name} must be string, list, or None")
 
 
+def build_pagination_metadata(
```
sergeykad, line 232: **5. Missing unit tests for shared utilities.** `build_pagination_metadata` and `coerce_int_param` are new shared utilities used by 3+ tools with no unit test coverage. Both are pure functions ideal for fast deterministic tests.

`build_pagination_metadata` computes `has_more` and `next_offset` — getting these wrong means AI agents stop paginating early or loop forever. `coerce_int_param` has clamping behavior that should be pinned. Key cases: exact boundary (offset+count == total), empty page (offset beyond total), clamping behavior, string coercion, invalid strings.

src/ha_mcp/tools/tools_search.py
```diff
@@ -28,16 +29,16 @@
...
-    }
+    """Build standardized pagination metadata for search responses.
+
+    Thin wrapper around the shared ``build_pagination_metadata`` helper that
+    keeps the existing call-site signature (accepts a *results* list and uses
+    ``total_matches`` as the key name expected by search tools).
+    """
+    meta = build_pagination_metadata(total_matches, offset, limit, len(results))
+    # Search tools use "total_matches" instead of "total_count"
+    meta["total_matches"] = meta.pop("total_count")
```
sergeykad: **6. Fragile key rename via `pop`.** This depends on the exact key name `total_count` in the shared helper's return dict. If someone renames that key in `build_pagination_metadata`, this breaks with a `KeyError` at runtime.

Consider having the search module construct its own dict from the helper's values rather than mutating keys, or accept a `total_key_name` parameter in the shared helper.

src/ha_mcp/tools/util_helpers.py
```diff
@@ -232,6 +228,28 @@ def parse_string_list_param(
...
+) -> dict[str, Any]:
+    """Build standardized pagination metadata for paginated responses.
+
+    Args:
+        total_count: Total number of items matching filters (before pagination).
+        offset: Current pagination offset.
+        limit: Maximum items per page.
+        count: Number of items in this page.
+    """
+    has_more = (offset + count) < total_count
```
sergeykad, line 245: **7. No defensive `limit` guard.** If a future caller passes `limit=0` (bypassing `coerce_int_param`), `next_offset` equals `offset`, which could cause an infinite pagination loop for automated consumers. A simple `if limit <= 0: raise ValueError("limit must be positive")` guard at the top of this function is cheap insurance.

kingpanther13 (commented):
All 7 round-3 review comments addressed in 39b8f27:

1. **Duplicate total field** — Removed. Only `total_count` from pagination metadata remains.

2. **Silent detail_level override** — Removed the auto-promotion. Filters no longer override explicit `detail_level` choices.

3. **coerce_int_param silent clamping** — Changed behavior: values below `min_value` now raise `ValueError` (indicates caller bug — negative offset/limit is never valid). Values above `max_value` are still clamped (soft cap for oversized requests like `limit=500` capped to 200). Added unit tests covering both behaviors.

4. **Unnecessary entity registry work in summary mode** — `device_to_entities` construction now gated on `need_entity_details` (true only for single-device lookup or full detail mode). `entity_to_device` map still built for entity_id parameter resolution.

5. **Missing unit tests** — Added `TestBuildPaginationMetadata` (7 test cases: first/last page, exact boundary, empty results, offset beyond total, zero/negative limit) and `TestCoerceIntParam` (12 test cases: None/default, string coercion, below-min raises, above-max clamped, boundary values).

6. **Fragile key rename via pop** — Replaced with explicit dict construction from the shared helper's values.

7. **No defensive limit guard** — Added `if limit <= 0: raise ValueError` at the top of `build_pagination_metadata`.

PR #873, opened 2026-04-05
refactor: consolidate 3 overlapping tool pairs 

### LINKED ISSUE #834: [FEATURE] Consolidate overlapping entity, state, and todo tools

Three pairs of tools have significant overlap and could be consolidated following existing patterns in the codebase:

**1. `ha_set_entity` + `ha_rename_entity` → merged `ha_set_entity`**

Both tools use the same WebSocket endpoint (`config/entity_registry/update`). `ha_rename_entity` just passes `new_entity_id` on the same API call, plus handles voice exposure migration. A merged tool would add an optional `new_entity_id` parameter to `ha_set_entity`.

`ha_set_entity` already handles: area, name, icon, enabled, hidden, aliases, labels, and voice exposure. Adding entity ID rename is a natural extension. The voice exposure migration logic (pre-fetch settings, rename, re-apply to new ID) would be triggered conditionally when `new_entity_id` is provided.

Note: If #813 merges first, `ha_rename_entity_and_device` would be removed, which simplifies this merge since there would be no downstream dependency on the rename internal function.

**2. `ha_get_state` + `ha_get_states` → merged `ha_get_states`**

`ha_get_states` is a parallel wrapper over `ha_get_state` — it calls `client.get_entity_state()` for each entity via `asyncio.gather()`. This pattern of accepting `str | list[str]` already exists in the codebase (`ha_get_entity` works this way). A merged `ha_get_states` would accept `entity_id: str | list[str]` and handle single/bulk internally.

**3. `ha_add_todo_item` + `ha_update_todo_item` → merged `ha_set_todo_item`**

These call different HA services (`todo.add_item` vs `todo.update_item`), but `ha_config_set_helper` already uses the same pattern — branching between create and update based on whether an identifier is provided. A merged `ha_set_todo_item` would route based on whether an existing `item` identifier is provided (update) or only a `summary` is given (create).

### Why do you need this?

Following the project's consolidation principle from AGENTS.md: fewer tools reduce cognitive load for AI agents. These merges would reduce the tool count by 3 while preserving all existing functionality. Each merge follows a pattern already established in the codebase.

### Additional context

This is a proposal — I'm open to adjusting the approach for any of these, keeping specific pairs separate if there's a good reason, or dropping it altogether. The planned implementation details are up for discussion.

These could also be implemented as separate PRs if preferred (one per merge).

kingpanther13:
## What does this PR do?

Consolidates 3 pairs of overlapping tools, reducing tool count by 3 per the discussion in #834. Each merge follows a pattern already established in the codebase.

### 1. ha_get_state now accepts str | list[str] (absorbs ha_get_states)

ha_get_states was asyncio.gather over ha_get_state. The merged tool accepts entity_id: str | list[str] — single string returns the same response as before, list uses the parallel fetch logic with dedup and MAX_ENTITIES=100.

**Why keep ha_get_state not ha_get_states:** ha_get_state is referenced in 50+ test files, wait_helpers, cleanup utilities, conftest, performance baselines, and docstrings across the codebase. Renaming to ha_get_states would cause massive churn for zero user-facing benefit. ha_get_entity already uses str | list[str] with a singular name — precedent.

### 2. ha_set_todo_item replaces ha_add_todo_item + ha_update_todo_item

Routes between todo.add_item and todo.update_item based on whether an item identifier is provided, following the ha_config_set_helper create/update pattern.

### 3. ha_set_entity absorbs ha_rename_entity

Adds new_entity_id and new_device_name parameters. Removes ha_rename_entity, _rename_entity_internal, and ~210 lines of code from tools_registry.py.

**Why the ~150-line voice exposure migration was safe to drop:** The migration logic was confirmed redundant by reading HA Core source. Voice exposure for registered entities is stored in RegistryEntry.options, which is preserved via attr.evolve during entity_id renames. The legacy ExposedEntities.entities store only applies to entities WITHOUT a registry entry — but those cannot be renamed. See [issue discussion](https://github.com/homeassistant-ai/ha-mcp/issues/834#issuecomment-4188908868) for the full HA Core source analysis.

## Type of change
- [x] Maintenance/refactor

## Testing
- [x] All automated tests pass
- [x] Code follows style guidelines

Tests updated: test_get_states.py, test_tools_get_states.py (unit), test_lifecycle.py (todo), test_entity_rename.py, test_rename_consolidation.py, homeassistant-addon/DOCS.md

## Checklist

Closes #834

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -21,7 +21,7 @@ env:
   PYTHON_VERSION: "3.13"
   UV_CACHE_DIR: /tmp/.uv-cache
   # renovate: datasource=docker depName=ghcr.io/home-assistant/home-assistant
-  HA_IMAGE_GHCR: "ghcr.io/home-assistant/home-assistant:2026.1.3"
+  HA_IMAGE_GHCR: "ghcr.io/home-assistant/home-assistant:2026.4.1"
 
 jobs:
   # Comprehensive E2E validation for main branch
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -11,7 +11,7 @@ env:
   PYTHON_VERSION: "3.13"
   UV_CACHE_DIR: /tmp/.uv-cache
   # renovate: datasource=docker depName=ghcr.io/home-assistant/home-assistant
-  HA_IMAGE_GHCR: "ghcr.io/home-assistant/home-assistant:2026.1.3"
+  HA_IMAGE_GHCR: "ghcr.io/home-assistant/home-assistant:2026.4.1"
 
 jobs:
   lockfile:
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -82,6 +82,55 @@ To disable it:
 
 See [#783](https://github.com/homeassistant-ai/ha-mcp/issues/783) for more details.
 
+### macOS: "All connection attempts failed" to local Home Assistant
+
+If ha-mcp connects to the demo server but fails to reach your local Home Assistant (`192.168.x.x`, `10.x.x.x`, etc.) on macOS, the most common causes are listed below. See [#867](https://github.com/homeassistant-ai/ha-mcp/issues/867) (Local Network Privacy), [#630](https://github.com/homeassistant-ai/ha-mcp/issues/630) (env vars not reaching ha-mcp), and [#773](https://github.com/homeassistant-ai/ha-mcp/issues/773) (Python version/read-only filesystem) for related reports.
+
+**1. macOS Local Network Privacy (Sequoia 15+)**
+
+macOS Sequoia silently blocks subprocess connections to local network IPs. Claude Desktop spawns `uvx` as a child process, and macOS may block its outbound LAN connections without showing a permission dialog.
+
+- Check **System Settings → Privacy & Security → Local Network** for Claude Desktop
+- If Claude Desktop is not listed, try restarting it to trigger the permission prompt
+
+**Workaround — SSH tunnel to localhost:**
+
+Since macOS does not restrict connections to `localhost`, an SSH port forward bypasses the restriction:
+
+```bash
+ssh -N -L 8123:localhost:8123 user@your-ha-server-ip
+```
+
+Then set `HOMEASSISTANT_URL` to `http://localhost:8123` in your config.
+
+**2. Firewall software (Little Snitch, Lulu, etc.)**
+
+Third-party firewalls may block `python` or `node` processes spawned by Claude Desktop from making network connections. Check your firewall rules and allow connections for these processes. See [#780](https://github.com/homeassistant-ai/ha-mcp/issues/780) for an example resolution.
+
+**3. http:// vs https://**
+
+Home Assistant running in container mode (Docker, K3s) uses HTTP by default. Using `https://` causes a TLS handshake error. Use `http://` unless you have explicitly configured SSL/TLS or a reverse proxy.
+
+**4. Python version too old**
+
+ha-mcp requires Python 3.13+. If you are on Python 3.12 or older, `uvx` installs an outdated version of ha-mcp that may have known bugs (including read-only filesystem errors). Upgrade Python:
+
+```bash
+brew install python@3.13
+```
+
+Then force a refresh:
+
+```bash
+uvx --refresh ha-mcp@latest
+```
+
+If `uvx` still uses the old Python after installing 3.13, explicitly pin it by adding `--python 3.13` to your config args:
+
+```json
+"args": ["--python", "3.13", "ha-mcp@latest"]
+```
+
 ### SSL certificate errors (self-signed certificates)
 
 If your Home Assistant uses HTTPS with a self-signed certificate or custom CA, you may see SSL verification errors.
--- a/docs/macOS-uv-guide.md
+++ b/docs/macOS-uv-guide.md
@@ -126,6 +126,26 @@ We'd love to hear how you're using ha-mcp!
 - **[GitHub Discussions](https://github.com/homeassistant-ai/ha-mcp/discussions)** — Share your automations, ask questions
 - **[GitHub Issues](https://github.com/homeassistant-ai/ha-mcp/issues)** — Report bugs or request features
 
+## Troubleshooting
+
+### Can't connect to local Home Assistant?
+
+If the demo server works but your local HA (`192.168.x.x`) doesn't:
+
+1. **macOS Local Network Privacy (Sequoia 15+)** — macOS may silently block Claude Desktop subprocesses from accessing local network IPs. Check **System Settings → Privacy & Security → Local Network**. As a workaround, use an SSH tunnel:
+   ```bash
+   ssh -N -L 8123:localhost:8123 user@your-ha-server-ip
+   ```
+   Then set `HOMEASSISTANT_URL` to `http://localhost:8123`.
+
+2. **Firewall software** — Little Snitch, Lulu, or similar tools may block `python`/`node` connections from Claude Desktop. Add firewall rules to allow them.
+
+3. **http:// vs https://** — HA in container mode (Docker, K3s) uses HTTP by default. Use `http://` unless you configured SSL/TLS.
+
+4. **Python 3.13 required** — ha-mcp requires Python 3.13+. Older versions get outdated ha-mcp with known bugs. Run `brew install python@3.13` then `uvx --refresh ha-mcp@latest`. If uvx still picks the old Python, add `"--python", "3.13"` to your config args.
+
+See the [FAQ](FAQ.md) for full details.
+
 ---
 
 Having issues? See the **[FAQ & Troubleshooting Guide](FAQ.md)**.
--- a/site/package-lock.json
+++ b/site/package-lock.json
@@ -6174,9 +6174,9 @@
       }
     },
     "node_modules/vite": {
-      "version": "6.4.1",
-      "resolved": "https://registry.npmjs.org/vite/-/vite-6.4.1.tgz",
-      "integrity": "sha512-+Oxm7q9hDoLMyJOYfUYBuHQo+dkAloi33apOPP56pzj+vsdJDzr+j1NISE5pyaAuKL4A3UD34qd0lx5+kfKp2g==",
+      "version": "6.4.2",
+      "resolved": "https://registry.npmjs.org/vite/-/vite-6.4.2.tgz",
+      "integrity": "sha512-2N/55r4JDJ4gdrCvGgINMy+HH3iRpNIz8K6SFwVsA+JbQScLiC+clmAxBgwiSPgcG9U15QmvqCGWzMbqda5zGQ==",
       "license": "MIT",
       "dependencies": {
         "esbuild": "^0.25.0",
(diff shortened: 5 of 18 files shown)


kingpanther13:
@Patch76 Great catch — you're right that the rename warnings from `ha_rename_entity` were lost in the consolidation. Fixed in 9b10d2b:

1. **Added RENAME LIMITATIONS block** to docstring — covers references not auto-updated, entity history preserved, unique_id and disabled-by-integration restrictions
2. **Added runtime `"warning"` field** in the rename response — LLM agents will see "Remember to update any automations, scripts, or dashboards that reference the old entity_id"
3. **Added rename-specific error suggestions** — when a rename fails, the error now suggests checking for duplicate entity_id and unique_id requirement (was generic "check area_id" before)

Also did a deeper audit and confirmed the voice exposure claim is accurate for current HA versions (test `test_rename_preserves_voice_exposure` passes), and the todo consolidation didn't lose any warnings.

src/ha_mcp/tools/tools_entities.py
```diff
@@ -188,21 +237,67 @@ async def _update_single_entity(
...
 
             entity_entry = result.get("result", {}).get("entity_entry", {})
 
+            # If entity was renamed, update entity_id for subsequent operations
+            if new_entity_id:
+                entity_id = new_entity_id
+
+        # Handle new_device_name — rename the associated device
+        device_rename_result: dict[str, Any] | None = None
+        if new_device_name is not None:
```
sergeykad, line 273: **1. Missing empty-string normalization for `new_device_name`**

The old `ha_rename_entity` normalized empty strings to `None`:
```python
if new_device_name is not None and not new_device_name.strip():
    new_device_name = None
```
This wasn't ported. An empty string `""` now enters this block, triggers a device registry lookup, and sets `name_by_user: None` (line 290) — clearing the device's custom name instead of being a no-op. The test `test_rename_with_empty_device_name_treated_as_entity_only` passes trivially because it asserts `"results" not in rename_data` but the new format never has a `results` key — it should check `"device_rename" not in rename_data`.

src/ha_mcp/tools/tools_entities.py
```diff
@@ -188,21 +237,67 @@ async def _update_single_entity(
...
+            # If no registry update was sent, fetch entity_entry to get device_id
+            if not entity_entry:
+                device_lookup_msg: dict[str, Any] = {
+                    "type": "config/entity_registry/get",
+                    "entity_id": entity_id,
+                }
+                get_result = await client.send_websocket_message(device_lookup_msg)
+                if get_result.get("success"):
```
sergeykad, line 281: **2. Device registry lookup failure indistinguishable from "no device"**

When this WebSocket call fails, the code silently falls through to `device_id = entity_entry.get("device_id")` → `None`, producing the warning "Entity has no associated device." The old code tracked `registry_lookup_failed` and surfaced a distinct message with a retry suggestion. A transient network error shouldn't be reported as "no device."

Add an `else` branch:
```python
else:
    logger.warning(f"Entity registry lookup failed for {entity_id}: {get_result.get('error')}")
    device_rename_result = {
        "warning": "Entity registry lookup failed — could not determine device. Retry may succeed.",
    }
```
Then skip the subsequent `device_id` check when `device_rename_result` is already set.

src/ha_mcp/tools/tools_registry.py
```diff
@@ -20,235 +16,21 @@
...
 )
 
-# Known voice assistant identifiers
-KNOWN_ASSISTANTS = ["conversation", "cloud.alexa", "cloud.google_assistant"]
-
 logger = logging.getLogger(__name__)
 
 
 def register_registry_tools(mcp: Any, client: Any, **kwargs: Any) -> None:
     """Register entity registry and device registry management tools."""
```
sergeykad: **3. Stale function docstring**

Module docstring was correctly updated to "Device Registry management" but this function docstring still says "entity registry and device registry." After removing `ha_rename_entity`, no entity registry tools remain in this module.

src/ha_mcp/tools/tools_search.py
```diff
@@ -841,78 +841,67 @@ async def ha_deep_search(
...
+                        "Check Home Assistant connection",
+                        "Use ha_search_entities() to find correct entity IDs",
+                    ],
+                )
+
+        # Multiple entities path
+        entity_ids: list[str] = entity_id
         MAX_ENTITIES = 100
 
         if not isinstance(entity_ids, list) or not entity_ids:
```
sergeykad, line 890: **4. Bulk validation errors return dicts instead of raising `ToolError`**

Pre-existing, but the consolidation perpetuates it: these three validation checks (empty list, non-string values, max exceeded) return error dicts with `isError=false` instead of raising `ToolError` via `raise_tool_error()`. Per CLAUDE.md: "All tool-level failures must raise ToolError." These are top-level validation failures, not batch-item failures.

src/ha_mcp/tools/tools_todo.py
```diff
@@ -177,289 +187,245 @@ async def ha_get_todo(
...
+                    suggestions=[
+                        "Use ha_get_todo() to find valid todo list entity IDs"
+                    ],
+                )
+            )
 
-            # Call the service
-            result = await client.call_service("todo", "update_item", service_data)
+        # Route: create mode (no item) vs update mode (item provided)
+        if item is None:
```
sergeykad, line 293: **5. Create mode silently ignores `rename` and `status`**

When `item is None` (create mode), `rename` and `status` parameters are silently dropped — the code never uses them. An LLM calling `ha_set_todo_item("todo.list", summary="Buy milk", status="completed")` would expect the item to be created as completed, but `status` is ignored. Either validate that update-only params aren't passed in create mode, or pass `status` through to the service call.

src/ha_mcp/tools/tools_entities.py
```diff
@@ -396,29 +508,68 @@ async def ha_set_entity(
...
     ) -> dict[str, Any]:
         """Update entity properties in the entity registry.
 
         Allows modifying entity metadata such as area assignment, display name,
-        icon, enabled/disabled state, visibility, aliases, labels, and voice
-        assistant exposure in a single call.
+        icon, enabled/disabled state, visibility, aliases, labels, voice
+        assistant exposure, and entity_id rename in a single call.
 
         BULK OPERATIONS:
```
sergeykad, line 546: **6. BULK OPERATIONS section omits `categories`**

Says "only labels and expose_to parameters are supported" for bulk, but `categories` is also supported in bulk (it's not in the `single_entity_params` dict at line 632). Pre-existing but the docstring was touched by this PR.

src/ha_mcp/tools/tools_todo.py
```diff
@@ -177,289 +187,245 @@ async def ha_get_todo(
...
-            Field(
-                description="Optional detailed description for the item",
-                default=None,
-            ),
-        ] = None,
-        due_date: Annotated[
             str | None,
             Field(
-                description="Optional due date in YYYY-MM-DD format (e.g., '2024-12-25')",
+                description="Item text/name. Required when creating a new item. "
```
sergeykad, line 207: **7. `summary` parameter description should clarify update mode behavior**

Description says "Not required when updating an existing item" but doesn't say what happens if it IS provided during update. It's silently ignored (the update path uses `rename` to change names). Consider adding "Ignored in update mode — use rename to change the item name."

src/ha_mcp/tools/tools_entities.py
```diff
@@ -396,29 +508,68 @@ async def ha_set_entity(
...
+
         Use ha_search_entities() or ha_get_device() to find entity IDs.
         Use ha_config_get_label() to find available label IDs.
 
         EXAMPLES:
         Single entity:
         - Assign to area: ha_set_entity("sensor.temp", area_id="living_room")
-        - Rename: ha_set_entity("sensor.temp", name="Living Room Temperature")
+        - Rename display name: ha_set_entity("sensor.temp", name="Living Room Temperature")
+        - Rename entity_id: ha_set_entity("light.old_name", new_entity_id="light.new_name")
```
sergeykad, line 579: **8. Rename examples missing combined `name` + `new_entity_id`**

The old `ha_rename_entity` examples showed combining `name` and `icon` with rename. A common workflow is renaming both entity_id and friendly name at once. Add an example:
`- Rename entity_id with friendly name: ha_set_entity("sensor.old", new_entity_id="sensor.new", name="New Name")`

tests/src/unit/test_tools_get_states.py
```diff
@@ -40,21 +41,25 @@ def mock_smart_tools(self):
 
     @pytest.fixture
     def get_states_tool(self, mock_mcp, mock_client, mock_smart_tools):
-        """Register tools and return the ha_get_states function."""
+        """Register tools and return the ha_get_state function."""
```
sergeykad, line 46: **9. No unit test for single-entity path**

The unit tests only cover the bulk (list) path. The PR introduces an `isinstance(entity_id, str)` branch with its own error handling and return structure. A unit test for the single-entity path would catch regressions in the most-used tool in the system.

src/ha_mcp/tools/tools_entities.py
kingpanther13: Fixed in 3ea9065 — added empty-string normalization for `new_device_name`. Empty/whitespace strings are now normalized to `None` (no-op) before the device rename block.

src/ha_mcp/tools/tools_entities.py
kingpanther13: Fixed in 3ea9065 — registry lookup failure now logs a warning and produces a distinct `device_rename_result` with retry suggestion, instead of falling through to the 'no device' path.

src/ha_mcp/tools/tools_registry.py
kingpanther13: Fixed in 3ea9065 — docstring now reads "Register device registry management tools."

src/ha_mcp/tools/tools_search.py
kingpanther13: Fixed in 3ea9065 — all three bulk validation checks now use `raise_tool_error()` instead of returning error dicts. Updated corresponding test assertions to expect `ToolError`.

src/ha_mcp/tools/tools_todo.py
kingpanther13: Fixed in 3ea9065 — create mode now validates that `rename` and `status` are not passed, raising `ToolError` with a clear message directing to the update path.

src/ha_mcp/tools/tools_entities.py
kingpanther13: Fixed in 3ea9065 — BULK OPERATIONS docstring now reads "only labels, expose_to, and categories parameters are supported."

src/ha_mcp/tools/tools_todo.py
kingpanther13: Fixed in 3ea9065 — description now says "Ignored in update mode — use 'rename' to change the item name."

src/ha_mcp/tools/tools_entities.py
kingpanther13: Fixed in 3ea9065 — added example: `ha_set_entity("sensor.old", new_entity_id="sensor.new", name="New Name")`

tests/src/unit/test_tools_get_states.py
kingpanther13: Fixed in 3ea9065 — added `TestHaGetStateSingleEntity` class with tests for successful single-entity fetch and not-found error path.

PR #878, opened 2026-04-05
fix: add missing get_entity_state mock to group unit tests

sergeykad:
## What does this PR do?

Fixes incomplete mock setup in group tool unit tests that caused 3 tests to silently wait for the full 10s timeout each.

The `mock_client` fixture was missing `get_entity_state` as an `AsyncMock`. When group tools called their internal `wait_for_entity_registered()` / `wait_for_entity_removed()`, the mock returned a non-awaitable `MagicMock`, which was caught by a broad `except Exception` and silently retried for 10 seconds.

**Changes:**
- Add `get_entity_state = AsyncMock(return_value={"state": "on"})` to the fixture for create/update tests
- Override with `side_effect=HomeAssistantAPIError("Not found", status_code=404)` for the remove test, matching real HA behavior

**Impact:** Group unit tests go from ~22s to ~12s (the remaining time is Python/module startup overhead).

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 💥 Breaking change

## Testing

## Checklist

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -87,6 +87,7 @@ jobs:
         echo "🚀 Running full E2E test suite with 3 workers..."
         uv run pytest tests/src/e2e/ \
           -n3 \
+          --dist loadscope \
           --tb=short \
           -v
         echo "✅ Full E2E test suite passed"
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -438,7 +438,10 @@ an isolated Docker HA instance — Docker daemon must be running.
 
 ```bash
 # Run FULL E2E suite (required before claiming all tests pass)
-cd tests && uv run pytest src/e2e/ -v --tb=short
+# -n2 is optimal locally (each worker spins up its own HA container;
+# more workers add memory pressure without proportional speedup).
+# CI uses -n3 tuned for 2-vCPU GitHub runners with 15GB RAM.
+cd tests && uv run pytest src/e2e/ -n2 --dist loadscope -v --tb=short
 
 # Run specific file (partial coverage only — never substitute for full suite)
 cd tests && uv run pytest src/e2e/workflows/automation/test_lifecycle.py -v
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,7 +7,7 @@ Thank you for your interest in contributing!
 1. **Fork and clone** the repository
 2. **Install**: `uv sync --group dev`
 3. **Install hooks**: `uv run lefthook install`
-4. **Test**: `uv run pytest tests/src/e2e/ -v` (requires Docker)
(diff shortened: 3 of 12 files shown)


kingpanther13 (changes requested):
## Review: fix: add missing get_entity_state mock to group unit tests

The mock fix is correct and well-scoped — the root cause analysis, mock values, and import paths all check out. But this PR fixes the symptom while the disease remains: the broad `except Exception` in the wait helpers is what made this bug invisible in the first place, and it will do the same for the next missing mock or programming error.

---

### Important

**1. The broad `except Exception` in wait helpers is the root cause — fix it here**
`src/ha_mcp/tools/util_helpers.py` lines ~296-297 and ~338-339

`wait_for_entity_registered` and `wait_for_entity_removed` both have:
```python
except Exception as e:
    logger.debug(f"Unexpected error polling {entity_id}: {e}")
```

This catch-all sits after specific handlers for `HomeAssistantAPIError` and connection/auth errors. It catches everything else — including `TypeError` (the exact bug this PR exposed), `AttributeError`, `KeyError`, `RuntimeError` — at `debug` level, then silently retries for 10 seconds.

This is what allowed the missing mock to hide: `await MagicMock()` raised `TypeError`, the broad catch swallowed it 33 times over 10 seconds, the function returned `False`, and the caller in `tools_groups.py` turned it into a vague "not yet queryable" warning on a `success: True` response.

**In production, this same pattern will mask real bugs.** If `get_entity_state` has a programming error, returns unexpected data, or the client is in a broken state, the user gets a 10-second delay and a misleading warning instead of a clear error.

Suggested fix: narrow the catch to expected transient exceptions only, and let programming errors (`TypeError`, `AttributeError`) propagate:
```python
except (requests.RequestException, OSError, asyncio.TimeoutError) as e:
    logger.debug(f"Transient error polling {entity_id}: {e}")
```

Or at minimum, log at `warning` level instead of `debug` so these are visible in default configurations.

**2. `wait_for_state_change` has the same pattern**
`src/ha_mcp/tools/util_helpers.py` line ~411-412

Same broad `except Exception` at `debug` level. Same fix applies.

---

### Suggestion (non-blocking)

**3. Consider adding a comment to the `mock_client` fixture**
`tests/src/unit/test_tools_groups.py` line ~23

A brief comment explaining `get_entity_state` is needed for the wait helpers would prevent the same regression if someone modifies the fixture later:
```python
client.get_entity_state = AsyncMock(return_value={"state": "on"})  # Required by wait_for_entity_registered/removed
```

---

### Strengths
- Correct root cause analysis of the 10s timeout per test
- Mock values accurately model real HA behavior (200 with state dict for exists, 404 for removed)
- Clean fixture-level default with per-test override pattern
- No other unit test files have this gap (verified: test_tools_config_scripts, test_wait_parameter, test_wait_helpers all already mock get_entity_state)
- Minimal, focused change

PR #882, opened 2026-04-05
fix: parallelize deep_search Tier 3 config fetches (closes #879)

### LINKED ISSUE #879: deep_search misses automations when bulk config fetch fails and Tier 3 fallback times out
## Description

`ha_deep_search` fails to find automations that reference a searched entity in their actions/conditions when:

1. Tier 1 (REST bulk) and Tier 2 (WebSocket bulk) config fetches fail (timeout or unsupported endpoint)
2. Tier 3 (individual fetch) prioritizes automations by **name-match score** — automations whose *name* doesn't match the query are deprioritized
3. The 15-second time budget expires before low-priority automations are fetched

### Reproduction

Search for an entity that is referenced only inside an automation's conditions/actions, but not in the automation's name or entity_id. If bulk fetch fails and Tier 3 kicks in, the automation is deprioritized by name score and its config is never fetched or searched — resulting in 0 matches despite the entity being present.

### Root Cause

The hardcoded timeout constants in `smart_search.py` (lines 20-27) are not configurable:

| Constant | Value |
|----------|-------|
| `BULK_REST_TIMEOUT` | 5.0s |
| `BULK_WEBSOCKET_TIMEOUT` | 3.0s |
| `INDIVIDUAL_CONFIG_TIMEOUT` | 5.0s |
| `AUTOMATION_CONFIG_TIME_BUDGET` | 15.0s |
| `SCRIPT_CONFIG_TIME_BUDGET` | 10.0s |

### Suggestions

1. **Make timeouts configurable** via environment variables
2. **Increase default budgets** — 15s is tight for instances with many automations
3. **Improve Tier 3 strategy** — fetch configs in parallel instead of sequentially, or remove the name-score prioritization (the whole point of deep_search is to find matches *inside* configs, not by name)
4. **Log/warn** when Tier 3 budget is exhausted and automations were skipped

### Environment

- ha-mcp version: 7.0.0

kingpanther13:
## What does this PR do?

Fixes `ha_deep_search` missing automations/scripts when bulk config fetch fails (Tier 1/2) and the Tier 3 fallback times out before fetching all configs.

**Root cause:** Tier 3 fetched configs sequentially, prioritized by name-match score. Automations whose names didn't match the query were deprioritized and skipped when the time budget expired — even if the entity was referenced inside their conditions/actions.

**Fix (3 changes):**

1. **Parallel batch fetching** — Replace sequential `for` loop with `asyncio.gather` batches of 10. This fetches 10x more configs in the same time window.

2. **Remove name-score prioritization for fetch order** — All configs are now fetched without priority ordering. Name score is still used for result ranking after configs are retrieved. Deep search's purpose is to find matches *inside* configs, so deprioritizing by name defeats the purpose.

3. **Configurable time budgets** — Budgets are now configurable via `HAMCP_AUTOMATION_CONFIG_TIME_BUDGET` (default: 30s, was 15s) and `HAMCP_SCRIPT_CONFIG_TIME_BUDGET` (default: 20s, was 10s). Higher defaults are safe since parallel fetching completes faster.

**Additional:** Logs a warning with fetch/skip counts when the budget is exhausted, so users can diagnose incomplete results.

Closes #879

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 💥 Breaking change

## Testing

## Checklist

### DIFF
--- a/src/ha_mcp/tools/smart_search.py
+++ b/src/ha_mcp/tools/smart_search.py
@@ -27,15 +27,19 @@ INDIVIDUAL_CONFIG_TIMEOUT = 5.0  # Timeout for individual config fetches
 # Time budgets for fallback individual fetching (in seconds).
 # Configurable via env vars for instances with many automations/scripts.
 def _env_float(key: str, default: float) -> float:
+    raw = os.environ.get(key)
+    if raw is None:
+        return default
     try:
-        return float(os.environ.get(key, default))
+        return float(raw)
     except (ValueError, TypeError):
+        logger.warning(f"Invalid value for {key}={raw!r}, using default {default}")
         return default
 
 AUTOMATION_CONFIG_TIME_BUDGET = _env_float("HAMCP_AUTOMATION_CONFIG_TIME_BUDGET", 30.0)
 SCRIPT_CONFIG_TIME_BUDGET = _env_float("HAMCP_SCRIPT_CONFIG_TIME_BUDGET", 20.0)
 
-# Batch size for parallel individual config fetches (Tier 3 fallback)
+# Batch size for parallel individual config fetches (Attempt C fallback)
 INDIVIDUAL_FETCH_BATCH_SIZE = 10
 
 
@@ -913,7 +917,7 @@ class SmartSearchTools:
                             )
 
                 # Attempt C: Parallel individual REST calls with time budget (LAST RESORT)
-                # Fetch ALL configs in parallel batches — don't prioritize by name score.
+                # Fetch configs in parallel batches (subject to time budget) — don't prioritize by name score.
                 # Name score is only used for result ranking, not fetch order, because
                 # deep_search's purpose is to find matches INSIDE configs (conditions/actions),
                 # not just by name. Prioritizing by name would skip the configs most likely
@@ -927,6 +931,7 @@ class SmartSearchTools:
                     ]
                     total_to_fetch = len(uids_to_fetch)
                     fetched_count = 0
+                    failed_count = 0
 
                     async def _fetch_automation_config(uid: str) -> tuple[str, dict[str, Any] | None]:
                         try:
@@ -948,23 +953,24 @@ class SmartSearchTools:
                             time.perf_counter() - budget_start
                             > AUTOMATION_CONFIG_TIME_BUDGET
                         ):
-                            skipped = total_to_fetch - fetched_count
+                            skipped = total_to_fetch - fetched_count - failed_count
                             logger.warning(
                                 f"Automation config fetch budget exhausted "
                                 f"({AUTOMATION_CONFIG_TIME_BUDGET}s). "
-                                f"Fetched {fetched_count}/{total_to_fetch}, "
-                                f"skipped {skipped} automations."
+                                f"Fetched {fetched_count}/{total_to_fetch} "
+                                f"({failed_count} failed), skipped {skipped} automations."
                             )
                             break
                         batch = uids_to_fetch[i : i + INDIVIDUAL_FETCH_BATCH_SIZE]
                         batch_results = await asyncio.gather(
                             *[_fetch_automation_config(uid) for uid in batch],
-                            return_exceptions=True,
                         )
-                        for br in batch_results:
-                            if isinstance(br, tuple) and br[1] is not None:
-                                all_automation_configs[br[0]] = br[1]
+                        for uid_result, config_result in batch_results:
+                            if config_result is not None:
+                                all_automation_configs[uid_result] = config_result
                                 fetched_count += 1
+                            else:
+                                failed_count += 1
 
                 # Phase 3: Score with whatever configs we have
                 for entity_id, friendly_name, name_score, unique_id in name_scored:
@@ -1079,6 +1085,7 @@ class SmartSearchTools:
                     ]
                     total_to_fetch = len(sids_to_fetch)
                     fetched_count = 0
+                    failed_count = 0
 
                     async def _fetch_script_config(sid: str) -> tuple[str, dict[str, Any] | None]:
                         try:
@@ -1098,23 +1105,24 @@ class SmartSearchTools:
                             time.perf_counter() - budget_start
                             > SCRIPT_CONFIG_TIME_BUDGET
                         ):
-                            skipped = total_to_fetch - fetched_count
+                            skipped = total_to_fetch - fetched_count - failed_count
                             logger.warning(
                                 f"Script config fetch budget exhausted "
                                 f"({SCRIPT_CONFIG_TIME_BUDGET}s). "
-                                f"Fetched {fetched_count}/{total_to_fetch}, "
-                                f"skipped {skipped} scripts."
+                                f"Fetched {fetched_count}/{total_to_fetch} "
+                                f"({failed_count} failed), skipped {skipped} scripts."
                             )
                             break
                         batch = sids_to_fetch[i : i + INDIVIDUAL_FETCH_BATCH_SIZE]
                         batch_results = await asyncio.gather(
                             *[_fetch_script_config(sid) for sid in batch],
-                            return_exceptions=True,
                         )
-                        for br in batch_results:
-                            if isinstance(br, tuple) and br[1] is not None:
-                                all_script_configs[br[0]] = br[1]
+                        for sid_result, config_result in batch_results:
+                            if config_result is not None:
+                                all_script_configs[sid_result] = config_result
                                 fetched_count += 1
+                            else:
+                                failed_count += 1
 
                 # Phase 3: Score scripts
                 for (
--- a/tests/src/unit/test_deep_search_tier3_parallel.py
+++ b/tests/src/unit/test_deep_search_tier3_parallel.py
@@ -1,9 +1,9 @@
-"""Unit tests for parallel Tier 3 config fetching in deep_search.
+"""Unit tests for parallel Attempt C config fetching in deep_search.
 
-Validates that when bulk config fetches fail (Tier 1 & 2), Tier 3 fetches
-configs in parallel batches without name-score prioritization. This ensures
-entities referenced only inside automation conditions/actions (not in the
-automation name) are still found. Regression test for #879.
+Validates that when bulk config fetches fail (Attempts A & B), Attempt C
+fetches configs in parallel batches without name-score prioritization. This
+ensures entities referenced only inside automation/script conditions/actions
+(not in the name) are still found. Regression test for #879.
 """
 
 import asyncio
@@ -44,8 +44,8 @@ def _make_automation_entities(count: int) -> list[dict]:
     ]
 
 
-class TestTier3ParallelFetch:
-    """Test that Tier 3 fetches configs in parallel without name-score prioritization."""
+class TestAttemptCParallelFetch:
+    """Test that Attempt C fetches configs in parallel without name-score prioritization."""
 
     @pytest.fixture
     def mock_client(self):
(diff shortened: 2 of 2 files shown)


kingpanther13 (commented):
✅ Fixed both in 19f4a1e: test sleep increased to 1.5s so batch execution exceeds budget, and env var parsing wrapped in try/except with fallback to defaults.

src/ha_mcp/tools/smart_search.py
```diff
@@ -903,39 +912,59 @@ async def deep_search(
...
+                                f"Automation config fetch budget exhausted "
+                                f"({AUTOMATION_CONFIG_TIME_BUDGET}s). "
+                                f"Fetched {fetched_count}/{total_to_fetch}, "
+                                f"skipped {skipped} automations."
+                            )
+                            break
+                        batch = uids_to_fetch[i : i + INDIVIDUAL_FETCH_BATCH_SIZE]
+                        batch_results = await asyncio.gather(
+                            *[_fetch_automation_config(uid) for uid in batch],
+                            return_exceptions=True,
```
sergeykad: **1. `return_exceptions=True` is redundant and silently swallows unexpected exceptions**

The inner function `_fetch_automation_config` already catches `Exception` and returns `(uid, None)`, so no exception will ever propagate out of it under normal circumstances. This makes `return_exceptions=True` dead code. Worse, if something truly unexpected escapes the inner catch (e.g. `CancelledError`), `isinstance(br, tuple)` silently skips it with no logging.

This deviates from existing patterns in this codebase — e.g. `tools_entities.py:687` explicitly checks `isinstance(result, BaseException)`, and the dashboard search in this same file checks `isinstance(dash_result, Exception)` and logs at debug level.

Recommendation: Remove `return_exceptions=True` since the inner functions already handle errors gracefully. Truly unexpected exceptions should propagate rather than be silently eaten. The same applies to the identical script section below.

src/ha_mcp/tools/smart_search.py
```diff
@@ -23,11 +24,19 @@
...
 INDIVIDUAL_CONFIG_TIMEOUT = 5.0  # Timeout for individual config fetches
 
-# Time budgets for fallback individual fetching (in seconds)
-AUTOMATION_CONFIG_TIME_BUDGET = (
-    15.0  # Max time for fetching automation configs individually
-)
-SCRIPT_CONFIG_TIME_BUDGET = 10.0  # Max time for fetching script configs individually
+# Time budgets for fallback individual fetching (in seconds).
+# Configurable via env vars for instances with many automations/scripts.
+def _env_float(key: str, default: float) -> float:
```
sergeykad, line 29: **2. `_env_float` silently ignores invalid environment variable values**

If a user sets `HAMCP_AUTOMATION_CONFIG_TIME_BUDGET=60s` (note the trailing `s`), this silently falls back to the default with zero feedback. The user believes they configured a custom budget but the system ignores their explicit action.

Also, `os.environ.get(key, default)` passes a float as the default to `os.environ.get`, which means `float(default)` is called unnecessarily when the env var is not set.

Suggested fix:
```python
def _env_float(key: str, default: float) -> float:
    raw = os.environ.get(key)
    if raw is None:
        return default
    try:
        return float(raw)
    except (ValueError, TypeError):
        logger.warning(f"Invalid value for {key}={raw!r}, using default {default}")
        return default
```

tests/src/unit/test_deep_search_tier3_parallel.py
```diff
@@ -0,0 +1,197 @@
...
+                "friendly_name": f"Automation {i}",
+                "id": f"uid_{i}",
+            },
+        }
+        for i in range(count)
+    ]
+
+
+class TestTier3ParallelFetch:
+    """Test that Tier 3 fetches configs in parallel without name-score prioritization."""
```
sergeykad: **3. Script section has zero test coverage**

The identical parallel-batch pattern was applied to both automations and scripts in the production code, but all tests only exercise the automation path. The script path has a structurally different tuple format — `(entity_id, friendly_name, script_id, name_score)` vs automations `(entity_id, friendly_name, name_score, unique_id)` — and uses `client.get_script_config(sid)` with a `.get("config", {})` unwrap instead of `client._request("GET", url)`. A bug in the script-specific tuple unpacking or API call would go undetected.

At minimum, one test mirroring `test_tier3_fetches_all_configs_not_just_name_matches` but using `search_types=["script"]` would catch script-specific regressions.

src/ha_mcp/tools/smart_search.py
```diff
@@ -903,39 +912,59 @@ async def deep_search(
                                 f"Automation WebSocket bulk fetch ({ws_type}) failed: {e}"
                             )
 
-                # Attempt C: Individual REST calls with time budget (LAST RESORT)
-                # Prioritize name-matched automations so we at least get their configs
+                # Attempt C: Parallel individual REST calls with time budget (LAST RESORT)
+                # Fetch ALL configs in parallel batches — don't prioritize by name score.
```
sergeykad: **4. Comment says "Fetch ALL configs" but time budget can truncate**

This comment says "Fetch ALL configs in parallel batches" which is misleading since the time budget a few lines below can stop fetching early. The word "ALL" implies completeness that isn't guaranteed.

Suggestion: "Fetch configs in parallel batches (subject to time budget)"

src/ha_mcp/tools/smart_search.py
```diff
@@ -23,11 +24,19 @@
...
+def _env_float(key: str, default: float) -> float:
+    try:
+        return float(os.environ.get(key, default))
+    except (ValueError, TypeError):
+        return default
+
+AUTOMATION_CONFIG_TIME_BUDGET = _env_float("HAMCP_AUTOMATION_CONFIG_TIME_BUDGET", 30.0)
+SCRIPT_CONFIG_TIME_BUDGET = _env_float("HAMCP_SCRIPT_CONFIG_TIME_BUDGET", 20.0)
+
+# Batch size for parallel individual config fetches (Tier 3 fallback)
```
sergeykad: **5. Terminology mismatch: "Tier 3" vs "Attempt C"**

This comment uses "Tier 3" but the inline code comments in this same file use "Attempt A/B/C" (see lines 915, 1072). The test file also uses "Tier" throughout. Future maintainers searching for "Tier" will not find the production code labels, and vice versa.

Pick one terminology and use it consistently — either update all references to "Attempt C" or define the Tier/Attempt mapping once at the top.

src/ha_mcp/tools/smart_search.py
```diff
@@ -903,39 +912,59 @@ async def deep_search(
...
+                            break
+                        batch = uids_to_fetch[i : i + INDIVIDUAL_FETCH_BATCH_SIZE]
+                        batch_results = await asyncio.gather(
+                            *[_fetch_automation_config(uid) for uid in batch],
+                            return_exceptions=True,
+                        )
+                        for br in batch_results:
+                            if isinstance(br, tuple) and br[1] is not None:
+                                all_automation_configs[br[0]] = br[1]
+                                fetched_count += 1
```
sergeykad, line 971: **6. Budget warning conflates "not attempted" with "attempted and failed"**

`fetched_count` is only incremented for successful fetches. The budget-exhaustion warning reports `skipped = total_to_fetch - fetched_count`, which conflates items that were never attempted (budget ran out) with items that were attempted but failed (HTTP errors, timeouts). A user seeing "Fetched 5/30, skipped 25" can't tell whether to increase the budget or investigate HA health.

Suggestion: Track `failed_count` separately and include it in the warning:
```python
f"Fetched {fetched_count}/{total_to_fetch} ({failed_count} failed), skipped {skipped}"
```

tests/src/unit/test_deep_search_tier3_parallel.py
```diff
@@ -0,0 +1,197 @@
...
+        call_count = 0
+
+        async def _slow_fetch(method: str, url: str) -> dict:
+            nonlocal call_count
+            uid = url.split("/")[-1]
+            # First call triggers bulk fetch failure
+            if url.rstrip("/") == "/config/automation/config":
+                raise Exception("Bulk unavailable")
+            call_count += 1
+            await asyncio.sleep(1.5)  # Simulate slow fetches
```
sergeykad: **7. Time-dependent test is fragile**

This test uses real `asyncio.sleep(1.5)` with a 2-second budget. The assertion `call_count >= 10` assumes the first batch of 10 parallel 1.5-second sleeps completes within the budget. On a slow CI machine, event loop scheduling overhead could push the batch past 2 seconds, causing a flaky failure.

Consider either making the sleep much shorter (e.g. 0.01s) with a proportionally shorter budget, or mocking `time.perf_counter` to make the test deterministic.

tests/src/unit/test_deep_search_tier3_parallel.py
```diff
@@ -0,0 +1,197 @@
...
+            "ha_mcp.tools.smart_search.AUTOMATION_CONFIG_TIME_BUDGET", 2.0
+        ):
+            result = await smart_tools.deep_search(
+                query="test",
+                search_types=["automation"],
+                limit=10,
+            )
+
+        # With 2s budget and 1.5s per fetch (parallel batches of 10),
+        # batch 1 (t=0→1.5s) completes under budget, batch 2 (t=1.5→3.0s)
```
sergeykad: **8. Assertion comment is imprecise**

The comment says "batch 2 (t=1.5→3.0s) may start" but batch 2 *will* start because 1.5 < 2.0 (the budget check happens *before* launching each batch). The hedging language is misleading.

Suggestion: "batch 1 (t=0→1.5s) and batch 2 (t=1.5→3.0s) both complete because the budget check occurs before each batch launch. Batch 3 at t=3.0s exceeds the 2.0s budget. Expect exactly 20 fetched."

src/ha_mcp/tools/smart_search.py
kingpanther13: Fixed in 848fbbb — removed `return_exceptions=True` from both automation and script sections. Inner functions already catch all exceptions, so this was dead code. Now unpacking tuples directly.

src/ha_mcp/tools/smart_search.py
kingpanther13: Fixed in 848fbbb — `_env_float` now returns early when env var is not set, and logs a warning with the invalid value on parse failure.

tests/src/unit/test_deep_search_tier3_parallel.py
kingpanther13: Fixed in 848fbbb — added `TestAttemptCScriptParallelFetch` with a test that verifies script configs are fetched via `get_script_config()` and matches inside sequences are found.

src/ha_mcp/tools/smart_search.py
kingpanther13: Fixed in 848fbbb — now reads "Fetch configs in parallel batches (subject to time budget)".

src/ha_mcp/tools/smart_search.py
kingpanther13: Fixed in 848fbbb — renamed all references to Attempt C to match the inline code comments. Test class and method names updated too.

src/ha_mcp/tools/smart_search.py
kingpanther13: Fixed in 848fbbb — tracking `failed_count` separately, warning now shows "Fetched X/Y (Z failed), skipped N".

tests/src/unit/test_deep_search_tier3_parallel.py
kingpanther13: Fixed in 848fbbb — replaced 1.5s sleep with 0.01s and budget from 2.0s to 0.005s. Same logic, deterministic on CI.

tests/src/unit/test_deep_search_tier3_parallel.py
kingpanther13: Fixed in 848fbbb — removed hedging, comment now states the exact batch execution behavior.

Issue #837, opened 2026-03-28
[BUG] OAuth disk persistence does not survive Docker container restarts


julienld:
### Related

Follow-up to #787 / #790. The disk persistence fix in #790 writes to `~/.ha-mcp/oauth_state.json` inside the container filesystem, which is ephemeral by default. Container restarts still lose all OAuth state.

### Root Cause

`HomeAssistantOAuthProvider` defaults `state_dir` to `Path.home() / ".ha-mcp"` (i.e. `/root/.ha-mcp` in Docker). No `VOLUME` is declared in the Dockerfile, and `__main__.py` does not pass a `state_dir` to the constructor. The HA add-on has the same issue unless `/root/.ha-mcp` happens to be on a persistent volume managed by Supervisor.

### Possible Fixes

**Option A — Fully stateless tokens (recommended)**
Make both access and refresh tokens stateless by encoding the HA LLAT in each (same base64 pattern already used for access tokens). Set long expiry aligned with the LLAT itself (10 years in HA). On refresh, decode the LLAT from the old refresh token and issue new tokens — no server-side map needed.

- Access token: `base64({"ha_token": "...", "iat": ..., "type": "access"})` — long expiry (e.g. 10 years)
- Refresh token: `base64({"ha_token": "...", "iat": ..., "type": "refresh"})` — long expiry
- Eliminates `_refresh_to_access_map`, `_save_state()`, `_load_state()`, and all disk I/O
- Survives container restarts by design — zero server-side state for token management
- Compatible with ChatGPT (which requires refresh tokens and refreshes on every tool call regardless of expiry)
- Security boundary unchanged: if the LLAT is revoked in HA, both token types fail immediately on the next API call

Clients still need to re-register (DCR) after container restart, but DCR is automatic and transparent.

**Option B — Volume mount + state_dir configuration**
Declare a `VOLUME` in Dockerfile and/or pass `state_dir=/data` for the HA add-on. Document volume mount for standalone Docker.

- Pros: Minimal code change
- Cons: Requires user to configure volume mount correctly; add-on needs to use Supervisor persistent storage path; doesn't eliminate the complexity of disk persistence code

### Recommendation

Option A aligns with the existing stateless access token design and eliminates all server-side persistence. The LLAT is already 10 years in HA — the OAuth token expiry should match rather than impose an artificial 1-hour boundary that forces complex refresh state management.

PR #893, opened 2026-04-06
fix: fully stateless OAuth tokens, drop HOMEASSISTANT_TOKEN requirement

### LINKED ISSUE #886: [BUG] OAuth still requires HOMEASSISTANT_TOKEN
### 💬 What Went Wrong?

Running the server in OAuth mode still requires the `HOMEASSISTANT_TOKEN` variable, contrary to the documentation which mentions:

`HOMEASSISTANT_TOKEN is NOT required in OAuth mode. Each user provides their own Long-Lived Access Token via the consent form.`

### 📋 AI-Generated Content / Logs / Config

```markdown
2026-04-06 11:00:48 ha_mcp.__main__ INFO: OAuth mode logging configured at INFO level
2026-04-06 11:00:48 ha_mcp.auth.provider INFO: HomeAssistantOAuthProvider initialized with base_url=https://home-assistant-mcp.example.com
2026-04-06 11:00:48 ha_mcp.__main__ ERROR: OAuth server error: 1 validation error for Settings
HOMEASSISTANT_TOKEN
  Value error, Home Assistant token must be provided [type=value_error, input_value='', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/value_error
```

### 🖥️ Client Application

Claude Desktop

### 📦 Installation Method

Docker (ghcr.io/homeassistant-ai/ha-mcp)

### 💻 Operating System

Linux

### OS Version

Debian 13

### 📖 Setup Guide Followed

https://github.com/homeassistant-ai/ha-mcp/blob/master/docs/OAUTH.md

### 🔄 Exact Steps You Followed

Follow setup docs on https://github.com/homeassistant-ai/ha-mcp/blob/master/docs/OAUTH.md

### 🐍 Python Version

_No response_

### 📦 ha-mcp Version

_No response_

### 🏠 Home Assistant Version

_No response_

### 💡 Additional Context

_No response_

kingpanther13:
## What does this PR do?

Fixes #886 and #837 by making the OAuth provider fully stateless — both access and refresh tokens are now self-contained base64 JSON, eliminating all server-side token storage and disk persistence.

**#886 — OAuth still requires HOMEASSISTANT_TOKEN:** `main_oauth()` now sets the `OAUTH_MODE_TOKEN` sentinel when `HOMEASSISTANT_TOKEN` is empty/unset, so `Settings` validation passes without a server-level token. This is a one-line fix in the entrypoint.

**#837 — OAuth state lost on container restart:** Refresh tokens are now stateless (encoded with `ha_token`, `client_id`, `scopes`, `exp`, and `type`), matching the existing stateless access token pattern. This removes `_save_state()`, `_load_state()`, `_refresh_to_access_map`, the `state_dir` parameter, and `oauth_state.json` entirely. Tokens survive container restarts by design — clients re-register via DCR automatically and transparently.

**Security model unchanged:** The LLAT is the authorization boundary. Revoking it in Home Assistant immediately invalidates all derived tokens. Server-side revocation was previously a no-op in practice (in-memory state lost on restart anyway).

**Backwards compatibility:** `load_access_token()` accepts tokens with `type=access` or no `type` field (old format). Refresh tokens from old format (random strings stored server-side) won't survive a restart — which was already the bug being fixed.

**Net result:** −226 lines, zero disk I/O, container-restart-safe.

## Type of change
- [x] 🐛 Bug fix

## Testing

## Checklist

### DIFF
--- a/tests/src/unit/test_oauth.py
+++ b/tests/src/unit/test_oauth.py
@@ -327,17 +327,16 @@ class TestHomeAssistantOAuthProvider:
         assert result is None
 
     @pytest.mark.asyncio
-    async def test_load_access_token_backwards_compat(self, provider):
-        """Test that tokens without a type field are accepted (backwards compat)."""
+    async def test_unsigned_token_rejected(self, provider):
+        """Unsigned tokens (no HMAC signature) must be rejected."""
         from base64 import urlsafe_b64encode
 
-        # Manually create a token without "type" field (old format)
-        payload = {"ha_token": "old_format_token", "iat": int(time.time())}
-        old_token = urlsafe_b64encode(json.dumps(payload).encode()).decode().rstrip("=")
+        # Manually create an unsigned token (old format, no sig envelope)
+        payload = {"ha_token": "forged_token", "iat": int(time.time())}
+        unsigned_token = urlsafe_b64encode(json.dumps(payload).encode()).decode().rstrip("=")
 
-        result = await provider.load_access_token(old_token)
-        assert result is not None
-        assert result.claims["ha_token"] == "old_format_token"
+        result = await provider.load_access_token(unsigned_token)
+        assert result is None
 
     @pytest.mark.asyncio
     async def test_load_invalid_access_token(self, provider):


src/ha_mcp/auth/provider.py
```diff
@@ -744,42 +632,27 @@ async def exchange_refresh_token(
...
             )
 
-        # Revoke old tokens
-        self._revoke_internal(refresh_token_str=refresh_token.token)
-
-        # Issue new stateless access token with recovered HA credentials
-        new_access_token_value = self._encode_credentials(ha_token)
-        new_refresh_token_value = f"ha_refresh_{secrets.token_hex(32)}"
-
+        ha_token = payload["ha_token"]
```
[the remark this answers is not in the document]
kingpanther13: Not applicable — `_decode_token()` already guarantees `ha_token` is present in any non-None return value (line 182: `if not payload.get("ha_token"): return None`). So when execution reaches `payload["ha_token"]` on line 643, the key is guaranteed to exist.

The suggested change would merge two distinct error conditions ("can't decode" vs "missing ha_token") into one message, making debugging harder. Keeping them separate is better.

src/ha_mcp/auth/provider.py
```diff
@@ -179,133 +177,15 @@ def _decode_credentials(self, token: str) -> str | None:
                 token += "=" * padding
 
             decoded = urlsafe_b64decode(token.encode()).decode()
-            payload = json.loads(decoded)
+            payload: dict[str, Any] = json.loads(decoded)
```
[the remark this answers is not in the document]
kingpanther13: Good catch on the json.loads typing. Fixed in 976b03d — added `isinstance(payload, dict)` check before `.get()`. A crafted token that base64-decodes to a JSON array or scalar would have raised `AttributeError` (not in the except clause).

The except clause already catches the specific exceptions (`binascii.Error`, `json.JSONDecodeError`, `UnicodeDecodeError`) — no broad `except Exception` here.

src/ha_mcp/auth/provider.py
```diff
@@ -129,48 +121,54 @@ def __init__(
...
-        # Persistent state file
-        self._state_file = Path(state_dir or Path.home() / ".ha-mcp") / "oauth_state.json"
-        self._load_state()
-
         logger.info(f"HomeAssistantOAuthProvider initialized with base_url={base_url}")
 
-    def _encode_credentials(self, ha_token: str) -> str:
-        """
-        Encode HA token into a stateless access token.
+    def _encode_token(
```
sergeykad, line 141: **1. Refresh tokens now expose the LLAT in plaintext base64.**

Previously refresh tokens were opaque random strings (`ha_refresh_{hex}`), so a leaked refresh token only allowed getting a new access token through the server — it did not directly expose the LLAT. Now refresh tokens contain the raw LLAT in base64, directly extractable by anyone.

Refresh tokens are longer-lived (7 days) and stored more persistently by clients (browser storage, token databases), making them higher-value leak targets than short-lived access tokens.

This is a conscious design tradeoff for statelessness, but it reduces defense-in-depth. Worth considering: could refresh tokens carry an HMAC of the LLAT (using a server-side secret) instead of the raw credential? This would preserve statelessness (server can recompute the HMAC) without embedding the LLAT in a second token type.

src/ha_mcp/auth/provider.py
```diff
@@ -789,56 +662,40 @@ async def exchange_refresh_token(
...
             token=token,
-            client_id="stateless",  # We don't store client_id in token
-            scopes=["homeassistant", "mcp"],
-            expires_at=None,  # Stateless tokens don't expire
-            claims={
-                "ha_token": ha_token,
-            },
+            client_id=payload.get("client_id", "stateless"),
+            scopes=payload.get("scopes", ["homeassistant", "mcp"]),
+            expires_at=None,  # LLAT revocation is the security boundary
```
sergeykad: **2. `expires_in` is advertised but never enforced on access tokens.**

`exchange_authorization_code` and `exchange_refresh_token` both return `expires_in=ACCESS_TOKEN_EXPIRY_SECONDS` (1 hour) in the OAuthToken response, telling clients the token expires. But here, `load_access_token` sets `expires_at=None` and never checks `exp` — the server will accept the token forever.

This actively misinforms clients. Either:
- Encode `exp` in access tokens and check it here (enforce expiry), or
- Set `expires_in=None` in the OAuthToken response (stop advertising expiry)

The current mismatch means clients will unnecessarily refresh tokens every hour while the server would have accepted the old one indefinitely.

src/ha_mcp/auth/provider.py
```diff
@@ -789,56 +662,40 @@ async def exchange_refresh_token(
...
-            # so there is no server-side cascade to revoke. The paired
-            # refresh token is only revoked when revoke_token() is called
-            # with a RefreshToken argument directly.
-            self.access_tokens.pop(access_token_str, None)
-
-        if refresh_token_str:
-            self.refresh_tokens.pop(refresh_token_str, None)
-            self._refresh_to_access_map.pop(refresh_token_str, None)
-
     async def revoke_token(self, token: AccessToken | RefreshToken) -> None:
```
sergeykad, line 743: **3. Token revocation is a complete no-op with no logging.**

Clients calling `/revoke` get a success response but the token remains fully valid. There is no audit trail — operators cannot see that revocation was requested and ignored.

At minimum, add logging so revocation attempts are visible:
```python
async def revoke_token(self, token: AccessToken | RefreshToken) -> None:
    logger.debug(
        "Token revocation requested (no-op for stateless tokens; "
        "revoke the LLAT in Home Assistant to invalidate)"
    )
```

Also worth noting the RFC 7009 implication in the docstring: the `/revoke` endpoint returns success per spec, but the token remains valid until the underlying LLAT is revoked in Home Assistant.

src/ha_mcp/__main__.py
```diff
@@ -710,6 +714,15 @@ def main_oauth() -> None:
     Note: HOMEASSISTANT_TOKEN is NOT required in this mode.
     Per-user tokens are collected via the OAuth consent form.
     """
+    # In OAuth mode, per-user tokens come from the consent form — no
+    # server-level HOMEASSISTANT_TOKEN is needed.  Set the sentinel so
+    # Settings validation passes even when the env var is empty (e.g.
+    # Dockerfile sets HOMEASSISTANT_TOKEN="").  Fixes #886.
+    if not os.getenv("HOMEASSISTANT_TOKEN"):
```
sergeykad, line 721: **4. No test coverage for the `main_oauth()` sentinel fix (#886).**

This is the core bug fix for issue #886 — setting `OAUTH_MODE_TOKEN` when `HOMEASSISTANT_TOKEN` is empty. It has zero test coverage. This is a production startup path; if it regresses, OAuth mode fails to start entirely in Docker deployments.

Please add a test that:
1. Sets `HOMEASSISTANT_TOKEN=""` in the environment
2. Calls the sentinel-setting logic
3. Verifies `os.environ["HOMEASSISTANT_TOKEN"] == OAUTH_MODE_TOKEN`
4. Also verifies that when `HOMEASSISTANT_TOKEN` already has a real value, it is not overwritten.

src/ha_mcp/auth/provider.py
```diff
@@ -789,56 +662,40 @@ async def exchange_refresh_token(
...
+        payload = self._decode_token(token)
+        if not payload:
+            return None
 
-        # Decode token to get HA token
-        ha_token = self._decode_credentials(token)
-        if not ha_token:
+        # Reject refresh tokens presented as access tokens
+        token_type = payload.get("type")
+        if token_type == "refresh":
```
sergeykad, line 721: **5. Silent `None` returns in security-sensitive token validation.**

This line rejects a refresh token presented as an access token — a security-relevant event that could indicate client misconfiguration or an attack. It produces no log output.

Similarly, `load_refresh_token` silently returns `None` for client_id mismatch (possible token theft) and `_decode_token` silently returns `None` for structurally invalid payloads (token corruption).

Recommendation:
```python
if token_type == "refresh":
    logger.warning("Refresh token presented as access token (rejected)")
    return None
```

And in `load_refresh_token`, log client_id mismatches at warning level.

src/ha_mcp/auth/provider.py
```diff
@@ -676,62 +559,67 @@ async def exchange_authorization_code(
...
         )
 
-        # Map refresh token to access token so we can recover HA credentials on refresh
-        self._refresh_to_access_map[refresh_token_value] = access_token_value
-
-        # Clean up temporary credentials storage (no longer needed after token issued)
-        if client.client_id in self.ha_credentials:
-            del self.ha_credentials[client.client_id]
+        # Clean up temporary credentials (no longer needed after token issued)
+        del self.ha_credentials[client.client_id]
```
sergeykad: **6. Bare `del` can raise `KeyError` in a race condition.**

The code checks `self.ha_credentials.get(client.client_id)` earlier and raises `TokenError` if not found. But between the `get` and this `del`, another concurrent request could consume the credentials. The previous code used `if client.client_id in self.ha_credentials: del ...` which was marginally safer (though still racy).

Use `self.ha_credentials.pop(client.client_id, None)` instead.

src/ha_mcp/auth/provider.py
```diff
@@ -789,56 +662,40 @@ async def exchange_refresh_token(
...
-            self._revoke_internal(refresh_token_str=token.token)
-        self._save_state()
+        """Revoke a token (no-op for stateless tokens).
+
+        With fully stateless tokens there is no server-side state to
+        remove.  The LLAT itself is the security boundary — revoking
+        it in Home Assistant immediately invalidates all derived tokens.
+        """
 
     def get_ha_credentials(self, client_id: str) -> HomeAssistantCredentials | None:
```
sergeykad: **7. `get_ha_credentials` docstring describes the old architecture.**

The docstring says "used by the MCP server to get the HA token for making API calls on behalf of the authenticated user." In the new stateless architecture, HA tokens are extracted from the bearer token itself (via `load_access_token` → `claims["ha_token"]`). The `ha_credentials` dict is now only a transient store between consent form submission and token exchange.

Suggested update:
```python
"""Get temporarily stored HA credentials for a client.

Only valid between consent form submission and token exchange.
After token issuance, credentials are deleted from this store.
"""
```

src/ha_mcp/auth/provider.py
```diff
@@ -181,131 +179,13 @@ def _decode_credentials(self, token: str) -> str | None:
             decoded = urlsafe_b64decode(token.encode()).decode()
             payload = json.loads(decoded)
 
-            ha_token: str | None = payload.get("ha_token")
-
-            if ha_token:
-                return str(ha_token)
-            return None
+            if not isinstance(payload, dict) or not payload.get("ha_token"):
+                return None
```
sergeykad, line 197: **8. No test for `_decode_token` with non-dict JSON payload.**

The `isinstance(payload, dict)` guard added in 976b03d (fixing Gemini's review feedback) is untested. A token like `base64('[1,2,3]')` or `base64('"just_a_string"')` would exercise this check. Please add a test case for this distinct code path.

src/ha_mcp/auth/provider.py
```diff
@@ -129,48 +121,54 @@ def __init__(
...
+        token_type: str = "access",
+        client_id: str | None = None,
+        scopes: list[str] | None = None,
+        expires_at: int | None = None,
+    ) -> str:
+        """Encode a stateless OAuth token as base64 JSON.
+
+        Both access and refresh tokens are self-contained: the HA LLAT,
+        token type, and (for refresh tokens) client/scope metadata are
+        encoded directly.  No server-side storage is needed.
```
sergeykad: **10. `_encode_token` docstring omits `exp` from the description.**

The docstring says "the HA LLAT, token type, and (for refresh tokens) client/scope metadata are encoded directly" but `expires_at` → `exp` is also specific to refresh tokens and not mentioned.

Suggested update: "(for refresh tokens) client/scope metadata and expiry"

src/ha_mcp/auth/provider.py
```diff
@@ -789,56 +662,40 @@ async def exchange_refresh_token(
         )
 
     async def load_access_token(self, token: str) -> AccessToken | None:
-        """
-        Load and validate access token.
+        """Decode and validate a stateless access token.
 
-        STATELESS: Decodes token to extract HA token.
-        No server-side storage needed - token is self-contained base64-encoded JSON.
+        Accepts tokens with ``type=access`` or no type field (backwards
```
sergeykad, line 710: **11. Backwards compatibility comment should include a version/date anchor.**

This note about accepting tokens without a `type` field is valuable now during the transition, but could mislead future maintainers long after all pre-type-field tokens have expired (access tokens are advertised as 1-hour, refresh tokens are 7 days).

Add a reference so a future maintainer can evaluate whether this path is still needed, e.g.:
```
Accepts tokens with ``type=access`` or no type field (backwards
compatibility with tokens issued before v6.x / April 2026).
```

sergeykad (changes requested):
**9. `to_dict` docstring says "for storage" but storage is removed** (`provider.py:55`)

This method's docstring reads `"""Convert to dictionary for storage."""` but the PR removes all disk persistence. The method appears unused outside `test_credentials_to_dict`. Either remove it or update the docstring to `"""Convert to dictionary representation."""`. Also: the `validated_at` field (line 53) is set but never read anywhere in the codebase — consider removing it.

src/ha_mcp/auth/provider.py
sergeykad: Tying items 1-3 together: they share a root cause — tokens are unsigned base64, so expiry, scopes, and revocation are all unenforceable. Any party with the LLAT can forge arbitrary tokens with any claims.

Adding HMAC signing (using a server-generated secret) would fix all three without sacrificing the stateless architecture:
- LLAT no longer directly extractable from tokens (wrapped in HMAC, not plaintext)
- Expiry becomes enforceable (forged `exp` values rejected)
- Scope forgery blocked
- Token revocation becomes possible via a lightweight deny-list keyed by token signature

This is a single change to `_encode_token` / `_decode_token` that makes the rest of the security model meaningful. Without it, the expiry check, scope validation, and type discrimination in this PR are all cosmetic — an attacker who intercepts any token can extract the LLAT and forge whatever they want.

kingpanther13: Fixed in 16c2ac6 — but with an important design note.

Your suggestion was to HMAC the LLAT so refresh tokens don't expose it. However, that approach doesn't work for stateless refresh: during `exchange_refresh_token`, the server needs the raw LLAT to issue new access tokens. If the refresh token only contains an HMAC (a one-way hash), the LLAT is unrecoverable — you'd need server-side storage mapping HMAC→LLAT, which defeats statelessness.

What I implemented instead: **HMAC-signed refresh tokens**. The refresh token is an envelope `{"payload": {...}, "sig": "..."}` where `sig` is `HMAC-SHA256(payload_json, server_secret)`. The raw LLAT remains in the payload (necessary for token exchange), but the signature prevents tampering with any field — an attacker can't modify `client_id`, `scopes`, `exp`, or the LLAT itself without invalidating the signature.

This preserves statelessness (server recomputes the signature to verify, no storage needed) and adds integrity protection. The server secret is generated per-instance at startup, so refresh tokens from previous instances are rejected (acceptable since clients re-authenticate via consent form).

Access tokens remain unsigned (short-lived, 1 hour) since they're the working credential anyway.

src/ha_mcp/auth/provider.py
kingpanther13: Fixed in 16c2ac6 — access tokens now encode `exp` and `load_access_token` enforces it. Both `exchange_authorization_code` and `exchange_refresh_token` set `expires_at=int(time.time() + ACCESS_TOKEN_EXPIRY_SECONDS)` on access tokens.

src/ha_mcp/auth/provider.py
kingpanther13: Fixed in 16c2ac6 — added debug logging to `revoke_token` and RFC 7009 note in docstring explaining that the token remains valid until the LLAT is revoked in HA.

src/ha_mcp/__main__.py
kingpanther13: Fixed in 16c2ac6 — added `TestOAuthMainSentinel` with two tests: empty token gets sentinel, existing token not overwritten. Tests exercise the sentinel logic directly.

src/ha_mcp/auth/provider.py
kingpanther13: Fixed in 16c2ac6 — added warning logs for refresh-token-as-access-token rejection and client_id mismatch in `load_refresh_token`.

src/ha_mcp/auth/provider.py
kingpanther13: Fixed in 16c2ac6 — replaced `del self.ha_credentials[client.client_id]` with `.pop(client.client_id, None)`.

src/ha_mcp/auth/provider.py
kingpanther13: Fixed in 16c2ac6 — docstring now reads "Get temporarily stored HA credentials for a client. Only valid between consent form submission and token exchange."

src/ha_mcp/auth/provider.py
kingpanther13: Fixed in 16c2ac6 — added `TestDecodeTokenEdgeCases` with tests for JSON array payload and JSON string payload, both correctly rejected.

src/ha_mcp/auth/provider.py
kingpanther13: Fixed in 16c2ac6 — docstring now describes signing and exp: "Refresh tokens are HMAC-signed... Access tokens contain the raw HA LLAT (short-lived)."

src/ha_mcp/auth/provider.py
kingpanther13: Fixed in 16c2ac6 — backwards compat comment now anchored: "tokens issued before v7.x / April 2026".

src/ha_mcp/auth/provider.py
kingpanther13: You're right — signing only refresh tokens was insufficient. Fixed in c88cfa8: ALL tokens are now HMAC-signed.

Both access and refresh tokens use the same signed envelope format: `{"payload": {...}, "sig": HMAC-SHA256(payload, secret)}`. This means:
- Expiry (`exp`) is enforceable — forged values invalidate the signature
- Scope and client_id forgery blocked
- LLAT extraction from a token is possible (base64, not encrypted) but forging a *new* token with modified claims is not

Tokens no longer survive provider restart (new secret each startup). Updated the restart tests to verify tokens are *rejected* by a new instance, and added a same-instance chained refresh test.

The backwards-compat path still accepts unsigned tokens (pre-v7.x) with a debug log, so existing clients aren't broken during the transition.

src/ha_mcp/auth/provider.py
```diff
@@ -179,133 +191,36 @@ def _decode_credentials(self, token: str) -> str | None:
...
+                    json.dumps(payload, sort_keys=True).encode()
+                )
+                if not hmac.compare_digest(outer["sig"], expected_sig):
+                    logger.warning("Token signature verification failed")
+                    return None
+                if not payload.get("ha_token"):
+                    return None
+                return payload
+
+            # Unsigned payload (backwards compat with tokens issued before
```
sergeykad: **12. Remove the unsigned token backwards-compat path entirely.**

The HMAC secret is regenerated on every server restart (`secrets.token_bytes(32)` in `__init__`), which means all pre-existing tokens — signed or unsigned — are already invalidated on restart. So the only scenario where an unsigned token reaches a running server is:

1. The server has been running continuously since before the upgrade (no restart) — but the upgrade itself requires a restart
2. An attacker crafts a plain base64 JSON blob with `ha_token` and no `sig`/`payload` envelope to bypass HMAC verification

Scenario 1 can't happen. Scenario 2 is the exact attack that HMAC signing was added to prevent.

This code path should be removed. Reject all tokens without a valid signature:
```python
# Remove lines 214-218 entirely. The `return None` at the end
# of the method already handles this case.
```

kingpanther13: Fixed in 5e8f4db — removed the unsigned backwards-compat path entirely. All tokens without a valid HMAC signature are now rejected.

PR #908, opened 2026-04-07
chore: bump HA test image to 2026.4.1 and improve test stabilization

kingpanther13:
## What does this PR do?

Bumps the Home Assistant Docker image used in all test suites from `2026.1.3` to `2026.4.1` (current stable) and fixes test infrastructure to handle HA 2026.4's slower entity initialization.

### Image bump
- Updates HA Docker image version across 7 files (2 workflow YAML + 5 Python test files)
- Centralizes the version in `tests/test_constants.py` as `HA_TEST_IMAGE` so future bumps only need one Python change (workflow YAML files still duplicate it since they can't import Python)

### Test stabilization for HA 2026.4
HA 2026.4 reports 80+ components loaded while individual integrations are still registering entities and WebSocket handlers. Tests that run immediately after component stabilization hit 404s on `sun.sun` or "Unknown command" on helper WebSocket calls.

- **Added entity registration wait** after component count stabilization: polls `/api/states` until 50+ entities are registered (demo integration alone creates 60+), ensuring lights, helpers, and sun.sun are all queryable before tests start
- **Increased stabilization timeout** from 20s to 30s
- **Made `test_simple_connection` resilient**: falls back to `zone.home` if `sun.sun` not available (known flaky across HA versions)

### Verified via HA Core changelogs
Reviewed all commits to `sun`, `demo`, `default_config`, `input_boolean/number/select/text`, `collection.py`, `core.py`, and `setup.py` between 2026.1 and 2026.4. No functional API changes — the failures are purely initialization timing from demo modernization commits (async conversion).

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [x] 🔧 Maintenance/refactor
- [ ] 💥 Breaking change

## Testing

## Checklist

### DIFF
--- a/tests/src/e2e/basic/test_connection.py
+++ b/tests/src/e2e/basic/test_connection.py
@@ -18,21 +18,10 @@ async def test_simple_connection(mcp_client):
     logger.info("☀️ Testing basic MCP connection with sun.sun entity")
 
     # Test a simple tool that doesn't use WebSocket - just get state.
-    # Try sun.sun first (default_config), fall back to zone.home (always exists).
-    from ..utilities.assertions import safe_call_tool
-
-    result_data = await safe_call_tool(
-        mcp_client, "ha_get_state", {"entity_id": "sun.sun"}
+    # sun.sun is part of default_config and guaranteed by conftest entity wait.
+    result = await mcp_client.call_tool(
+        "ha_get_state", {"entity_id": "sun.sun"}
     )
-    if result_data.get("success"):
-        result = await mcp_client.call_tool(
-            "ha_get_state", {"entity_id": "sun.sun"}
-        )
-    else:
-        logger.info("sun.sun not available, falling back to zone.home")
-        result = await mcp_client.call_tool(
-            "ha_get_state", {"entity_id": "zone.home"}
-        )
 
     # Parse and verify the result using standard assertion utility
     data = assert_mcp_success(result, "Get state request")
--- a/tests/src/e2e/conftest.py
+++ b/tests/src/e2e/conftest.py
@@ -45,7 +45,7 @@ from .utilities.assertions import parse_mcp_result
 
 # Import test constants
 sys.path.insert(0, str(Path(__file__).parent.parent.parent))
-from test_constants import TEST_TOKEN
+from test_constants import HA_TEST_IMAGE, TEST_TOKEN
 
 # Configure logging for tests
 logging.basicConfig(level=logging.INFO)
@@ -262,8 +262,6 @@ def ha_container_with_fresh_config():
     )
 
     # Create testcontainer with port configuration
-    from test_constants import HA_TEST_IMAGE
-
     container = DockerContainer(HA_TEST_IMAGE)
 
     # Check for custom port via environment variable
@@ -434,14 +432,15 @@ def ha_container_with_fresh_config():
                             f"waiting for more..."
                         )
                         last_entity_count = entity_count
-            except requests.exceptions.RequestException:
-                pass
+            except (requests.exceptions.RequestException, json.JSONDecodeError) as exc:
+                logger.debug(f"Entity registration check failed: {exc}")
             time.sleep(1)
         else:
-            logger.warning(
-                f"⚠️ Entity registration timed out after "
-                f"{ENTITY_STABILIZATION_TIMEOUT}s with only "
-                f"{last_entity_count} entities — some tests may fail"
+            pytest.fail(
+                f"Entity registration timed out after "
+                f"{ENTITY_STABILIZATION_TIMEOUT}s. "
+                f"Only {last_entity_count} entities registered "
+                f"(minimum: {MIN_ENTITIES}). Check Docker logs."
             )
 
         # Store connection info for other fixtures


tests/src/e2e/basic/test_connection.py
```diff
@@ -17,11 +17,22 @@ async def test_simple_connection(mcp_client):
...
+
+    result_data = await safe_call_tool(
+        mcp_client, "ha_get_state", {"entity_id": "sun.sun"}
     )
+    if result_data.get("success"):
+        result = await mcp_client.call_tool(
+            "ha_get_state", {"entity_id": "sun.sun"}
+        )
+    else:
+        logger.info("sun.sun not available, falling back to zone.home")
```
sergeykad: **1. Double `ha_get_state` call on success path**

`safe_call_tool` makes a full API call for `sun.sun`, then its result is discarded and `mcp_client.call_tool` is called again for the same entity. This is wasteful and introduces a TOCTOU race (first call succeeds, second could fail if entity momentarily disappears).

The root cause is that `assert_mcp_success` expects a raw MCP result object, not the parsed dict from `safe_call_tool`. Consider restructuring to avoid the double call:

```python
try:
    result = await mcp_client.call_tool("ha_get_state", {"entity_id": "sun.sun"})
except Exception:
    logger.info("sun.sun not available, falling back to zone.home")
    result = await mcp_client.call_tool("ha_get_state", {"entity_id": "zone.home"})
```

tests/src/e2e/basic/test_connection.py
```diff
@@ -17,11 +17,22 @@ async def test_simple_connection(mcp_client):
...
+        result = await mcp_client.call_tool(
+            "ha_get_state", {"entity_id": "zone.home"}
+        )
```
sergeykad: **2. Fallback from `sun.sun` to `zone.home` is redundant and masks problems**

The conftest entity wait already ensures 50+ entities are registered before tests run. If it works correctly, `sun.sun` (part of `default_config`) will be available and this fallback is dead code. If it doesn't work (timeout only warns — see separate comment), this fallback masks an infrastructure problem while other tests fail with scattered entity-not-found errors.

A connection test should be simple and deterministic. Pick the right entity and assert it exists — don't add probe-then-fallback logic. If the conftest wait is the real fix (it is), this fallback should be removed.

tests/src/e2e/conftest.py
```diff
@@ -404,6 +404,46 @@ def ha_container_with_fresh_config():
...
+                    if entity_count != last_entity_count:
+                        logger.info(
+                            f"⏳ {entity_count} entities registered, "
+                            f"waiting for more..."
+                        )
+                        last_entity_count = entity_count
+            except requests.exceptions.RequestException:
+                pass
+            time.sleep(1)
+        else:
```
sergeykad, line 438: **3. Entity wait loop timeout warns instead of failing**

The component stabilization loop 40 lines above uses `pytest.fail()` on timeout. This new entity registration loop only uses `logger.warning()`, letting tests proceed into a broken environment. This will cause dozens of confusing failures across the suite instead of one clear message pointing to the root cause.

Should match the existing pattern:

```python
else:
    pytest.fail(
        f"Entity registration timed out after {ENTITY_STABILIZATION_TIMEOUT}s. "
        f"Only {last_entity_count} entities registered (minimum: {MIN_ENTITIES}). "
        f"Check Docker logs."
    )
```

tests/src/e2e/conftest.py
```diff
@@ -404,6 +404,46 @@ def ha_container_with_fresh_config():
...
+                states_resp = requests.get(
+                    f"{base_url}/api/states",
+                    timeout=5,
+                    headers=headers,
+                )
+                if states_resp.status_code == 200:
+                    entity_count = len(states_resp.json())
+                    if entity_count >= MIN_ENTITIES:
+                        logger.info(
+                            f"✅ {entity_count} entities registered "
```
sergeykad, line 425: **4. Entity wait loop silently swallows exceptions**

The component stabilization loop directly above catches exceptions into a variable and logs at debug level, and also catches `json.JSONDecodeError`:

```python
except (requests.exceptions.RequestException, json.JSONDecodeError) as exc:
    logger.debug(f"Stabilization check failed: {exc}")
```

This new loop does bare `pass` with no logging and doesn't catch `JSONDecodeError`. If `/api/states` returns malformed JSON, the loop will crash with an unhandled exception instead of retrying. Should match the existing pattern:

```python
except (requests.exceptions.RequestException, json.JSONDecodeError) as exc:
    logger.debug(f"Entity registration check failed: {exc}")
```

tests/src/e2e/conftest.py
```diff
@@ -262,9 +262,9 @@ def ha_container_with_fresh_config():
     )
 
     # Create testcontainer with port configuration
-    # renovate: datasource=docker depName=ghcr.io/home-assistant/home-assistant
-    HA_IMAGE = "ghcr.io/home-assistant/home-assistant:2026.1.3"
-    container = DockerContainer(HA_IMAGE)
+    from test_constants import HA_TEST_IMAGE
```
sergeykad: **5. Import inside function body inconsistent with existing pattern**

`HA_TEST_IMAGE` is imported inside the fixture function body, but `TEST_TOKEN` is already imported at module level (line 48 of this file). Should add `HA_TEST_IMAGE` to the existing module-level import:

```python
from test_constants import TEST_TOKEN, HA_TEST_IMAGE
```

Same applies to the `safe_call_tool` import in `test_connection.py` — it's inside the test function but `assert_mcp_success` is imported at module level.

tests/src/e2e/basic/test_connection.py
kingpanther13: Fixed in 7294b23 — removed the double call entirely. Now uses a single direct `mcp_client.call_tool` for `sun.sun` with no probe.

tests/src/e2e/basic/test_connection.py
kingpanther13: Fixed in 7294b23 — removed the fallback. Conftest entity wait guarantees 50+ entities including `sun.sun`. If entity wait fails, that's the right place to surface the error (now `pytest.fail()` per comment #3).

tests/src/e2e/conftest.py
kingpanther13: Fixed in 7294b23 — entity wait loop now uses `pytest.fail()` on timeout, matching the component stabilization loop.

tests/src/e2e/conftest.py
kingpanther13: Fixed in 7294b23 — now catches `(RequestException, json.JSONDecodeError)` and logs at debug level, matching the component stabilization pattern.

tests/src/e2e/conftest.py
kingpanther13: Fixed in 7294b23 — `HA_TEST_IMAGE` moved to module-level import alongside `TEST_TOKEN`.

PR #932, opened 2026-04-09
feat: replace SequenceMatcher fuzzy search with BM25 scoring

### LINKED ISSUE #851: [FEATURE] Replace SequenceMatcher fuzzy search with BM25 scoring for search tools

Replace the current `difflib.SequenceMatcher`-based fuzzy matching in search tools with BM25 (Best Matching 25) scoring for the fuzzy search path (`exact_match=False`). This would affect the 4 tools that currently do text-based corpus search:

1. **`ha_search_entities`** — biggest win. Multi-word queries like "kitchen temperature sensor" would get proper term weighting.
2. **`ha_deep_search`** — the recursive `_search_in_dict` would benefit from term-aware scoring instead of `calculate_partial_ratio` on every string leaf.
3. **`ha_hacs_search`** — moderate benefit. Would add typo tolerance and better ranking (currently pure substring, zero fuzzy).
4. **`ha_list_services`** — minor benefit. Would add ranking and typo tolerance to what is currently binary substring matching.

Tools that do structural matching (e.g., `ha_dashboard_find_card` with wildcard/exact entity matching) would **not** be affected — those are doing the right thing already.

### Current implementation problems

The current fuzzy engine (`utils/fuzzy_search.py`) uses Python's `difflib.SequenceMatcher` — a character-level LCS algorithm. For each entity it computes 7 SequenceMatcher calls (`calculate_ratio`, `calculate_partial_ratio`, `calculate_token_sort_ratio` × entity_id and friendly_name, plus domain ratio).

**Problem 1: Additive scoring inflates everything.** Scores are additive across match categories. A search for "light" against `light.kitchen_light` ("Kitchen Light") accumulates: +100 (exact domain) + 85 (substring entity_id) + 80 (substring name) + weighted fuzzy ratios. Scores easily exceed 200+. With a threshold of 60, nearly everything with a vague resemblance passes.

**Problem 2: No term importance (IDF).** Searching "kitchen light" weights "light" the same as "kitchen" — but "light" appears in hundreds of entities while "kitchen" is the discriminating term. The algorithm has no concept of which terms are rare/important.

**Problem 3: `calculate_partial_ratio` is expensive and noisy.** It slides a window of query-length across the target and runs SequenceMatcher for each position — O(n³) per entity. Short queries match short substrings easily, producing high scores for irrelevant entities.

**Problem 4: The `exact_match=True` default (added in #814) was a band-aid.** The fuzzy search returned too many false positives, so exact substring was made the default. This effectively disables fuzzy capability unless the caller explicitly opts in — which means the LLM loses typo tolerance and synonym handling in the default path.

**Problem 5: `ha_hacs_search` and `ha_list_services` have no fuzzy at all.** HACS search uses its own additive substring scoring (name: +100, full_name: +50, description: +30, authors: +20). Services search is pure binary substring — no scoring, no ranking. A typo in either tool returns zero results.

### What BM25 solves

- **Built-in IDF** — "kitchen" automatically gets more weight than "light" because it appears in fewer entity names. Eliminates the need for hardcoded keyword boosting (though those lists should be kept as additional signal).
- **Term frequency saturation** — BM25's `k1` parameter controls diminishing returns for repeated terms.
- **Document length normalization** — short entity names and long friendly names compared fairly via the `b` parameter.
- **Predictable score range** — scores are bounded and comparable, making threshold-based filtering meaningful.
- **Much cheaper** — tokenization + inverted index lookup. O(n) per query term per document, not O(n³).

## Why do you need this?

When an LLM uses `ha_search_entities` with `exact_match=False` (or when it should be the default), the current fuzzy search returns too many irrelevant results and ranks them poorly. For example, searching "kitchen temperature" returns every sensor and every entity with "kitchen" in the name scored equally, because there's no concept of which terms are discriminating.

This directly impacts LLM tool-calling accuracy — the agent either gets too much noise to pick the right entity, or we default to exact substring matching and lose all typo/synonym tolerance.

BM25 is already proven in this codebase: FastMCP's `BM25SearchTransform` powers `ha_search_tools` (the tool catalogue search). Extending BM25 to entity/config/HACS/service search would bring the same quality to HA data search.

### Current search inventory

| Tool | Current Matching | Fuzzy? | BM25 Benefit |
|------|-----------------|--------|-------------|
| `ha_search_entities` | SequenceMatcher fuzzy (opt-in) + exact substring (default) | Yes (opt-in) | **High** — multi-word queries, term weighting |
| `ha_deep_search` | SequenceMatcher fuzzy (opt-in) + recursive `calculate_partial_ratio` on config dicts | Yes (opt-in) | **High** — config search with term awareness |
| `ha_hacs_search` | Additive substring scoring (100/50/30/20) | No | **Medium** — adds typo tolerance, better ranking |
| `ha_list_services` | Binary substring on concatenated string, no scoring | No | **Low** — adds ranking and typo tolerance |
| `ha_dashboard_find_card` | Exact + wildcard + heading substring | No | Not needed — structural matching |
| `get_entities_by_area()` | `calculate_partial_ratio`, threshold 80 | Yes | Low — short strings, strict threshold works |

### Implementation options

**Option A: Reuse FastMCP's `BM25SearchTransform`**

FastMCP already ships `BM25SearchTransform` which powers `ha_search_tools`. This could potentially be adapted for entity search. Pros: no new dependencies, proven code. Cons: it's designed for tool descriptions (longer text), not entity names (short, delimiter-separated). Tokenization may need customization since entity IDs use `_` and `.` as delimiters rather than spaces.

**Option B: Lightweight BM25 scorer in `fuzzy_search.py`**

Implement a standalone BM25 scorer (~40-50 lines) in `utils/fuzzy_search.py`. Tokenize entity IDs on `.` and `_`, friendly names on spaces. Build an inverted index, score with the standard BM25 formula. Pros: full control over tokenization (critical for entity ID format), no coupling to FastMCP internals, can tune `k1`/`b` parameters for short-document search. Cons: new code to maintain (though BM25 is a well-understood algorithm).

### Proposed architecture (either option)

| Layer | Purpose | When |
|-------|---------|------|
| **Tier 1: Exact substring** | Fast path for known IDs | `exact_match=True` (default, unchanged) |
| **Tier 2: BM25 keyword search** | Primary fuzzy search for multi-word queries | `exact_match=False` (replaces SequenceMatcher) |
| **Tier 3: Token-level edit distance** | Typo correction for unmatched tokens | Fallback when BM25 returns nothing |

- **Keep `exact_match` parameter** — exact substring remains the default, BM25 activates on `exact_match=False`
- **Keep hardcoded keyword boost lists** — apply as BM25 query expansion or post-scoring boost
- **Keep existing defaults** (limit 10 for entity search, 5 for deep search, threshold 60)
- **Keep `calculate_partial_ratio` / `calculate_ratio`** as tier-3 fallback for single-token typo correction when BM25 returns nothing

### Related

- #814 — Added `exact_match=True` default to search tools (the band-aid this would properly fix)
- FastMCP's `BM25SearchTransform` — existing BM25 in the dependency tree

kingpanther13:
## What does this PR do?

Replaces the `difflib.SequenceMatcher`-based fuzzy matching in `ha_search_entities` and `ha_deep_search` with BM25 (Best Matching 25) scoring for the fuzzy search path (`exact_match=False`). This is the implementation for #851.

**Key improvements:**
- **Multi-word queries work when terms are non-adjacent** — e.g., `"dryer override"` now finds automations where both terms exist in different config fields (the strongest use case from issue discussion)
- **IDF term weighting** — rare/discriminating terms like "kitchen" get more weight than common terms like "light"
- **Noise reduction** — completely unrelated queries return 0 results instead of hundreds of false positives from character-level sliding window matches
- **Unified tokenization** — `"tesla_ble"` and `"tesla ble"` produce identical token lists, eliminating underscore/space inconsistencies

**What changed:**
- `src/ha_mcp/utils/fuzzy_search.py`: Added `BM25Scorer` class (~50 lines) and `tokenize()` function. `FuzzyEntitySearcher.search_entities()` now uses BM25 as primary scorer with SequenceMatcher as tier-3 typo fallback
- `src/ha_mcp/tools/smart_search.py`: `_search_in_dict()` fuzzy path collects all string leaves from config dicts, tokenizes into a single BM25 document, and scores against query tokens. Exact match path is completely unchanged

**Scope:** Limited to `ha_search_entities` and `ha_deep_search` per maintainer feedback — HACS/services search deferred pending benchmarks.

**Architecture (3-tier, as proposed in #851):**
| Tier | Method | When |
|------|--------|------|
| 1 | Exact substring | `exact_match=True` (default, unchanged) |
| 2 | BM25 keyword search | `exact_match=False` (replaces SequenceMatcher) |
| 3 | Token-level SequenceMatcher | Fallback when BM25 returns 0 (typo correction) |

Closes #851

## Type of change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 💥 Breaking change

## Testing

**Tests added:**
- 24 unit tests (`tests/src/unit/test_bm25_search.py`): tokenizer, BM25Scorer, FuzzyEntitySearcher BM25 integration, `_search_in_dict` BM25 path
- 5 e2e tests (`tests/src/e2e/tools/test_bm25_fuzzy_search.py`): multi-word fuzzy search, noise reduction, underscore/space equivalence, deep search fuzzy multi-word, exact match regression

## Checklist


src/ha_mcp/tools/smart_search.py
```diff
@@ -1429,53 +1434,97 @@ def _search_in_dict(
...
+            # Normalise: max possible score ≈ sum of IDF for all query tokens
+            # With 1 document and all tokens present, IDF = log(1.5) ≈ 0.405 each.
+            # Scale to 0-100 range relative to a perfect match.
+            max_possible = sum(scorer._idf.get(t, 0.0) for t in query_tokens)
+            if max_possible > 0:
+                return min(100, round(raw / max_possible * 100))
+            return 100  # all query tokens have IDF=0 but scored > 0 — shouldn't happen
+
+        # Tier-3 fallback: token-level SequenceMatcher for typos
+        from ..utils.fuzzy_search import calculate_ratio
```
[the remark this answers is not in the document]
kingpanther13: Fixed in 45da20a — moved `calculate_ratio` to the top-level import block.

kingpanther13:
Here are BAT test results using gemini. Please note that the story 1 stories are affected negatively by the container still starting up as the story/test was running, so that one can't really be trusted. Overall there's an improvement across the board:  

                                                                                          
  # Master vs PR932 — Filtered Story-by-Story Comparison                                                                                                                                                                                          
                                                                                                                                                                                                                                                  
  **PR #932**: `feat: replace SequenceMatcher fuzzy search with BM25 scoring` (changes `ha_search_entities` and `ha_deep_search` fuzzy path — tool signatures unchanged, only internal scoring changed)
                                                                                                                                                                                                                                                  
  **Only stories that passed on both master AND pr932 for the same model are included.**                                                                                                                                                          
                                                                                                                                                                                                                                                  
  ### gemini-3-flash-preview                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                  
  Stories included: s01, s02, s03, s04, s05, s10, s12, s13 (8 stories)
  Excluded: s11 (master failed)                                                                                                                                                                                                                   
                                              
  | Story | Title | Master dur | PR932 dur | Δ dur | Master tools | PR932 tools | Δ tools | Master billable | PR932 billable | Δ billable |
  |---|---|---|---|---|---|---|---|---|---|---|
  | s01 | Create sunset/sunrise lights automation | 47.4s | 82.1s | +34.7s | 5 | 5 | +0 | 103,077 | 56,198 | −46,879 |                                                                                                                            
  | s02 | Create motion-activated light with timeout | 40.4s | 44.7s | +4.3s | 5 | 5 | +0 | 65,485 | 50,377 | −15,108 |                                                                                                                           
  | s03 | Debug why automation didn't fire | 62.1s | 39.2s | −22.9s | 10 | 4 | −6 | 64,895 | 18,657 | −46,238 |                                                                                                                                   
  | s04 | Create room overview dashboard | 106.2s | 32.3s | −73.9s | 10 | 5 | −5 | 62,559 | 19,098 | −43,461 |                                                                                                                                    
  | s05 | Discover and explore available entities | 45.6s | 46.9s | +1.3s | 7 | 11 | +4 | 58,159 | 57,983 | −176 |                                                                                                                                
  | s10 | Create areas and organize entities | 32.6s | 33.6s | +1.0s | 7 | 8 | +1 | 45,790 | 20,506 | −25,284 |                                                                                                                                   
  | s12 | Create and assign labels to entities | 29.6s | 21.6s | −8.1s | 6 | 4 | −2 | 48,346 | 14,319 | −34,027 |                                                                                                                                 
  | s13 | Find automations and scripts that reference an entity | 26.6s | 18.8s | −7.8s | 1 | 1 | +0 | 13,151 | 12,962 | −189 |                                                                                                                   
  | **TOTAL (8 stories)** | | **390.5s** | **319.0s** | **−71.5s** | **51** | **43** | **−8** | **461,462** | **250,100** | **−211,362** |                                                                                                        
                                                                                                                                                                                                                                                  
  - Δ duration: **−71.5s** (−18.3%)                                                                                                                                                                                                               
  - Δ tool calls: **−8** (−15.7%)                                                                                                                                                                                                                 
  - Δ billable tokens: **−211,362** (**−45.8%**)                                                                                                                                                                                                  
                                                                                                                                                                                                                                                  
  ### gemini-3-pro-preview                                                                                                                                                                                                                        
                  
  Stories included: all 9 (none failed on either side)                                                                                                                                                                                            
                                          
  | Story | Title | Master dur | PR932 dur | Δ dur | Master tools | PR932 tools | Δ tools | Master billable | PR932 billable | Δ billable |                                                                                                       
  |---|---|---|---|---|---|---|---|---|---|---|                                                                                                                                                                                                   
  | s01 | Create sunset/sunrise lights automation | 51.5s | 116.6s | +65.1s | 1 | 3 | +2 | 46,810 | 55,363 | +8,553 |
  | s02 | Create motion-activated light with timeout | 70.8s | 72.4s | +1.6s | 5 | 5 | +0 | 58,683 | 66,081 | +7,398 |                                                                                                                            
  | s03 | Debug why automation didn't fire | 46.1s | 33.3s | −12.8s | 5 | 4 | −1 | 53,297 | 48,345 | −4,952 |
  | s04 | Create room overview dashboard | 98.5s | 48.1s | −50.4s | 4 | 7 | +3 | 93,436 | 62,934 | −30,502 |                                                                                                                                      
  | s05 | Discover and explore available entities | 48.2s | 40.3s | −8.0s | 5 | 12 | +7 | 52,826 | 39,151 | −13,675 |                                                                                                                             
  | s10 | Create areas and organize entities | 53.6s | 32.5s | −21.1s | 5 | 5 | +0 | 51,191 | 44,144 | −7,047 |                                                                                                                                   
  | s11 | Analyze entity history and patterns | 35.0s | 31.8s | −3.2s | 3 | 3 | +0 | 48,515 | 48,116 | −399 |                                                                                                                                     
  | s12 | Create and assign labels to entities | 33.8s | 30.1s | −3.7s | 4 | 4 | +0 | 27,115 | 48,073 | +20,958 |                                                                                                                                 
  | s13 | Find automations and scripts that reference an entity | 22.3s | 20.4s | −1.9s | 1 | 1 | +0 | 45,144 | 13,980 | −31,164 |                                                                                                                
  | **TOTAL (9 stories)** | | **459.8s** | **425.4s** | **−34.5s** | **33** | **44** | **+11** | **477,017** | **426,187** | **−50,830** |                                                                                                        
                                                                                                                                                                                                                                                  
  - Δ duration: **−34.5s** (−7.5%)                                                                                                                                                                                                                
  - Δ tool calls: **+11** (+33.3%)                                                                                                                                                                                                                
  - Δ billable tokens: **−50,830** (−10.7%)                                                                                                                                                                                                       
                                                                                                                                                                                                                                                  
  Quick read                                                                                                                                                                                                                                      
                  
  Flash (8 of 9 stories):                                                                                                                                                                                                                         
  - All metrics improved: −18.3% duration, −15.7% tool calls, −45.8% billable tokens
  - Biggest wins: s04 (−74s, dashboard creation — BM25 search found entities in fewer tries), s03 (−23s, troubleshoot), s01 token drop                                                                                                            
  - s01 is interesting: same tool count but 35s slower yet 47k fewer tokens. Longer latency, smaller payload.                                                                                                                                     
                                                                                                             
  Pro (all 9 stories):                                                                                                                                                                                                                            
  - Mixed but net improvement: −7.5% duration, +33.3% tool calls, −10.7% billable tokens
  - More tool calls but fewer tokens total — pro ran more queries under BM25 but each query's results were shorter/more relevant                                                                                                                  
  - s05 is the standout: tool count +7 (5→12) but ran faster AND used fewer tokens. The new fuzzy path is cheaper per call.                                                                                                                       
  - s13 (the "find automations by entity" story that's pure search) went from 45k to 14k tokens on pro — that's a 69% reduction on a single-call search story, which is exactly where BM25 should shine.

src/ha_mcp/utils/fuzzy_search.py
```diff
@@ -24,59 +124,110 @@ def __init__(self, threshold: int = 60):
...
+
+        # Normalise BM25 scores to 0-100 range for threshold comparison
+        max_raw = max(raw_scores) if raw_scores else 0.0
+        matches: list[dict[str, Any]] = []
+
+        if max_raw > 0:
+            for i, raw in enumerate(raw_scores):
+                if raw <= 0:
+                    continue
+                score = round(raw / max_raw * 100)
```
sergeykad: **Relative normalization makes threshold meaningless as a quality gate**

`raw / max_raw * 100` means the best match **always** scores 100, regardless of actual relevance. If only one entity vaguely shares a single token with the query, it gets score 100 and passes any threshold.

The old `_calculate_entity_score` produced absolute scores; this changes the threshold from "has at least 60% absolute similarity" to "within 60% of the best match." A single weak match inflated to 100 contradicts the PR's stated goal of noise reduction.

**Suggested fix**: Apply a minimum absolute BM25 score floor before normalization. If `max_raw` is below a calibrated minimum (e.g., where less than half of query tokens match), return empty results instead of normalizing weak matches to 100. Alternatively, normalize against theoretical maximum (`sum of IDF * perfect TF`) rather than empirical max.

src/ha_mcp/utils/fuzzy_search.py
```diff
@@ -1,20 +1,120 @@
...
+            tf[t] = tf.get(t, 0) + 1
+
+        total = 0.0
+        for qt in query_tokens:
+            idf = self._idf.get(qt, 0.0)
+            f = tf.get(qt, 0)
+            if f == 0:
+                continue
+            numer = f * (self.k1 + 1)
+            denom = f + self.k1 * (1 - self.b + self.b * dl / self._avgdl)
```
sergeykad, line 105: **`nan` propagation when `_avgdl = 0.0`**

If `fit()` receives entities with empty entity_ids and no friendly_name, `tokenize("")` returns `[]`, making `_avgdl = 0.0`. Float division `0 / 0.0` yields `nan` (not `ZeroDivisionError`), which silently propagates through scoring.

While the `if f == 0: continue` guard prevents this for all-empty corpora, a **mixed** corpus (some empty, some not) with near-zero `_avgdl` produces wildly inflated length normalization.

In `search_entities()`, entities with empty IDs/names produce empty token lists in `docs` — nothing filters them before `scorer.fit(docs)`.

**Suggested fix** — guard in `fit()`:
```python
if self._avgdl == 0.0:
    self._avgdl = 1.0  # neutral length normalization
```

src/ha_mcp/tools/smart_search.py
```diff
@@ -1429,53 +1435,95 @@ def _search_in_dict(
...
+        # Use BM25 with a 1-document corpus (the config dict as a single doc)
+        scorer = BM25Scorer()
+        scorer.fit([doc_tokens])
+        raw = scorer.score(query_tokens, 0)
+
+        if raw > 0:
+            # Normalise: max possible score ≈ sum of IDF for all query tokens
+            # With 1 document and all tokens present, IDF = log(1.5) ≈ 0.405 each.
+            # Scale to 0-100 range relative to a perfect match.
+            max_possible = sum(scorer._idf.get(t, 0.0) for t in query_tokens)
```
sergeykad: **Accessing private `_idf` across module boundaries**

This reaches into `BM25Scorer._idf`, coupling `smart_search.py` to internal implementation details of a class in another module. If `BM25Scorer` refactors its IDF storage, this silently breaks.

**Suggested fix**: Add a public method to `BM25Scorer`:
```python
def max_possible_score(self, query_tokens: list[str]) -> float:
    """Theoretical max BM25 score for these query tokens."""
    return sum(self._idf.get(t, 0.0) for t in query_tokens)
```
Then call `scorer.max_possible_score(query_tokens)` here.

src/ha_mcp/utils/fuzzy_search.py
```diff
@@ -24,59 +124,110 @@ def __init__(self, threshold: int = 60):
...
+                    "score": best_token_score,
+                    "match_type": "typo_fallback",
+                })
+        return results
+
     def _calculate_entity_score(
         self, entity_id: str, friendly_name: str, domain: str, query: str
     ) -> int:
-        """Calculate comprehensive fuzzy score for an entity."""
+        """Calculate comprehensive fuzzy score for an entity (legacy, used by deep_search name scoring)."""
```
sergeykad: **Labeled "legacy" but still has 3 active production callers**

This method is called at lines ~868, ~1024, ~1195 in `smart_search.py` for `ha_deep_search` name scoring. The "legacy" label misleads future contributors into thinking it's safe to remove.

Also creates a scoring mismatch: `ha_search_entities` now uses BM25 (relative 0-100), while `ha_deep_search` name scoring uses additive SequenceMatcher (can exceed 100). The `fuzzy_threshold` setting means different things for each code path.

**Suggested fix**: Either remove the "legacy" label or migrate the remaining callers. At minimum, document that this method is actively used by deep search and intentionally preserved.

src/ha_mcp/tools/smart_search.py
```diff
@@ -1429,53 +1435,95 @@ def _search_in_dict(
...
+            return 0
+
+        # Use BM25 with a 1-document corpus (the config dict as a single doc)
+        scorer = BM25Scorer()
+        scorer.fit([doc_tokens])
+        raw = scorer.score(query_tokens, 0)
+
+        if raw > 0:
+            # Normalise: max possible score ≈ sum of IDF for all query tokens
+            # With 1 document and all tokens present, IDF = log(1.5) ≈ 0.405 each.
```
sergeykad: **Incorrect IDF value in comment**

Comment says `IDF = log(1.5) ≈ 0.405` but the Robertson IDF with n=1, freq=1 yields:
```
(1 - 1 + 0.5) / (1 + 0.5) + 1.0 = 0.5/1.5 + 1.0 = 1.333
log(1.333) ≈ 0.288
```

Incorrect math comments in scoring code mislead anyone debugging unexpected score values.

**Fix**: Correct to `≈ 0.288` or describe normalization conceptually without specific numbers.

src/ha_mcp/tools/smart_search.py
```diff
@@ -1429,53 +1435,95 @@ def _search_in_dict(
...
+            # Scale to 0-100 range relative to a perfect match.
+            max_possible = sum(scorer._idf.get(t, 0.0) for t in query_tokens)
+            if max_possible > 0:
+                return min(100, round(raw / max_possible * 100))
+            return 100  # all query tokens have IDF=0 but scored > 0 — shouldn't happen
+
+        # Tier-3 fallback: token-level SequenceMatcher for typos
+        best = 0
+        for qt in query_tokens:
```
sergeykad, line 1496: **Unbounded O(Q×D) SequenceMatcher fallback**

This iterates all query tokens × all unique doc tokens with `calculate_ratio()` (creates a `SequenceMatcher` per pair). For large config dicts during `ha_deep_search`, this fires for every automation/script/helper where BM25 returns 0.

With 200 automations and large configs, a typo query could trigger thousands of `SequenceMatcher` instantiations with no logging or time budget.

**Suggested fix**: Add `logger.debug` when fallback fires, and consider capping `doc_tokens` or adding an early exit if the loop exceeds a reasonable count.

src/ha_mcp/tools/smart_search.py
```diff
@@ -1429,53 +1435,95 @@ def _search_in_dict(
...
+        raw = scorer.score(query_tokens, 0)
+
+        if raw > 0:
+            # Normalise: max possible score ≈ sum of IDF for all query tokens
+            # With 1 document and all tokens present, IDF = log(1.5) ≈ 0.405 each.
+            # Scale to 0-100 range relative to a perfect match.
+            max_possible = sum(scorer._idf.get(t, 0.0) for t in query_tokens)
+            if max_possible > 0:
+                return min(100, round(raw / max_possible * 100))
+            return 100  # all query tokens have IDF=0 but scored > 0 — shouldn't happen
```
sergeykad: **Unreachable `return 100` with zero observability**

The comment says "shouldn't happen" but if this IS reached (e.g., due to a bug in IDF computation), it silently returns a perfect score — the worst possible fallback.

**Suggested fix**: Add a `logger.warning` so the impossible case becomes observable:
```python
logger.warning(
    "BM25 scored > 0 but max_possible IDF is 0; "
    "query_tokens=%s, doc_tokens_len=%d",
    query_tokens, len(doc_tokens),
)
return 100
```

tests/src/e2e/tools/test_bm25_fuzzy_search.py
```diff
@@ -0,0 +1,233 @@
...
+    )
+    raw_data = assert_mcp_success(result, "Multi-word fuzzy search")
+    data = raw_data.get("data", raw_data)
+
+    assert data.get("success") is True
+    results = data.get("results", [])
+    # BM25 should return results where both terms appear (tokenized matching)
+    # Even if "light kitchen" is not a contiguous substring
+    logger.info(
+        f"Multi-word fuzzy search returned {len(results)} results "
```
sergeykad, line 47: **This test has no meaningful assertion and can never fail**

The test logs the result count but never asserts anything about it. After `assert_mcp_success`, any result (including 0 results) passes. This provides no regression protection for the BM25 multi-word improvement.

**Suggested fix**: Add at minimum:
```python
assert len(results) > 0, "BM25 should return results for multi-word query with matching tokens"
```

tests/src/unit/test_bm25_search.py
```diff
@@ -0,0 +1,273 @@
...
+            {
+                "entity_id": "binary_sensor.garage_door",
+                "attributes": {"friendly_name": "Garage Door"},
+                "state": "closed",
+            },
+        ]
+
+    def test_multi_word_query_ranks_correctly(self, entities):
+        """'kitchen light' should rank kitchen ceiling light first."""
+        searcher = FuzzyEntitySearcher(threshold=30)
```
sergeykad, line 137: **Unit tests use `threshold=30` vs production default of `60`**

Most `TestFuzzyEntitySearcherBM25` tests use threshold=30, much lower than production default (60). Matches that pass in tests may be filtered out in production.

Consider adding at least one test with `threshold=60` (or the default) to validate that the BM25 scoring actually produces results that survive the production threshold.

tests/src/e2e/tools/test_bm25_fuzzy_search.py
```diff
@@ -0,0 +1,233 @@
...
+
+    finally:
+        # Cleanup
+        try:
+            await mcp_client.call_tool(
+                "ha_config_remove_automation",
+                {"entity_id": "automation.bm25_test_load_sharing"},
+            )
+        except Exception:
+            logger.debug("Cleanup of BM25 test automation failed (may not exist)")
```
sergeykad: **Cleanup exception details discarded**

`except Exception:` discards the exception object entirely. If cleanup fails for an unexpected reason (server down, auth expired), subsequent tests may fail with confusing state and this log gives no hint.

**Suggested fix**:
```python
except Exception as e:
    logger.debug("Cleanup of BM25 test automation failed: %s", e)
```

(Same applies to line 233.)

sergeykad (commented):
## PR Review: 10 findings across 4 files

See inline comments for details.

src/ha_mcp/tools/smart_search.py
kingpanther13: Fixed in d0988b9. Corrected the math in the comment (`log(1.333) ≈ 0.288`) and reframed it to describe normalization behavior rather than relying on a specific numeric value.

src/ha_mcp/utils/fuzzy_search.py
kingpanther13: Fixed in d0988b9. `BM25Scorer.fit()` now guards `_avgdl == 0.0` by setting it to 1.0, neutralizing length normalization when the entire corpus tokenizes to empty lists.

src/ha_mcp/utils/fuzzy_search.py
kingpanther13: Fixed in d0988b9. Replaced the misleading 'legacy' label on `_calculate_entity_score` with an explicit docstring noting that `ha_deep_search` name scoring actively depends on it and warning against removal without migrating those call sites.

src/ha_mcp/utils/fuzzy_search.py
kingpanther13: Fixed in d0988b9. `search_entities` now normalizes raw BM25 scores against a theoretical max from `BM25Scorer.max_possible_score()` (sum of per-token IDFs, with the corpus's max IDF as a penalty baseline for query tokens absent from the corpus). A single weak match can no longer be inflated to 100 by empirical normalization, and partial matches with missing tokens are scaled down proportionally. Added unit tests exercising the production threshold=60 at `test_bm25_search.py`.

src/ha_mcp/tools/smart_search.py
kingpanther13: Fixed in d0988b9. Added `BM25Scorer.max_possible_score(query_tokens)` as a public API and replaced the private-attribute access in `smart_search.py`.

tests/src/unit/test_bm25_search.py
kingpanther13: Fixed in d0988b9. Added two tests using the production threshold=60: one asserts that a full multi-token match survives the threshold (`test_production_threshold_passes_full_match`), and the other verifies that a partial match on a common token is not normalized to 100 (`test_production_threshold_filters_partial_match`). The second test actually caught a real bug — the typo fallback was firing on BM25-partial matches and inflating them to 100. Now gated on BM25 returning zero raw scores across the entire corpus.

src/ha_mcp/tools/smart_search.py
kingpanther13: Fixed in d0988b9. Added a `logger.debug` line when `_search_in_dict`'s SequenceMatcher fallback fires, so the O(Q×D) pass is observable in logs. I left the loop uncapped for now — it only fires when BM25 returns zero and the doc_tokens set is bounded by the config's unique tokens, which is typically small. If we see it becoming a hot path in production, we can add an early exit.

src/ha_mcp/tools/smart_search.py
kingpanther13: Fixed in d0988b9. The 'shouldn't happen' branch now emits a `logger.warning` with the query tokens and doc length before returning 100, so if the impossible case ever fires it is visible instead of silent.

tests/src/e2e/tools/test_bm25_fuzzy_search.py
kingpanther13: Fixed in d0988b9. Both cleanup blocks now capture the exception as `e` and log it via `logger.debug`.

tests/src/e2e/tools/test_bm25_fuzzy_search.py
kingpanther13: Fixed in d0988b9. Added `assert len(results) > 0` to `test_fuzzy_search_multi_word_non_adjacent`, so the test now fails if the BM25 path stops returning results for a multi-word query against the test HA instance.

sergeykad (approved):

Issue #925, opened 2026-04-08
[FEATURE] Reduce cyclomatic complexity in tools/ by replacing closure pattern with classes


sergeykad:
## What would you like to see?

Migrate the 32 tool files in `src/ha_mcp/tools/` from the closure-based `register_*_tools(mcp, client)` pattern to a class-based approach, enabling C901 complexity enforcement across the entire codebase.

Currently 75% of the codebase (23k of 31k lines) is excluded from complexity checking because 92 functions exceed the threshold. The root cause is the registration pattern — every tool is a nested function inside a single `register_*_tools()` function, so ruff counts the entire nested structure as one function's complexity:

```python
# Current pattern — all complexity counted as one function
def register_foo_tools(mcp, client, **kwargs):
    @mcp.tool(...)
    async def ha_do_thing(param: str) -> dict:
        result = await client.get_states()  # captured via closure
        ...
```

Worst offenders: `register_config_helper_tools` (134), `register_entity_tools` (97), `register_config_dashboard_tools` (82), `register_registry_tools` (80).

Two files already use the target pattern (`device_control.py`, `smart_search.py`):

```python
# Target pattern — complexity distributed across methods
class FooTools:
    def __init__(self, client):
        self._client = client

    async def do_thing(self, param: str) -> dict:
        result = await self._client.get_states()
        ...

def register_foo_tools(mcp, client, **kwargs):
    tools = FooTools(client)

    @mcp.tool(...)
    async def ha_do_thing(param: str) -> dict:
        return await tools.do_thing(param)
```

## Why do you need this?

The current closure pattern diverges from MCP best practices and prevents complexity enforcement:

1. **MCP standard pattern** is one `@mcp.tool` per standalone function, with each tool as an independent unit ([MCP Python SDK](https://github.com/modelcontextprotocol/python-sdk), [FastMCP docs](https://gofastmcp.com/getting-started/welcome)). The closure pattern bundles everything into one function.

2. The [MCP Python Implementation Guide](https://github.com/modelcontextprotocol/python-sdk) emphasizes:
   - **"Extract Common Functionality"** — reusable helpers, shared API clients, centralized error handling rather than bundling logic inside registration functions
   - **"Use specific exception types"** (not generic `Exception`) — the closure pattern tends to accumulate broad `except Exception` blocks
   - **Pydantic models for input validation** rather than inline validation logic

[...]

PR #944, opened 2026-04-11
refactor: migrate 7 tool files to class-based pattern (batch 3)

sergeykad:
## Summary

- Migrate 7 tool files from closure-based `@mcp.tool()` to class-based `@tool(name=)` pattern: `tools_filesystem`, `tools_history`, `tools_resources`, `tools_traces`, `tools_config_automations`, `tools_hacs`, `tools_updates`
- Extract helper functions to reduce complexity and enable C901 enforcement
- Remove 7 C901 ignores from `pyproject.toml`
- Update 3 unit test files for class-based instantiation
- Fix lefthook mypy hook: replace bitwise-OR exit code pattern with `set -e` to fail immediately on mypy errors

Tracks #925

## Extracted helpers

| File | Helpers |
|------|---------|
| `tools_history` | `_parse_entity_ids`, `_parse_time_range` |
| `tools_resources` | `_process_resource_list`, `_check_ws_error`, `_extract_resource_id`, `_upsert_resource` |
| `tools_traces` | `_populate_trigger_info`, `_populate_condition_results`, `_populate_action_trace` |
| `tools_config_automations` | `_parse_and_validate_config`, `_validate_required_fields` |
| `tools_hacs` | `_filter_and_score_repos`, `_resolve_hacs_repo_id` |
| `tools_updates` | `_fetch_release_notes` |

## Test plan

- [x] All 1005 unit tests pass
- [x] Tool annotation tests pass (86 tools detected)
- [x] `extract_tools.py` finds all tools correctly
- [x] Ruff passes on all migrated files
- [x] Pre-commit hooks pass (ruff, mypy, unit tests)


src/ha_mcp/tools/tools_resources.py
```diff
@@ -583,3 +551,56 @@ async def ha_config_delete_dashboard_resource(
...
+
+def _check_ws_error(result: Any) -> str | None:
+    """Check a WebSocket result for errors. Returns error message or None."""
+    if isinstance(result, dict) and not result.get("success", True):
+        error = result.get("error", {})
+        if isinstance(error, dict):
+            msg: str = error.get("message", str(error))
+            return msg
+        return str(error)
+    return None
```
kingpanther13, line 598: Silent-success-on-failure regression from `upstream/master`. This helper returns `str | None` — `None` on success, message on failure. But all three callers (`:349`, `:395`, `:512`) then check `if error_msg:`, which is `False` for both `None` *and* `""`.

Old code at `master:tools_resources.py:377` unconditionally raised whenever `not result.get("success", True)`, regardless of message content. So if HA ever returns `{"success": False, "error": {"message": ""}}`, the pre-refactor code raised `SERVICE_CALL_FAILED`; post-refactor it returns a fake success dict.

Low probability in practice (HA usually populates the message), but silent success on a real failure is the worst failure mode. Two equivalent fixes:

- Return `"Unknown error"` (or `result.get("error") or "Unknown error"`) from this helper when the message is empty, so callers can keep `if error_msg:`.
- Or change the three call sites to `if error_msg is not None:`.

src/ha_mcp/tools/tools_resources.py
```diff
@@ -309,139 +289,110 @@ class MyCard extends HTMLElement {
...
+
+        resource_url = f"{WORKER_BASE_URL}/{encoded}?type={resource_type}"
 
-            encoded, _, encoded_size = _encode_content(content)
+        try:
+            result, action = await self._upsert_resource(resource_id, resource_url, resource_type)
 
-            if encoded_size > MAX_ENCODED_LENGTH:
+            error_msg = _check_ws_error(result)
+            if error_msg:
```
kingpanther13, line 349: Same issue as `_check_ws_error` — `if error_msg:` treats `""` as success. Repeats at `:395` and `:512`. See the comment on `_check_ws_error` for fix options.

src/ha_mcp/tools/tools_history.py
```diff
@@ -331,6 +289,72 @@ async def ha_get_history(
...
+            parsed_ids = parse_string_list_param(entity_ids, "entity_ids")
+            if parsed_ids is None:
+                raise_tool_error(create_error_response(
+                    ErrorCode.VALIDATION_MISSING_PARAMETER,
+                    "entity_ids is required",
+                    suggestions=["Provide at least one entity ID"],
+                ))
+            return parsed_ids
+        elif "," in entity_ids:
+            return [e.strip() for e in entity_ids.split(",") if e.strip()]
```
kingpanther13: Minor validation drift: old code at `master:tools_history.py:252` had an `if not entity_id_list:` guard *after* the parse branches that caught literal `","` / `",,,"` input → raised `VALIDATION_MISSING_PARAMETER`. The extracted helper only guards the non-str branch (`:313`), so comma-only garbage now silently returns `[]` and falls through to `_fetch_history` / `_fetch_statistics` with an empty entity list.

Arguably "empty input → empty output" is fine, but since the upstream contract was to reject it with a clear error, easiest fix is one extra check on this line:

```python
elif "," in entity_ids:
    result = [e.strip() for e in entity_ids.split(",") if e.strip()]
    if not result:
            ErrorCode.VALIDATION_MISSING_PARAMETER,
            "entity_ids is required",
            suggestions=["Provide at least one entity ID"],
        ))
    return result
```

kingpanther13 (changes requested):
Nice work on batch 3 — clean mechanical refactor for the most part. Found two behavioral regressions vs `upstream/master` that I'd like fixed before merge. Both are narrow and both have trivial fixes. Inline comments on the relevant lines.

Everything else checks out: all 17 tool names / annotations / docstrings / schemas preserved, `@tool` correctly outermost, `except ToolError: raise` guards present, `pyproject.toml` C901 removals match, `lefthook.yml set -e` switch is correct, test updates are pure boilerplate removal with assertions ported 1:1, and `_resolve_hacs_repo_id` actually improves on upstream by raising `RESOURCE_NOT_FOUND` when `hacs/repositories/list` returns `success: false`.

kingpanther13 (approved):
All feedback addressed cleanly, plus nice self-initiated cleanup. Verified:

- **`_check_ws_error`** (`tools_resources.py:595`) — returns `"Unknown error"` for empty error messages, applied to both dict and non-dict branches. Silent-success-on-failure is closed.
- **`_parse_entity_ids`** (`tools_history.py:310`) — comma-only input now raises `VALIDATION_MISSING_PARAMETER`. Contract restored.
- **`_resolve_hacs_repo_id`** (`tools_hacs.py:677`) — returning `(id, name)` eliminates the duplicate `hacs/repositories/list` call in `ha_hacs_download`. Real simplification.
- **Dead helper removal** (`tools_config_automations.py`) — `_strip_empty_automation_fields` inlined into `_validate_required_fields` under the `use_blueprint` branch. Semantically equivalent, tighter.
- **`lefthook.yml`** — splitting into two parallel mypy jobs is cleaner than the `bash -c` wrapper.

CI all green. LGTM.

PR #951, opened 2026-04-11
fix: route supervisor add-on log fetches through HA Core REST proxy

### LINKED ISSUE #950: [BUG] ha_get_logs(source="supervisor") always fails + error suggests nonexistent ha_list_addons()
## 💬 What Happened?

Two bugs in `ha_get_logs` when called with `source="supervisor"`:

1. **The call always fails with an empty error message.** Every attempt returns `"Command failed: "` with no detail, for every installed addon slug I tried — both ha-mcp addons and unrelated core addons like `core_mosquitto`. The slug is valid, the addon is installed and running, and the Supervisor is responding; the tool just can't decode the response.
2. **The failure suggestions tell you to use `ha_list_addons()`** — a tool that no longer exists in ha-mcp. That string is hardcoded in three places in `tools_utility.py` but no tool by that name is registered. The tool was consolidated into `ha_get_addon()` (which lists installed addons when called with no args), and the suggestions were never updated.

Both bugs are 100% reproducible.

### Reproduction

Every call of this shape fails identically, regardless of slug:

```
ha_get_logs(source="supervisor", slug="<any installed addon>")
```

Tried slugs (all returned the same error):
- `81f33d0f_ha_mcp_webhook_proxy`
- `81f33d0f_ha_mcp_dev`
- `ha_mcp_webhook_proxy`
- `core_mosquitto`  ← non-ha-mcp, official core addon

All return:

```json
{
  "success": false,
  "error": {
    "code": "SERVICE_CALL_FAILED",
    "message": "Command failed: ",
    "suggestion": "Verify add-on slug '<slug>' is correct",
    "suggestions": [
      "Verify add-on slug '<slug>' is correct",
      "Use ha_list_addons() to find available add-on slugs"
    ]
  },
  "slug": "<slug>"
}
```

### Root cause — Bug 1 (always fails)

`_get_supervisor_log` in `src/ha_mcp/tools/tools_utility.py` sends a websocket message with `type: "supervisor/api"` and `endpoint: /addons/{slug}/logs`. The Supervisor returns addon logs as `text/plain`, and the HA Core websocket proxy that services `supervisor/api` fails to decode that response body as JSON before ha-mcp ever sees it.

Same-endpoint evidence from this instance's HA system log (unrelated failures from Core's own hassio handler hitting the same endpoint):

```
homeassistant.components.hassio.handler: Client error on /addons/81f33d0f_ha_mcp_webhook_proxy/logs request 200,
  message='Attempt to decode JSON with unexpected mimetype: text/plain',
  url='http://172.30.32.2/addons/81f33d0f_ha_mcp_webhook_proxy/logs'
homeassistant.components.hassio.handler: Client error on /addons/81f33d0f_ha_mcp_dev/logs request 200,
  message='Attempt to decode JSON with unexpected mimetype: text/plain',
  url='http://172.30.32.2/addons/81f33d0f_ha_mcp_dev/logs'
```

The Supervisor returns HTTP 200 with the log body; only the JSON deserialization inside the websocket proxy fails. Because `result["error"]` is an empty string, `_get_supervisor_log` raises `"Command failed: "` with no actionable detail.

**Suggested fix:** don't route supervisor log fetches through `supervisor/api` (which forces JSON parsing). Fetch them via a plain HTTP GET against `http://supervisor/addons/{slug}/logs` with `Accept: text/plain`, exactly the way `homeassistant-addon-webhook-proxy/start.py::_supervisor_get_text` already does it in this same repo.

### Root cause — Bug 2 (stale suggestion)

`src/ha_mcp/tools/tools_utility.py` hardcodes `"Use ha_list_addons() to find available add-on slugs"` in three places: lines 185, 534, and 608. Grepping `upstream/master` for `ha_list_addons` turns up exactly those three strings plus a historical `CHANGELOG.md` entry — no tool by that name is registered anywhere. #99 originally added `ha_list_addons`, and it was later consolidated into `ha_get_addon()`, but these suggestion strings were never updated.

**Suggested fix:** replace all three occurrences with `"Use ha_get_addon() to list installed add-on slugs"` (or a similar string referencing the real tool).

### Environment

- **ha-mcp Version:** 7.2.0.dev209 (running via dev addon `81f33d0f_ha_mcp_dev`)
- **Installation Method:** addon
- **Platform:** Linux 6.12.77-haos (x86_64)
- **Python Version:** 3.13.12
- **Home Assistant Version:** 2026.4.1

### ha_report_issue diagnostic excerpt

```
=== Recent Tool Calls ===
2026-04-11T15:25:32 | ha_get_logs | FAIL | slug=81f33d0f_ha_mcp_webhook_proxy | "Command failed: "
2026-04-11T15:21:32 | ha_get_logs | FAIL | slug=ha_mcp_webhook_proxy          | "Command failed: "
2026-04-11T15:19:08 | ha_get_logs | FAIL | slug=81f33d0f_ha_mcp_dev           | "Command failed: "
2026-04-11T15:19:01 | ha_get_logs | FAIL | slug=81f33d0f_ha_mcp_webhook_proxy | "Command failed: "
```

Plus a subsequent run against `core_mosquitto` with identical failure.

kingpanther13:
## What does this PR do?

Fixes `ha_get_logs(source="supervisor")` which has been 100% broken for every slug on every install.

**Bug 1 — always fails.** The old code routed through HA Core's `supervisor/api` websocket proxy, which tries to JSON-decode Supervisor's `text/plain` log body and raises. ha-mcp only sees an empty error string and surfaces it as `"Command failed: "`.

**Fix:** fetch via HA Core's `/api/hassio/addons/{slug}/logs` REST endpoint, which HA Core returns as `text/plain` — no JSON hop. Works for all install types, not just addon.

Added `HomeAssistantClient.get_addon_logs(slug)` that centralizes the text fetch with proper error mapping (401 → auth, 4xx → API error with status, network → connection error).

**Bug 2 — stale suggestion text.** Three places in `tools_utility.py` still referenced `ha_list_addons()`, a tool that was removed during consolidation. Replaced with `ha_get_addon()`.

**Manually verified** against a live HA OS instance with 7 different add-ons (`core_mosquitto`, `81f33d0f_ha_mcp_dev`, `81f33d0f_ha_mcp_webhook_proxy`, `a0d7b954_nodered`, `a0d7b954_appdaemon`, `core_ssh`, `49686a9f_evcc`) — all return real logs. Unknown slugs return clean 404s with actionable suggestions.

**Independent of #934.** PR #934 (addon logs in bug reports) adds its own `_fetch_addon_logs` helper that uses `http://supervisor/addons/self/logs` directly. That path is self-only and addon-only, so the two PRs cover different use cases without overlap. When both land, a follow-up can extract a shared helper.

Closes #950

## Type of change
- [x] 🐛 Bug fix

## Testing
- [x] All automated tests pass (\`uv run pytest\`) — CI will run unit + e2e

**New unit tests in \`tests/src/unit/test_tools_utility_supervisor_logs.py\`:**
- \`TestGetAddonLogs\` (7 tests) — 200 returns text, correct endpoint + \`Accept: text/plain\`, 401 → auth error, 404 → API error with status, connect/timeout → connection error, no JSON decoding path.
- \`TestStaleToolNameReferences\` (1 test) — source grep to prevent regression of stale \`ha_list_addons()\` references.

## Checklist

### DIFF
--- a/src/ha_mcp/client/rest_client.py
+++ b/src/ha_mcp/client/rest_client.py
@@ -367,6 +367,52 @@ class HomeAssistantClient:
         response = await self._request("GET", "/error_log")
         return response if isinstance(response, str) else str(response)
 
+    async def get_addon_logs(self, slug: str) -> str:
+        """Fetch an add-on's container logs via HA Core's Supervisor REST proxy.
+
+        Uses `/api/hassio/addons/{slug}/logs`, which HA Core proxies to
+        Supervisor and returns as `text/plain`. This avoids the
+        `supervisor/api` websocket path that tries to JSON-decode the text
+        body and always fails (see #950).
+
+        Raises:
+            HomeAssistantAuthError: 401 from HA Core.
+            HomeAssistantAPIError: Non-2xx response (e.g. 404 unknown slug,
+                400 addon not installed). `status_code` is set so callers
+                can map to specific suggestions.
+            HomeAssistantConnectionError: Network, timeout, or transport error.
+        """
+        logger.debug(f"Fetching addon logs for slug={slug}")
+        endpoint = f"/hassio/addons/{slug}/logs"
+        try:
+            response = await self.httpx_client.request(
+                "GET",
+                endpoint,
+                headers={"Accept": "text/plain"},
+            )
+
+            if response.status_code == 401:
+                raise HomeAssistantAuthError("Invalid authentication token")
+
+            if response.status_code >= 400:
+                body = response.text[:500] if response.text else ""
+                raise HomeAssistantAPIError(
+                    f"Addon log request failed: {response.status_code} - {body}",
+                    status_code=response.status_code,
+                    response_data={"message": body},
+                )
+
+            return response.text
+
+        except httpx.ConnectError as e:
+            raise HomeAssistantConnectionError(
+                f"Failed to connect to Home Assistant: {e}"
+            ) from e
+        except httpx.TimeoutException as e:
+            raise HomeAssistantConnectionError(f"Request timeout: {e}") from e
+        except httpx.HTTPError as e:
+            raise HomeAssistantConnectionError(f"HTTP error: {e}") from e
+
     async def test_connection(self) -> tuple[bool, str | None]:
         """
         Test connection to Home Assistant.
--- a/src/ha_mcp/client/rest_client.py
+++ b/src/ha_mcp/client/rest_client.py
@@ -97,9 +97,53 @@ class HomeAssistantClient:
         await self.httpx_client.aclose()
         logger.debug("Closed Home Assistant client")
 
+    async def _raw_request(
+        self, method: str, endpoint: str, **kwargs: Any
+    ) -> httpx.Response:
+        """Authenticated request that returns the raw httpx.Response.
+
+        Handles auth, HTTP 4xx/5xx, and transport errors in one place.
+        Callers parse the body themselves (JSON via `_request`, text via
+        `get_addon_logs`, etc.).
+
+        Raises:
+            HomeAssistantAuthError: 401 response.
+            HomeAssistantAPIError: Non-2xx response (with status_code and
+                response_data set from JSON body when possible).
+            HomeAssistantConnectionError: Network, timeout, or transport error.
+        """
+        try:
+            response = await self.httpx_client.request(method, endpoint, **kwargs)
+
+            if response.status_code == 401:
+                raise HomeAssistantAuthError("Invalid authentication token")
+
+            if response.status_code >= 400:
+                try:
+                    error_data = response.json()
+                except Exception:
+                    error_data = {"message": response.text}
+
+                raise HomeAssistantAPIError(
+                    f"API error: {response.status_code} - {error_data.get('message', 'Unknown error')}",
+                    status_code=response.status_code,
+                    response_data=error_data,
+                )
+
+            return response
+
+        except httpx.ConnectError as e:
+            raise HomeAssistantConnectionError(
+                f"Failed to connect to Home Assistant: {e}"
+            ) from e
+        except httpx.TimeoutException as e:
+            raise HomeAssistantConnectionError(f"Request timeout: {e}") from e
+        except httpx.HTTPError as e:
+            raise HomeAssistantConnectionError(f"HTTP error: {e}") from e
+
     async def _request(self, method: str, endpoint: str, **kwargs: Any) -> dict[str, Any]:
         """
-        Make authenticated request to Home Assistant API.
+        Make authenticated request to Home Assistant API and parse JSON body.
 
         Args:
             method: HTTP method (GET, POST, etc.)
@@ -114,42 +158,13 @@ class HomeAssistantClient:
             HomeAssistantAuthError: Authentication failed
             HomeAssistantAPIError: API error
         """
+        response = await self._raw_request(method, endpoint, **kwargs)
         try:
-            response = await self.httpx_client.request(method, endpoint, **kwargs)
-
-            # Handle authentication errors
-            if response.status_code == 401:
-                raise HomeAssistantAuthError("Invalid authentication token")
-
-            # Handle other HTTP errors
-            if response.status_code >= 400:
-                try:
-                    error_data = response.json()
-                except Exception:
-                    error_data = {"message": response.text}
-
-                raise HomeAssistantAPIError(
-                    f"API error: {response.status_code} - {error_data.get('message', 'Unknown error')}",
-                    status_code=response.status_code,
-                    response_data=error_data,
-                )
-
-            # Parse JSON response
-            try:
-                result: dict[str, Any] = response.json()
-                return result
-            except json.JSONDecodeError:
-                # Some endpoints return empty responses
-                return {}
-
-        except httpx.ConnectError as e:
-            raise HomeAssistantConnectionError(
-                f"Failed to connect to Home Assistant: {e}"
-            ) from e
-        except httpx.TimeoutException as e:
-            raise HomeAssistantConnectionError(f"Request timeout: {e}") from e
-        except httpx.HTTPError as e:
-            raise HomeAssistantConnectionError(f"HTTP error: {e}") from e
+            result: dict[str, Any] = response.json()
+            return result
+        except json.JSONDecodeError:
+            # Some endpoints return empty responses
+            return {}
 
     async def get_config(self) -> dict[str, Any]:
         """Get Home Assistant configuration."""
@@ -383,35 +398,12 @@ class HomeAssistantClient:
             HomeAssistantConnectionError: Network, timeout, or transport error.
         """
         logger.debug(f"Fetching addon logs for slug={slug}")
-        endpoint = f"/hassio/addons/{slug}/logs"
-        try:
-            response = await self.httpx_client.request(
-                "GET",
-                endpoint,
-                headers={"Accept": "text/plain"},
-            )
-
-            if response.status_code == 401:
-                raise HomeAssistantAuthError("Invalid authentication token")
-
-            if response.status_code >= 400:
-                body = response.text[:500] if response.text else ""
-                raise HomeAssistantAPIError(
-                    f"Addon log request failed: {response.status_code} - {body}",
-                    status_code=response.status_code,
-                    response_data={"message": body},
-                )
-
-            return response.text
-
-        except httpx.ConnectError as e:
-            raise HomeAssistantConnectionError(
-                f"Failed to connect to Home Assistant: {e}"
-            ) from e
-        except httpx.TimeoutException as e:
-            raise HomeAssistantConnectionError(f"Request timeout: {e}") from e
-        except httpx.HTTPError as e:
-            raise HomeAssistantConnectionError(f"HTTP error: {e}") from e
+        response = await self._raw_request(
+            "GET",
+            f"/hassio/addons/{slug}/logs",
+            headers={"Accept": "text/plain"},
+        )
+        return response.text
 
     async def test_connection(self) -> tuple[bool, str | None]:
         """
--- a/src/ha_mcp/client/rest_client.py
+++ b/src/ha_mcp/client/rest_client.py
@@ -124,8 +124,12 @@ class HomeAssistantClient:
                 except Exception:
                     error_data = {"message": response.text}
 
+                message = error_data.get("message")
+                if not message or not message.strip():
+                    message = response.reason_phrase or "<empty body>"
+
                 raise HomeAssistantAPIError(
-                    f"API error: {response.status_code} - {error_data.get('message', 'Unknown error')}",
+                    f"API error: {response.status_code} - {message}",
                     status_code=response.status_code,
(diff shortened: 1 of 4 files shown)


kingpanther13:
Addressed both Gemini review threads in 4920b06.

**rest_client.py** — extracted `_raw_request` helper so `_request` (JSON path) and `get_addon_logs` (text path) share one auth/HTTP/transport error block. No more duplicate httpx exception handling; `_request` is now a 5-line wrapper around `_raw_request`.

**tools_utility.py** — consolidated the two `HomeAssistantAPIError` branches in `_get_supervisor_log` into a single `exception_to_structured_error` call. Only the first suggestion text branches on 400/404 vs other.

src/ha_mcp/tools/tools_utility.py
```diff
@@ -594,18 +558,34 @@ async def _get_supervisor_log(
...
             exception_to_structured_error(
                 e,
                 context={"source": "supervisor", "slug": slug},
                 suggestions=[
-                    f"Verify add-on slug '{slug}' is correct",
-                    "Use ha_list_addons() to find available add-on slugs",
+                    "Check Home Assistant connection",
                     "Ensure Supervisor is available (HA OS or Supervised install)",
                 ],
             )
```
sergeykad, line 610: The prior shared `except` passed `f"Verify add-on slug '{slug}' is correct"` as the first suggestion. After splitting `HomeAssistantAPIError` into its own branch, this one replaced that with `"Check Home Assistant connection"`, so a timeout or connect error on a valid-looking slug no longer hints at slug verification. Worth keeping the slug suggestion here too, since a transient Supervisor failure on a wrong slug can surface as `HomeAssistantConnectionError` before the HTTP layer returns a recognizable status.

src/ha_mcp/tools/tools_utility.py
```diff
@@ -594,18 +558,34 @@ async def _get_supervisor_log(
...
+            )
+            exception_to_structured_error(
+                e,
+                context={"source": "supervisor", "slug": slug},
+                suggestions=[
+                    first_suggestion,
+                    "Use ha_get_addon() to list installed add-on slugs",
+                    "Ensure Supervisor is available (HA OS or Supervised install)",
+                ],
+            )
```
sergeykad, line 595: `is_not_found = status in (400, 404)` lumps both codes under "Add-on not found or not installed", but a 400 from `/api/hassio/addons/{slug}/logs` is typically a malformed request or upstream Supervisor rejection rather than an unknown slug, so the user sees a misleading first suggestion. Splitting to `if status == 404` for the not-found message and a 400-specific message like "Supervisor rejected the request; verify slug format or check Supervisor availability" gives clearer diagnostics. `exception_to_structured_error` also maps 400 to `VALIDATION_INVALID_PARAMETER`, which reads as "caller passed bad input" rather than "proxy returned 400".

src/ha_mcp/tools/tools_utility.py
```diff
@@ -515,53 +515,17 @@ async def _get_supervisor_log(
...
-                    log_text = log_text["data"]
-                else:
-                    logger.warning(
-                        "Supervisor log for '%s' returned unexpected dict structure",
-                        slug,
-                    )
-                    log_text = ""
-            if not isinstance(log_text, str):
-                log_text = str(log_text)
+            log_text = await client.get_addon_logs(slug)
```
sergeykad, line 528: The 8 new tests cover `HomeAssistantClient.get_addon_logs()` directly but skip the `_get_supervisor_log` wrapper entirely. That leaves the response shape (`source`, `slug`, `log`, `total_lines`, `returned_lines`, `limit`, `filters_applied`), tail slicing (`lines[-effective_limit:]`), the `search` filter, and the API-error to suggestion translation with no coverage. A future refactor flipping `[-effective_limit:]` to `[:effective_limit]` or dropping `total_lines` would ship without any test failing. Four or five unit tests mocking `client.get_addon_logs` (happy-path shape, 404 translation, tail slicing, search filter) would close the gap cheaply.

src/ha_mcp/client/rest_client.py
```diff
@@ -97,31 +97,27 @@ async def close(self) -> None:
...
             response = await self.httpx_client.request(method, endpoint, **kwargs)
 
-            # Handle authentication errors
             if response.status_code == 401:
                 raise HomeAssistantAuthError("Invalid authentication token")
 
-            # Handle other HTTP errors
             if response.status_code >= 400:
                 try:
                     error_data = response.json()
```
sergeykad, line 123: When the 4xx body is empty or whitespace-only (which the hassio proxy does produce in some failure modes), `response.json()` raises, the fallback yields `error_data = {"message": ""}`, and the raised `HomeAssistantAPIError` message becomes `f"API error: {status} - "` with an empty tail. That reproduces the same silent-failure signature #950 describes, just one layer down. Falling back to `response.reason_phrase` or a literal `"<empty body>"` when `response.text` is empty would ensure users always see something actionable instead of a naked status code. Affects every caller of `_raw_request`, so worth fixing here rather than in `_get_supervisor_log`.

tests/src/unit/test_tools_utility_supervisor_logs.py
```diff
@@ -0,0 +1,153 @@
...
+            Path(__file__).resolve().parents[3]
+            / "src"
+            / "ha_mcp"
+            / "tools"
+            / "tools_utility.py"
+        ).read_text()
+        assert "ha_list_addons" not in source, (
+            "tools_utility.py still references the removed `ha_list_addons` tool. "
+            "Replace suggestions with `ha_get_addon()` — see #950."
+        )
```
sergeykad: The grep is scoped to `tools_utility.py` only. If a stale `ha_list_addons()` reference lands in another tool module's suggestions or docstring, this test stays green. Broadening the read to `src/ha_mcp/tools/**/*.py` matches the bug class rather than this one file. The plain substring check also matches inside comments, changelog notes, or coincidental variable names; switching to a word-boundary regex (`\bha_list_addons\b`) tightens the guard without weakening coverage.

src/ha_mcp/tools/tools_utility.py
```diff
@@ -515,53 +515,17 @@ async def _get_supervisor_log(
         limit: int | str | None = None,
         search: str | None = None,
     ) -> dict[str, Any]:
-        """Fetch add-on container logs via the Supervisor API."""
+        """Fetch add-on container logs via HA Core's Supervisor REST proxy.
+
+        Routes through `/api/hassio/addons/{slug}/logs` (returned as
+        text/plain) instead of the `supervisor/api` websocket path, which
+        always fails because HA Core's proxy tries to JSON-decode the
+        text body. See #950.
```
sergeykad, line 523: AGENTS.md asks for e2e coverage when `src/` adds or significantly rewrites functionality, and this supervisor path was 100% broken before the fix, so the full `ha_get_logs(source="supervisor")` flow has no integration-level guard against future wiring regressions. The existing `test_logs_supervisor_missing_slug` only covers the slug-required validation path, which fires before any HTTP call. The testcontainer runs HA Core (`ghcr.io/home-assistant/home-assistant:2026.4.1` per `test_constants.py:16`), which has no Supervisor, so a happy-path e2e isn't reachable without mocking. An error-path e2e is though: calling `ha_get_logs(source="supervisor", slug="anything")` against Core will hit `/api/hassio/addons/anything/logs`, get a non-2xx from HA, and must return a `ToolError` with the expected suggestions. That pins the full `_raw_request` to `HomeAssistantAPIError` to `exception_to_structured_error` chain.

src/ha_mcp/tools/tools_utility.py
kingpanther13: Restored in b1828ac. The connection-error path now carries both `Verify add-on slug '<slug>' is correct` and `Use ha_get_addon() to list installed add-on slugs` alongside the existing connection/Supervisor hints, so a transient failure on a wrong slug still points at the slug. `test_connection_error_keeps_slug_hint` in the new unit tests pins it.

src/ha_mcp/tools/tools_utility.py
kingpanther13: Split in b1828ac:
- `404` → `Add-on '<slug>' not found or not installed`
- `400` → `Supervisor rejected the request for '<slug>' — verify slug format or that the add-on is installed and running`
- else → `Verify add-on slug '<slug>' is correct`

On the `VALIDATION_INVALID_PARAMETER` mapping: fair point, fixed in e9c3849 by building the 400 response manually with `SERVICE_CALL_FAILED` so the code reflects a downstream proxy rejection rather than caller bad input. `test_400_uses_distinct_suggestion_and_service_error_code` now asserts on both the suggestion text and the error code.

src/ha_mcp/tools/tools_utility.py
kingpanther13: Added a `TestGetSupervisorLogWrapper` class in b1828ac covering the gaps:
- happy-path response shape (`success`/`source`/`slug`/`log`/`total_lines`/`returned_lines`/`limit`; `filters_applied` omitted when no search)
- tail slicing — 20 lines with `limit=5` returns `line 16`..`line 20`
- case-insensitive `search` filter with `filters_applied` recorded
- 404 → `ToolError` with the not-installed suggestion and slug/source context
- 400 → distinct `Supervisor rejected` suggestion (also asserts `SERVICE_CALL_FAILED` after e9c3849)
- connection-error path keeps the slug hint

Exercised through the registered `ha_get_logs` tool via a `SimpleNamespace(tool=...)` collector so the `@mcp.tool` + `@log_tool_usage` stack runs as in production.

src/ha_mcp/client/rest_client.py
kingpanther13: Fixed in b1828ac. `_raw_request` now falls back to `response.reason_phrase` or `"<empty body>"` when the parsed message is empty or whitespace-only, so the error tail is never bare.

Three regression tests in `TestRawRequestEmptyBodyFallback`:
- empty text body + 502 → `Bad Gateway` in the error message
- whitespace-only JSON `{"message": "   "}` + 500 → `Internal Server Error` in the error message
- empty body + empty `reason_phrase` → `<empty body>` placeholder

tests/src/unit/test_tools_utility_supervisor_logs.py
kingpanther13: Done in b1828ac. The guard now walks `src/ha_mcp/tools/**/*.py` and matches on `\bha_list_addons\b`, so it catches regressions in any tool module and won't match substrings inside longer identifiers or coincidental strings.

src/ha_mcp/tools/tools_utility.py
kingpanther13: Added `test_logs_supervisor_propagates_api_error_to_structured_tool_error` in b1828ac. The testcontainer's Supervisor-less HA Core returns a non-2xx for `/api/hassio/addons/<anything>/logs`, which must travel `_raw_request → HomeAssistantAPIError → exception_to_structured_error` and surface as a `ToolError` whose payload carries the slug context plus the `ha_get_addon()` / Supervisor hints. That pins the full chain end-to-end.

sergeykad (approved):

PR #968, opened 2026-04-13
feat: add python_transform support to automations and scripts

### LINKED ISSUE #962: Add python_transform support to ha_config_set_automation
## Problem

When modifying complex automations programmatically, the only option is to pass the entire automation config through the `config` parameter. For large automations with many nested action blocks (choose/if), this is error-prone and impractical — especially for targeted changes like adding/removing a single entity from a target list or inserting a new action block at a specific position.

`ha_config_set_dashboard` already supports `python_transform` and `jq_transform`, which allow in-place modifications with optimistic locking via `config_hash`. This capability is missing for automations.

## Current Workaround

For targeted modifications to large automations, agents must fall back to direct REST/WebSocket API calls with custom scripts — bypassing ha-mcp entirely.

## Proposal

Add `python_transform` (and optionally `jq_transform`) support to `ha_config_set_automation`, analogous to how `ha_config_set_dashboard` already works. The transform would receive the current automation config and allow in-place modifications with `config_hash`-based optimistic locking.

### LINKED ISSUE #963: Add python_transform support to ha_config_set_script
## Problem

When modifying complex scripts programmatically, the only option is to pass the entire script config through the `config` parameter. For large scripts with many nested action blocks (choose/if/repeat), this is error-prone and impractical — especially for targeted changes like adding/removing a single entity from a target list or inserting a new action at a specific position.

`ha_config_set_dashboard` already supports `python_transform` and `jq_transform`, which allow in-place modifications with optimistic locking via `config_hash`. This capability is missing for scripts.

Related: #962 — same request for `ha_config_set_automation`.

## Current Workaround

For targeted modifications to large scripts, agents must fall back to direct REST/WebSocket API calls with custom scripts — bypassing ha-mcp entirely.

## Proposal

Add `python_transform` (and optionally `jq_transform`) support to `ha_config_set_script`, analogous to how `ha_config_set_dashboard` already works. The transform would receive the current script config and allow in-place modifications with `config_hash`-based optimistic locking.

bzelch:
## Summary

Adds `python_transform` and `config_hash` parameters to `ha_config_set_automation` and `ha_config_set_script`, enabling surgical in-place modifications with optimistic locking — the same proven pattern `ha_config_set_dashboard` already uses.

- `ha_config_get_automation` / `ha_config_get_script` now return `config_hash`
- `ha_config_set_automation` / `ha_config_set_script` accept `python_transform` + `config_hash`
- Mutual exclusivity enforced between `config` and `python_transform`
- `config_hash` required for `python_transform`, optional for full config updates
- Hash conflict detection prevents silent overwrites from concurrent edits
- Automation hash computed after normalization (`_normalize_config_for_roundtrip`) to avoid phantom conflicts from plural/singular key differences

Closes #962, closes #963

## Test plan

- [x] E2E: Simple property update via python_transform (automation + script)
- [x] E2E: Pattern-based update with loops (automation)
- [x] E2E: config_hash required for python_transform
- [x] E2E: identifier required for python_transform (automation)
- [x] E2E: Mutual exclusivity of config and python_transform
- [x] E2E: Hash conflict detection with stale hash
- [x] E2E: Blocked import expressions (security)
- [x] All 12 new E2E tests passing


Patch76:
Just to be upfront about where I'm coming from: I'm another contributor here, not a maintainer. sergeykad, kingpanther13, and julienld call the shots — what follows are my observations, not decisions.

This is a clean, well-structured PR. The pattern mirrors ha_config_set_dashboard nicely, and the test coverage is solid.

A few things I noticed while reading through:

**1. Sandbox: the risk surface is narrower than it looks**

I verified a couple of the obvious attack vectors against the actual python_sandbox.py:

- Lambda recursion (`f = lambda x: f(x+1); f(0)`) → NameError at runtime because `f` isn't in the locals dict. The sandbox passes the expression as a string to exec, so self-referential lambdas don't resolve.
- range() comprehensions → NameError, because `__builtins__: {}` strips range from the runtime environment. The AST check passes it, but it can't execute.

So the `__builtins__: {}` + AST whitelist combination does more work than it might appear. The existing docstring note — "Not a security boundary — callers are already authenticated MCP users with full HA access" — is accurate and appropriate framing.

One genuine gap worth flagging: there's no timeout on `exec()`. A deeply nested dict comprehension operating on large existing config data could still block the worker. Not a showstopper given the auth model, but worth a comment or a note in the docstring.

**2. Hash truncation to 16 hex chars (64 bits)**

For optimistic locking in a single-user HA instance, 64-bit is fine — accidental collisions are astronomically unlikely at this scale. The only meaningful attack requires crafting a hash collision while already being authenticated, which is a contradiction. No objection here, but `[:32]` (128 bits) would be a low-cost improvement if anyone wants belt-and-suspenders.

**3. Normalization asymmetry between automations and scripts**

This is the one I'd flag to the maintainers as the real design question:

- Automations: `_compute_config_hash` runs on the output of `_normalize_config_for_roundtrip` in both `get` and `_fetch_and_verify_hash`. Consistent.
- Scripts: `get` hashes the raw REST wrapper (`config_result`). `_fetch_and_verify_hash` hashes `current_wrapper = current["config"]` — same level, consistent within the PR. `json.dumps(sort_keys=True)` handles key ordering.

But scripts don't have an equivalent of `_normalize_config_for_roundtrip`. If the HA API performs any normalization on script configs during a roundtrip (type coercion, key additions, default injection), the hash will change even if the user made no logical edit. The existing E2E tests won't catch this because they test the happy path, not the "get → no change → set" stability.

A simple smoke test: two consecutive `ha_config_get_script` calls on a real HA instance — if the hashes differ, the normalization gap is the culprit.

**4. _compute_config_hash is defined twice**

Identical implementation in `tools_config_automations.py` and `tools_config_scripts.py`. Minor maintenance debt — if the hashing logic ever changes, it needs to be updated in two places. A shared `utils/config_hash.py` would clean this up.

---

None of the above are blockers IMO — the core logic is sound. Items 3 and 4 are the ones I'd want resolved before merge; items 1 and 2 are FYI observations. Happy to be corrected if I've misread the code.

Apologies if I've overstepped — this pattern touches tooling I rely on, so I wanted to share what I found. Final call is entirely with the maintainers.

bzelch:
@Patch76 Thanks for the thorough review — addressed points 3 and 4:

**Shared hash utility (point 4):** Extracted `compute_config_hash` to `utils/config_hash.py`. Both automation and script tools now import from there. Dashboard has its own copy — migrating that is out of scope for this PR but noted in the module docstring.

**Hash stability (point 3):** Added a smoke test (`test_config_hash_stable_across_reads`) that does two consecutive `ha_config_get_script` calls and asserts the hashes match. Passes on HA 2026.4.1 — no roundtrip normalization jitter observed.

Points 1 (sandbox timeout) and 2 (hash truncation) are noted — both reasonable future improvements but not blockers for this PR as you said.

kingpanther13 (changes requested):
First — thank you for tackling this, and welcome! This is a really
strong first contribution to the project. The dashboard pattern is
mirrored faithfully, the shared `compute_config_hash` extraction in
response to Patch76's review is exactly the right call, and your
responsiveness to feedback so far has been excellent. We're glad to
have you here.

I went through the diff carefully and there's a handful of things
I'd like resolved before merge. Most are small; a couple are real
bugs that the current tests don't catch. Since follow-up PRs in
this area are uncommon, I'd rather we sweep up the minor stuff
while we're already in the file.

## Blockers

### 1. `category` is not popped in the automation python_transform path

`ha_config_get_automation` injects `normalized_config["category"] = cat_id`
(line ~264) when the entity has a category in the registry. The
transform branch then sends that dict straight to
`upsert_automation_config` without popping `category`. The
non-transform branch pops it explicitly at line ~601 with the
comment *"HA REST API rejects unknown keys."* Any user calling
`python_transform` against a categorized automation will hit a
400. The new tests don't create categorized automations so CI is
green.

(Scripts are fine here — `config_result["category"]` lives in the
outer REST wrapper rather than inside `actual_config`, so it never
reaches `upsert_script_config`. But see #3 below for a related
concern.)

**Fix:** pop `category` from `transformed_config` before upsert in
the automation transform branch, and re-apply it via
`apply_entity_category` the same way the regular path does so
categorization round-trips correctly.

### 2. Validation is bypassed in the transform path

The transform branches skip `_validate_required_fields` and
`_check_best_practices` (automations) and `_validate_script_config`
+ best-practice checks (scripts). A transform like
`del config['action']` or one that introduces a best-practice
issue ships straight to HA with no friendly error and no warnings
attached to the response.

**Fix:** after `safe_execute`, run the same validators on the
transformed dict that the regular config path runs, before
calling `upsert_*`.

### 3. Script hash is brittle — hashes response envelope, not script body

`ha_config_get_script` computes
`compute_config_hash(config_result)` where `config_result` is the
REST wrapper `{"success": True, "script_id": ..., "config": <body>,
"category"?}`. So the hash depends on envelope keys — not just the
script body. `_fetch_and_verify_hash` then returns a
`(wrapper, actual_config)` tuple and the transform branch
synthesizes `new_wrapper = {**current_wrapper, "config":
transformed_config}` for re-hashing. This is internally consistent
within a single call, but:

- Any future addition to the REST wrapper (a timestamp, an
  `_meta` field) silently invalidates every outstanding hash
  and produces phantom conflicts
- It's inconsistent with the automation side, which hashes the
  bare config dict — making the new shared `compute_config_hash`
  utility behave differently depending on caller

**Fix:** hash only the inner script dict (mirror the automation
approach). Drop the wrapper/actual_config tuple and treat scripts
and automations the same way.

### 4. Returned `config_hash` may not match the next `get`

After `upsert_*_config`, HA may add or normalize fields (defaults,
field coercion, `id` injection, etc.). Returning
`compute_config_hash(transformed_config)` from the pre-upsert
dict means the documented "use the returned config_hash for the
next call" pattern could produce phantom conflicts on the next
chained operation. I haven't been able to definitively reproduce
this against our test HA, but it's the kind of thing we really
want a test for before we merge — see "tests required" below.

**Fix (if the test confirms the concern):** after the upsert,
re-fetch with `ha_config_get_*` (or re-run the same normalization
path) and compute the new hash from that. The returned hash
should be exactly the hash the caller will see on the next read.

### 5. Hash stability under transient registry failures

`fetch_entity_category` is best-effort — it can succeed once and
return None on the next call (transient REST hiccup, WS error,
etc.). Since `category` is conditionally injected into the hashed
dict (both auto and script sides), two consecutive reads can
produce different hashes without any user modification, generating
phantom conflicts. Your `test_config_hash_stable_across_reads`
passes because the lookup is reliable in your test env, but it
doesn't simulate the failure boundary.

**Fix:** compute the hash before category injection, or hash a
normalized projection that excludes registry-derived best-effort
fields so the hash depends only on the actual config payload.

## Smaller things to fix while you're in there

- **`Expression: ...` truncation always shows the trailing `...`**
  even when the expression is shorter than 100 chars
  (`tools_config_automations.py:~563`,
  `tools_config_scripts.py:~441`). Use
  `expr[:100] + ('...' if len(expr) > 100 else '')`.

- **`del config['action'][2]` in the docstring will IndexError on
  most automations**. Use `config['action'].pop()` or another
  example that doesn't assume an out-of-bound index.

- **Sandbox DoS** — Patch76 raised this and you acknowledged it.
  Agreed it's not a security boundary, but a `while True: pass`
  one tool call deep pins a worker. At minimum, add an explicit
  warning to the tool description so callers know infinite loops
  hang the server. A real timeout (wrapping `exec` in a
  `run_in_executor` + `asyncio.wait_for`, or a signal-based
  alarm) would be ideal but I'd take the docstring warning as
  the minimum bar here.

- **`ErrorCode.SERVICE_CALL_FAILED` for hash conflict** is
  semantically off — `RESOURCE_LOCKED` (`errors.py:79`) or a
  new `STALE_HASH` code would be more accurate. Dashboard
  already uses `SERVICE_CALL_FAILED` for the same case so
  you're being consistent; leave as-is for this PR and I'll
  file a follow-up issue covering all three sites so they can
  be fixed together.

- **`safe_execute` mutates AND returns the dict** — dual
  semantics make aliasing bugs easy. Pick one: either return
  `None` and document in-place mutation, or deep-copy the input
  and return the new dict.

- **`**{k: v for k, v in result.items() if k not in ("success",)}`**
  in both transform return statements — `if k != "success"` is
  cleaner. Also worth a comment explaining intent since
  `upsert_*` doesn't currently return `"success"` anyway.

- **Internal `self.ha_config_get_automation` / `ha_config_get_script`
  calls inside `_fetch_and_verify_hash`** go through
  `@log_tool_usage`, so each transform generates a phantom
  log entry the user didn't invoke. Factor an undecorated
  `_get_*_config_internal` that both the public tool and the
  helper call into.

- **`_compute_config_hash` still lives in
  `tools_config_dashboards.py`** as a byte-identical third
  copy (line ~29). Your shared util's docstring notes this as
  a future PR, but while you're here let's just do the
  migration in this PR so we don't carry three copies. It's a
  ~5-line change.

## Tests required for the changes above

A few tests would have caught the issues above and the ones the
PR claims to cover but doesn't quite. Please add at least:

1. **Categorized automation transform** (catches #1) — create an
   automation in a category, transform a property, assert
   success and that the category is preserved on the next get.

2. **Transform that produces an invalid config** (catches #2) —
   a transform like `del config['action']` should be rejected
   by the validator before reaching HA, with a clear error.

3. **Chained transforms** (catches #3 and #4) — transform once,
   capture `result["config_hash"]`, transform again using that
   hash, assert success. This is the single most important new
   test — it exercises the whole locking contract end-to-end
   for both automation and script.

4. **Returned hash matches next get** (catches #4) — transform,
   capture `result["config_hash"]`, call `ha_config_get_*`,
   assert the two hashes are equal. Cheap and catches an
   entire class of bug.

5. **Plural-key automation hash stability** — your PR
   description notes that the post-normalization hashing
   avoids `triggers`/`actions` phantom conflicts, but there's
   no test for it. Create an automation with `triggers`/
   `actions` plural, get hash twice, assert equal; then
   transform with the first hash, assert success.

6. **Full-config update with `config_hash`** — both auto and
   script support `config + config_hash` for optimistic
   locking on full replacements, but no test exercises this
   path (the existing hash tests only hit the python_transform
   branch). Add one positive test (matching hash, succeeds)
   and one conflict test (stale hash, fails) for each.

7. **`PythonSandboxError → ToolError` mapping** — the current
   blocked-import test asserts on substring `"import"`/
   `"forbidden"`. Tighten to assert
   `ErrorCode.VALIDATION_FAILED` and the structured
   `context`/`suggestions` payload so a regression that
   downgrades the error doesn't slip past.

8. **AGENTS.md "poll after creating entities"** — neither test
   file uses `wait_for_*` helpers from
   `tests/src/e2e/utilities/wait_helpers.py`. They rely on the
   tool's internal `wait` default, which is racy on slow CI.
   Worth importing the helpers and using them around the
   create→get sequences so CI stays stable.

---

Again — this is genuinely strong work for a first contribution,
and most of the above is "we're already in the file, let's sweep
it up" rather than "this is broken." The core architecture is
right; we just want the edges tightened before it lands.

**Pushback is completely welcome and expected** — especially on
the smaller items, everything is up for discussion. If any of the
blockers are wrong (I verified them by reading the code but I
may have missed context), or the suggested fixes don't sit right
with you, push back and we'll talk it through. Same goes for the
test list — if you think any of those are overkill or already
covered elsewhere, say so. Take the time you need; happy to
answer questions on any of the above. Thanks again for the
contribution, and welcome to the project!

kingpanther13:
Sorry, I forgot to run CI before I did my review, please address all of the CI failures too. This can be done by asking your Claude or whatever agent you use to check the CI status and implement fixes for the failures. Sometimes you can even run the tests locally first to make sure they will past before posting here, but that isn't a requirement.

bzelch:
@kingpanther13 Thanks for the detailed review — all blockers and smaller items addressed:

**Blockers:**
1. Category is now popped before upsert and re-applied via `apply_entity_category` in the transform path
2. Validation (`_validate_required_fields`, `_check_best_practices`) runs on the transformed config before upsert
3. Script hash now computed on the inner config body, consistent with automations
4. Re-fetch after save via `_get_*_config_internal()` for authoritative hash
5. Hash computed before category injection so transient registry failures don't affect it

**Smaller fixes:**
- Expression truncation conditional, docstring examples use `.pop()`, DoS warning added
- `_get_automation_config_internal` / `_get_script_config_internal` avoid phantom log entries
- Dashboard `_compute_config_hash` migrated to shared `utils/config_hash.py` (all three tools now use the same function)
- Result merge cleanup

**Tests (25 total):** Added chained transforms, returned hash == next get, invalid transform rejection, hash stability, plural-key stability, and full config update with config_hash (positive + conflict) — for both automations and scripts.

Also fixed the Mypy CI failure (`Returning Any` in `_fetch_and_verify_hash`) via the internal method refactor.

kingpanther13:
Really impressive turnaround. Verified all 5 blockers are fixed correctly: category pop + re-apply via `apply_entity_category`, validation + best-practice checks on the transformed config, script hashing the inner body, re-fetch after save for authoritative hash, and hash computed before category injection. Every smaller item is addressed. The dashboard `_compute_config_hash` migration is clean, the stale `hashlib` import is gone, and the extracted `_get_*_config_internal` methods are a nicer refactor than I asked for. The chained-transform, returned-hash-matches-next-get, plural-key, and full-config-hash (positive + conflict) tests are exactly what the locking contract needed.

One thing I missed on my last pass — apologies for the churn: the category bug from blocker #1 doesn't have a regression test. All the new automation tests create uncategorized automations, so the `transformed_config.pop("category", None)` line in the transform branch is never actually exercised. A future refactor that drops it would still go CI-green. Can you add one test that:

1. Creates an automation with a `category`
2. Gets the config_hash
3. Runs a `python_transform` that changes some property
4. Calls `ha_config_get_automation` and asserts the category is preserved

~15 lines, and it closes the last regression gap I can see. The other two test items I listed in my review (structured error code assertion, `wait_for_*` poll helpers) — drop those, they're repo-wide cleanups that shouldn't ride this PR.

After that test lands, we'll want to run BAT (Bot Acceptance Testing) on the branch before merging — that's the framework at `tests/uat/` which runs real agent CLIs (Claude, Gemini, OpenAI) against a Home Assistant testcontainer using dynamically-generated scenarios. See `tests/uat/README.md` for the full picture. The calling agent generates a scenario JSON (`setup_prompt` / `test_prompt` / `teardown_prompt`), pipes it to `tests/uat/run_uat.py`, and the runner handles HA container startup + MCP config + agent invocation. Two Claude Code skills drive it:

- `.claude/skills/bat-adhoc/SKILL.md` — ad-hoc PR validation with dynamically-generated scenarios
- `.claude/skills/bat-story-eval/SKILL.md` — pre-built stories in `tests/uat/stories/catalog/` plus diff-based regression eval

Quick usage example:

```bash
echo '{"test_prompt":"Create an automation with python_transform that updates brightness. Report success."}' | \
  python tests/uat/run_uat.py --agents gemini
```

It looks very mergeable otherwise. If you want to run your own BAT passes in the meantime, go for it — the skill READMEs cover everything you need. Otherwise we'll handle BAT on our end once the category test lands. Either way works.

Thanks again for the work on this — one of the cleanest responses to review feedback I've seen.

bzelch:
@kingpanther13 Added the categorized automation transform test — creates automation with category, transforms a property via python_transform, asserts category is preserved on the next get. 26 tests total now, all green. Ready for BAT whenever you are.

kingpanther13:
Looks good to me. I just tested in my live HA instance and it worked flawlessly. I don't think we need to do a full round of BAT tests, that's really meant for other types of tool changes. If your changes were to fail for whatever reason the original parameters of the tool should work fine. I'm going to go ahead and approve it. Nice work!

kingpanther13 (approved):

PR #945, opened 2026-04-11
test(e2e): add negative-input tests for ha_get_history and ha_get_automation_traces

Patch76:
## What does this PR do?

Adds three hard-asserting negative-input tests for `ha_get_history` and `ha_get_automation_traces`, covering paths with no prior hard coverage in either the unit or E2E suite. This continues the pattern established in the [#914 discussion](https://github.com/homeassistant-ai/ha-mcp/discussions/914) and follows up on the scope decision at the end of #924.

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 💥 Breaking change
- [x] 🧪 Tests only

## Background

After #924 was closed, I mapped the remaining uncovered negative-input paths for these two tools using the same source-level methodology: identify distinct failure branches in the tool code, verify coverage across both test layers, then write exactly one test per branch.

`ha_get_history` is relevant here because #911 introduced `test_tools_history.py` (2 unit tests, happy-path only) — the negative-input paths were explicitly out of scope there and remain uncovered.

### `ha_get_history` — `TestGetHistoryNegativeInputs` in `test_history.py`

**Test 1: `entity_ids=""`** — WS-validation path

| Step | Source | Evidence |
|---|---|---|
| Single-string branch: `entity_id_list = [entity_ids.strip()]` = `[""]` | `tools_history.py` | `"".strip()` = `""` → non-empty list |
| `if not entity_id_list` → False — no pre-flight, WS call proceeds | `tools_history.py` | `not [""]` = False |
| HA rejects: `valid_entity_id("") = False` → `send_error("invalid_entity_ids")` | `homeassistant/components/history/websocket_api.py` | HA core |
| `websocket_client.py`: `success=False` → `raise Exception("Command failed: Invalid entity_ids")` | `websocket_client.py:442` | live read |
| `exception_to_structured_error`: generic `Exception` matches no specific pattern → `INTERNAL_ERROR` | `helpers.py:222` | live read — not `SERVICE_CALL_FAILED` |

Expected: `success=False`, `error.code == "INTERNAL_ERROR"`
Prior coverage: none (`test_tools_history.py` has no negative-input tests; E2E `test_get_history_nonexistent_entity` uses `parse_mcp_result` + logging only — no assertion)

**Test 2: `entity_ids=[]`** — MCP pre-flight path

| Step | Source | Evidence |
|---|---|---|
[...]

PR #911, opened 2026-04-07
feat: consolidate ha_get_statistics into ha_get_history via source parameter

Patch76:
Closes #900.

Merges `ha_get_statistics` into `ha_get_history` following the `ha_get_logs` `source: Literal[...]` pattern already in the codebase.

## What changes

**`src/ha_mcp/tools/tools_history.py`**
- Add `source: Literal["history", "statistics"] = "history"` -- default preserves full backward compatibility
- - Source-dependent default look-back: 24h for history, 30d for statistics
- - Statistics-specific params (`period`, `statistic_types`) added -- silently ignored for `source="history"`
- - History-specific params (`minimal_response`, `significant_changes_only`, `limit`) silently ignored for `source="statistics"`
- - Extract `_fetch_history()` and `_fetch_statistics()` as module-level async helpers
- - Remove `ha_get_statistics` as standalone tool
**`tests/src/e2e/workflows/core/test_history.py`**
- Rename `TestGetStatistics` to `TestGetHistoryStatisticsSource`
- - Replace 5 direct `ha_get_statistics` calls with `ha_get_history(source="statistics", ...)`
## Precedent

`ha_get_logs` uses the identical pattern: `source: Literal["logbook", "system", "error_log", "supervisor"] = "logbook"`.

## Notes

- `tools.json`, `README.md`, `DOCS.md` are auto-regenerated by `sync-tool-docs.yml` on merge -- no manual updates needed.
- - Draft pending naming discussion in #900.
## Type of change
- [ ] Bug fix
- [ ] - [x] New feature
- [ ] - [ ] Documentation
- [ ] - [x] Maintenance/refactor
- [ ] - [ ] Breaking change
## Testing
- [x] I have tested these changes with a LLM agent
- [ ] - [ ] All automated tests pass (`uv run pytest`)
- [ ] - [x] Code follows style guidelines (`uv run ruff check`)

PR #975, opened 2026-04-13
fix(history): add "year" to valid statistics periods

Patch76:
## What

Adds `"year"` to the valid statistics periods in `_fetch_statistics`.

HA's `recorder/statistics_during_period` WebSocket API explicitly accepts `"year"` as a valid period (verified in `homeassistant/core` — `homeassistant/components/recorder/websocket_api.py`):

```python
vol.Required("period"): vol.Any("5minute", "hour", "day", "week", "month", "year")
```

The tool validator was missing it — an oversight from the original implementation that was carried over unchanged into #911. Passing `period="year"` currently raises `VALIDATION_INVALID_PARAMETER` even though HA would accept it.

Also fixes `test_get_statistics_different_periods` in the E2E suite (added in #945), which mirrored the incomplete `valid_periods` list.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

- [x] `test_get_statistics_different_periods` updated to include `"year"`
- [x] Existing `test_get_statistics_invalid_period` continues to reject non-listed values
- [x] All automated tests pass

## Checklist

- [x] I have updated documentation if needed (`tools.json` period description updated)

### DIFF
--- a/site/src/data/tools.json
+++ b/site/src/data/tools.json
@@ -1626,7 +1626,7 @@
           "default": null
         },
         "period": {
-          "type": "Annotated[str, Field(description='Aggregation period: \"5minute\", \"hour\", \"day\", \"week\", \"month\". Default: \"day\". Ignored when source=\"history\"', default='day')]",
+          "type": "Annotated[str, Field(description='Aggregation period: \"5minute\", \"hour\", \"day\", \"week\", \"month\", \"year\". Default: \"day\". Ignored when source=\"history\"', default='day')]",
           "default": "day"
         },
         "statistic_types": {
--- a/src/ha_mcp/tools/tools_history.py
+++ b/src/ha_mcp/tools/tools_history.py
@@ -186,7 +186,7 @@ class HistoryTools:
         period: Annotated[
             str,
             Field(
-                description='Aggregation period: "5minute", "hour", "day", "week", "month". Default: "day". Ignored when source="history"',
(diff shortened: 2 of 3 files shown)


kingpanther13 (approved):
Nice find, and thanks again for the contribution. Verified against HA core (`homeassistant/components/recorder/websocket_api.py:272-273`) — `vol.Any("5minute", "hour", "day", "week", "month", "year")` — so this is exactly the gap you describe.

The change is consistent across both source touch points (`tools_history.py:189` docstring and `:484` `valid_periods`) and the test at `test_history.py:442` extends the positive-path loop. Since `test_get_statistics_different_periods` uses `mcp_client.call_tool` directly (not `safe_call_tool`), a future regression that dropped `"year"` from `valid_periods` would propagate `ToolError` out of the iteration and fail the test — behavioral coverage is intact without needing a dedicated positive assertion.

One nit you can take or leave: `site/src/data/tools.json` is auto-regenerated on merge by `sync-tool-docs.yml` (AGENTS.md "CI/CD Workflows"), so the hand-edit at line 1629 is harmless but unnecessary — it'll be overwritten with the same value on merge. No strong preference on dropping vs keeping the hunk.

Approving.

PR #1000, opened 2026-04-16
refactor(internal): reduce C901 complexity in 5 non-tool files

sergeykad:
## Summary

- Extract helper functions to bring cyclomatic complexity below threshold in `backup.py`, `best_practice_checker.py`, `device_control.py`, `helpers.py`, and `registry.py`
- Remove 5 C901 ignores from `pyproject.toml` (down from 21 to 9)

## Behavior changes

While refactoring `bulk_device_control`, two behaviors were intentionally changed for consistency:

1. **Sequential bulk execution is now fail-soft** (previously fail-fast). Before: a `ToolError` from op #2 aborted the batch via the outer `except ToolError: raise` (op #3+ never ran). After: each op is wrapped in its own try/except, matching the parallel branch's `return_exceptions=True` semantics. Both execution modes now behave consistently.

2. **Structured error codes from `ToolError` are preserved in bulk results** (previously flattened). Before: `ToolError` (subclass of `Exception`) was caught by the generic branch and flattened to `f"Exception during execution: {str(e)}"`, losing the `ErrorCode` (e.g., `ENTITY_NOT_FOUND`, `VALIDATION_INVALID_JSON`) inside an opaque string. After: `_tool_error_to_dict` decodes the JSON-encoded error, preserving the structured code. Consumers pattern-matching on `error.code == "SERVICE_CALL_FAILED"` for per-op bulk failures will now see the original codes.

Tests added: `test_sequential_continues_after_tool_error` (op #2 raises, op #3 still executes), `test_parallel_preserves_tool_error_code` (asserts original `ErrorCode` survives).

## Details

**backup.py** (complexity 15, 12):
- `_poll_backup_completion`: extracted polling loop from `create_backup`
- `_create_safety_backup`: extracted safety backup creation from `restore_backup`
- Promoted polling constants to module-level (`_BACKUP_MAX_WAIT_S`, `_BACKUP_POLL_INTERVAL_S`)

**best_practice_checker.py** (complexity 11, 11):
- `_check_choose_actions`, `_check_repeat_actions`: extracted from `_check_action_tree`
- `_has_delay_or_wait_in_nested`: extracted from `_has_delay_or_wait`

**device_control.py** (complexity 13, 25, 15, 21):
- `_parse_parameters`, `_validate_entity_exists`: extracted from `control_device_smart`
- `_resolve_service_name`, `_normalize_light_color_temp`, `_add_domain_params`: extracted from `_build_service_call`
- `_predict_state_from_action`, `_predict_attributes_from_params`: extracted from `_predict_expected_state`
- `_validate_bulk_operations`, `_execute_parallel`, `_execute_sequential`, `_build_bulk_response`, `_tool_error_to_dict`: extracted from `bulk_device_control`
- `_DOMAIN_PARAMS` ClassVar replaces repeated inline parameter lists
- Removed unused `handler` parameter from `_build_service_call`

**helpers.py** (complexity 22):
- `_classify_exception`, `_classify_api_status`, `_classify_by_message`, `_append_macos_hints`: extracted from `exception_to_structured_error`

**registry.py** (complexity 11):
- `_import_and_register_module`: consolidated duplicated import-and-register logic

## Test plan

- [x] All 1045 unit tests pass (added 2 for the behavior changes above)
- [x] Ruff lint clean (including C901)
- [x] Mypy type check clean
- [x] ast-grep clean

Tracks #925

### DIFF
--- a/src/ha_mcp/tools/backup.py
+++ b/src/ha_mcp/tools/backup.py
@@ -27,6 +27,9 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
+_BACKUP_MAX_WAIT_S = 120
+_BACKUP_POLL_INTERVAL_S = 2
+
 
 def _get_backup_hint_text() -> str:
     """
@@ -85,6 +88,80 @@ async def _get_backup_password(
     return cast(str, default_password)
 
 
+async def _poll_backup_completion(
+    ws_client: HomeAssistantWebSocketClient,
+    name: str,
+    backup_job_id: str,
+    max_wait_seconds: int,
+    poll_interval: int,
+) -> dict[str, Any]:
+    """Poll backup/info until the named backup completes, fails, or times out.
+
+    Raises ToolError on backup failure or timeout.
+    """
+    waited = 0
+
+    while waited < max_wait_seconds:
+        await asyncio.sleep(poll_interval)
+        waited += poll_interval
+
+        info_result = await ws_client.send_command("backup/info")
+        if info_result.get("success"):
+            state = info_result.get("result", {}).get("state")
+            last_event = info_result.get("result", {}).get("last_action_event", {})
+            event_state = last_event.get("state")
+
+            logger.debug(
+                f"Backup state: {state}, event_state: {event_state}, waited: {waited}s"
+            )
+
+            if state == "idle" and event_state == "completed":
+                backups = info_result.get("result", {}).get("backups", [])
+                created_backup = None
+                for backup in backups:
+                    if backup.get("name") == name:
+                        created_backup = backup
+                        break
+
+                if created_backup:
+                    logger.info(
+                        f"Backup completed successfully: {created_backup.get('backup_id')}"
+                    )
+                    return {
+                        "success": True,
+                        "backup_id": created_backup.get("backup_id"),
+                        "backup_job_id": backup_job_id,
+                        "name": name,
+                        "date": created_backup.get("date"),
+                        "size_bytes": created_backup.get("agents", {})
+                        .get("hassio.local", {})
+                        .get("size"),
+                        "status": "Backup completed successfully",
+                        "duration_seconds": waited,
+                        "note": "Backup uses your Home Assistant's default backup password",
+                    }
+                else:
+                    logger.warning(
+                        "Backup completed but not found in backup list yet, waiting..."
+                    )
+                    continue
+
+            elif event_state == "failed":
+                raise_tool_error(create_error_response(
+                    ErrorCode.SERVICE_CALL_FAILED,
+                    "Backup creation failed",
+                    context={"backup_job_id": backup_job_id},
+                ))
+
+    logger.warning(f"Backup did not complete within {max_wait_seconds} seconds")
+    raise_tool_error(create_error_response(
+        ErrorCode.TIMEOUT_OPERATION,
+        f"Backup creation timed out after {max_wait_seconds} seconds",
+        context={"backup_job_id": backup_job_id, "name": name},
+        suggestions=["Backup may still be in progress. Check Home Assistant backup status."],
+    ))
+
+
 async def create_backup(
     client: HomeAssistantClient, name: str | None = None
 ) -> dict[str, Any]:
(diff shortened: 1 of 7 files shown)


src/ha_mcp/tools/device_control.py
```diff
@@ -641,6 +582,122 @@ def validate_operation(
...
+                    results.append(create_error_response(
+                        ErrorCode.SERVICE_CALL_FAILED,
+                        f"Exception during execution: {result!s}",
+                    ))
+                elif isinstance(result, dict):
+                    results.append(result)
+                    if "operation_id" in result:
+                        operation_ids.append(result["operation_id"])
+
+    async def _execute_sequential(
```
kingpanther13, line 629: **Behavior change** — not pure refactor.

Before this PR, `_execute_sequential` had no per-op `try/except`. A `ToolError` raised by op #2 propagated to the outer `except ToolError: raise` and aborted the whole batch (op #3+ never ran).

After the refactor, each op is wrapped in `try/except ToolError` and the batch continues — fail-fast becomes fail-soft.

This is arguably an improvement (it matches the parallel branch's `return_exceptions=True` semantics) but the PR description claims no behavioral change. Suggest either:
- Acknowledging this as an intentional consistency fix in the PR description and adding a sequential-mode test where op #2 raises a `ToolError` and op #3 still executes, or
- Reverting to the original passthrough behavior.

src/ha_mcp/tools/device_control.py
```diff
@@ -641,6 +582,122 @@ def validate_operation(
...
+                validate_first=op.get("validate_first", True),
+            )
+            tasks.append(task)
+
+        if tasks:
+            task_results = await asyncio.gather(*tasks, return_exceptions=True)
+
+            for result in task_results:
+                if isinstance(result, ToolError):
+                    results.append(self._tool_error_to_dict(result))
```
kingpanther13, line 618: **Behavior change** — same theme as the sequential branch.

Previously, `ToolError` (a subclass of `Exception`) was caught by the generic `isinstance(result, Exception)` branch and flattened to `f"Exception during execution: {result!s}"`, so the structured `ErrorCode` (e.g. `ENTITY_NOT_FOUND`, `VALIDATION_INVALID_JSON`) was lost inside an opaque outer string.

The new `_tool_error_to_dict` decode preserves the structured code in `results`. Genuine improvement, but consumers pattern-matching on `error.code == "SERVICE_CALL_FAILED"` for bulk failures will now see different codes. Worth a brief note in the PR description and a parallel-mode test (e.g. one op with malformed JSON parameters) that asserts the original code survives.

src/ha_mcp/tools/device_control.py
```diff
@@ -641,6 +582,122 @@ def validate_operation(
                 suggestions=["Check operation parameters and try again"],
             )
 
+    @staticmethod
+    def _tool_error_to_dict(e: ToolError) -> dict[str, Any]:
+        """Extract structured error dict from ToolError without double-encoding."""
+        try:
+            result: dict[str, Any] = json.loads(str(e))
+        except (json.JSONDecodeError, TypeError):
+            result = create_error_response(ErrorCode.SERVICE_CALL_FAILED, str(e))
```
kingpanther13, line 593: If `str(e)` is ever not JSON (a future change to `ToolError` serialization or an unusual subclass), this silently re-codes to `SERVICE_CALL_FAILED` with no diagnostic signal. A `logger.warning(f"Could not decode ToolError as structured response: {e!r}")` before the fallback would make a future regression here visible.

src/ha_mcp/tools/device_control.py
```diff
@@ -224,17 +190,64 @@ async def control_device_smart(
...
+                    ],
+                    context={"entity_id": entity_id, "action": action},
+                ))
+        return parameters
+
+    async def _validate_entity_exists(
         self,
         entity_id: str,
+        action: str,
+    ) -> dict[str, Any] | None:
```
kingpanther13: Return type is `dict[str, Any] | None`, but in practice this never returns `None` — the success path returns `current_state`, the `except ToolError: raise` re-raises, and `exception_to_structured_error(...)` in the outer except always raises. Tightening to `-> dict[str, Any]` would document intent and remove a phantom branch from type-checker output.

src/ha_mcp/tools/device_control.py
```diff
@@ -224,17 +190,64 @@ async def control_device_smart(
...
 
-    def _build_service_call(
+    def _parse_parameters(
+        self,
+        parameters: dict[str, Any] | None,
+        entity_id: str,
+        action: str,
+    ) -> dict[str, Any] | None:
+        if parameters and isinstance(parameters, str):
+            try:
```
kingpanther13, line 200: Stray blank line (presumably leftover from removing an inline `import json`). Safe to drop.

src/ha_mcp/tools/backup.py
```diff
@@ -85,6 +85,76 @@ async def _get_backup_password(
     return cast(str, default_password)
 
 
+async def _poll_backup_completion(
```
kingpanther13, line 91: The original inline polling loop had useful comments next to the magic numbers (`# 2 minutes max wait` next to `120`, `# Check every 2 seconds` next to `2`) and around the status checks. Those are gone, and the helper has no docstring. Suggest either:
- Promoting the polling constants to module-level (`_BACKUP_MAX_WAIT_S = 120`, `_BACKUP_POLL_INTERVAL_S = 2`), or
- Adding a one-liner to the helper, e.g. `"""Poll backup/info until the named backup completes, fails, or max_wait_seconds elapses; raises ToolError on failure/timeout."""`

src/ha_mcp/tools/backup.py
```diff
@@ -229,6 +232,38 @@ async def create_backup(
                 pass  # Ignore errors during cleanup
 
 
+async def _create_safety_backup(
```
kingpanther13, line 246: Non-obvious contract: `password is None → return None` means "safety backup intentionally skipped," not "failure." Without a docstring, callers can't tell the two `None` cases apart. A one-liner would help, e.g. `"""Create a pre-restore safety backup; returns None when password is None (skipped). Raises ToolError on creation failure."""`

src/ha_mcp/tools/registry.py
```diff
@@ -132,6 +132,35 @@ def _discover_tool_modules(self) -> list[str]:
 
         return discovered
 
+    def _import_and_register_module(
```
kingpanther13, line 135: Two non-obvious behaviors worth a one-line docstring: (1) explicit-`func_name` mode vs auto-discovery via `register_*_tools` convention, and (2) the `import_module` failure path is fail-fast (logs + re-raises), so registry init aborts on a broken module. Something like `"""Import a tools submodule and call its register function. Re-raises on import or registration failure."""`

kingpanther13 (changes requested):
Thanks for the cleanup — the 5-file extraction reads cleanly and the C901 ignore removals correspond exactly to the refactored functions. Verified `_DOMAIN_PARAMS` is identical to the four inline lists it replaces, and the `exception_to_structured_error` decomposition preserves classification precedence.

One meaningful concern: `bulk_device_control` has two observable behavior changes that the PR description bills as "no behavioral changes." Both are arguably improvements, but they should be either acknowledged + tested, or reverted. Inline comments below.

A few smaller suggestions on docstrings/cleanup as well.

src/ha_mcp/tools/device_control.py
sergeykad: Good catch. Acknowledged as an intentional consistency fix in the PR description and added `test_sequential_continues_after_tool_error` in d560afb.

src/ha_mcp/tools/device_control.py
sergeykad: Acknowledged in PR description. Added `test_parallel_preserves_tool_error_code` in d560afb to assert the original `ErrorCode` survives (using `VALIDATION_INVALID_JSON`).

src/ha_mcp/tools/device_control.py
sergeykad: Fixed in d560afb: added `logger.warning(f"Could not decode ToolError as structured response: {e!r}")` before the fallback.

src/ha_mcp/tools/device_control.py
sergeykad: Fixed in d560afb: tightened return type to `-> dict[str, Any]` and added a docstring.

src/ha_mcp/tools/device_control.py
sergeykad: Fixed in d560afb.

src/ha_mcp/tools/backup.py
sergeykad: Fixed in d560afb: promoted constants to module-level (`_BACKUP_MAX_WAIT_S`, `_BACKUP_POLL_INTERVAL_S`) and added a docstring to `_poll_backup_completion`.

src/ha_mcp/tools/backup.py
sergeykad: Fixed in d560afb: added docstring clarifying the `None` vs failure distinction.

src/ha_mcp/tools/registry.py
sergeykad: Fixed in d560afb: added docstring covering both modes and the fail-fast behavior.

kingpanther13 (approved):
Thanks for the thorough follow-through! All 8 comments addressed:

- Bulk execution behavior changes acknowledged in PR description under "Boy scout fixes" and locked in with two new unit tests (`test_sequential_continues_after_tool_error`, `test_parallel_preserves_tool_error_code`).
- `_tool_error_to_dict` now logs on JSON-decode fallback.
- `_validate_entity_exists` return type tightened to `dict[str, Any]`.
- Backup polling constants promoted to module-level with clear names.
- Docstrings added to `_poll_backup_completion`, `_create_safety_backup`, `_import_and_register_module`, and `_validate_entity_exists`.
- Stray blank line removed.

CI green across all 17 checks. LGTM.

PR #1006, opened 2026-04-17
docs: fix and clarify PR thread-resolution workflow in AGENTS.md

sergeykad:
## What does this PR do?

Fixes and clarifies the PR Review Comments section in `AGENTS.md` after discovering three issues while resolving a Gemini Code Assist thread on PR #1004:

1. **`resolveReviewThread` input field fix.** The mutation example used `pullRequestReviewThreadId`, but the input object takes `threadId`. GitHub rejects the old field with `missingRequiredInputObjectAttribute` and `argumentNotAccepted`.

2. **Inline reply workflow.** The previous example used `gh pr review --comment`, which posts a general PR comment. For responding to a specific review thread, `gh api repos/.../pulls/{pr}/comments/{id}/replies` posts the reply inline on the thread, which is what "post a comment explaining the resolution" intends.

3. **Thread ID lookup.** Resolving a thread requires the thread node ID (`PRRT_...`), which is not exposed in the REST comment payload. Added a GraphQL query that returns thread IDs alongside the `databaseId` of their first comment, so you can match back to a REST comment you already have.

Also: fixed `.author.login` → `.user.login` in the inline-comment check (REST uses `user`, GraphQL uses `author`), and exposed the `id` field so the reply URL segment is readable from the listing.

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

Doc-only change; no runtime impact. Every command shown was verified against PR #1004's review threads during this session.

## Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -134,24 +134,27 @@ gh issue list --state open --json number,title,labels --jq '.[] | select(.labels
 gh pr view <PR> --json comments --jq '.comments[] | {author: .author.login, created: .createdAt}'
 
 # Check inline review comments (specific to code lines)
-gh api repos/homeassistant-ai/ha-mcp/pulls/<PR>/comments --jq '.[] | {path: .path, line: .line, author: .author.login, created_at: .created_at}'
+gh api repos/homeassistant-ai/ha-mcp/pulls/<PR>/comments --jq '.[] | {id: .id, path: .path, line: .line, author: .user.login, created_at: .created_at}'
+
+# Get review-thread IDs (needed to resolve threads). Match by comment databaseId:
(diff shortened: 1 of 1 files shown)


kingpanther13:
Heads up I believe #949 addresses some of this

sergeykad:
Closing in favor of #949 by @kingpanther13, which covers the same fixes plus the dual-reply pattern and the merge-blocking behavior I missed. Will leave a note on #949 with two small extras that didn't overlap (`.user.login` vs `.author.login`, and exposing the comment `id` in the listing).

PR #1008, opened 2026-04-17
fix: classify Supervisor schema errors as VALIDATION_FAILED

### LINKED ISSUE #993: [BUG] ha_manage_addon config mode: Supervisor schema errors surface as SERVICE_CALL_FAILED / AUTH_INVALID_TOKEN instead of VALIDATION_FAILED
### 💬 What Happened?

During live testing of ha_manage_addon config mode in #978, Supervisor schema rejections surfaced with the wrong error codes. A partial nested option update (e.g. ssh.sftp only) causes Supervisor to reject the write with "Missing option 'authorized_keys' in ssh" -- but the tool returns AUTH_INVALID_TOKEN or INTERNAL_ERROR, not VALIDATION_FAILED.

The actionable detail is buried in the details field; the top-level code misleads the agent into retrying auth instead of fixing the payload.

```markdown
> No `ha_report_issue` output — this bug was observed during code review of #978,
> not from a running instance. Reproducing requires a live add-on with a nested
> option schema (SSH, Grocy, Node-RED confirmed by KP13 in #978 review).

---

### 🔍 Root Cause

`_supervisor_api_call` (`tools_addons.py` L103–115) already pattern-matches
`error_msg` for `"not_found"` / `"unknown"` → `RESOURCE_NOT_FOUND`.
Schema errors fall through to the generic `SERVICE_CALL_FAILED` block.
The reclassification guard in the config-mode caller is unreachable because
`_supervisor_api_call` raises `ToolError` internally before returning.

### 💡 Suggested Fix

Add a pattern-match block alongside the existing `not_found` check:


if any(p in error_msg.lower() for p in ("missing option", "invalid option")):
        ErrorCode.VALIDATION_FAILED,
        "Supervisor rejected configuration: schema validation failed",
        details=str(result),
        suggestions=["Fetch current options via ha_get_addon(slug) to see required fields"],
    ))


> **Note:** Depends on Supervisor error strings being stable across versions.
> If a structured error payload or HTTP status code is available from Supervisor,
> that would be preferable over string matching.

### 🔗 References

KP13 in #978 review (2026-04-16):
> "if there's a clean way to catch Supervisor schema errors in the exception
> handler and map them to `VALIDATION_FAILED`, that would help"
```

Patch76:
## What does this PR do?

Supervisor schema validation errors on `POST /addons/{slug}/options` surfaced as `AUTH_INVALID_TOKEN` or `INTERNAL_ERROR` instead of `VALIDATION_FAILED`, misleading LLM agents into retrying authentication instead of fixing the payload.

**Root cause.** Schema errors arrive in `_supervisor_api_call`'s exception handler as plain `Exception` (raised by `ws_client.send_command` when Supervisor rejects the payload with a `vol.Invalid`). The generic classifier in `exception_to_structured_error` routes messages via greedy substring matching, and misclassifies strings like `"Missing option 'authorized_keys' in ssh"` as `AUTH_INVALID_TOKEN` — because `"authorized_keys"` contains `"auth"`.

The reclassification guard in the config-mode caller (the old `if not result.get("success")` block in `ha_manage_addon`) was unreachable for schema errors: `send_command` raises on `success=False`, so the error never arrived as a dict.

**Why the exception handler.** This matches your suggestion in the #978 review ("if there's a clean way to catch Supervisor schema errors in the exception handler and map them to `VALIDATION_FAILED`"). That's exactly where schema errors land, so intercepting them there catches the symptom at the source rather than patching the downstream callers.

**Fix.** A pre-classifier in `_supervisor_api_call`'s exception handler intercepts schema validation errors before they fall into the greedy classifier. Gated on three conditions:

1. `method == "POST"`
2. `endpoint.startswith("/addons/")` and `endpoint.endswith("/options")`
3. `type(e) is Exception` — plain exception only, not a typed subclass

Only `POST /addons/*/options` triggers schema validation on the Supervisor side (confirmed against [supervisor/addons/options.py](https://github.com/home-assistant/supervisor/blob/main/supervisor/addons/options.py)), so endpoint-gating is deterministic — no string heuristics needed. `startswith` is used instead of `"/addons/" in endpoint` to reject unrelated paths that might happen to contain the substring (e.g. a future `/api/addons/...` shape). The plain-exception filter keeps typed exceptions (connection, auth, timeout) on their own classification path.

**Why not the approach in the issue body.** The issue suggested matching on substrings like `"missing option"` / `"invalid option"`, but the Supervisor error corpus is heterogeneous (`"Unknown secret"`, `"Unknown type"`, `"Device '...' does not exist"`, `"Fatal error for option"`, etc.) — a marker list would have gaps. Endpoint-gating catches all of them without false positives.

**What is NOT in this PR.**

- The greedy `"auth"` substring in `_classify_by_message` (`helpers.py`) is the underlying root cause for the specific `AUTH_INVALID_TOKEN` symptom. That classifier is used by 30+ tools; tightening it carries cross-cutting regression risk. Will open a follow-up issue after this lands.
- The analogous reclassification block for `GET /info` in `ha_manage_addon` has the same structural issue (reclassifies connection failures as `RESOURCE_NOT_FOUND`). Will fold into the follow-up.
- The outer `if not result.get("success")` block in `ha_manage_addon` (config-mode `POST` path) that KP13 flagged in #978 stays as-is: after this fix it becomes effectively dead for schema errors (the inner pre-classifier raises first), but removing it would let a Connection-Fail return dict flow past the success-check into the response path. Correcting it properly requires re-raising with the original error code preserved, which is a separate change.

## Type of change
- [x] 🐛 Bug fix

## Testing

- [x] All automated tests pass (`uv run pytest tests/src/unit/test_tools_addons.py`)

Three test changes. The existing `test_config_mode_supervisor_error_raises` sits directly on the code path this PR fixes; leaving it as a false-positive would continue to obscure the real failure mode, so correcting it is part of the fix rather than a separate change:

- `test_config_mode_supervisor_error_raises` mocked `_supervisor_api_call` returning a `{success: False}` dict, which is unreachable in production (`send_command` raises). Rewritten to match the real failure mode and renamed to `test_config_mode_supervisor_schema_error_raises`.
- New `TestSupervisorApiCall` class with two branch tests:
  - `test_post_options_plain_exception_classified_as_validation_failed` — match case.
  - `test_non_options_endpoint_passes_through_to_generic_classifier` — miss case via endpoint gate (bidirectional assertion per [merge principle #2](https://github.com/homeassistant-ai/ha-mcp/pull/876)).

A third test for the `type(e) is Exception` gate (typed exception on the same endpoint) was deliberately omitted: it would exercise `exception_to_structured_error`'s own paths rather than a distinct branch in the new code. Kept to two tests per "different branch, not different spelling" discipline.

## Checklist

Closes #993

### DIFF
--- a/src/ha_mcp/client/rest_client.py
+++ b/src/ha_mcp/client/rest_client.py
@@ -43,32 +43,15 @@ class HomeAssistantAPIError(HomeAssistantError):
         self.response_data = response_data
 
 
-
 class HomeAssistantCommandError(HomeAssistantError):
     """WebSocket command returned success=False.
 
     Raised by ``WebSocketClient.send_command`` when Home Assistant responds
-    with ``{type: "result", success: False}``. Carries the structured
-    ``{code, message}`` payload from the response so downstream
-    classification can dispatch on type + message without resorting to
-    ``str(exc)`` parsing.
-
-    Note: Supervisor API calls routed through HA Core's ``hassio/api`` WS
-    bridge always surface with ``code == "unknown_error"`` regardless of
-    the underlying Supervisor error kind (see
-    homeassistant/components/hassio/websocket_api.py). The discriminative
-    signal is the message payload, not the code.
+    with ``{type: "result", success: False}``. Used as a type marker in
+    ``_classify_exception``'s match dispatch; classification then falls
+    through to ``_classify_by_message`` for pattern matching on the
+    error message.
     """
-
-    def __init__(
-        self,
-        message: str,
-        code: str | None = None,
-        payload: dict[str, Any] | None = None,
-    ):
-        super().__init__(message)
-        self.code = code
-        self.payload = payload
 
 
 class HomeAssistantClient:
--- a/src/ha_mcp/client/rest_client.py
+++ b/src/ha_mcp/client/rest_client.py
@@ -43,6 +43,34 @@ class HomeAssistantAPIError(HomeAssistantError):
         self.response_data = response_data
 
 
+
+class HomeAssistantCommandError(HomeAssistantError):
+    """WebSocket command returned success=False.
+
+    Raised by ``WebSocketClient.send_command`` when Home Assistant responds
+    with ``{type: "result", success: False}``. Carries the structured
+    ``{code, message}`` payload from the response so downstream
+    classification can dispatch on type + message without resorting to
+    ``str(exc)`` parsing.
+
+    Note: Supervisor API calls routed through HA Core's ``hassio/api`` WS
+    bridge always surface with ``code == "unknown_error"`` regardless of
+    the underlying Supervisor error kind (see
+    homeassistant/components/hassio/websocket_api.py). The discriminative
+    signal is the message payload, not the code.
+    """
+
+    def __init__(
+        self,
+        message: str,
+        code: str | None = None,
+        payload: dict[str, Any] | None = None,
+    ):
+        super().__init__(message)
+        self.code = code
+        self.payload = payload
+
+
 class HomeAssistantClient:
     """Authenticated HTTP client for Home Assistant API."""
 
--- a/src/ha_mcp/tools/helpers.py
+++ b/src/ha_mcp/tools/helpers.py
@@ -182,7 +182,30 @@ def _classify_by_message(
         result = create_timeout_error("operation", 30, details=error_msg, context=context)
     elif "connection" in error_str or "connect" in error_str:
         result = create_connection_error(error_msg, context=context)
-    elif "auth" in error_str or "token" in error_str or "401" in error_str:
+    elif "command failed:" in error_str and any(
+        marker in error_str
+        for marker in (
+            "missing option",
+            "extra keys not allowed",
+            "expected",
+            "unknown secret",
+            "unknown type",
+        )
+    ):
+        # Supervisor schema validation: vol.Invalid message arriving as a
+        # HomeAssistantCommandError via HA Core's hassio WS bridge. The
+        # markers cover the heterogeneous vol.Invalid vocabulary without
+        # relying on an error code (always unknown_error from the bridge).
+        result = create_validation_error(error_msg, context=context)
+    elif any(
+        phrase in error_str
+        for phrase in (
+            "unauthorized",
+            "authentication",
+            "invalid token",
+            "access denied",
+        )
+    ) or "401" in error_str:
         result = create_auth_error(error_msg, context=context)
     else:
         result = create_error_response(
--- a/src/ha_mcp/tools/helpers.py
+++ b/src/ha_mcp/tools/helpers.py
@@ -7,6 +7,7 @@ Centralized utilities that can be shared across multiple tool implementations.
 import functools
 import json
 import logging
+import re
 import sys
 import time
 from typing import Any, Literal, NoReturn, overload
@@ -172,7 +173,29 @@ def _classify_by_message(
 ) -> dict[str, Any]:
     """Classify exception by error message patterns."""
     result: dict[str, Any]
-    if "not found" in error_str or "404" in error_str:
+    # Schema-branch must precede the "not found" / 404 branch (most-specific-first):
+    # a vol.Invalid message phrased like "Command failed: key X not found" would
+    # otherwise misclassify as RESOURCE_NOT_FOUND. The "command failed:" prefix
+    # gates the branch so non-schema WS errors fall through.
+    if "command failed:" in error_str and (
+        any(
+            marker in error_str
+            for marker in (
+                "missing option",
+                "extra keys not allowed",
+                "unknown secret",
+                "unknown type",
+            )
+        )
+        or re.search(r"expected (?:a |str|int|bool|dict|list|float|type|one of)", error_str)
+    ):
+        # Supervisor schema validation: vol.Invalid message arriving as a
+        # HomeAssistantCommandError via HA Core's hassio WS bridge. The
+        # markers plus the "expected <type>" anchor regex cover the
+        # heterogeneous vol.Invalid vocabulary without relying on an
+        # error code (always unknown_error from the bridge).
+        result = create_validation_error(error_msg, context=context)
+    elif "not found" in error_str or "404" in error_str:
         entity_id = context.get("entity_id") if context else None
         if entity_id:
             result = create_entity_not_found_error(entity_id, details=error_msg)
@@ -182,21 +205,6 @@ def _classify_by_message(
         result = create_timeout_error("operation", 30, details=error_msg, context=context)
     elif "connection" in error_str or "connect" in error_str:
         result = create_connection_error(error_msg, context=context)
-    elif "command failed:" in error_str and any(
-        marker in error_str
-        for marker in (
-            "missing option",
-            "extra keys not allowed",
-            "expected",
-            "unknown secret",
-            "unknown type",
-        )
-    ):
-        # Supervisor schema validation: vol.Invalid message arriving as a
-        # HomeAssistantCommandError via HA Core's hassio WS bridge. The
-        # markers cover the heterogeneous vol.Invalid vocabulary without
-        # relying on an error code (always unknown_error from the bridge).
-        result = create_validation_error(error_msg, context=context)
     elif any(
         phrase in error_str
         for phrase in (
@@ -207,6 +215,13 @@ def _classify_by_message(
         )
     ) or "401" in error_str:
         result = create_auth_error(error_msg, context=context)
+    elif error_str.startswith("command failed:"):
+        # HomeAssistantCommandError fallback: WS ``success=False`` with a
+        # message that doesn't match any specific marker above. This is a
+        # known failure mode (the WS command itself failed), not an
+        # unexpected internal error — route to SERVICE_CALL_FAILED,
+        # mirroring the 4xx fallback in _classify_api_status.
+        result = create_error_response(ErrorCode.SERVICE_CALL_FAILED, error_msg, context=context)
     else:
         result = create_error_response(
             ErrorCode.INTERNAL_ERROR, "An unexpected error occurred", details=error_msg, context=context
--- a/src/ha_mcp/tools/helpers.py
+++ b/src/ha_mcp/tools/helpers.py
@@ -16,6 +16,7 @@ from fastmcp.exceptions import ToolError
 from ..client.rest_client import (
     HomeAssistantAPIError,
     HomeAssistantAuthError,
+    HomeAssistantCommandError,
     HomeAssistantConnectionError,
 )
 from ..client.websocket_client import HomeAssistantWebSocketClient
@@ -142,6 +143,14 @@ def _classify_exception(
             )
         case HomeAssistantAPIError():
             result = _classify_api_status(error, error_msg, context)
+        case HomeAssistantCommandError():
+            # WebSocket command-failure. The ``error.code`` on Supervisor
+            # calls routed through HA Core's hassio WS bridge is always
+            # ``unknown_error`` (see homeassistant/components/hassio/
+            # websocket_api.py), so discrimination must come from the
+            # message. Fall through to ``_classify_by_message`` which
+            # pattern-matches schema, auth, not-found and timeout cases.
+            result = None
         case TimeoutError():
             operation = context.get("operation", "request") if context else "request"
             timeout_seconds = context.get("timeout_seconds", 30) if context else 30
--- a/src/ha_mcp/tools/tools_addons.py
+++ b/src/ha_mcp/tools/tools_addons.py
@@ -144,7 +144,7 @@ async def _supervisor_api_call(
             and type(e) is Exception
             and str(e).startswith("Command failed:")
         ):
-            slug = endpoint.split("/")[2] if endpoint.startswith("/addons/") else ""
+            slug = endpoint.split("/")[2]
             raise_tool_error(
                 create_error_response(
                     ErrorCode.VALIDATION_FAILED,
--- a/src/ha_mcp/tools/tools_addons.py
+++ b/src/ha_mcp/tools/tools_addons.py
@@ -126,6 +126,34 @@ async def _supervisor_api_call(
     except ToolError:
         raise
     except Exception as e:
+        # Pre-classify Supervisor schema validation errors.
+        # Schema errors from POST /addons/{slug}/options arrive here as plain
+        # Exception (raised by ws_client.send_command when Supervisor rejects
+        # the payload with a vol.Invalid). The generic classifier in
+        # exception_to_structured_error routes these by greedy substring match
+        # and misclassifies messages like "Missing option 'authorized_keys'
+        # in ssh" as AUTH_INVALID_TOKEN (due to "auth" in "authorized_keys").
+        # Endpoint-gating is used instead of string heuristics because
+        # vol.Invalid messages across Supervisor are heterogeneous, but only
+        # POST on /options triggers schema validation. Plain Exception filter
+        # keeps typed exceptions (connection, auth, timeout) on their own path.
+        if (
+            method == "POST"
+            and "/addons/" in endpoint
+            and endpoint.endswith("/options")
+            and type(e) is Exception
+        ):
+            raise_tool_error(
+                create_error_response(
+                    ErrorCode.VALIDATION_FAILED,
+                    "Supervisor rejected configuration: schema validation failed",
+                    details=str(e),
+                    suggestions=[
+                        "Fetch current options via ha_get_addon(slug) to see the schema",
+                        "Re-submit all required option fields together",
+                    ],
+                )
+            )
         logger.error(f"Error calling Supervisor API {endpoint}: {e}")
         exception_to_structured_error(
             e,
--- a/src/ha_mcp/tools/tools_addons.py
+++ b/src/ha_mcp/tools/tools_addons.py
@@ -142,6 +142,7 @@ async def _supervisor_api_call(
             and endpoint.startswith("/addons/")
             and endpoint.endswith("/options")
             and type(e) is Exception
+            and str(e).startswith("Command failed:")
         ):
             raise_tool_error(
                 create_error_response(
--- a/src/ha_mcp/tools/tools_addons.py
+++ b/src/ha_mcp/tools/tools_addons.py
@@ -144,13 +144,14 @@ async def _supervisor_api_call(
             and type(e) is Exception
             and str(e).startswith("Command failed:")
         ):
+            slug = endpoint.split("/")[2] if endpoint.startswith("/addons/") else ""
             raise_tool_error(
                 create_error_response(
                     ErrorCode.VALIDATION_FAILED,
                     "Supervisor rejected configuration: schema validation failed",
                     details=str(e),
                     suggestions=[
-                        "Fetch current options via ha_get_addon(slug) to see the schema",
+                        f"Fetch current options via ha_get_addon(slug='{slug}') to see the schema",
                         "Re-submit all required option fields together",
                     ],
                 )
--- a/src/ha_mcp/tools/tools_addons.py
+++ b/src/ha_mcp/tools/tools_addons.py
@@ -139,7 +139,7 @@ async def _supervisor_api_call(
         # keeps typed exceptions (connection, auth, timeout) on their own path.
         if (
             method == "POST"
-            and "/addons/" in endpoint
+            and endpoint.startswith("/addons/")
             and endpoint.endswith("/options")
             and type(e) is Exception
         ):
--- a/src/ha_mcp/tools/tools_addons.py
+++ b/src/ha_mcp/tools/tools_addons.py
@@ -126,36 +126,6 @@ async def _supervisor_api_call(
     except ToolError:
         raise
     except Exception as e:
-        # Pre-classify Supervisor schema validation errors.
-        # Schema errors from POST /addons/{slug}/options arrive here as plain
-        # Exception (raised by ws_client.send_command when Supervisor rejects
-        # the payload with a vol.Invalid). The generic classifier in
-        # exception_to_structured_error routes these by greedy substring match
-        # and misclassifies messages like "Missing option 'authorized_keys'
-        # in ssh" as AUTH_INVALID_TOKEN (due to "auth" in "authorized_keys").
-        # Endpoint-gating is used instead of string heuristics because
-        # vol.Invalid messages across Supervisor are heterogeneous, but only
-        # POST on /options triggers schema validation. Plain Exception filter
-        # keeps typed exceptions (connection, auth, timeout) on their own path.
-        if (
-            method == "POST"
-            and endpoint.startswith("/addons/")
-            and endpoint.endswith("/options")
-            and type(e) is Exception
-            and str(e).startswith("Command failed:")
-        ):
-            slug = endpoint.split("/")[2]
-            raise_tool_error(
-                create_error_response(
-                    ErrorCode.VALIDATION_FAILED,
-                    "Supervisor rejected configuration: schema validation failed",
-                    details=str(e),
-                    suggestions=[
-                        f"Fetch current options via ha_get_addon(slug='{slug}') to see the schema",
-                        "Re-submit all required option fields together",
-                    ],
-                )
-            )
         logger.error(f"Error calling Supervisor API {endpoint}: {e}")
         exception_to_structured_error(
             e,
--- a/tests/src/unit/test_tool_error_signaling.py
+++ b/tests/src/unit/test_tool_error_signaling.py
@@ -220,3 +220,64 @@ class TestIntegrationWithMCPProtocol:
         error_data = json.loads(error_message)
         assert error_data["success"] is False
         assert "VALIDATION" in error_data["error"]["code"]
+
+
+class TestSchemaAndAuthClassification:
+    """Tests for _classify_by_message schema and auth branches (issue #993).
+
+    Pins two behaviours at the classifier boundary:
+    1. Supervisor vol.Invalid messages route to VALIDATION_FAILED via
+       the "Command failed: ... missing option" branch.
+    2. Messages that merely contain the substring "auth" (e.g.
+       "Did not receive auth_required message", "authorized_keys")
+       are NOT misclassified as AUTH_INVALID_TOKEN. Only the phrase
+       list (unauthorized, authentication, invalid token, access
+       denied) matches the auth branch.
+    """
+
+    def test_command_error_with_schema_marker_is_validation_failed(self):
+        """HomeAssistantCommandError carrying vol.Invalid markers => VALIDATION_FAILED."""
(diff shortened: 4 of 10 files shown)


tests/src/unit/test_tools_addons.py
```diff
@@ -1082,17 +1082,41 @@ async def test_config_mode_network(self, manage_addon_tool):
...
         with patch(
             "ha_mcp.tools.tools_addons._supervisor_api_call",
-            return_value={"success": False, "error": "boot_config locked"},
+            side_effect=schema_reject,
         ), pytest.raises(ToolError) as exc_info:
-            await manage_addon_tool(slug="test_addon", boot="auto")
+            await manage_addon_tool(
+                slug="test_addon",
+                options={"ssh": {"sftp": True}},
+            )
```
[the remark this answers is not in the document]
Patch76: Fixed in 827b28d. The parameter is `options`, not `config`.

src/ha_mcp/tools/tools_addons.py
```diff
@@ -126,6 +126,34 @@ async def _supervisor_api_call(
...
+        # Endpoint-gating is used instead of string heuristics because
+        # vol.Invalid messages across Supervisor are heterogeneous, but only
+        # POST on /options triggers schema validation. Plain Exception filter
+        # keeps typed exceptions (connection, auth, timeout) on their own path.
+        if (
+            method == "POST"
+            and endpoint.startswith("/addons/")
+            and endpoint.endswith("/options")
+            and type(e) is Exception
+        ):
```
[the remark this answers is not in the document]
Patch76: Addressed in 8b59e2e by adding `str(e).startswith("Command failed:")` to the guard. This scopes the pre-classifier to exceptions raised by `ws_client.send_command` itself, so internal errors like `"Command timeout"` or `"WebSocket not authenticated"` keep their normal classification path.

src/ha_mcp/tools/tools_addons.py
```diff
@@ -126,6 +126,35 @@ async def _supervisor_api_call(
...
+                create_error_response(
+                    ErrorCode.VALIDATION_FAILED,
+                    "Supervisor rejected configuration: schema validation failed",
+                    details=str(e),
+                    suggestions=[
+                        "Fetch current options via ha_get_addon(slug) to see the schema",
+                        "Re-submit all required option fields together",
+                    ],
+                )
+            )
```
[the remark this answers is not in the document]
Patch76: Addressed the slug suggestion in 94e6a88 — `endpoint.split("/")[2]` yields the slug deterministically since the gate already requires `startswith("/addons/")`, and the concrete value lands in the suggestion.

Leaving `str(e)` as-is for `details=`. `type(e) is Exception` restricts the branch to plain `Exception` (subclasses take their own classifier path), and plain `Exception` has no `.message` attribute — `getattr(e, "message", str(e))` would always fall through to `str(e)` here. No parsing happens; the string is passed through verbatim.

src/ha_mcp/tools/tools_addons.py
```diff
@@ -126,6 +126,36 @@ async def _supervisor_api_call(
...
+        # POST on /options triggers schema validation. Plain Exception filter
+        # keeps typed exceptions (connection, auth, timeout) on their own path.
+        if (
+            method == "POST"
+            and endpoint.startswith("/addons/")
+            and endpoint.endswith("/options")
+            and type(e) is Exception
+            and str(e).startswith("Command failed:")
+        ):
+            slug = endpoint.split("/")[2] if endpoint.startswith("/addons/") else ""
```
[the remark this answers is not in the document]
Patch76: Removed in aa3ef48. Correct — the outer gate already guarantees `startswith("/addons/")`, the fallback was dead code.

kingpanther13 (changes requested):
Thanks for the thorough write-up and clean iteration on Gemini's feedback — the diff itself is tight. But stepping back: this intercepts the symptom at one endpoint rather than fixing the underlying bug, and the fix for the root cause is small enough that there's no real reason to prefer a workaround.

`test_non_options_endpoint_passes_through_to_generic_classifier` literally documents that the misclassification still fires on `/info` and every other path (test comment at tests/src/unit/test_tools_addons.py:1343-1347: "asserts the known misclassification still occurs"). The greedy `"auth" in error_str` at `helpers.py:176` that causes the whole problem is untouched. Anywhere else this substring trips, we're still shipping the bug.

Two small options, both fix #993 on every endpoint:

**Option A — Phrase-list fix (~5–8 LOC in `_classify_by_message`):**
1. Replace `"auth" in error_str or "token" in error_str` with a phrase list: `any(p in error_str for p in ("unauthorized", "authentication", "invalid token", "access denied"))`.
2. Add a schema branch *before* the auth one: `elif "Command failed:" in error_str and ("Missing option" in error_str or "extra keys not allowed" in error_str or "expected" in error_str)` → `create_validation_error`.

`test_non_options_endpoint_passes_through_to_generic_classifier` flips from documenting the bug to verifying the fix.

**Option B — Typed exception (~15 LOC, structurally cleaner):**
`websocket_client.py:442` and `:519` raise bare `Exception(f\"Command failed: {error_msg}\")` and discard the structured `{error: {code, message}}` dict Supervisor actually sends. Raise a typed `HomeAssistantCommandError` carrying the payload; add a `case` in `_classify_exception`. Eliminates string-matching for Supervisor errors entirely.

Either way, the 30-line endpoint-gated pre-classifier in `tools_addons.py` becomes unnecessary.

On the stated regression-risk argument: typed auth errors (`HomeAssistantAuthError`, 401/403 via `HomeAssistantAPIError`) go through `_classify_exception`'s type dispatch at helpers.py:124-156 *before* `_classify_by_message`. Only untyped plain `Exception`s reach the fallback, and `grep` across `tests/` turns up zero tests asserting `AUTH_INVALID_TOKEN` from a string-matched path. The blast radius is smaller than it looks — there isn't a strong case for patching around the bug when a handful of lines fixes it at the source.

Happy to approve either approach once it lands on the root cause. Would you be up for rolling the fix into this PR?

Patch76:
Thanks @kingpanther13  — you're right, and the history helps explain how I ended up at the wrong layer:

The original suggestion in #993 was a phrase-match in `_supervisor_api_call`'s `not_found` block. When I went to implement it, I found that the `success=False` branch in `_supervisor_api_call` is **effectively dead code** — `send_command` converts `{success: False}` responses into a plain `Exception` at `websocket_client.py:442` before the caller ever sees the dict, so anything after `if not result.get("success")` in `_supervisor_api_call` is unreachable in production. From there I moved one layer further up in the same file instead of one layer down to `helpers.py` / `websocket_client.py`. Wrong direction.

---

### One additional finding worth flagging before I implement

Option B alone doesn't fully close the greedy substring issue at `helpers.py:176`. To illustrate the pattern:

> `websocket_client.py:221` raises `Exception("Did not receive auth_required message")` — not an auth failure, but `"auth" in error_str` matches and classifies it as `AUTH_INVALID_TOKEN`.

Same pattern applies to any future untyped exception whose message contains `"auth"` or `"token"`. Option A's phrase list closes that:

```python
any(p in error_str for p in (
    "unauthorized", "authentication", "invalid token", "access denied"
))
```

Your blast-radius point holds — type-dispatched exceptions skip `_classify_by_message` entirely, and `grep` across `tests/` confirms **no test asserts `AUTH_INVALID_TOKEN` from a string-matched path**.

---

### Plan for one PR

| # | Change | Where | Purpose |
|---|---|---|---|
| 1 | **Option B** — typed `HomeAssistantCommandError` carrying the Supervisor payload, match-case branch in `_classify_exception` | `websocket_client.py` + `helpers.py` | Fixes #993 at the source |
| 2 | **Option A** — replace `"auth"/"token"` with the phrase list above | `helpers.py:176` | Closes latent misclassification for remaining untyped `Exception` paths |
| 3 | **Revert** the endpoint-gated pre-classifier (all 30 LOC); flip `test_non_options_endpoint_passes_through_to_generic_classifier` from documenting the bug to verifying the fix | `tools_addons.py` + `test_tools_addons.py` | Removes the wrong-layer workaround |

Commit split is your call — happy to do one combined or split B and A for review clarity.

kingpanther13:
Since Claude seems to believe it won't be very many lines, I'd say just go ahead and do both options and we'll go from there.

Patch76:
@kingpanther13 — plan from your review implemented in three commits (`be97f86` → `2a5ac5e` → `504159c`):

1. **Option B** — `raise Exception(f"Command failed: ...")` at `websocket_client.py:442` and `:524` replaced with a typed `HomeAssistantCommandError` that carries the structured `{code, message}` payload. Matching `case` in `_classify_exception`. One note on the design: Supervisor errors routed through HA Core's `hassio` WS bridge always surface with `code == "unknown_error"` ([core `websocket_api.py` L149-151](https://github.com/home-assistant/core/blob/dev/homeassistant/components/hassio/websocket_api.py)), so the discriminative signal is the message payload — the `case` delegates to `_classify_by_message` rather than doing code-based dispatch. Docstring on the class spells this out.

2. **Option A** — `"auth" in error_str or "token" in error_str` replaced with the phrase list (`unauthorized`, `authentication`, `invalid token`, `access denied`). Schema branch added ahead of the auth branch, gated on `"command failed:" in error_str` plus markers (`missing option`, `extra keys not allowed`, `expected`, `unknown secret`, `unknown type`). `401` numeric signal preserved on the auth branch.

3. **Revert** of the endpoint-gated pre-classifier in `tools_addons.py` (the 30 LOC you flagged). `test_non_options_endpoint_passes_through_to_generic_classifier` flipped from documenting the bug to verifying the fix — now asserts `VALIDATION_FAILED` on `/info` with the same schema-shaped message. Renamed to `test_schema_error_on_non_options_endpoint_also_classified`. `test_config_mode_supervisor_schema_error_raises` reworked to the merge-based mock pattern.

**New tests** in `test_tool_error_signaling.py` (`TestSchemaAndAuthClassification`, 5 tests) pin the classifier directly at the root cause:
- `HomeAssistantCommandError` with vol.Invalid markers → `VALIDATION_FAILED`
- `"authorized_keys"` substring no longer triggers `AUTH_INVALID_TOKEN`
- `"Did not receive auth_required message"` no longer matches auth (handshake path at `websocket_client.py:221` — same greedy-substring bug, different call site)
- genuine auth vocabulary still classified
- `401` numeric signal still classified

**Verification:**
- `uv run pytest tests/src/unit/` — 1070 passed, 1 skipped (numpy), 0 failures
- `uv run mypy src/ scripts/` — clean
- `uv run ruff check` (full CI scope) — clean
- `uv run ast-grep scan` — clean
- CI on HEAD `504159c` — 13/13 success
- Live BAT against a real HA Supervisor (Mosquitto add-on, partial payload): happy path regression-free, no state drift. The schema-reject path wasn't reached in that run because `_merge_options` correctly completes the partial payload — consistent with the PR #978 design, and confirms the classifier fix is the right layer for the residual bug.

src/ha_mcp/tools/helpers.py
```diff
@@ -173,7 +182,30 @@ def _classify_by_message(
...
+        result = create_validation_error(error_msg, context=context)
+    elif any(
+        phrase in error_str
+        for phrase in (
+            "unauthorized",
+            "authentication",
+            "invalid token",
+            "access denied",
+        )
+    ) or "401" in error_str:
```
kingpanther13, line 216: The phrase list drops three existing raise sites that the old `"auth"/"token"` substring caught. All three propagate to this classifier in production (verified):

- `websocket_client.py:415, 492` — `raise Exception("WebSocket not authenticated")`. `"authenticated"` doesn't contain `"authentication"` as a substring, so this now lands in INTERNAL_ERROR.
- `__main__.py:72` — `raise RuntimeError("No OAuth token in request context")`. `"token"` removed from the list.
- `__main__.py:81` — `raise RuntimeError("No Home Assistant credentials in OAuth token claims")`. Same.

Old AUTH classification was approximate, but `INTERNAL_ERROR` ("An unexpected error occurred") is a strictly worse signal than the old AUTH code with suggestions. Two fixes:

1. Extend the phrase list with `"not authenticated"` and `"oauth token"`.
2. Convert those four raise sites to typed exceptions (`HomeAssistantAuthError` for the OAuth ones, `HomeAssistantConnectionError` for the handshake/state ones).

(2) is the principled fix — the greedy-substring problem only exists because we're doing string classification on plain Exceptions. Prefer (2).

src/ha_mcp/tools/helpers.py
```diff
@@ -173,7 +182,30 @@ def _classify_by_message(
         result = create_timeout_error("operation", 30, details=error_msg, context=context)
     elif "connection" in error_str or "connect" in error_str:
         result = create_connection_error(error_msg, context=context)
-    elif "auth" in error_str or "token" in error_str or "401" in error_str:
+    elif "command failed:" in error_str and any(
```
kingpanther13: Ordering fragility: the `"not found"`/`404` branch at line 175 fires **before** this schema branch. A future vol.Invalid phrased `"Command failed: ... key X not found in dict"` (vol does emit "not found"-style text in some invariants) would misclassify as `RESOURCE_NOT_FOUND`.

Either gate line 175 with `and "command failed:" not in error_str`, or hoist the schema branch above the `"not found"` one.

src/ha_mcp/tools/helpers.py
```diff
@@ -173,7 +182,30 @@ def _classify_by_message(
...
+        for phrase in (
+            "unauthorized",
+            "authentication",
+            "invalid token",
+            "access denied",
+        )
+    ) or "401" in error_str:
         result = create_auth_error(error_msg, context=context)
     else:
         result = create_error_response(
```
kingpanther13, line 226: When a `HomeAssistantCommandError` hits no schema/auth/not-found/timeout/connection marker, it lands here at INTERNAL_ERROR ("An unexpected error occurred"). For an HA command that explicitly returned `success=False`, that's masking a known failure mode behind a generic "unexpected" label.

Add a terminal `elif error_str.startswith("command failed:")` branch routing to `ErrorCode.SERVICE_CALL_FAILED` before the else. Matches the 400-path fallback semantics in `_classify_api_status` at line 121.

src/ha_mcp/tools/helpers.py
```diff
@@ -173,7 +182,30 @@ def _classify_by_message(
         result = create_timeout_error("operation", 30, details=error_msg, context=context)
     elif "connection" in error_str or "connect" in error_str:
         result = create_connection_error(error_msg, context=context)
-    elif "auth" in error_str or "token" in error_str or "401" in error_str:
+    elif "command failed:" in error_str and any(
+        marker in error_str
+        for marker in (
+            "missing option",
+            "extra keys not allowed",
+            "expected",
```
kingpanther13: `"expected"` is broader than the other four markers. The `"command failed:"` prefix mostly contains it, but any future non-schema WS error containing that word would hit VALIDATION_FAILED.

Tighten to the vol.Invalid patterns: `"expected a "`, `"expected str"`, `"expected int"`, `"expected dict"`, etc. Or keep `"expected"` with an anchor regex: `re.search(r"expected (?:a |str|int|bool|dict|list|float)", error_str)`.

tests/src/unit/test_tool_error_signaling.py
```diff
@@ -220,3 +220,64 @@ def simulated_tool_call():
         error_data = json.loads(error_message)
         assert error_data["success"] is False
         assert "VALIDATION" in error_data["error"]["code"]
+
+
+class TestSchemaAndAuthClassification:
```
kingpanther13, line 225: Three coverage gaps in this class:

1. **Only 1 of 5 schema markers and 1 of 4 auth phrases exercised.** `"missing option"` and `"unauthorized"` have tests; `"extra keys not allowed"`, `"expected"`, `"unknown secret"`, `"unknown type"` and `"authentication"`, `"invalid token"`, `"access denied"` are untested. Drop any of them from the source tuple in helpers.py and no test fails. Parametrize the two "still classified" tests over the full tuples.

2. **Missing negative test for the outer `"command failed:"` gate.** No test asserts that `Exception("Missing option 'foo' in bar")` (without the prefix) does NOT classify as VALIDATION_FAILED. Drop the gate and no test catches it. Add:

```python
def test_schema_phrase_without_command_prefix_not_validation(self):
    exc = Exception("Missing option 'foo' in bar")
    result = exception_to_structured_error(exc, raise_error=False)
    assert result["error"]["code"] != "VALIDATION_FAILED"
```

3. **`HomeAssistantCommandError` contract is untested.** No test pins that `send_command` actually constructs and raises it with `code` and `payload` preserved — `TestSupervisorApiCall` mocks `send_command` directly, so the raise sites at `websocket_client.py:443-447` (dict-error branch) and `:524-528` (`send_command_with_event`) are uncovered. Add a test with a mock WebSocket returning `{type: "result", success: False, error: {code, message}}` and assert the raised exception's `.code` and `.payload`. Cover the string-error branch too.

src/ha_mcp/client/rest_client.py
```diff
@@ -43,6 +43,34 @@ def __init__(
...
+    the underlying Supervisor error kind (see
+    homeassistant/components/hassio/websocket_api.py). The discriminative
+    signal is the message payload, not the code.
+    """
+
+    def __init__(
+        self,
+        message: str,
+        code: str | None = None,
+        payload: dict[str, Any] | None = None,
```
kingpanther13: Two things:

**`.code` and `.payload` are currently dead weight.** The sole consumer in `_classify_by_message` parses `str(exc)` with substring markers and ignores the attributes. Either wire the consumer to dispatch on `.code`/`.payload` (e.g., when `payload is not None and "message" in payload`, use that directly; use `.code` as a discriminator if HA Core ever starts emitting specific codes), or drop them.

**`payload: dict[str, Any] | None` understates the shape.** HA's WS error payload always has `{code: str, message: str}`. A TypedDict expresses the invariant:

```python
class WSErrorPayload(TypedDict):
    code: str
    message: str
```

Also nit: 3 blank lines between `HomeAssistantAPIError` and `HomeAssistantCommandError` where the rest of this file uses 2.

kingpanther13 (changes requested):
Rework lands the #993 root-cause fix cleanly — Option A phrase list, Option B typed `HomeAssistantCommandError`, pre-classifier fully reverted, `test_schema_error_on_non_options_endpoint_also_classified` flipped to positive assertion. But the phrase-list tightening silently downgrades three existing auth-state error messages to `INTERNAL_ERROR`, plus a few classifier-ordering, test-coverage, and type-design issues. All tractable in this PR. Details inline.

src/ha_mcp/tools/helpers.py
Patch76: Thanks — went with option (2), plus one more raise site your logic applies to.

Converted in 61a6e8f:
  websocket_client.py:415, :492   HomeAssistantConnectionError("WebSocket not authenticated")
  __main__.py:72                  HomeAssistantAuthError("No OAuth token in request context")
  __main__.py:81                  HomeAssistantAuthError("No Home Assistant credentials in OAuth token claims")

Plus websocket_client.py:222 (handshake) — raise Exception("Did not receive auth_required message"). Before the phrase list, it matched "auth" in error_str and landed as AUTH; after the tightening it would fall into INTERNAL_ERROR. Transport handshake failure semantically → HomeAssistantConnectionError. Your state-vs-credentials distinction applies directly, so I folded it in rather than leaving a known classification regression. Happy to split it into a follow-up if you prefer a tighter scope.

Test updates:
  - test_auth_required_handshake_not_auth_error → test_auth_required_handshake_is_connection_error with a positive assertion (== CONNECTION_FAILED) instead of the earlier != AUTH_INVALID_TOKEN
  - test_oauth_proxy_client_no_token_raises_error / test_oauth_proxy_client_missing_claims_raises_error now expect HomeAssistantAuthError instead of RuntimeError
  - New test_typed_auth_error_classified_as_auth and test_typed_connection_error_classified_as_connection pin the type-dispatch paths (63c12e8)

Grep across tests/ confirmed no other assertion asserts the old string-matched AUTH path.

src/ha_mcp/tools/helpers.py
Patch76: Hoisted the schema branch above the not-found branch in 4fb13e3 — most-specific-first, and the "command failed:" prefix gate keeps non-schema WS errors from being swallowed. Running through the existing branches afterwards: no test shifted classification (no existing "not found" test contains a "command failed:" prefix).

Went with hoist rather than gating the not-found branch with and "command failed:" not in error_str — adding negative-conditions couples two semantically distinct branches and doesn't scale if another branch later needs the same shield. Hoist keeps each branch self-contained.

src/ha_mcp/tools/helpers.py
Patch76: Added a terminal elif error_str.startswith("command failed:") → SERVICE_CALL_FAILED before the else (4fb13e3). Mirrors the 4xx fallback pattern in _classify_api_status at helpers.py:121. Pinned by test_command_error_unknown_message_is_service_call_failed in 63c12e8.

One downstream effect worth flagging: the e2e test test_empty_string_entity_id_rejected in test_history.py had pinned the old INTERNAL_ERROR output (the empty entity_id reaches the WS handler and returns success=False). Updated the assertion to SERVICE_CALL_FAILED in d9f7b3e — the new classification is semantically what the test was really reaching for.

src/ha_mcp/tools/helpers.py
Patch76: Replaced the bare "expected" marker with an anchor regex in 4fb13e3:
re.search(r"expected (?:a |str|int|bool|dict|list|float|type|one of)", error_str)

Covers the vol.Invalid variants ("expected a string", "expected str", "expected int for ...", "expected type", "expected one of [...]"). The anchor prevents false-positive matches on unrelated phrases that happen to contain the word. 9 positive cases + 2 negative cases parametrized in TestSchemaAndAuthClassification::test_schema_marker_classified_as_validation_failed (63c12e8).

tests/src/unit/test_tool_error_signaling.py
Patch76: All three addressed in 63c12e8:

1. Parametrized the schema branch over 13 cases (4 fixed markers + 9 "expected" regex anchors) and the auth branch over the full 4-phrase tuple. Each parametrized id maps 1:1 to a source tuple entry — removing any entry fails the corresponding case, so the "mutation test" motivation holds.

2. Added test_schema_phrase_without_command_prefix_not_validation — a plain Exception("Missing option 'foo' in bar") without the "Command failed:" prefix must NOT route to VALIDATION_FAILED. Dropping the outer gate in helpers.py now fails this test.

3. Added TestSendCommandErrorContract in test_websocket_client.py covering the two HomeAssistantCommandError raise sites (websocket_client.py:443 and :524) across the dict-error branch, string-error branch, and the send_command_with_event variant. Mock strategy resolves the pending-response future from inside a stubbed send_json_message — avoids depending on the private message-id counter and keeps the test robust to internal state changes.

src/ha_mcp/client/rest_client.py
Patch76: Dropped .code and .payload entirely in cde1d73 — no consumer reads them (verified via grep), and wiring them in wouldn't change behaviour for the primary use case: Supervisor errors routed through HA Core's hassio WS bridge always surface with code == "unknown_error", so .code as a discriminator adds zero information today and payload["message"] is identical to str(exc). Keeping reserved attributes for hypothetical future HA-Core code changes felt like exactly the dead-weight pattern you flagged.

The class stays as a pure type marker — that's what gives the match-case dispatch in _classify_exception its value (routes to _classify_by_message for the schema/auth/command-failed branches instead of being caught by the generic fallback).

Callers updated in cde1d73:
  - websocket_client.py:443, :524 — raise HomeAssistantCommandError(f"Command failed: {error_msg}")
  - test_tools_addons.py:1310, :1346 and test_tool_error_signaling.py:242 — dropped code="unknown_error" kwargs

Blank-line nit: reduced from 3 to 2 before the class. Left the rest of the file's inconsistent spacing as-is (the convention is broken throughout) — happy to clean the whole file in a follow-up if you'd like, but kept the change scoped to your finding.

Patch76:
@kingpanther13 — all six threads addressed. Summary of what landed and why for the three findings that had non-trivial choices:

**Scope 1: B1 widened to 5 raise sites (vs. the 4 you named).** Site `websocket_client.py:222` — `raise Exception("Did not receive auth_required message")` — was in scope for the same reason as the other four: it matched the old greedy `"auth"` via `"auth_required"` and falls into INTERNAL_ERROR under the tightened phrase list. Transport handshake failure → `HomeAssistantConnectionError`. I folded it in rather than leave a known regression, but I can split it into a follow-up if you'd rather keep this PR strictly bounded to the four you called out.

**Ordering fix (B2) went with hoist, not local gate.** Gating the not-found branch with `and "command failed:" not in error_str` would couple two branches every time a new specific type enters the chain. Hoisting the schema branch — most-specific first — keeps each branch self-contained and scales. No existing test shifted classification from the reorder.

**B6 went with drop, not wire.** Wiring `.payload["message"]` into `_classify_by_message` would require changing its signature to receive the exception object, and for the current HA-Core-via-hassio path `code == "unknown_error"` always, so the discriminative signal is the message — which `str(exc)` already provides. Keeping dead attributes with a "reserved for future" docstring is the exact pattern you flagged in the original thread. The typed class itself stays because the match-case dispatch is where its value lives.

**Secondary effect on an e2e test** (d9f7b3e): `test_empty_string_entity_id_rejected` in `test_history.py` had pinned the old INTERNAL_ERROR output for a WS `success=False`. With B3 (terminal `command failed:` branch) that now classifies as SERVICE_CALL_FAILED — semantically more accurate. Updated the assertion and docstring to pin the new behaviour.

**What's not in this PR** — follow-up candidates if you want them:
- The other bare `raise Exception(...)` sites in `websocket_client.py` (lines 236, 237, 269, 290, 372, 376, 466, 560, 723, 769, 799) — transport/state guards that could become typed, but none surface via the classifier path this PR touches.
- The `or`-fallback branches in the WS raise sites' message derivation (`error.get("message", str(error)) if isinstance(error, dict) else str(error)`) — same pattern you flagged in #974 as schema-drift risk. Narrow scope; a separate pass would read cleaner.

Commits:
- `61a6e8f` — B1 — 5 typed exceptions
- `4fb13e3` — B2+B3+B4 — hoist schema, expected-regex, command-failed fallback
- `cde1d73` — B6 — drop .code/.payload
- `63c12e8` — B5 — parametrize matrix, contract tests, fallback test
- `d9f7b3e` — downstream e2e assertion update

kingpanther13 (approved):
Six-for-six. Option 2 on the typed exceptions (my preferred path), schema branch hoisted ahead of not-found, terminal `command failed:` fallback to SERVICE_CALL_FAILED, regex-anchored `expected` marker with a broader type list than I'd suggested, `.code`/`.payload` dropped, parametrized matrix plus the `TestSendCommandErrorContract` that actually exercises both WS raise sites. The e2e assertion flip in `test_history.py` is the right call and nicely documents the classification rationale.

The Boy Scout 5th site (`websocket_client.py:222`) was the right call to fold in — same root cause, no reason to leave a known regression behind.

LGTM.

PR #1017, opened 2026-04-19
perf(internal): inline-mode openai agent for BAT runner

sergeykad:
## What does this PR do?

Add an inline execution path for the BAT openai agent: one `fastmcp.Client` plus one `openai.AsyncOpenAI` per agent run, shared across all stories, instead of spawning a subprocess and warming the model per story. Cuts wall-clock for a full openai BAT run by roughly 15-20%. Claude and Gemini still take the subprocess path; the inline path is gated by an `INLINE_AGENTS` frozenset so adding another inline-capable agent later is a one-line edit.

Also in scope:
- Switch `openai_agent.py` from `openai.OpenAI` to `openai.AsyncOpenAI` so chat completions, `models.list`, and the warmup call no longer block the event loop. Addresses Gemini Code Assist high-priority review. Concrete type hints added: `mcp_client: fastmcp.Client`, `openai_client: openai.AsyncOpenAI`, `tool: mcp.types.Tool`.
- Fail-fast preflight for Docker and the `--base-url` endpoint so outages exit in ~1s with a clear `FATAL:` line rather than stalling inside container startup or model warmup.
- Surface the max-iteration cap (`hit_iteration_limit`) as a non-zero exit in both subprocess and inline modes. Previously a model stuck in a tool-call loop was reported as a clean pass.
- Per-agent setup failures (OpenAI auth, MCP server crash) write a stub failure row per filtered story instead of dropping them from the JSONL log and the process exit code.
- Inline failures always produce a summary with the traceback attached.
- `parse_mcp_env` and `build_stdio_mcp_config` extracted from `run_uat.py` so `run_story.py` can share them.

Validated against LM Studio + qwen3.5-9b: 13/14 openai BAT stories pass. The one failure is s07 hitting `hit_iteration_limit` on HA backend 500 errors during `ha_config_set_automation`, unrelated to the async conversion and correctly surfaced by the new error handling.

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [x] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

## Checklist

### DIFF
--- a/tests/uat/openai_agent.py
+++ b/tests/uat/openai_agent.py
@@ -18,6 +18,7 @@ import argparse
 import asyncio
 import json
 import sys
+import traceback
 from pathlib import Path
 
 import openai
@@ -368,7 +369,7 @@ async def _main_async(args: argparse.Namespace) -> None:
         log(f"ERROR: Model warmup failed (BadRequestError): {e}")
         sys.exit(1)
     except Exception as e:
-        log(f"ERROR ({type(e).__name__}): {e}")
+        log(f"ERROR ({type(e).__name__}): {e}\n{traceback.format_exc()}")
         sys.exit(1)
 
     log(f"MCP config: {args.mcp_config}")
@@ -379,7 +380,7 @@ async def _main_async(args: argparse.Namespace) -> None:
         finally:
             await client.close()
     except Exception as e:
-        log(f"ERROR ({type(e).__name__}): {e}")
+        log(f"ERROR ({type(e).__name__}): {e}\n{traceback.format_exc()}")
         sys.exit(1)
 
     json.dump(result, sys.stdout, indent=2)
--- a/tests/uat/openai_agent.py
+++ b/tests/uat/openai_agent.py
@@ -374,7 +374,10 @@ async def _main_async(args: argparse.Namespace) -> None:
     log(f"MCP config: {args.mcp_config}")
 
     try:
-        result = await run_agent(client, model, args)
+        try:
+            result = await run_agent(client, model, args)
+        finally:
+            await client.close()
     except Exception as e:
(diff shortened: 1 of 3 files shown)


kingpanther13 (changes requested):
One thing to tighten before this lands:

`preflight_check_base_url` (`tests/uat/run_uat.py:343-349`) only catches `RequestException`, so a backend that's up but returning 500/401 on `/models` passes the preflight and then fails later in warmup with a less clear error. That breaks the "outages exit in ~1s with a clear FATAL line" guarantee in the PR body for the up-but-broken case (which is the common LM Studio "wrong port / wrong endpoint" footgun). Adding `resp.raise_for_status()` (or at minimum a `status_code < 500` gate) and folding any `HTTPError` into the same FATAL branch would close that.

While you're in there, the outer `except Exception` in `_main_async` (`tests/uat/openai_agent.py:252-254`) logs `type(e).__name__` and the message but drops the traceback. On the inline path in `run_story.py` you already capture `traceback.format_exc()` into the summary — worth doing the same here so a hang in `run_agent` isn't a guessing game at the CLI.

Scope and everything else looks good — `AsyncExitStack` ordering is right, `hit_iteration_limit` surfacing is consistent across paths, and the setup-failure stub rows close a real observability gap.

kingpanther13 (approved):
Both addressed cleanly in `d27ab468` — `preflight_check_base_url` now folds HTTP errors into the same FATAL path via `raise_for_status()` (and `RequestException` is the right base to catch, since `HTTPError` inherits from it), and the two outer handlers in `_main_async` now log the traceback. LGTM.

PR #1032, opened 2026-04-21
Windows - Do not kill the terminal

gcormier:
When following the install guide for Windows, the installation will kill the terminal by executing `exit 0`. This is poor UX, since I can't see anything it did nor read the output.

If an exit 0 is truly needed, I can modify the PR to provide a delay of a few seconds, which will allow the user to abort the script before the window is killed.

### DIFF
--- a/scripts/install-windows.ps1
+++ b/scripts/install-windows.ps1
@@ -37,7 +37,7 @@ if ($uvInstalled) {
         Write-Host "  Please install manually:" -ForegroundColor Red
         Write-Host "  winget install astral-sh.uv" -ForegroundColor Cyan
         Write-Host "  OR download from: https://docs.astral.sh/uv/" -ForegroundColor Cyan
-        exit 1
+        throw
     }
 }
 Write-Host ""
--- a/scripts/install-windows.ps1
+++ b/scripts/install-windows.ps1
@@ -154,5 +154,3 @@ Write-Host "  Replace HOMEASSISTANT_TOKEN with your token"
 Write-Host "  (Generate token in HA: Profile > Security > Long-lived tokens)"
 Write-Host ""
 
-# Exit successfully
-exit 0


kingpanther13 (changes requested):
Good catch on the root cause — confirmed that `irm … | iex` runs the script body in the caller's session, so the trailing `exit 0` terminates the user's shell. The installer CI (`test-installer-scripts.yml`) invokes via `& ./script`, so removing `exit 0` doesn't regress CI (implicit exit 0 on clean completion).

Gemini's point is correct though: the `exit 1` on line 40 has the same bug — if `winget` fails, the caught exception path still kills the user's terminal before they can read the manual-install instructions you print just above it. Since you're already here, could you extend the fix to cover that path?

One nuance: Gemini suggested `return` or `throw`. `return` would regress CI — the `exit 1` sits inside a `catch` block, so `return` would exit the script cleanly with `$LASTEXITCODE = 0` and mask winget failures in `test-installer-scripts.yml`. Use `throw` (or bare `throw` to re-raise `\$_`) instead — that propagates the error, keeps the interactive terminal open in the `iex` case, and still fails the step in the `& ./script` CI case.

Suggested change at line 40:

\`\`\`diff
-        exit 1
+        throw
\`\`\`

Once that's in, happy to approve.

gcormier:
Changed line 40 per your comments! Thx

kingpanther13 (approved):
Thanks for the quick turnaround! Verified both behaviors locally:

- Bare `throw` from the catch under `pwsh -File` produces exit 1, so `test-installer-scripts.yml` still fails on a real winget error.
- Removing the trailing `exit 0` is fine — natural PowerShell script end is exit 0 regardless of any internal `$LASTEXITCODE`, so CI's success path stays green.

Under `irm | iex` the `throw` terminates the piped scriptblock as a terminating error without killing the host session, which is exactly what was reported.

LGTM.

PR #1040, opened 2026-04-21
Installer (Windows) - remove unused variable

gcormier:
The variable `$HaMcpConfig` was defined but never used.


kingpanther13 (approved):
Thanks for the cleanup! Verified `$HaMcpConfig` has no references anywhere in the repo, and `$JsonConfig` (the variable actually written to the config file) is untouched. LGTM.

PR #1048, opened 2026-04-24
feat: add ha_manage_energy_prefs tool for Energy Dashboard CRUD

### LINKED ISSUE #1033: [FEATURE] expose Energy Dashboard preferences as an MCP tool

A thin wrapper around the `energy/save_prefs` websocket API (get + set):

```
ha_energy_get_prefs()
ha_energy_set_prefs(config={...})
```

`config` would mirror the existing websocket schema: `energy_sources` (grid, solar, battery, gas, water) and `device_consumption` (individual devices). Full-replace semantics match the underlying API and keep the tool surface small.

### Why do you need this?

The built-in Energy Dashboard is configured via the `energy/save_prefs` websocket API. Its configuration (grid consumption source, individual devices, gas/water, cost tariffs) is stored in `.storage/energy` and isn't reachable through any service, REST endpoint, or helper flow.

As a result, agents using this MCP can create helpers, build full custom dashboards, call services, and set up automations — but cannot add a single sensor to the Energy Dashboard itself. Users have to click through the Settings → Dashboards → Energy UI manually for every device.

Concrete use cases:

- Agent that bootstraps a new HA install adds the user's smart-plug energy sensors to the Energy Dashboard automatically after creating them.
- Agent that discovers a newly paired smart plug offers to "also add to Energy Dashboard" and can actually do it.
- Audit/migration scripts that rebuild the Energy Dashboard from a known spec.

### Additional context

- The API is stable and already used by the frontend.
- Read-only access (`ha_energy_get_prefs`) is also useful on its own — lets agents reason about what's currently tracked.
- Full-replace write matches the existing websocket behavior, so no new merge/patch semantics needed.

Patch76:
Closes #1033.

## What

Adds `ha_manage_energy_prefs`, a single tool with `mode="get"` / `mode="set"` wrapping Home Assistant's `energy/get_prefs` and `energy/save_prefs` WebSocket commands. The Energy Dashboard configuration is otherwise unreachable via REST, services, or helper flows — agents currently cannot add a newly-paired smart plug to the Energy Dashboard programmatically.

## Design choices

**Single `manage` tool, not a `get`+`set` pair.** Read-modify-write is the only safe workflow given the API's destructive per-key full-replace semantics (sending `{"device_consumption": [<one>]}` silently deletes every other device). Mirrors the `ha_manage_addon` consolidation (#978).

**Required `config_hash` on write, no force-write escape hatch.** Primary mitigation for the full-replace footgun; also covers concurrent-editing. Follows `compute_config_hash` usage in `tools_config_dashboards.py`.

**Local shape check + post-save server-side validate.** HA Core's `energy/validate` does not accept an unsubmitted payload — it validates the currently-persisted config — so pre-write validation of the proposed payload is not possible. Post-save residual issues (missing stats, wrong unit classes) surface as `post_save_validation_errors`.

**`dry_run=True`: shape check + current-state validate.** Returns `shape_errors` (for the proposed config) and `current_state_validation_errors` (labelled as reflecting the *persisted* state, since validate can't do the proposal).

**Fresh-install handling.** On an HA instance that has never had the Energy Dashboard configured, `energy/get_prefs` returns `ERR_NOT_FOUND "No prefs"`. The tool maps that to the documented empty default so the workflow is uniform across fresh and configured instances. Caught by the E2E test against a fresh testcontainer.

## Tests

- **Unit (30 tests):** every state-machine branch — success and error paths, `config_hash` mismatch, shape errors before any WS call, post-save validate failures treated as non-fatal, fresh-install `No prefs` handling in both `get` and `set`.
- **E2E (2 tests):** read-only `mode="get"` against the freshly-initialised testcontainer — tool registration, WebSocket round-trip, response shape, hash determinism. Write paths are unit-only to avoid coupling CI to container-persisted energy state.

ruff + mypy clean.

## Deferred as follow-ups

- **Per-key hashes:** partial updates with one hash per top-level key. Architecturally cleaner but introduces a new pattern the codebase doesn't use elsewhere — separate design discussion.
- **Convenience helpers** (e.g. `add_device_to_energy_dashboard` doing read-modify-write internally): violates the tool-consolidation guideline. Primitive-tool approach is deliberate.

---

Non-maintainer contribution. Merge decision stays with the maintainers.

Coverage gap: `gas` energy sources and `device_consumption_water` entries are shape-checked but not exercised against live HA — neither is configured on my instance.

### DIFF
--- a/src/ha_mcp/tools/tools_energy.py
+++ b/src/ha_mcp/tools/tools_energy.py
@@ -1,0 +1,539 @@
+"""
+Energy Dashboard preference management tools for Home Assistant.
+
+This module provides a single tool to read and write Home Assistant's Energy
+Dashboard configuration through the ``energy/get_prefs`` / ``energy/save_prefs``
+WebSocket commands. The underlying API has destructive full-replace semantics
+per top-level key (``energy_sources``, ``device_consumption``,
+``device_consumption_water``) — sending a key with a partial list silently
+deletes everything else the user had configured. Optimistic locking via
+``config_hash`` prevents concurrent-modification data loss; a local shape
+check catches the most common agent-side errors; and a server-side
+``energy/validate`` call after every write surfaces residual issues
+(missing stats, wrong unit classes, etc.) in the response.
+
+Note: ``energy/validate`` in Home Assistant Core takes no payload — it
+validates the currently-persisted config. Pre-write validation of an
+unsubmitted payload is therefore not possible; this tool validates the
+post-save state instead.
+
+Note: On a fresh Home Assistant instance that has never had the Energy
+Dashboard configured, ``energy/get_prefs`` returns
+``ERR_NOT_FOUND "No prefs"`` rather than an empty default. The tool
+transparently maps that case to the documented default preferences
+structure (all three top-level keys present, empty lists) so agents
+get uniform behavior on fresh and configured instances alike.
+"""
+
+import logging
+from typing import Annotated, Any, Literal
+
+from fastmcp.exceptions import ToolError
+from fastmcp.tools import tool
+from pydantic import Field
+
+from ..errors import ErrorCode, create_error_response
+from ..utils.config_hash import compute_config_hash
+from .helpers import (
+    exception_to_structured_error,
+    log_tool_usage,
+    raise_tool_error,
+    register_tool_methods,
+)
+
+logger = logging.getLogger(__name__)
+
+
+# Top-level keys in the energy prefs payload. Each is an independent
+# full-replace slot in ``energy/save_prefs``.
+_PREFS_TOP_LEVEL_KEYS = (
+    "energy_sources",
+    "device_consumption",
+    "device_consumption_water",
+)
+
+
+def _default_prefs() -> dict[str, Any]:
+    """Return the default empty prefs structure used by HA Core.
+
+    Mirrors ``EnergyManager.default_preferences()`` in
+    ``homeassistant/components/energy/data.py``. A Home Assistant instance
+    that has never had the Energy Dashboard configured returns
+    ``ERR_NOT_FOUND "No prefs"`` from ``energy/get_prefs``; this helper
+    provides the canonical empty structure so the tool can transparently
+    treat the two cases (never-configured vs. configured-but-empty) the
+    same way.
+    """
+    return {
+        "energy_sources": [],
+        "device_consumption": [],
+        "device_consumption_water": [],
+    }
+
+
+def _is_no_prefs_error(error_msg: str) -> bool:
+    """True if an error string from send_websocket_message indicates
+    ``ERR_NOT_FOUND "No prefs"`` from HA Core's energy/get_prefs handler.
+
+    HA Core wraps the error as ``f"Command failed: {message}"``; the
+    underlying sentinel we key on is the literal ``"No prefs"`` message
+    emitted by ``ws_get_prefs`` when ``manager.data is None``.
+    """
+    return "No prefs" in error_msg
+
+
+def _flatten_validation_errors(raw: Any) -> list[dict[str, str]]:
+    """Convert the raw ``energy/validate`` response into a flat error list.
+
+    The raw response mirrors the prefs structure: a dict with the three
+    top-level keys, each mapping to a list of per-entry error lists (empty
+    inner list = that entry is valid). This function walks that structure and
+    returns a flat list of ``{"path", "message"}`` dicts, suitable for agent
+    consumption.
+
+    A successful validation returns an empty list.
+    """
+    if not isinstance(raw, dict):
+        return []
+
+    errors: list[dict[str, str]] = []
+    for key in _PREFS_TOP_LEVEL_KEYS:
+        entries = raw.get(key, [])
+        if not isinstance(entries, list):
+            continue
+        for idx, entry_errors in enumerate(entries):
+            if not entry_errors:
+                continue
+            if isinstance(entry_errors, list):
+                errors.extend(
+                    {"path": f"{key}[{idx}]", "message": str(msg)}
+                    for msg in entry_errors
+                )
+            elif isinstance(entry_errors, dict):
+                for field, msgs in entry_errors.items():
+                    msg_list = msgs if isinstance(msgs, list) else [msgs]
+                    errors.extend(
+                        {"path": f"{key}[{idx}].{field}", "message": str(msg)}
+                        for msg in msg_list
+                    )
+    return errors
+
+
+def _shape_check(config: dict[str, Any]) -> list[dict[str, str]]:
+    """Cheap local shape check before sending to the server.
+
+    Validates that top-level keys have the expected list-of-dicts shape and
+    that required identifying fields are present. Does NOT validate semantic
+    correctness (stat IDs existing, units matching, etc.) — that's surfaced
+    by the post-save server-side ``energy/validate`` call.
+    """
+    errors: list[dict[str, str]] = []
+
+    if not isinstance(config, dict):
+        return [{"path": "config", "message": "must be a dict"}]
+
+    for key in _PREFS_TOP_LEVEL_KEYS:
+        if key not in config:
+            continue
+        value = config[key]
+        if not isinstance(value, list):
+            errors.append({"path": key, "message": "must be a list"})
+            continue
+        for idx, entry in enumerate(value):
+            if not isinstance(entry, dict):
+                errors.append({
+                    "path": f"{key}[{idx}]",
+                    "message": "entry must be a dict",
+                })
+                continue
+            if key == "energy_sources" and "type" not in entry:
+                errors.append({
+                    "path": f"{key}[{idx}]",
+                    "message": "energy_sources entries require 'type' (grid|solar|battery|gas)",
+                })
+            if key == "device_consumption" and "stat_consumption" not in entry:
+                errors.append({
+                    "path": f"{key}[{idx}]",
+                    "message": "device_consumption entries require 'stat_consumption'",
+                })
+            if key == "device_consumption_water" and "stat_consumption" not in entry:
+                errors.append({
+                    "path": f"{key}[{idx}]",
+                    "message": "device_consumption_water entries require 'stat_consumption'",
+                })
+
+    return errors
+
+
+class EnergyTools:
+    """Energy Dashboard preference management tools for Home Assistant."""
+
+    def __init__(self, client: Any) -> None:
+        self._client = client
+
+    @tool(
+        name="ha_manage_energy_prefs",
+        tags={"Energy"},
+        annotations={
+            "destructiveHint": True,
+            "idempotentHint": False,
+            "title": "Manage Energy Dashboard Preferences",
+        },
+    )
+    @log_tool_usage
+    async def ha_manage_energy_prefs(
+        self,
+        mode: Annotated[
+            Literal["get", "set"],
+            Field(description="Operation mode: 'get' reads the current prefs; 'set' writes a new prefs payload."),
+        ],
+        config: Annotated[
+            dict[str, Any] | None,
+            Field(
+                description=(
+                    "Full prefs payload for mode='set'. Must contain the "
+                    "top-level keys you intend to replace: 'energy_sources', "
+                    "'device_consumption', 'device_consumption_water'. Any "
+                    "top-level key present in this payload REPLACES the "
+                    "existing list entirely; any omitted key is preserved. "
+                    "Call with mode='get' first, mutate the returned config, "
+                    "then pass the whole object back."
+                ),
+                default=None,
+            ),
+        ] = None,
+        config_hash: Annotated[
+            str | None,
+            Field(
+                description=(
+                    "Hash returned by the previous mode='get' call. REQUIRED "
+                    "for mode='set' unless dry_run=True. Rejected if the "
+                    "server-side config has changed since that read — re-read "
+                    "and retry."
+                ),
+                default=None,
+            ),
+        ] = None,
+        dry_run: Annotated[
+            bool,
+            Field(
+                description=(
+                    "For mode='set' only. If True, runs a local shape check "
+                    "on the proposed config AND calls the server's "
+                    "energy/validate against the CURRENT persisted state "
+                    "(Home Assistant's validate endpoint cannot validate "
+                    "an unsubmitted payload). Returns both error lists "
+                    "without writing. Default False."
+                ),
+                default=False,
+            ),
+        ] = False,
+    ) -> dict[str, Any]:
+        """
+        Read or write the Home Assistant Energy Dashboard preferences.
+
+        The Energy Dashboard configuration (grid/solar/battery/gas sources,
+        individual device consumption sensors, cost tariffs, water) is stored
+        in ``.storage/energy`` and not otherwise reachable via REST, services,
+        or helper flows — this tool is the only way for agents to inspect or
+        modify it.
+
+        WHEN TO USE:
+        - To inspect or modify the Energy Dashboard config programmatically.
+
+        WHEN NOT TO USE:
+        - To create the underlying statistics themselves — they must already
+          exist as HA entities before being referenced here; create them via
+          the relevant integration's config flow first.
+
+        CAVEATS:
+        - ``energy/save_prefs`` has per-key FULL-REPLACE semantics. Passing
+          ``{"device_consumption": [<one entry>]}`` deletes every other device
+          the user had configured — silently, with no error. Always call
+          mode='get' first, mutate the returned config, pass the whole object
+          back, and include the returned ``config_hash`` so the tool can
+          reject concurrent modifications.
+        - A local shape check runs before every write; malformed payloads
+          are rejected with a ``shape_errors`` list.
+        - After a successful write, the tool calls ``energy/validate`` and
+          returns any residual issues as ``post_save_validation_errors`` in
+          the response. These reflect semantic problems (missing stats, unit
+          mismatches) that shape checks can't catch; the save persists
+          regardless — correct the config and write again if needed.
+        - The underlying save endpoint is admin-only. Non-admin tokens will
+          receive an authorization error from Home Assistant.
+        """
+        if mode == "get":
+            return await self._get_prefs()
+
+        # mode == "set"
+        if config is None:
+            raise_tool_error(create_error_response(
+                ErrorCode.VALIDATION_MISSING_PARAMETER,
+                "'config' is required when mode='set'",
+                context={"mode": mode},
+                suggestions=[
+                    "Call ha_manage_energy_prefs(mode='get') first, mutate the returned config, pass it back",
+                ],
+            ))
+
+        if dry_run:
+            return await self._dry_run(config)
+
+        if config_hash is None:
+            raise_tool_error(create_error_response(
+                ErrorCode.VALIDATION_MISSING_PARAMETER,
+                "'config_hash' is required when mode='set' and dry_run=False",
+                context={"mode": mode},
+                suggestions=[
+                    "Call ha_manage_energy_prefs(mode='get') to obtain a fresh config_hash",
+                    "Or call again with dry_run=True to validate without a hash",
+                ],
+            ))
+
+        return await self._set_prefs(config, config_hash)
+
+    # ------------------------------------------------------------------
+    # Internal handlers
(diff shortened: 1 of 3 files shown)


kingpanther13:
Normally I would say not to combine read and write tools, manage_addon is an exception just due to its nature as a generic addon tool. However I think that this is the right move in this case since this is more of a niche tool anyways.... Doing a combined manage style tool will help keep the tool counts down for something that likely won't be used frequently.

src/ha_mcp/tools/tools_energy.py
```diff
@@ -0,0 +1,539 @@
...
+                errors.append({
+                    "path": f"{key}[{idx}]",
+                    "message": "entry must be a dict",
+                })
+                continue
+            if key == "energy_sources" and "type" not in entry:
+                errors.append({
+                    "path": f"{key}[{idx}]",
+                    "message": "energy_sources entries require 'type' (grid|solar|battery|gas)",
+                })
```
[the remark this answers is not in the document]
Patch76: Split response for the three sub-points.

**Path-scoped `create_validation_error`:** declined. The helper at
`errors.py:318` takes `(message, parameter=None, details=None,
invalid_json=False, context=None)` — no `path` kwarg. Passing one would raise
`TypeError`. Different semantics too: `create_validation_error` builds a
tool-level error for `raise_tool_error`, not a shape-error entry the agent can
iterate over. The inline dict literals keep the current shape stable for
consumers.

**Enum-style check for `type`:** adopted. Replaced the membership test with
an explicit `{"grid", "solar", "battery", "gas"}` set; missing vs. invalid
now reported on separate paths (`[idx]` vs. `[idx].type`).

**`stat_energy_from` as a required field:** partial accept. Verified against
HA core (`homeassistant/components/energy/data.py`):

- `GridSourceType.stat_energy_from: str | None`, schema
  `vol.Optional("stat_energy_from", default=None)` — grid entries do not
  require it (flow-based sources).
- `SolarSourceType`, `BatterySourceType`, `GasSourceType` all declare it
  `str` and schema `vol.Required(...)`.

Required-check now applied only to solar/battery/gas.

Implemented in a8ba82f, covered by new unit test
`test_shape_errors_energy_sources_enum_and_conditional`.v

src/ha_mcp/tools/tools_energy.py
```diff
@@ -0,0 +1,539 @@
...
+                    "energy/validate against the CURRENT persisted state "
+                    "(Home Assistant's validate endpoint cannot validate "
+                    "an unsubmitted payload). Returns both error lists "
+                    "without writing. Default False."
+                ),
+                default=False,
+            ),
+        ] = False,
+    ) -> dict[str, Any]:
+        """
```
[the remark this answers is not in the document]
Patch76: Adopted. Docstring opener changed to "Manage the Home Assistant Energy
Dashboard preferences." — lines up with the tool name and the `ha_manage_*`
verb convention.

Implemented in a8ba82f.

src/ha_mcp/tools/tools_energy.py
```diff
@@ -0,0 +1,539 @@
...
+                    # unreachable; appeases type checkers
+                    current_prefs = {}
+            else:
+                current_prefs = current_result.get("result", {}) or {}
+
+            current_hash = compute_config_hash(current_prefs)
+
+            if current_hash != config_hash:
+                raise_tool_error(create_error_response(
+                    ErrorCode.SERVICE_CALL_FAILED,
```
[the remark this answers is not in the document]
Patch76: Adopted. `ErrorCode.RESOURCE_LOCKED` is defined at `errors.py:79` in the
resource-errors group and currently unused elsewhere — hash-mismatch on a
read-modify-write cycle is a closer fit than `SERVICE_CALL_FAILED`.
Unit test `test_hash_mismatch_rejects_and_does_not_save` updated to assert
on the new code.

Implemented in a8ba82f.

src/ha_mcp/tools/tools_energy.py
```diff
@@ -0,0 +1,595 @@
...
+                        create_error_response(
+                            ErrorCode.SERVICE_CALL_FAILED,
+                            f"Failed to re-read prefs for hash check: {current_result.get('error', 'Unknown error')}",
+                            context={"mode": "set"},
+                        )
+                    )
+                    # unreachable; appeases type checkers
+                    current_prefs = {}
+            else:
+                current_prefs = current_result.get("result", {}) or {}
```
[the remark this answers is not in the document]
Patch76: Adopted in c1fb187:

- Success-first branch ordering for readability.
- Error string captured once: `error = current_result.get("error") or
  "Unknown error"` — the `or` fallback also handles a None value, not just
  a missing key.

The `# unreachable; appeases type checkers` pattern with `current_prefs = {}`
was kept rather than replaced with `_default_prefs()`. Grepping the codebase
shows seven pre-existing `# unreachable` comments (tools_hacs.py,
device_control.py, others) using the same idiom, so this preserves
established convention. If the project prefers `_default_prefs()` here, a
separate sweep across those call sites would be the cleaner way to switch.

Patch76:
## Implementation Summary

**Choices Made:**

- **Single `manage` tool with `mode=get|set`.** Read-modify-write is the only
  safe workflow given the destructive per-key full-replace semantics of
  `energy/save_prefs`. Rationale already in the PR body; cited here because
  maintainer feedback in this PR validated the decision explicitly:
  niche-tool + tool-count economics (#1048 comment 4312766461).

- **`create_validation_error(path=...)` not used for shape errors.** Gemini
  suggested routing per-entry errors through the structured error helper,
  but `errors.py:318` has no `path` kwarg — passing one would raise
  `TypeError`. The helper also builds a tool-level error for
  `raise_tool_error`, not an iterable shape-error entry. Inline dict literals
  with `path` + `message` keep the shape agents already consume.

- **`stat_energy_from` required-check narrowed to solar/battery/gas.**
  Gemini's broader claim (required for all `energy_sources`) contradicted
  HA core's `data.py` schema: `GridSourceType` declares the field
  `Optional` (flow-based grid sources). Required-check now matches core's
  three `vol.Required(...)` declarations.

- **`ErrorCode.RESOURCE_LOCKED` for hash-mismatch conflict.** The enum value
  exists in `errors.py:79` and was previously unused; optimistic-locking
  semantics fit it better than the generic `SERVICE_CALL_FAILED`.

- **`# unreachable; appeases type checkers` idiom preserved in refactor.**
  Gemini's round-2 suggestion would have replaced the dead `current_prefs =
  {}` with `_default_prefs()`. Seven pre-existing call sites (tools_hacs.py,
  device_control.py, others) use the `# unreachable` idiom; staying with
  convention here rather than introducing a mixed pattern. If the project
  later prefers `_default_prefs()` in this position, the better path is a
  sweep across all call sites.

- **Coverage gap acknowledged rather than hidden.** `gas` sources and
  `device_consumption_water` are shape-checked but not exercised against a
  live HA instance because neither is configured on the test system. Noted
  in PR body for reviewers.

**Problems Encountered:**

- **Two Gemini review rounds.** Round 1 returned three medium-priority
  findings (validation structure, docstring verb, error-code specificity).
  Round 2 on the fix commit returned one additional medium on the
  `current_prefs` retrieval branch. All five points now either adopted or
  declined with rationale posted in-thread.

- **Gemini helper-signature mismatch.** Round 1's `create_validation_error`
  suggestion would not compile — the helper's signature doesn't accept the
  suggested kwargs. Flagged as decline with cross-reference to
  `errors.py:318`.

- **No unrelated test failures encountered** — unit + E2E stayed green
  across feature, fix, and refactor commits.

- Already filed as separate feature issues:
  - #1049 — per-key hashes for partial updates
  - #1050 — convenience helpers (add_device / remove_device) on top of this
    primitive

- Not filed but noticed during this work:
  - The `# unreachable; appeases type checkers` idiom is used seven times
    across the codebase. A follow-up could consolidate these into a typed
    helper (e.g. `_unreachable() -> NoReturn`) if the maintainers see value;
    out of scope for this PR.

src/ha_mcp/tools/tools_energy.py
```diff
@@ -0,0 +1,595 @@
...
+
+def _is_no_prefs_error(error_msg: str) -> bool:
+    """True if an error string from send_websocket_message indicates
+    ``ERR_NOT_FOUND "No prefs"`` from HA Core's energy/get_prefs handler.
+
+    HA Core wraps the error as ``f"Command failed: {message}"``; the
+    underlying sentinel we key on is the literal ``"No prefs"`` message
+    emitted by ``ws_get_prefs`` when ``manager.data is None``.
+    """
+    return "No prefs" in error_msg
```
kingpanther13: Minor/optional: substring `"No prefs" in error_msg` is a little brittle — any future wrapper that happens to echo the config blob or a custom message containing that phrase would match. HA Core's wrapper is stable as `"Command failed: No prefs"`, so an `endswith("No prefs")` or explicit prefix check would tighten this without giving up robustness to the exact wrapping. Low urgency, flagging while we're here.

src/ha_mcp/tools/tools_energy.py
```diff
@@ -0,0 +1,595 @@
...
+                    }
+                raise_tool_error(
+                    create_error_response(
+                        ErrorCode.SERVICE_CALL_FAILED,
+                        f"Failed to get energy prefs: {result.get('error', 'Unknown error')}",
+                        context={"mode": "get"},
+                    )
+                )
+
+            prefs = result.get("result", {})
```
kingpanther13: Nit-but-symmetric: this path falls to `{}` on a degenerate response (success but no `result` key), while the `No prefs` branch at line 352–358 falls to `_default_prefs()`. `_set_prefs` has the same split (line 485 `or {}` vs line 489 `_default_prefs()`). If the two reads hit different fallbacks across a single get→set cycle, the hash check falsely reports a conflict. Using `_default_prefs()` on both empty-result branches keeps the hash symmetric — cheap fix.

src/ha_mcp/tools/tools_energy.py
```diff
@@ -0,0 +1,595 @@
...
+                {
+                    "type": "energy/validate",
+                }
+            )
+            if validate_result.get("success"):
+                current_state_errors = _flatten_validation_errors(
+                    validate_result.get("result", {})
+                )
+            else:
+                current_state_errors = []
```
kingpanther13, line 418: **Silent drop of a real validate-endpoint failure.** If `energy/validate` returns `success=false` (older HA without that command, auth glitch, anything), `current_state_errors` is set to `[]` with no log and no signal in the response. The caller sees an empty list indistinguishable from "validated, everything clean."

Suggest logging and surfacing:

```python
else:
    logger.warning(
        "energy/validate (current state) failed: %s",
        validate_result.get("error"),
    )
    current_state_errors = []
```

Plus something like `current_state_validation_unavailable: True` (or a `current_state_validation_error` field) in the returned dict so agents can tell "didn't run" from "ran clean."

src/ha_mcp/tools/tools_energy.py
```diff
@@ -0,0 +1,595 @@
...
+            # 2. Fresh read for hash comparison. Map "No prefs" (never
+            # configured) to empty default so the hash-check works on
+            # fresh installations too.
+            current_result = await self._client.send_websocket_message(
+                {
+                    "type": "energy/get_prefs",
+                }
+            )
+            if current_result.get("success"):
+                current_prefs: dict[str, Any] = current_result.get("result") or {}
```
kingpanther13: Symmetric with the nit above at line 372 — the `or {}` here and `_default_prefs()` at line 489 can disagree. Using `_default_prefs()` on both empty-result branches keeps the hash computation consistent across the degenerate-result path and the `No prefs` path.

src/ha_mcp/tools/tools_energy.py
```diff
@@ -0,0 +1,595 @@
...
+            try:
+                validate_result = await self._client.send_websocket_message(
+                    {
+                        "type": "energy/validate",
+                    }
+                )
+                if validate_result.get("success"):
+                    post_save_errors = _flatten_validation_errors(
+                        validate_result.get("result", {})
+                    )
```
kingpanther13, line 564: **Same silent-drop as the dry_run case — post-save validate `success=false` is invisible to the caller.** The `except Exception` on line 549–552 logs a warning, but the `if validate_result.get("success")` branch here doesn't — so a validate call that returns structured failure produces the same empty `post_save_errors` list as a validate that actually ran clean. The `response` dict at line 563 then looks identical in both cases.

Suggest:

```python
if validate_result.get("success"):
    post_save_errors = _flatten_validation_errors(
        validate_result.get("result", {})
    )
else:
    logger.warning(
        "energy/validate (post-save) failed: %s",
        validate_result.get("error"),
    )
```

And surface it in the response (`post_save_validation_skipped: True` plus the error message) so agents don't assume the newly-persisted state is clean when we actually just don't know.

tests/src/e2e/tools/test_tools_energy.py
```diff
@@ -0,0 +1,91 @@
...
+
+Scope: read-only mode="get" against the freshly-initialised test container.
+The container starts with no ``.storage/energy`` file, so HA returns the
+empty default — confirming tool registration, WebSocket connectivity, and
+response shape without mutating state.
+
+Write paths (mode="set", including dry_run) are covered by the unit tests
+under tests/src/unit/test_tools_energy.py. They are deliberately not
+exercised here to avoid coupling CI runs to container-persisted energy
+prefs state.
```
kingpanther13: Push-back on the "write paths are deliberately not exercised to avoid coupling CI to container-persisted energy state" rationale: the testcontainer is ephemeral, so there's nothing persistent for CI to couple to. A single `mode='set'` round-trip against the container (save a small prefs payload, re-read, assert the new hash matches) would cover what mocks genuinely cannot — e.g. if HA ever renames `energy/save_prefs` or changes the expected payload shape, unit tests wouldn't notice. Not a merge blocker, but I'd take this over the current unit-only write coverage if the cost is one extra test.

kingpanther13 (changes requested):
Reviewed against `AGENTS.md` + `.gemini/styleguide.md`. Overall this is a well-scoped PR — the `manage` pattern is the right call, the state machine is sound, and the unit suite is genuinely exhaustive across the documented branches. Thanks for the thorough writeup in the PR body; made the review a lot easier.

A couple of changes before merge, plus a few lower-priority suggestions.

**Important (before merge):**
- `energy/validate` `success=false` is silently dropped on two paths (`_dry_run` and post-save in `_set_prefs`). Caller can't distinguish "validated clean" from "validation didn't run." See inline comments.

**Suggestions (not strictly blocking, but worth folding in while we're here):**
- Hash-shape asymmetry between the `_get_prefs` empty-result fallback and the `_set_prefs` `No prefs` branch.
- Tighten the `_is_no_prefs_error` substring match.
- One `mode='set'` round-trip E2E — the testcontainer is ephemeral, so "don't couple CI to persisted state" doesn't really bite, and mocks can't catch a WebSocket command-name change in `energy/save_prefs`.

On the two follow-up issues — I'll land more detail on #1050 separately, but tl;dr: keep #1049 and #1050 as separate work, don't fold them in here.

src/ha_mcp/tools/tools_energy.py
Patch76: Adopted. `endswith("No prefs")` replaces the substring match — HA Core wraps the sentinel as `"Command failed: No prefs"`, so the suffix form is exact without losing robustness to wrapper-prefix changes.

Implemented in ebf59ef.

src/ha_mcp/tools/tools_energy.py
Patch76: Adopted. Empty-result fallback switched to `_default_prefs()`, symmetric with the `No prefs` branch at L352–358. Same fix applied to `_set_prefs` at L485.

Implemented in ebf59ef.

src/ha_mcp/tools/tools_energy.py
Patch76: Adopted — with a deviation on the response-field shape, flagged in the review-response summary above.

Short version: `logger.warning` added, and the caller can distinguish "didn't run" from "ran clean" via `partial: True` + `warning: "..."` — matching the existing degraded-success convention (`AGENTS.md` §Return Values, 8+ precedent sites in `src/`). No new `*_unavailable` boolean flag introduced.

If an explicit boolean field is preferred despite the convention, it's a one-line change on both this call site and L548.

Implemented in ebf59ef.

src/ha_mcp/tools/tools_energy.py
Patch76: Adopted. Same symmetry fix as L372 — `_default_prefs()` on both empty-result branches.

Implemented in ebf59ef.

src/ha_mcp/tools/tools_energy.py
Patch76: Adopted — same response-field-shape deviation as L413, see the review-response summary above.

`logger.warning` added in the `if validate_result.get("success")` branch, and the post-save silent-drop is surfaced via `partial: True` + `warning: "..."` in the response. Paired with L413 so both validate call sites behave identically.

Implemented in ebf59ef.

tests/src/e2e/tools/test_tools_energy.py
Patch76: Adopted. One `mode='set'` roundtrip added: save a minimal `device_consumption: []` payload, re-read, assert the returned hash matches. Catches the `energy/save_prefs` rename case the unit tests can't. Docstring rationale for omitting write paths removed — the ephemeral-container argument doesn't hold, as you noted.

Implemented in ebf59ef.

Patch76:
## Review response summary

Thanks for the pass. All six points adopted. Two deviations from the exact wording of the suggestions, flagged here so the reasoning is in one place rather than duplicated across threads.

### Adopted as proposed

| Thread | Change |
|---|---|
| L82 — `_is_no_prefs_error` | `endswith("No prefs")` replaces the substring match. |
| L372 — `_get_prefs` empty-result fallback | `_default_prefs()` replaces `or {}`, symmetric with the `No prefs` branch at L352–358. |
| L485 — `_set_prefs` empty-result fallback | Same symmetry fix as L372. |
| test_tools_energy.py | One `mode='set'` roundtrip added; docstring rationale for omitting write paths removed (ephemeral-container argument doesn't hold). |

### Adopted with deviation — response field shape (L413, L548)

Both silent-drop cases are surfaced in the response, but via the repo's existing degraded-success convention rather than a dedicated boolean flag.

Grep on `src/`:
- `"warning": "<text>"` in success dicts — 8+ sites (`tools_service.py`, `tools_entities.py`, `tools_system.py`, `tools_integrations.py`, `backup.py`)
- `"partial": True` as degraded marker — 3 sites (`tools_service.py`, `tools_registry.py`, `tools_search.py`)
- `*_unavailable: True` / `*_skipped: True` boolean flags — 0 sites

`AGENTS.md` §Return Values documents this as the canonical degraded-success shape:

```python
{"success": True, "partial": True, "warning": "..."}
```

Implementation for both cases:

```python
else:
    logger.warning(f"energy/validate (current state) failed: {validate_error}")
    current_state_errors = []
    validate_warning = (
        f"energy/validate failed: {validate_error} — "
        "current-state validation skipped"
    )
# ...
if validate_warning is not None:
    response["partial"] = True
    response["warning"] = validate_warning
```

The semantic goal from the review is preserved — the caller can distinguish "didn't run" from "ran clean" via `partial: True` plus the `warning` string. If an explicit boolean flag is preferred despite the convention, I'll switch both call sites in one follow-up commit.

Unit tests added for both new else-branches (`test_validate_failure_surfaced_as_warning`, `test_post_save_validate_failure_surfaced_as_warning`), plus the existing `test_post_save_validation_failure_non_fatal` extended to cover the new response shape from the exception path.

### Cross-references

- #1049 (per-top-level-key hashes) benefits from the L372/L485 symmetry fix — `_default_prefs()` as stable baseline on both read and write paths.
- #1050 (convenience helpers) — noted your comment on additional `mode=` values rather than new top-level tools; that's consistent with `AGENTS.md` §Tool Consolidation and doesn't affect this PR's scope.

kingpanther13 (approved):
All six points addressed — including the principled deviation on response field shape, which correctly uses the repo's existing `partial: True` + `warning` convention from `AGENTS.md` §Return Values rather than introducing a new boolean flag. New unit tests cover the dry_run and post-save else-branches with `caplog` assertions, and the `mode='set'` E2E roundtrip catches the `energy/save_prefs` rename case. CI green across all platforms. Nice work.

PR #1053, opened 2026-04-24
feat: expose allowlist_external_dirs in ha_get_overview full system_info

Patch76:
## What does this PR do?

Adds `allowlist_external_dirs` to the `system_info` block returned by
`ha_get_overview(detail_level="full")`.

Home Assistant permits file operations only under an explicit list of
filesystem paths (e.g. for `media_source`, `notify.file`, image uploads).
Agents reasoning about file I/O need visibility into this list to know
whether a proposed path will be reachable — otherwise a service call
fails after the decision has already been made.

The field name matches Home Assistant core (`allowlist_*`), not the
legacy `whitelist_*` alias.

Follow-up to discussion #997.

## Type of change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

## Checklist

### DIFF
--- a/src/ha_mcp/tools/tools_search.py
+++ b/src/ha_mcp/tools/tools_search.py
@@ -666,6 +666,9 @@ def register_search_tools(mcp: Any, client: Any, **kwargs: Any) -> None:
                         "safe_mode": config.get("safe_mode", False),
                         "internal_url": config.get("internal_url"),
                         "external_url": config.get("external_url"),
+                        "allowlist_external_dirs": config.get(
+                            "allowlist_external_dirs", []
+                        ),
                     }
                 )
             result["system_info"] = system_info
--- a/src/ha_mcp/tools/tools_search.py
+++ b/src/ha_mcp/tools/tools_search.py
@@ -666,8 +666,10 @@ def register_search_tools(mcp: Any, client: Any, **kwargs: Any) -> None:
                         "safe_mode": config.get("safe_mode", False),
                         "internal_url": config.get("internal_url"),
                         "external_url": config.get("external_url"),
+                        # No default: distinguish HA-not-exposing-the-key (None)
+                        # from empty-allowlist ([]) — security-relevant for agents.
                         "allowlist_external_dirs": config.get(
-                            "allowlist_external_dirs", []
+                            "allowlist_external_dirs"
                         ),
                     }
                 )
--- a/tests/src/e2e/workflows/system/test_system_tools.py
+++ b/tests/src/e2e/workflows/system/test_system_tools.py
@@ -299,10 +299,16 @@ class TestSystemTools:
             "location_name",
             "time_zone",
             "components_loaded",
+            "allowlist_external_dirs",
         ]
 
         for field in expected_fields:
             assert field in system_info, f"Missing expected field in system_info: {field}"
+
+        # allowlist_external_dirs is always a list (possibly empty)
+        assert isinstance(system_info["allowlist_external_dirs"], list), (
+            "allowlist_external_dirs should be a list"
+        )
 
         # Log key information
         logger.info(f"Home Assistant version: {system_info.get('version')}")
--- a/tests/src/e2e/workflows/system/test_system_tools.py
+++ b/tests/src/e2e/workflows/system/test_system_tools.py
@@ -310,7 +310,7 @@ class TestSystemTools:
             "allowlist_external_dirs should be a list"
         )
 
-        # Negative guard: field must not leak at lower detail levels
+        # Negative guard: full-only fields must not leak at lower detail levels
         minimal_result = await mcp_client.call_tool(
             "ha_get_overview", {"detail_level": "minimal"}
         )
@@ -318,8 +318,22 @@ class TestSystemTools:
         assert minimal_data.get("success") is True, (
             f"Minimal overview failed: {minimal_data.get('error')}"
         )
-        assert "allowlist_external_dirs" not in minimal_data.get("system_info", {}), (
-            "allowlist_external_dirs should only appear at detail_level='full'"
+        full_only_fields = {
+            "country",
+            "currency",
+            "unit_system",
+            "latitude",
+            "longitude",
+            "elevation",
+            "components_loaded",
+            "safe_mode",
+            "internal_url",
+            "external_url",
+            "allowlist_external_dirs",
+        }
+        leaked = full_only_fields & minimal_data.get("system_info", {}).keys()
+        assert not leaked, (
+            f"full-only fields leaked at detail_level='minimal': {leaked}"
         )
 
         # Log key information
--- a/tests/src/e2e/workflows/system/test_system_tools.py
+++ b/tests/src/e2e/workflows/system/test_system_tools.py
@@ -310,6 +310,18 @@ class TestSystemTools:
             "allowlist_external_dirs should be a list"
         )
 
+        # Negative guard: field must not leak at lower detail levels
+        minimal_result = await mcp_client.call_tool(
+            "ha_get_overview", {"detail_level": "minimal"}
+        )
+        minimal_data = parse_mcp_result(minimal_result)
+        assert minimal_data.get("success") is True, (
+            f"Minimal overview failed: {minimal_data.get('error')}"
+        )
+        assert "allowlist_external_dirs" not in minimal_data.get("system_info", {}), (
+            "allowlist_external_dirs should only appear at detail_level='full'"
+        )
+
         # Log key information
         logger.info(f"Home Assistant version: {system_info.get('version')}")
         logger.info(f"Location: {system_info.get('location_name')}")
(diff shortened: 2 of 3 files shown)


Patch76:
## Implementation Summary

**Choices Made:**
- Field name `allowlist_external_dirs` matches HA core's current canonical name. The deprecated alias `whitelist_external_dirs` is not added to avoid encoding legacy terminology.
- Scoped to `detail_level="full"` only, consistent with the existing pattern for optional/diagnostic fields in the same block. No change to `minimal` / `standard` response shape or size.
- Defensive default `config.get("allowlist_external_dirs", [])` matches the repo convention already in use for other fields in this block (`config.get("unit_system", {})`, `config.get("safe_mode", False)`).
- Test added to the existing `test_get_system_overview` expected-field list + one separate `isinstance` assertion for the type contract, matching the file's established pattern rather than creating a standalone test.

**Problems Encountered:**
- None.

**Suggested Improvements:**
- Other `/api/config` fields not covered here: `radius`, `debug`, `recovery_mode`, `config_source`, `allowlist_external_urls`. Left out because they are diagnostic rather than agent-actionable — can revisit if a concrete use case surfaces.
- Context: this PR closes the last narrow gap identified in discussion #997. The broader gaps raised there were already addressed by #728 (`ha_get_overview` refactor, 2026-04-03) which landed 13 days before the discussion opened.

Patch76:
Quick follow-up scan for analogous legacy-alias cases in ha-mcp after the review comment about `allowlist_*` vs `whitelist_*`:

**Methodology:**
- Code search across the repo for `whitelist`, `base_url`, `legacy`, `deprecated`, `renamed`, and dual-fallback patterns (`.get("old") or .get("new")`, `pop("legacy")`, etc.)
- Live check of `/api/config` on an HA instance (2026.4.3)

**Findings:**

| Candidate | Status |
|---|---|
| `whitelist_external_dirs` → `allowlist_external_dirs` | Addressed by this PR |
| `whitelist_external_urls` → `allowlist_external_urls` | HA Core no longer exposes the legacy form — only `allowlist_external_urls` is returned |
| `color_temp` / `kelvin` → `color_temp_kelvin` | Already normalized in `src/ha_mcp/tools/device_control.py` (`_normalize_light_color_temp`) |
| `base_url` hits in src/ | All local OAuth/WS parameter names, not the HA-Core http-config legacy key |
| `api_password`, `async_add_job`, `discovery_info`, `hass.helpers`, `icon_template` | 0 hits |
| Dual-fallback patterns | 0 hits |

Live `/api/config` confirms the asymmetry: `whitelist_external_dirs` and `allowlist_external_dirs` both present and identical, but only `allowlist_external_urls` (no legacy twin).

**Conclusion:** No other actively-exposed parallel-alias fields found in the current code. HA Core's general deprecation policy (deprecate → remove, rarely maintain parallel aliases long-term) makes this consistent with expectations. Nothing further to correct — posting for awareness, not as a change request.

Patch76:
## Self-review follow-up

Ran a code review against the project styleguide and flagged one **LOW**-severity gap:

The original test only asserted `allowlist_external_dirs` is *present* at `detail_level="full"`. There was no negative guard for `minimal`/`standard` — a future refactor could silently hoist the field out of the `if detail_level == "full":` block in `tools_search.py` without the test catching it.

Addressed in b7ef598 — added a sibling assertion in `test_get_system_overview` calling `ha_get_overview` with `detail_level="minimal"` and asserting `"allowlist_external_dirs" not in system_info`. Both E2E jobs (ubuntu-latest, ubuntu-24.04-arm) green against a real HA instance.

Scope kept tight: did not extend the same negative guard to the other `full`-only fields (`version`, `location_name`, `time_zone`, `components_loaded`) — that's a separate hardening pass on the existing test, not in scope for this PR.

kingpanther13:
I haven't done a formal review on this yet, I'm still on vacation, but just wanted to note that in my personal experience I've noticed Claude referring to "hardening passes" a lot, when really it's just being lazy and not wanting to do the work. If it's something minor, it's better to just fix it while you're in there. That actually goes back to our "boy scout rule" leave it better than how you found it. If it's a large change then absolutely do a new PR, but something that is just adding a few lines is well within scope to add on imo.

Patch76:
@kingpanther13 fair point, boy-scout-rule applies here. Two corrections to my earlier self-review:

1. **Field list was wrong** — `version`, `location_name`, `time_zone` are base `system_info` fields (always present), not `full`-only. The actual `full`-only set is: `country`, `currency`, `unit_system`, `latitude`, `longitude`, `elevation`, `components_loaded`, `safe_mode`, `internal_url`, `external_url`, `allowlist_external_dirs`.
2. **Extended scope, not deferred** — pushed in `dd3e0ff`: the negative guard now covers the whole `full`-only set via a single set-intersection assertion.

kingpanther13 (changes requested):
Just a few asks before this lands — nice scope, and the broadened negative guard in dd3e0ff is a great call.

### 1. `[]` default conflates "HA didn't expose it" with "HA reported empty allowlist"

`tools_search.py:669-671` currently returns `[]` whether HA actually has zero allowlisted dirs *or* HA didn't surface the key at all (older version, schema change, redacted response). For a security-relevant field, that's a meaningful distinction — an agent reasoning about file paths sees `[]` and concludes "no dirs allowed" when HA may simply not have reported it.

The dominant pattern in the surrounding block defaults to `None` (`country`, `currency`, `internal_url`, `external_url`, `latitude`, `longitude`, `elevation`). I know `unit_system={}` / `safe_mode=False` go the other way, but those aren't security-load-bearing. Suggest:

```python
"allowlist_external_dirs": config.get("allowlist_external_dirs"),
```

Test type assertion would need a tweak (`isinstance(..., list) or value is None`).

### 2. Add a unit test for the missing-key path

The whole `system_info` builder is currently only exercised by E2E. With the default change above, a one-liner mocking `client.get_config()` to return `{}` and asserting `system_info["allowlist_external_dirs"] is None` (or absent, depending on what we land on) locks in the contract and prevents a future refactor from silently switching the default back.

### 3. Split the minimal-tier negative guard into its own test

`test_get_system_overview` now does double duty — full-tier presence checks AND a minimal-tier leak guard. Sibling tests in the same file like `test_overview_notifications_opt_out` follow the narrower "one behavior, name says what it checks" pattern. A regression in the leak guard surfacing under a name about the full overview slows diagnosis. Suggest pulling the minimal-tier block into its own `test_overview_minimal_excludes_full_only_fields` (or similar).

Happy to re-review once these are in.

Patch76:
## Implementation Summary — Review Round 2

**Choices Made:**
- **Default `[]` → `None` in `tools_search.py:669-674`** — adopted. Distinguishes HA-omitted-the-key from HA-reported-empty-allowlist; matches the dominant `None`-default pattern in the surrounding `system_info` block (`country`, `currency`, `internal_url`, `external_url`, `latitude`, `longitude`, `elevation`). 2-line WHY comment added inline because the asymmetry vs. neighbouring `unit_system={}` / `safe_mode=False` is non-obvious.
- **New unit test file `tests/src/unit/test_overview_system_info.py`** — adopted KP2. Three tests pin the contract: missing-key → `None`, list value passes through, field omitted at `detail_level="minimal"`. Uses the `mock_mcp.tool` decorator-capture pattern from `test_tools_get_states.py`; mocks `client.get_config`, `client.send_websocket_message`, `client.base_url`, and `smart_tools.get_system_overview`. Two companion tests beyond the missing-key path are Boy-Scout: lock the contract end-to-end so a future refactor cannot silently revert any of the three behaviours.
- **Negative guard split into `test_overview_minimal_excludes_full_only_fields`** — adopted KP3. Sibling-test naming follows the `test_overview_*` pattern (e.g. `test_overview_notifications_opt_out`). Type assertion in `test_get_system_overview` loosened to `assert allowlist is None or isinstance(allowlist, list)` per the new contract.
- **Boy-Scout skip on `unit_system={}` (line 661)** — same missing-vs-empty ambiguity, but the maintainer’s review explicitly framed this field as not security-load-bearing and out of scope. Left as-is.

**Problems Encountered:**
- None during implementation. Local `uv run pytest` not executable in this environment (Alpine container, musl libc; `requires-python >=3.13` and `uv` provides no musl distros — see `astral-sh/uv#6890`). Validation deferred to CI: Ruff, AST Lint, Mypy, Unit Tests, E2E (ubuntu-latest + arm), Docker & Add-on Validation, Performance Tests, Build (mac/linux/windows) all green on push commit `378f623`.

- The `full_only_fields` set in the new negative-guard test is a manually-maintained mirror of the `system_info.update({...})` keys in `tools_search.py:657-672`. A new full-only field added to the dict won't be caught by the leak guard unless the set is updated too. Could be tightened by extracting to a module-level constant the tool itself iterates from; out of scope for a review-round PR.

kingpanther13 (approved):
All addressed. The `None`-vs-`[]` distinction is in, the unit tests lock the contract, and the broadened set-intersection guard for the full-only fields is a nice cleanup. Good to go.

PR #1055, opened 2026-04-25
fix(internal): improve error guidance for small-model failure modes

sergeykad:
## What does this PR do?

Targeted improvements to error messages and docstring examples that reduce retry loops on smaller models (9B/14B class), based on patterns observed across multiple UAT runs.

**Validation (pre-HA round-trip, only where HA returns no actionable error):**
- `ha_call_{read,write,delete}_tool`: accept `arguments` passed as a JSON string — parse and forward as dict; reject non-object JSON with a clear error. Small models sometimes serialize the nested arguments param before sending.
- `ha_config_set_automation`: when `config` contains `sequence` but no `trigger`/`action`, suggest `ha_config_set_script` instead of the generic missing-fields error.
- `ha_config_set_automation`: condition blocks with a `platform` key (trigger syntax) are now caught client-side before HA crashes with an unhelpful 500 instead of a 400.
- `ha_config_set_automation`: `VALIDATION_INVALID_JSON` on the `config` param now suggests passing `config` as a dict to avoid JSON escaping issues.

**Docstring examples (prevent systematic first-try mistakes on 9B models):**
- `ha_call_read_tool`, `ha_call_write_tool`, `ha_call_delete_tool`: added a concrete `EXAMPLE:` line showing the `name=`/`arguments=` keyword form — model consistently omitted `name=` when calling proxies without it.
- `ha_config_set_automation`, `ha_config_set_script`, `ha_config_set_helper`: all create examples now use explicit keyword form (`config={...}`, `script_id=`, `helper_type=`, `name=`). Previously these used positional style, causing 9B models to consistently pass fields as flat top-level kwargs on first try.
- `ha_config_set_script`: added note that scripts use `sequence`, not `trigger`/`action` (automation syntax).
- `ha_config_set_area`: added create/update examples — none existed before; model was passing `area_id=` instead of `name=` for new areas.

**Shared utility:**
- `coerce_to_list()` added to `util_helpers` and used in the condition validation loop, replacing duplicated inline normalization.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [x] 🧪 Tests only

## Testing

Unit tests added for all new validation paths (102 passing). Validated against multiple 9B and 27B UAT runs — docstring and validation changes address observed systematic failure modes.

## Checklist

### DIFF
--- a/src/ha_mcp/tools/tools_areas.py
+++ b/src/ha_mcp/tools/tools_areas.py
@@ -222,8 +222,15 @@ class AreaTools:
         """
         Create or update a Home Assistant area (room).
 
-        Provide name only to create a new area. Provide area_id to update existing.
         Areas organize entities by physical location for room-based control.
+
+        Create: provide name only.
+        Update: provide area_id (from ha_config_list_areas) plus any fields to change.
+
+        EXAMPLES:
+        ha_config_set_area(name="Kitchen")
+        ha_config_set_area(name="Living Room", icon="mdi:sofa")
+        ha_config_set_area(area_id="kitchen", name="Kitchen Renamed", floor_id="ground_floor")
         """
         try:
             # Parse aliases if provided as string
--- a/src/ha_mcp/tools/tools_config_automations.py
+++ b/src/ha_mcp/tools/tools_config_automations.py
@@ -41,6 +41,7 @@ from .reference_validator import validate_config_references
 from .util_helpers import (
     apply_entity_category,
     coerce_bool_param,
+    coerce_to_list,
     fetch_entity_category,
     merge_validation_meta,
     parse_json_param,
@@ -413,7 +414,7 @@ class AutomationConfigTools:
         BASIC EXAMPLES:
 
         Simple time-based automation:
-        ha_config_set_automation({
+        ha_config_set_automation(config={
             "alias": "Morning Lights",
             "description": "Turn on bedroom lights at 7 AM to help wake up",
             "trigger": [{"platform": "time", "at": "07:00:00"}],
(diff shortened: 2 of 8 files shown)


kingpanther13 (approved):
LGTM — verified against the PR branch:

Validation correctness
- Condition-`platform` pre-check: only fires when `platform` is present without `condition` (`test_platform_with_condition_key_not_flagged` covers the both-keys case correctly).
- Sequence-key hint: scoped to missing `trigger`/`action` so it doesn't fire spuriously when only `alias` is missing.
- JSON-string `arguments` fallback in proxies: parses once, validates the parsed value is an object, logs a warning, and chains the `JSONDecodeError`. Doesn't swallow anything.
- `_parse_and_validate_config` now passes `details` through `create_error_response`, which is supported.
- `coerce_to_list` is fine for the single caller (condition list, where empty dict isn't a valid block anyway).

Docstring/example pass
- All updated kw-form examples match current signatures (`ha_get_history(entity_ids=, start_time=)`, `ha_config_set_area(name=)`, `ha_config_remove_area(area_id=)`).

Two tiny nits — optional, fine to land as-is:

1. `tests/src/unit/test_config_automation_validation.py` module docstring still says `_validate_required_fields: sun trigger event pre-validation` — looks like leftover wording; the third bullet now covers the condition-vs-platform check (`TestValidateConditionBlocks`).
2. `coerce_to_list` docstring says "dict/other → [value]" but `[value] if value else []` returns `[]` for empty containers. Inconsequential at the call site, but worth a one-line tweak if the helper picks up more callers.

PR #1060, opened 2026-04-25
fix(site): allow text selection in Tool Explorer tool names

sergeykad:
## What does this PR do?

Tool names in the Tool Explorer could not be selected and copied because they were rendered inside a `<button>` element, which has `user-select: none` by default in browsers.

- Add `select-text` Tailwind class to the button to override `user-select: none`
- Guard the click handler to skip the expand/collapse toggle when the user is selecting text (`Selection.type === 'Range'`)

## Type of change
- [x] 🐛 Bug fix

## Testing

## Checklist

### DIFF
--- a/site/src/pages/tools.astro
+++ b/site/src/pages/tools.astro
@@ -181,7 +181,7 @@ function getBadgeInfo(annotations: Record<string, any>) {
                 data-tool-size={totalChars}
                 data-original-desc={tool.description}
               >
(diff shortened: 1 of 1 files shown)


kingpanther13 (approved):
LGTM. Standard fix for the `<button>` + `user-select: none` clash — `select-text` for the visual override and the `Selection.type === 'Range'` guard for the click handler is the right pair. Scope is confined to `.tool-header`, no spillover risk.

PR #1069, opened 2026-04-27
fix(addons): route addon API calls through HA Core ingress proxy

### LINKED ISSUE #1001: [BUG] ha_get_logs (supervisor) and ha_call_addon_api fail in v7.3.0 (regression from 7.2.0)
## 📋 Bug Description
After upgrading to ha-mcp 7.3.0, `ha_get_logs` with `source="supervisor"` fails for all addon slugs, 
and `ha_call_addon_api` fails to connect to any addon. Both worked in 7.2.0.

## 🔄 Steps to Reproduce
1. Upgrade ha-mcp from 7.2.0 to 7.3.0
2. Call `ha_get_logs(source="supervisor", slug="core_mosquitto")`
3. Call `ha_call_addon_api(slug="ccab4aaf_frigate", path="/api/stats", port=5000)`

**Expected:** Addon logs and addon API responses returned as in 7.2.0

**Actual:** 
- `ha_get_logs` supervisor returns `SERVICE_CALL_FAILED: Command failed` for all addon slugs
- `ha_call_addon_api` returns `CONNECTION_FAILED: All connection attempts failed` for all addons (times out after ~21 seconds)
- HA system log shows: `Client error on /addons/[slug]/logs request 200, message='Attempt to decode JSON with unexpected mimetype: text/plain'`

## 🔧 Environment
- **ha-mcp Version:** 7.3.0
- **Platform:** Windows 11 (AMD64)
- **Python Version:** 3.14.3
- **Home Assistant Version:** 2026.4.2
- **Connection Status:** Connected
- **Entity Count:** 1145

```markdown
## 📋 Bug Description
After upgrading to ha-mcp 7.3.0, `ha_get_logs` with `source="supervisor"` fails for all addon slugs, 
and `ha_call_addon_api` fails to connect to any addon. Both worked in 7.2.0.

## 🔄 Steps to Reproduce
1. Upgrade ha-mcp from 7.2.0 to 7.3.0
2. Call `ha_get_logs(source="supervisor", slug="core_mosquitto")`
3. Call `ha_call_addon_api(slug="ccab4aaf_frigate", path="/api/stats", port=5000)`

**Expected:** Addon logs and addon API responses returned as in 7.2.0

**Actual:** 
- `ha_get_logs` supervisor returns `SERVICE_CALL_FAILED: Command failed` for all addon slugs
- `ha_call_addon_api` returns `CONNECTION_FAILED: All connection attempts failed` for all addons (times out after ~21 seconds)
- HA system log shows: `Client error on /addons/[slug]/logs request 200, message='Attempt to decode JSON with unexpected mimetype: text/plain'`

## 🔧 Environment
- **ha-mcp Version:** 7.3.0
- **Platform:** Windows 11 (AMD64)
- **Python Version:** 3.14.3
- **Home Assistant Version:** 2026.4.2
- **Connection Status:** Connected
- **Entity Count:** 1145
```

hhopke:
Refs #1001 — addresses the `ha_manage_addon` (formerly `ha_call_addon_api`) symptom; see "Scope" below.

## Problem

`ha_manage_addon` proxy mode opened direct HTTP/WebSocket connections to the add-on's container IP on HA's `hassio` Docker bridge. That works when the MCP host shares HA's network namespace (the HAOS / Supervised add-on variant), but is unreachable from PyPI/uvx installs or any off-host setup sitting on a regular LAN — TCP just goes nowhere, hence the ~21 s `CONNECTION_FAILED` reported.

## Fix

Route the default (Ingress) path through HA Core's existing proxy:

1. Mint a Supervisor ingress session via the WS `supervisor/api` hop (same path `ha_get_addon` already uses, so this works equally on HAOS, Supervised, and off-host installs).
2. Send the request to `<HA_URL>/api/hassio_ingress/<addon_token>/<path>` with `Cookie: ingress_session=<session>`. HA Core proxies to the add-on container; we never touch `172.30.x.x` directly.

Both HTTP and WebSocket modes are covered. The `port=…` direct-container path is preserved unchanged (it's the only way to reach non-Ingress ports like Node-RED's 1880 or ESPHome's 6052), but its `ConnectError` now includes a suggestion to drop `port=…` so off-host callers get a useful next step instead of a silent 21 s timeout.

## Scope

Issue #1001 reports two failing tools. This PR addresses one:

- ✅ `ha_manage_addon` (proxy mode, ingress) — fixed.
- ⚠️ `ha_get_logs(source="supervisor")` — works already off-host (#950 retired the broken WS path)

## Validation

Tested from off-host PyPI/uvx (macOS) → real HAOS instance:

| Add-on | Slug | Path | Result |
|---|---|---|---|
| File Editor | `core_configurator` | `/` | ✅ 200, HTML body |
| Matter Server | `core_matter_server` | `/api/version` | ✅ 404 from add-on |
| Music Assistant | `d5369777_music_assistant` | `/api/info` | ✅ 404 from add-on |

All three round-trip cleanly through `<HA_URL>/api/hassio_ingress/<token>/...`. Pre-fix, all three would have hung ~21 s and errored out — never reaching the add-on. The 404s are proof of routing, not failure: a real HTTP response from the add-on container means the proxy chain worked.

## Test plan

- [x] Unit tests cover both modes:
  - `test_http_ingress_routes_through_ha_core` — asserts URL, cookie, no `X-Ingress-Path` header (HA Core adds those itself when proxying)
  - `test_http_direct_port_skips_ingress_session` — asserts direct-container URL + no session minted
  - `test_http_direct_port_offhost_error_hints_at_ingress` — asserts `ConnectError` suggestion mentions Ingress mode
  - WebSocket equivalents for all three
- [x] Existing addon unit tests updated to mock `_create_ingress_session`
- [x] Full E2E suite passes locally (no addon-specific E2E exists; testcontainer doesn't run Supervisor, so add-ons aren't reachable in that environment — unit tests are the appropriate coverage layer)
- [x] `ruff check` clean, `mypy` clean

### DIFF
--- a/src/ha_mcp/tools/tools_addons.py
+++ b/src/ha_mcp/tools/tools_addons.py
@@ -13,7 +13,7 @@ import logging
 import re
 import time
 from typing import Annotated, Any
-from urllib.parse import unquote
+from urllib.parse import unquote, urlsplit
 
 import httpx
 import websockets
@@ -674,11 +674,15 @@ async def _call_addon_ws(
                 )
             )
         session = await _create_ingress_session(client)
-        ws_scheme = "wss" if client.base_url.startswith("https") else "ws"
-        ws_host = client.base_url.split("://", 1)[1]
-        ws_url = f"{ws_scheme}://{ws_host}{ingress_entry}/{normalized}"
+        # urlsplit handles trailing slashes and reverse-proxied path
+        # prefixes that bare string slicing would mangle.
+        parsed = urlsplit(client.base_url)
+        ws_scheme = "wss" if parsed.scheme == "https" else "ws"
+        ws_path_prefix = parsed.path.rstrip("/")
+        ws_url = f"{ws_scheme}://{parsed.netloc}{ws_path_prefix}{ingress_entry}/{normalized}"
+        # Cookie-only matches the HA frontend. A bearer here would be
+        # forwarded to the add-on upstream — needless LLAT leak.
         headers["Cookie"] = f"ingress_session={session}"
-        headers["Authorization"] = f"Bearer {client.token}"
 
     # 7. Connect and exchange messages
     collected: list[str] = []
@@ -759,14 +763,25 @@ async def _call_addon_ws(
                 total_size += len(clean)
 
     except websockets.exceptions.InvalidHandshake as e:
+        suggestions = [
+            "Check that the add-on supports WebSocket on this path",
+            f"Use ha_get_addon(slug='{slug}') to inspect available endpoints",
+        ]
+        # 401/403 means auth was rejected, not a path-shape problem.
+        if isinstance(e, websockets.exceptions.InvalidStatus):
+            status = e.response.status_code
+            if status in (401, 403):
+                suggestions = [
+                    "The ingress session may have expired or your HA token "
+                    "may lack the required scope. Verify the token has admin "
+                    "rights and try again.",
+                    f"Status {status} from the WebSocket handshake.",
+                ]
         raise_tool_error(
             create_error_response(
                 ErrorCode.SERVICE_CALL_FAILED,
                 f"WebSocket handshake failed with '{addon_name}': {e!s}",
-                suggestions=[
-                    "Check that the add-on supports WebSocket on this path",
-                    f"Use ha_get_addon(slug='{slug}') to inspect available endpoints",
-                ],
+                suggestions=suggestions,
                 context={"slug": slug, "path": path},
             )
         )
@@ -1024,9 +1039,10 @@ async def _call_addon_api(
                 )
             )
         session = await _create_ingress_session(client)
-        url = f"{client.base_url}{ingress_entry}/{normalized}"
+        url = f"{client.base_url.rstrip('/')}{ingress_entry}/{normalized}"
+        # Cookie-only matches the HA frontend. A bearer here would be
+        # forwarded to the add-on upstream — needless LLAT leak.
         headers["Cookie"] = f"ingress_session={session}"
-        headers["Authorization"] = f"Bearer {client.token}"
 
     # 6. Set content type based on body type
     if isinstance(body, dict):
--- a/src/ha_mcp/tools/tools_addons.py
+++ b/src/ha_mcp/tools/tools_addons.py
@@ -20,6 +20,7 @@ import websockets
 from fastmcp.exceptions import ToolError
 from pydantic import Field
 
+from .._version import is_running_in_addon
 from ..client.rest_client import HomeAssistantClient
 from ..errors import (
     ErrorCode,
@@ -290,9 +291,10 @@ def _addon_connection_failure_suggestions(
 ) -> list[str]:
     """Suggestions for connect/timeout failures against an add-on.
 
-    Direct-port mode hits the addon's container IP, which off-host installs
-    can't route to — the actionable hint is to drop `port`. Ingress mode
-    hits HA Core, so failures usually mean HA Core is unreachable.
+    Three modes — direct-port hits a container IP, the addon-variant ingress
+    route hits a sibling container's ingress port, the off-host ingress route
+    hits HA Core. Each mode fails for different reasons, so suggest different
+    next steps.
     """
     if port:
         return [
@@ -300,6 +302,14 @@ def _addon_connection_failure_suggestions(
             "Direct-port access requires the MCP host to share Home "
             "Assistant's container network. On PyPI/uvx installs, drop "
             "the 'port' parameter to route through Ingress instead.",
+        ]
+    if is_running_in_addon():
+        return [
+            "The target add-on container may not be reachable from this "
+            "MCP add-on. Check that the target add-on is running.",
+            "If the failure persists, the addon Docker network may be "
+            "unhealthy — try restarting the target add-on, then this "
+            "MCP add-on.",
         ]
     return [
         f"Verify Home Assistant is reachable at {client.base_url}",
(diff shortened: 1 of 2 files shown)


kingpanther13 (changes requested):
Thanks for digging into this — the off-host failure mode is real and routing through `/api/hassio_ingress/<token>/...` is the right shape.

One concern dominates the rest: this swaps the default routing wholesale rather than keeping both paths. HAOS-addon installs that previously took the direct `<addon_ip>:<ingress_port>` shortcut now take an extra hop through HA Core plus a per-call Supervisor session mint. We can't validate the HAOS-addon variant in CI (testcontainer doesn't run Supervisor, no addon-bearing e2e exists), so behavior on that variant is "should be equivalent" without proof — which is how regressions ship.

**Concrete ask:** keep both routes and let the new one fail over to the old one. Sketch:

```python
async def _call_addon_api(client, slug, path, ...):
    ...
    if port:
        # direct-mapped-port path — unchanged
        ...
        return await _do_request(url, headers, ...)

    # ingress mode: try HA Core proxy first, fall back to direct ingress_port
    try:
        session = await _create_ingress_session(client)
        proxy_url = f"{client.base_url}{ingress_entry}/{normalized}"
        proxy_headers = {"Cookie": f"ingress_session={session}"}
        return await _do_request(proxy_url, proxy_headers, ...)
    except (httpx.ConnectError, httpx.ConnectTimeout, ToolError) as proxy_err:
        # HAOS-addon shortcut: skip HA Core, hit the addon container directly
        if not addon.get("ip_address") or not addon.get("ingress_port"):
            raise  # nothing to fall back to
        direct_url = f"http://{addon['ip_address']}:{addon['ingress_port']}/{normalized}"
        direct_headers = {
            "X-Ingress-Path": ingress_entry,
            "X-Hass-Source": "core.ingress",
        }
        try:
            return await _do_request(direct_url, direct_headers, ...)
        except Exception:
            raise proxy_err  # surface the original (more informative) error
```

(Same shape for `_call_addon_ws`.) Off-host installs: proxy succeeds, fallback never runs. HAOS-addon installs: if proxy hits any snag, the direct path takes over — matching old behavior. Optional refinement: cache which path worked per `(base_url, token)` so steady-state calls skip the failed attempt.

Alternative if you'd rather not have implicit retry: expose it as a `routing` param (`"ingress" | "ingress_port" | "direct"`), default `"ingress"`, and let the LLM pick. Same two code paths, explicit instead of automatic. I prefer the fallback because the LLM doesn't reliably know which install variant it's on, but either is fine.

A few smaller items worth bundling while you're in there:

**Tests** (unit-level — testcontainer can't reach addons, so this is the right layer)
- `_create_ingress_session` has no direct tests — the fixture mocks it everywhere. Cover the three failure modes: supervisor error response, `result` without `session`, and empty/non-string session token.
- The `wss://` scheme branch in `_call_addon_ws` is never exercised — both ingress WS tests use `http://localhost:8123`. A `client.base_url = "https://..."` test would lock the contract.
- `test_ws_direct_port_skips_ingress_session` doesn't assert `Cookie` is absent — its HTTP twin does. Mirror the assertion.
- No WS analogue to `test_http_direct_port_offhost_error_hints_at_ingress`. Same regression risk on the WS branch.
- If you go the fallback route: tests for the failover (proxy fails → direct succeeds; proxy fails → direct fails → original error surfaces).

**Error UX**
- `ConnectError` in ingress mode currently gets `suggestions=["Check that the add-on is running"]` because the ingress hint only fires inside `if port:`. The actual failure is usually HA Core unreachable from the MCP host — add an `else` suggestion pointing at `client.base_url`.
- WS `InvalidHandshake` 401 surfaces as a generic "handshake failed" — worth special-casing to suggest "ingress session may have expired or token may lack admin scope" when the response is a 4xx.

**Smaller**
- `Authorization: Bearer {client.token}` on the ingress path is redundant — the `ingress_session` cookie authenticates to HA Core, and the bearer then gets forwarded to the add-on's upstream where it's mostly stripped but occasionally inspected. Cookie-only matches what a browser sends.
- `client.base_url.split("://", 1)[1]` breaks on a trailing slash or path-prefixed reverse-proxied HA. `urllib.parse.urlsplit` is cleaner.
- Per-call session minting adds a WS round trip per `ha_manage_addon` call. A small cache keyed on `(base_url, token)` with ~10-min TTL would amortize that without adding lifetime management complexity. Optional.
- Docstring on `_create_ingress_session` claims "sessions valid ~15 minutes" (tied to a private Supervisor constant) and "minted via WS `supervisor/api` proxy" (actually REST when on-host per `_supervisor_api_call`). Worth softening both.
- `_call_addon_api` and `_call_addon_ws` lost their `Args:` sections in the rewrite — `offset`/`limit`/`wait_for_close` semantics aren't documented anymore.

The fallback (or the routing param) is the only structural ask. Everything else is bundle-while-you're-in-there.

kingpanther13:
Once you can address all of those things please make the PR ready for review and I'll do one more review pass over it, and hopefully we can get it merged.

hhopke:
Thanks for the careful read. I am not so certain if you're structural request is going in the right direction. I would propose not to rush this. This can wait until you're back. To me it seems to be uneccasary complexity in the code, more at the bottom. 

## Bundled

**Tests** (70 passing)
- `_create_ingress_session`: happy path + 4 failure modes
- `wss://` branch and trailing-slash regression tests (HTTP + WS)
- WS analogues for the off-host hint test (OSError + TimeoutError)
- `Cookie` and `Authorization` absence asserted on direct-port and ingress paths
- `InvalidStatus` 401/403 → token/scope hint; 404 → keeps path hint

**Error UX**
- Ingress-mode connect/timeout failures point at HA Core reachability, not "check the add-on"
- Same fix extended to `TimeoutException`/`TimeoutError` — Haiku stress run on my off-host install showed direct-port failures often surface as timeouts (silent TCP drop), not `ConnectError`. Both share `_addon_connection_failure_suggestions`.
- WS handshake 401/403 now suggests token/scope; other 4xx keeps the existing "WebSocket on this path" hint.

**Routing / headers**
- Dropped `Authorization: Bearer` on the ingress path. Verified against HA Core source: `/api/hassio_ingress/<token>/...` is `requires_auth = False`, the `ingress_session` cookie is sufficient, and the HA frontend sends cookie-only. The bearer was added in this PR's commit (not pre-existing) and would otherwise be forwarded to the add-on upstream — needless LLAT-leak vector. Test assertions lock the absence.
- `urlsplit` for WS, `rstrip("/")` for HTTP — handles trailing slashes and reverse-proxied path prefixes
- Restored `Args:` sections; softened `_create_ingress_session` docstring on the "~15 minutes" claim

## On the fallback / `routing` param

I see the appeal but I think it costs more than it saves:

- **Implicit fallback masks errors.** Proxy failure for a real reason (auth, addon down, bad path) gets buried behind a retry that either succeeds with different semantics or surfaces a confusing "we tried everything" error.
- **Two paths = two security models.** Direct `<addon_ip>:<ingress_port>` skips the Supervisor session validation the proxy enforces. Falling back to a less-restrictive route on failure is a quiet security regression.
- **The proxy path isn't unproven, it's just untested in our harness.** `/api/hassio_ingress/<token>/...` is the same endpoint the HA frontend hits on every Ingress click, from any topology. The "should be equivalent" framing reads as a CI-coverage gap, which I agree with — but the answer is to verify, not hedge in code.
- **`routing` param is worse.** The LLM doesn't reliably know which install variant it's on; exposing implementation detail as a knob most callers shouldn't touch is a bad API.
- **Perf regression on HAOS-addon is real.** Intra-bridge hop + one WS round trip per call. I haven't measured it under load — single calls in my Haiku run felt fine but that's vibes, not data. If HAOS users report slowness post-merge, an explicit addon-only fast-path inside the addon-detection branch would target it surgically without the failover-shape concerns above.

## Not done

- **Session caching.** Per-call mint is one WS round trip; I don't have measurements showing it's a bottleneck. Unlike the items above, this one introduces *new state* and *new failure modes* (cache + invalidation + the "cached session got 401, re-mint and retry" recovery path). The retry-on-stale pattern is structurally the same shape as the routing fallback we discussed — implicit retry that can mask real failures. "Easy to add later" is genuine here: it's pure additive optimization with no contract change. Finally its your call just wanted to pause here for a second.

Enjoy the vacation.

kingpanther13:
Tested PR head against a real HAOS-addon install (`d6450bd1_ha_mcp_dev` running PR branch `dev91`). The HAOS-addon variant regression my prior review flagged is real.

**Default ingress mode results (same calls that work on master):**

| Addon | Master | PR | Direct-port escape hatch? |
|---|---|---|---|
| Node-RED | 403 (Nginx) | **401** | yes (1880) |
| ESPHome | 403 (Nginx) | **401** | yes (6052) |
| Frigate | **200** ✅ | **401** ❌ | yes (5000) |
| evcc | **200** ✅ | **401** ❌ | **no** (7070 not in `ports` map) |

Frigate and evcc are the smoking gun: they were working on master via the direct-IP-to-`ingress_port` path that this PR removed. On PR they 401. evcc has no entry in `ports` for its ingress port, so users have no `port=` workaround — the tool is fully broken for that addon.

**Root cause** (debug output below):

```
URL:     http://supervisor/core/api/hassio_ingress/<token>/
Headers: Cookie: ingress_session=<...>
Result:  401 Unauthorized
Server:  Python/3.14 aiohttp/3.13.5  (← Supervisor/HA Core, not the addon's Nginx)
```

Inside the HAOS-addon variant, `client.base_url` is `http://supervisor/core` (Supervisor's proxy mount), not direct HA Core. The Supervisor `/core/*` proxy requires `Authorization: Bearer $SUPERVISOR_TOKEN` to authenticate the *outer* hop — the `ingress_session` cookie alone gets 401 before the request ever reaches HA Core's `/api/hassio_ingress/...` handler.

The `a16034e0 fix(addons): drop bearer on ingress path` commit assumed `base_url` points directly at HA Core. That's true off-host (`http://192.168.1.150:8123`) but false on the addon variant.

(Side note: the resulting "Nginx IP restriction" suggestion text fires on the 401 too — the HTTP-side 401-vs-403 distinction my prior review asked about would prevent users being misdirected here.)

**Two ways to fix:**

1. **Install-variant detection.** At startup or per-call, sniff whether `client.base_url` starts with `http://supervisor/` (or check `SUPERVISOR_TOKEN` env). If yes, use the old direct-IP-to-`ingress_port` route (which always worked on the addon variant). If no, use the new HA Core proxy route (which fixes off-host). Could also be wrapped as a FastMCP transform if you'd rather not thread the decision through the tool body. No per-call retry cost, no ambiguity in error messages.
2. **Fallback** (my original ask). Try the new HA Core proxy route first; on `ConnectError`/auth failure, fall back to the direct-IP-to-`ingress_port` route. Off-host installs: proxy succeeds, fallback never runs. HAOS-addon installs: proxy 401s, direct succeeds. Costs an extra round-trip on the addon variant unless the choice is cached per `(base_url, token)`.

Either fixes the regression. I'd lean #1 since the routing decision is really a property of the install variant and doesn't change between calls — detect-and-pin is cleaner than retry.

kingpanther13:
Oh and keep in mind there are some other PRs right now targeting the same area, so there may be more merge conflicts

hhopke:
@kingpanther13 — addressed the addon-variant regression and your HTTP 401/403 side note. Switched to install-variant detection (your option #1, env-var sniff via existing `is_running_in_addon()` helper, no fallback retry).

**Routing now:**
- `is_running_in_addon()` true → direct `<addon_ip>:<addon_ingress_port>` with `X-Ingress-Path` + `X-Hass-Source: core.ingress` (matches master pre-PR).
- `is_running_in_addon()` false → `<base_url>/api/hassio_ingress/<token>/...` with `Cookie: ingress_session=<token>` (cookie-only, no bearer).
- `port=…` direct-port branch unchanged.

**HTTP 4xx response handling — split per your side note (a1b2767):**
- 401 → auth/scope/session hint, `addon_config` dropped (token issue, not config).
- 403 → existing "Nginx IP restriction / leave_front_door_open" hint + `addon_config` preserved (correct for ACL/IP block).
- Mirrors the WS handshake split that already shipped in a16034e.

Off-host re-validated end-to-end against File Editor + Music Assistant — 200 with cookie, no `Authorization`. Branch rebased onto current master.

The HAOS-addon branch is exercised by new unit tests (URL/header shape, missing-port errors, `Cookie`/`Authorization` absence on the variant route, addon-network suggestion text, `wss://` not used on variant). I don't have HAOS install — would you mind re-running the same matrix you used last time on `d6450bd1_ha_mcp_dev`? 

**as before not done:**
- **Session caching** — additive optimization, no contract change. Adds state + invalidation + retry-on-stale failure modes; "easy to add later" is genuine here.

kingpanther13:
Thank you! Please test on your end to be sure it accurately detects your installation and can still access the add-ons. Also please jog my memory, which add-ons are you testing on? I've been using frigate evcc node red and esphome. They all have slightly different ways for the tool to access them. 

I will test again on my side either later today or tomorrow when I am able. 

Thanks again for putting all of the work in on this, and thanks for the help, I don't think I would have had any good way to test your specific installation method myself.

hhopke:
**Detection confirmed both ways from macOS / PyPI:**
- Direct check: `SUPERVISOR_TOKEN` unset → `is_running_in_addon()` returns False.
- Routing: `ha_manage_addon` debug output verified — `Cookie: ingress_session=...` present, `Authorization` absent, `X-Ingress-Path` absent. That's the off-host signature.

**Add-ons I tested**:
- **File Editor** (`core_configurator`) — Node-based — 200
- **Matter Server** (`core_matter_server`) — Python — 200
- **Music Assistant** (`d5369777_music_assistant`) — Python SPA — 200
- **ESPHome** (`5c53de3b_esphome`) — auth-aware (`auth_api: true`, `leave_front_door_open`) — `/devices` 200 with full JSON

I don't have Frigate / evcc / Node-RED installed.

kingpanther13 (approved):
Verified the install-variant fix on a real HAOS-addon install — Frigate and evcc default-ingress mode return 200 (the regression that pulled the prior CHANGES_REQUESTED is gone), direct-port path still works for Node-RED/ESPHome.

Withdrawing one observation from my prior review: the `success` check in `_create_ingress_session` isn't dead — `_supervisor_api_call` returns a `success=False` dict on connection errors rather than raising, so the guard is live and correct.

Approved.

PR #1070, opened 2026-04-28
feat: add Microsoft 365 calendar and To Do tools

andlo:
Adds tools_ms365.py with 7 new tools for interacting with Microsoft 365 via the Graph API. Auto-discovered by the existing tools_* registry — no changes to server.py or registry.py needed.

New tools:
- ms365_get_calendar_events    — list Outlook calendar events (date range)
- ms365_create_calendar_event  — create event with location/attendees
- ms365_update_calendar_event  — PATCH subject/time/location (partial update)
- ms365_delete_calendar_event  — delete event by ID
- ms365_get_todo_tasks         — list To Do tasks by list/status
- ms365_add_todo_task          — add task with due date and importance
- ms365_complete_todo_task     — mark task as completed

Design:
- Token management with in-memory cache + automatic refresh token rotation
- Multi-user support via MS365_REFRESH_TOKEN_<NAME> env vars
- Per-user secondary calendar IDs via MS365_CALENDAR_ID_<NAME>
- Tags: {MS365, Calendar} / {MS365, Todo} for client-side filtering
- httpx for Graph API calls (already a transitive dep via fastmcp)

Tests: 20 unit tests — mocked HTTP, no real credentials needed.
Docs:  docs/ms365-integration.md — Azure setup, device code flow, examples.

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

## Checklist

### DIFF
--- a/docs/ms365-integration.md
+++ b/docs/ms365-integration.md
@@ -1,0 +1,147 @@
+# Microsoft 365 Integration
+
+This document describes how to configure the Microsoft 365 tools in ha-mcp,
+enabling Claude and other AI assistants to read and manage your Outlook
+calendar and Microsoft To Do tasks directly from Home Assistant.
+
+## Available tools
+
+| Tool | Description |
+|------|-------------|
+| `ms365_get_calendar_events` | List events in a date range |
+| `ms365_create_calendar_event` | Create a new calendar event |
+| `ms365_update_calendar_event` | Update subject, time, location etc. |
+| `ms365_delete_calendar_event` | Delete an event by ID |
+| `ms365_get_todo_tasks` | List To Do tasks (by list / status) |
+| `ms365_add_todo_task` | Add a task with optional due date |
+| `ms365_complete_todo_task` | Mark a task as completed |
+
+All tools are auto-discovered by the existing `tools_*` registry — no changes
+to `server.py` or `registry.py` are needed.
+
+---
+
+## Prerequisites — Azure App Registration
+
+You need an Azure app with delegated Graph API permissions.
+
+### Step 1 — Create the app
+
+1. Go to [portal.azure.com](https://portal.azure.com) →
+   **Azure Active Directory** → **App registrations** → **New registration**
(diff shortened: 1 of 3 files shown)


kingpanther13:
Thank you for the contribution. 

In its current state, this is out of scope for the project. We have been actively trying to consolidate and pare down tools, so adding 7 new tools is not the direction we would like to take. Additionally, new tools should be specific to HA, and not for individual add-ons/apps/integrations. This would be better off as a standalone MCP, or we could add on functionality to some existing tools.

Have you tried using the manage addon / call addon api tool to do these things? I'm not familiar with if the Microsoft 365 calendar counts as an addon or integration. Either way our generic tools should cover it; if not, then we can consider a PR adding the functionality to our existing tools instead of making new tools. 

I'll mark this as draft for now till you can decide what direction you want to take it.

andlo:
Thanks for the quick review and the honest feedback — I appreciate it.

I understand the reasoning. Keeping the tool set focused on HA-native functionality makes sense, and adding 7 new tools for a third-party service is clearly going against that direction.

To answer your question: the Microsoft 365 calendar isn't a traditional HA add-on — it's the [MS365 Home Assistant integration](https://github.com/RogerSelwyn/MS365-Calendar) by RogerSelwyn, which exposes calendar events as HA entities and services. The generic HA tools could technically cover some of it, but the UX for an LLM agent would be quite rough compared to purpose-built tools.

Given that, I'll close this PR and publish it as a standalone MCP server instead — that way it's useful for those who want the integration without affecting this project's scope.

Thanks again for taking the time to look at it.

kingpanther13:
No problem! If you want to make an issue or PR to enhance/extend the existing tools to cover the Microsoft calendar integration we will definitely consider that, that would be well within scope for our project.

PR #1096, opened 2026-05-03
feat: add Linux support for install & docs

JohnBarton27:
## What does this PR do?

Adds a one-command installer and guided setup for Linux users, bringing Linux to parity with the existing macOS and Windows support.

## Type of change
- [ ] 🐛 Bug fix
- [X] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

## Checklist
- [X] I have updated documentation if needed


kingpanther13:
@JohnBarton27 it appears that Claude Desktop isn't available on Linux, how are you running it? When I check the official Claude docs it states that Claude is only available via CLI on Linux. The docs that you set up seem to only be instructions for setting up using Claude Code, but someone using these instructions would be stuck at a dead end when they get to the step that says "download Claude Desktop from Claude.ai."

JohnBarton27:
@kingpanther13 There's an "unofficial" build script that works pretty well for getting Claude Desktop to run on Linux (at least for Debian-based systems): https://github.com/aaddrick/claude-desktop-debian. I can definitely add that link into the docs

kingpanther13 (changes requested):
Thanks for taking this on — Linux parity is something users have asked for, and the skeleton you've put together (the macOS+Linux installer refactor, the new guide page, the wizard integration) is in good shape structurally. There's a fair number of items below, but the headline is the foundation is solid; mostly we need to reshape the Linux scope and harden some shell-script error-handling patterns. Once these are addressed, we'll merge.

## Headline: reshape the primary Linux flow to Claude Code; keep Claude Desktop as a community alternative

What tripped me up while reviewing: Anthropic doesn't actually ship Claude Desktop for Linux. `claude.com/download` renders "Not available for Linux" in plain text — so as currently written, the README and guide step 3 ("Download Claude Desktop") send Linux users to a dead-end page.

The good news is the officially-supported Linux MCP path is **Claude Code** (the CLI) — packaged for Ubuntu 20.04+, Debian 10+, and Alpine with apt/dnf/apk repos. That's the structural parallel to "Claude Desktop on macOS/Windows" for Linux users.

I took a look at `aaddrick/claude-desktop-debian` you mentioned. It's a legitimate, well-maintained project (~4k stars, multi-contributor, deb/rpm/AUR/Nix/AppImage coverage, Apache-2.0) — definitely reasonable to document as a community alternative. The reason it's not the right shape for the *primary* quick-start is that it's an unofficial bridge between Claude Desktop versions and Linux: when (not if) it lags an Anthropic Desktop update, ha-mcp inherits the support burden during the gap.

Concrete reshape:

1. **Primary Linux flow → Claude Code.** Install Claude Code, configure MCP via `claude mcp add` (or the canonical equivalent), test. Parallels Desktop on macOS/Windows structurally.
2. **Secondary callout for Claude Desktop on Linux** via `aaddrick/claude-desktop-debian` — clearly framed as community/unofficial, with a brief caveat that Anthropic doesn't officially support it.
3. `install.sh` Linux branch defaults to Claude Code; the Desktop path can stay as an opt-in alternate (separate script or `--desktop` flag).
4. `guide-linux.astro` restructured with Claude Code as the linear flow, Desktop as a callout near the bottom.
5. README Linux entry should match the official-support tier framing of the macOS/Windows entries.

## Shell-script hardening (worth fixing while we're here)

Installer scripts have a lot of subtle traps that even experienced shell devs hit. The current shape has a few patterns that swallow real failures and print green success in their place. Each one is small to fix:

1. **Wrappers `curl | sh` without pipefail.** `install-linux.sh:6` and `install-macos.sh:6` are `set -e` + `curl … | sh`. POSIX `sh` has no `pipefail`; if curl 404s, the empty body pipes to `sh` which exits 0. The wrapper "succeeds" silently. This is what's causing the current `test-macos` CI failure — the wrapper curls `master/install.sh`, which doesn't exist on master yet, returns 404, `sh` runs nothing, "success." Same trap fires for any future repo rename or transient GitHub 5xx. Fix: explicit failure capture, e.g.
   ```sh
   SCRIPT=$(curl -LsSf https://.../install.sh) || { echo "Failed to download installer" >&2; exit 1; }
   [ -n "$SCRIPT" ] || { echo "Empty installer body" >&2; exit 1; }
   printf '%s\n' "$SCRIPT" | sh
   ```

2. **JSON parse failure silently discards user's other MCP servers.** `install.sh:377` catches `json.JSONDecodeError` and falls back to `config = {}`. If the user has a malformed-but-recoverable existing config (BOM, trailing comma, JS-style comments — all common in hand-edited JSON) with other MCP servers configured (Filesystem, GitHub, etc.), the installer silently replaces it with an ha-mcp-only config. They have the backup file, but they're never told *why* it happened. Fix: print a clear warning identifying the parse error and pointing at `$BACKUP_FILE`.

3. **`|| true` on pre-download masks all failures, prints green success.** `install.sh:422` runs `uvx … --version > /dev/null 2>&1 || true` then unconditionally prints `Dependencies cached` in green. Network failure, PyPI 5xx, broken release — all silently OK, then the user hits a confusing failure on first Claude Desktop launch with no breadcrumb. Fix: replace with `if … ; then green-OK; else yellow-degraded with recovery hint; fi`.

4. **JSON write isn't atomic.** `install.sh:393` opens with `'w'` (truncates first), then `json.dump`. Crash mid-write leaves a half-written config. Fix: write to `${CONFIG_FILE}.tmp` then `os.replace(tmp, file)`.

## CI workflow updates

- The `test-installer-scripts.yml` paths filter only watches `install-macos.sh` and `install-windows.ps1`. Add `install.sh` and `install-linux.sh` so future edits trigger CI.
- Update `test-macos` to invoke `sh scripts/install.sh` directly — testing the wrapper just tests whatever's on master, not the PR's actual changes.
- Add a `test-linux: runs-on: ubuntu-latest` job mirroring `test-macos` once Linux is in scope.

(The other CI failure, `check-demo-env`, is environmental — TLS exit 60 on the demo server — not PR-related.)

## Linux edge cases

- **WSL silently writes a useless config.** `uname -s` returns `Linux` on WSL2, but Claude Desktop on Windows reads from `%APPDATA%\Claude\`. The script writes to `~/.config/Claude/` — appears to succeed, does nothing. Either detect WSL (`grep -qi microsoft /proc/version`) and refuse with a pointer to the Windows installer, or document the limitation in the guide.
- **`python3` is a hard runtime dependency** at `install.sh:365` but isn't checked beforehand. Minimal Debian/Ubuntu cloud images and Alpine don't ship it by default. With `set -e` the script aborts mid-config-write with `command not found` and no actionable message. Add a preflight `command -v python3` check with package-manager hints.

## Doc inconsistencies (small)

- `--python 3.13` appears in the installer's args and in the troubleshooting note, but is missing from the manual config snippets in the guide (step 2 and step 6). Either include it in all places or omit from all.
- The manual config in step 2 uses bare `"command": "uvx"` — Claude Desktop doesn't inherit shell PATH, so this will fail on a fresh system. Should match the installer's pattern of using the absolute path. (Pre-existing pattern in the macOS guide too — technically not introduced here — but on Linux `~/.local/bin` is even less likely to be in PATH than `/opt/homebrew/bin` is on macOS, so worth fixing.)
- `xdg-open "$HOME/.config/Claude/claude_desktop_config.json"` (guide step 6) routes by MIME, often opens in a browser. `${EDITOR:-nano}` or instructing the user to open with their text editor is more reliable. (Only relevant if Desktop path stays as a documented option.)
- The demo URL `https://ha-mcp-demo-server.qc-h.net` and creds `mcp/mcp` are now duplicated across `install.sh` (twice), `README.md`, and `guide-linux.astro` (twice). If the demo rotates, four surfaces go stale. Worth centralizing as a constant.

## Nits (only if you're already touching the area)

- Linux badge in README uses `logoColor=black` while macOS/Windows badges use `logoColor=white`. Visual asymmetry.
- `grid md:grid-cols-3` at `max-w-3xl` (in `setup.astro`) may feel cramped on tablet sizes — worth eyeballing with `npm run dev` after the reshape.
- `set -e` in the wrappers is essentially dead code (only command is the pipe). Cosmetic.
- Backup files accumulate without rotation — pre-existing pattern from the old macOS script, not a regression.

Filed #1097 separately for an unrelated issue I noticed about unrendered content collection bodies in `clients/*.md` etc. — not your concern, just noting it exists.

Once the reshape and the must-fixes are addressed we'll merge. Happy to discuss any of the above if anything's unclear — first-time-contributor-friendly, ask anything.

kingpanther13:
Nice progress this round — solid engineering on the wrappers and CI:

**Addressed:**

- ✅ **Wrappers** (`install-linux.sh` and `install-macos.sh`) — the `mktemp` + `curl -o $TMPFILE` + `sh "$TMPFILE"` pattern with `trap` cleanup is exactly the right shape. Curl failure now aborts the wrapper before `sh` runs, and the temp file is cleaned up. Good fix.
- ✅ **CI workflow** — paths filter now watches `install.sh` and `install-linux.sh`, `test-macos` invokes `install.sh` directly, and the new `test-linux: runs-on: ubuntu-latest` job mirrors test-macos. All three CI asks are in.
- ✅ **Install-script end-message** now points at `aaddrick/claude-desktop-debian` for Linux users — partial address of the reshape (gives the user who's already running the script a real destination instead of a dead link).

**Still outstanding:**

Headline reshape:
- README quick-start step 3 and `guide-linux.astro` step 3 still send Linux users to `claude.ai/download` (the "Not available for Linux" page). The install-script message points at aaddrick, but that fires only after the user has already started running the script. The primary entry points (README, guide page) need to either route the Linux flow to Claude Code as the officially-supported path, or front-and-center the aaddrick community alternative with a clear "unofficial" framing.

Silent-failure cluster (still in `install.sh`):
- L124 `except (json.JSONDecodeError, FileNotFoundError): config = {}` — still silently replaces a malformed user config (along with any other MCP servers it had configured). Needs a stderr warning + `$BACKUP_FILE` pointer.
- L141 `with open(file, 'w') as f: json.dump(...)` — non-atomic. Crash mid-write leaves a partial config. Should write to a tmp file then `os.replace(tmp, file)`.
- L169 `... --version > /dev/null 2>&1 || true` — still masks all pre-download failures and prints green "Dependencies cached" unconditionally. Needs an `if/then/else` shape with a yellow degraded message on failure.

Linux edge cases:
- WSL detection not added — script still happily writes `~/.config/Claude/...` when run on WSL2 (where Claude Desktop on Windows reads from `%APPDATA%\Claude\`).
- `python3` preflight not added — still invoked unconditionally at L365; minimal Linux distros (Alpine, headless cloud images) abort with `command not found`.

Doc inconsistencies (in `guide-linux.astro`):
- `--python 3.13` still missing from the manual JSON in step 2 and step 6.
- Manual JSON still uses bare `"command": "uvx"` — fresh-system users will hit "command not found" when Claude Desktop spawns the subprocess.
- `xdg-open` still in step 6 — opens in a browser on most distros.
- Demo URL/creds still duplicated across `install.sh`, README, and guide-linux.astro.

Nits from the original review unchanged (badge color, grid-cols-3, etc.) — those stay optional.

Take your time on these. The pipefail and CI fixes were the trickiest items technically; once the silent-failure cluster and the doc/reshape items land we'll merge. Happy to clarify anything that's unclear.

JohnBarton27:
Sweet, appreciate the feedback - let me work on moving this over to Claude Code for the Linux path and addressing some of the other points there.

kingpanther13:
Hi @JohnBarton27 are you planning on continuing work on this? If not the maintainers can take it over. Just let me know.

JohnBarton27:
Yeah, sorry, busy week! I should have a new version ready tomorrow

kingpanther13:
OK, no worries. And no rush! Just checking in :)

kingpanther13:
Hi @JohnBarton27 just checking back in. If you're not able to complete it that's fine, just let me know and we can complete it for you. You'll still get credit as contributor. The problem is the longer this sits here the more merge conflicts it could get and the harder it'll be to merge cleanly.

kingpanther13 (approved):
Took this over to land it (thanks for the original work, @JohnBarton27 — you're credited as co-author on the commits). What changed since the last round:

- Brought the branch current with master.
- Reshaped the Linux flow: the guide now offers both Claude Code (official, paid) and the community aaddrick Claude Desktop build (free, clearly labeled unofficial), instead of dead-ending at claude.ai/download.
- Hardened install.sh: WSL guard, python3 preflight, atomic config writes, JSON-parse warning, plus the silent-failure fixes from the original review.
- Surfaced Linux + the HA OS add-on on the landing/setup pages.
- Ran a full review pass and addressed the findings (env-var heredoc for the merge, fresh-config error handling/perms), and added installer tests for the new --claude-code and WSL paths.

CI is green across the board. Approving.

Issue #1086, opened 2026-05-01
[FEATURE] Decouple add_source local shape check from pre-existing energy_sources entries


Patch76:
### What would you like to see?

In `_add_source` (PR #1073), the convenience-mode write path goes through `_mutate_atomic`, which builds a `partial_config = {"energy_sources": existing + [source]}` and submits it to `_set_prefs`. `_set_prefs` then runs `_shape_check` on the full combined payload before writing ([`tools_energy.py:574`](https://github.com/homeassistant-ai/ha-mcp/blob/802ac6691c1c0ce0b7680307cb8dca8ebfc19a3c/src/ha_mcp/tools/tools_energy.py#L574)).

This means a **pre-existing entry** must pass the local shape check on every `add_source` call — even though that entry is HA-validated state we should be able to trust.

The asymmetry is currently invisible because the local `_shape_check` for `energy_sources` is a strict subset of HA's voluptuous schema (only `type ∈ {grid, solar, battery, gas}` plus `stat_energy_from` for non-grid types — both also required by HA). So any entry HA accepted will pass our check.

But if `_shape_check` is ever tightened to be stricter than HA in some respect — mandatory `name`, type-specific cost-config rules, unit normalisation, etc. — `add_source` would fail when appending an *unrelated* new source, because a *pre-existing* entry would now fail the stricter local check. The user would see their valid HA config rejected with no insight into why.

The structural fix is to scope `add_source`'s shape check to *just the new source*, not the union with the existing list.

### Why do you need this?

Forward-looking robustness. No reachable bug today (verified live on HA 2026.4 against current `_shape_check`). But the asymmetric coupling means any future tightening of `_shape_check` quietly breaks the ability to append until the user manually fixes their existing config — which is poor UX.

Filed as a forward-looking note from the [PR #1073 self-review](https://github.com/homeassistant-ai/ha-mcp/pull/1073#issuecomment-4360817307) (O2). Not a blocker for that PR.

### Additional context

`_add_source` already does a per-entry shape check on the new source alone before delegating to `_mutate_atomic` ([`tools_energy.py:843`](https://github.com/homeassistant-ai/ha-mcp/blob/802ac6691c1c0ce0b7680307cb8dca8ebfc19a3c/src/ha_mcp/tools/tools_energy.py#L843)) — so the new source is already validated independently. The redundant validation happens later in `_set_prefs`, on the union.

**Maintainer choice required — at least three candidate shapes:**

[...]

PR #1098, opened 2026-05-03
feat: per-top-level-key config_hash for ha_manage_energy_prefs (#1049)

### LINKED ISSUE #1049: [FEATURE] Per-top-level-key hashes for ha_manage_energy_prefs partial updates

A per-top-level-key hash extension to `ha_manage_energy_prefs` (added in #1048 / #1033) that lets agents update a single section of the Energy Dashboard preferences without having to re-send the entire config blob.

Proposed shape:

```python
ha_manage_energy_prefs(
    mode="set",
    config={"device_consumption": [...]},        # only one top-level key
    config_hash={
        "device_consumption": "<hash of current device_consumption>",
    },
)
```

- `config_hash` accepts either a single string (current behaviour — hash of the full prefs) or a dict mapping top-level keys to their individual hashes.
- Dict form → tool re-reads current prefs, checks that each submitted top-level key matches its per-key hash, then submits `save_prefs` with only those keys.
- Current full-blob behaviour stays as default; no breaking change.

### Why do you need this?

The current mitigation for the destructive full-replace semantics of `energy/save_prefs` requires the agent to read the entire prefs object, mutate it, and send it back. That works, but:

- An agent that only wants to add one device has to transport the full `energy_sources`, `device_consumption`, and `device_consumption_water` arrays through its context window on every write. For large setups this is wasteful.
- The hash-check protects against concurrent edits on the whole blob, but an agent modifying `device_consumption` is conceptually unaffected by a concurrent change to `energy_sources` — per-key hashing captures that more precisely and reduces false-positive "modified since last read" rejections.

### Additional context

Noted as a follow-up in the #1048 PR body. The full-blob approach is fine as MVP; per-key is an optimisation that introduces a pattern the rest of the codebase does not currently use, so it needs a separate design pass before implementation.

Related:
- #1033 — original feature request
- #1048 — PR adding the tool with full-blob hashing

Patch76:
## What does this PR do?

Closes #1049 by widening `ha_manage_energy_prefs`'s `config_hash` parameter to accept `str | dict[str, str]`. The dict form locks each submitted top-level key independently — agents can push a partial update (e.g. only `device_consumption`) without re-sending the full prefs blob, and without false-positive `RESOURCE_LOCKED` when an unrelated top-level key was concurrently changed.

- `mode='get'` response gains `config_hash_per_key: dict[str, str]` alongside the existing full-blob `config_hash`. Helper `_compute_per_key_hashes` wraps each top-level value in a single-key dict before hashing, so each hash is bound to its key.
- Dict-form validation is fail-closed: set-equality between `config` keys and `config_hash` keys (intersected with the canonical `_PREFS_TOP_LEVEL_KEYS`); unknown keys are silently dropped on both sides (mirrors the existing `_shape_check` / save-loop semantics). Per-key mismatch surfaces every offending key in the response's top-level `mismatched_keys` (`create_error_response` flattens `context` onto the response root).
- The `str` form is the original full-blob contract — unchanged. Convenience modes still pass `str` through `_mutate_atomic`; the dict path is only reachable from direct `mode='set'` callers.
- Pydantic `Annotated[str | dict[str, str] | None, Field(...)]` matches the existing precedent at `tools_config_dashboards.py:756`.

Touches the same module as the open #1086 (decouple `add_source` shape check) but no line overlap; either can land first.

## Type of change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing
- [x] All automated tests pass (`uv run pytest`) — full unit suite locally: 1467 passed, 1 skipped (`test_numpy_version_constraint` — numpy is a transitive optional dependency, not resolved in this venv), 36.72s
- [x] Code follows style guidelines (`uv run ruff check`) — repo-pinned ruff 0.15.12 clean on changed files

Tests added:

- Unit: `config_hash_per_key` on `get` (configured + No-prefs default), happy-path partial save with per-key hash, mismatch surfaces `mismatched_keys`, set-equality fail-closed (missing key, extra key), unknown-top-level-key drop, response carries fresh per-key hashes after write, str backward-compat.

## Checklist
- [x] I have updated documentation if needed — tool docstring CAVEATS gains a per-key bullet; `_set_prefs` docstring documents both forms.

### DIFF
--- a/src/ha_mcp/tools/tools_energy.py
+++ b/src/ha_mcp/tools/tools_energy.py
@@ -47,8 +47,14 @@ logger = logging.getLogger(__name__)
 
 
 # Top-level keys in the energy prefs payload. Each is an independent
-# full-replace slot in ``energy/save_prefs``.
-_PREFS_TOP_LEVEL_KEYS = (
+# full-replace slot in ``energy/save_prefs``. ``_PrefsKey`` is the
+# corresponding ``Literal`` alias so MCP-wire callers (Pydantic-validated)
+# get typo-rejection at the boundary; runtime guards in `_set_prefs` cover
+# the unit-test path that bypasses Pydantic.
+_PrefsKey = Literal[
+    "energy_sources", "device_consumption", "device_consumption_water"
+]
+_PREFS_TOP_LEVEL_KEYS: tuple[_PrefsKey, ...] = (
     "energy_sources",
     "device_consumption",
     "device_consumption_water",
@@ -73,7 +79,7 @@ def _default_prefs() -> dict[str, Any]:
     }
 
 
-def _compute_per_key_hashes(prefs: dict[str, Any]) -> dict[str, str]:
+def _compute_per_key_hashes(prefs: dict[str, Any]) -> dict[_PrefsKey, str]:
     """Per-top-level-key hashes for partial-update optimistic locking.
 
     Each top-level key is wrapped in its own single-key dict before hashing,
@@ -259,25 +265,15 @@ class EnergyTools:
             ),
         ] = None,
         config_hash: Annotated[
-            str | dict[str, str] | None,
+            str | dict[_PrefsKey, str] | None,
             Field(
                 description=(
-                    "Hash returned by the previous mode='get' call. REQUIRED "
-                    "for mode='set' unless dry_run=True. Rejected if the "
-                    "server-side config has changed since that read — re-read "
-                    "and retry. Two forms accepted: pass the full-blob "
-                    "config_hash (str) to lock against the entire prefs object, "
-                    "or pass the config_hash_per_key dict from a mode='get' "
-                    "response (dict[str, str], keyed by 'energy_sources' / "
-                    "'device_consumption' / 'device_consumption_water') to "
-                    "lock only the top-level keys you are submitting in "
-                    "'config'. The per-key form is fail-closed: every "
-                    "top-level key submitted in 'config' must have a matching "
-                    "entry in the dict, and vice versa. Unknown keys (outside "
-                    "the canonical top-level set) are silently dropped on "
-                    "both sides, mirroring the existing local-shape-check "
-                    "behaviour. Ignored by convenience modes (they read "
-                    "fresh internally)."
+                    "Hash from a previous mode='get' call. REQUIRED for "
+                    "mode='set' unless dry_run=True. Two forms: str (full-"
+                    "blob lock) or dict (per-key lock, taken from the "
+                    "config_hash_per_key field of mode='get'). See the tool "
+                    "docstring for fail-closed semantics. Ignored by "
+                    "convenience modes."
                 ),
                 default=None,
             ),
@@ -403,15 +399,22 @@ class EnergyTools:
           requires a fresh ``config_hash`` for optimistic locking; convenience
           modes hide this entirely.
         - ``config_hash`` accepts both a single ``str`` (full-blob lock) and
-          a ``dict[str, str]`` keyed by top-level keys (per-key lock, taken
-          from the ``config_hash_per_key`` field of the mode='get' response).
-          The per-key form lets an agent submit only the top-level key it
-          wants to change — set-equality between ``config`` keys and dict
-          keys is enforced, so a per-key submission still fully replaces
-          that key's value as the save endpoint requires. Mismatch on any
-          locked key returns ``RESOURCE_LOCKED`` with the offending keys
-          in the response's top-level ``mismatched_keys`` (context fields
-          are flattened to the response root by ``create_error_response``).
+          a ``dict[_PrefsKey, str]`` keyed by top-level keys (per-key lock,
+          taken from the ``config_hash_per_key`` field of the mode='get'
+          response). The per-key form lets an agent submit only the top-
+          level key it wants to change — set-equality between ``config``
+          keys and dict keys is enforced, and any key outside the canonical
+          set (typo, etc.) on either side is rejected with
+          ``VALIDATION_FAILED`` rather than silently dropped (so an empty
+          submission cannot succeed as a no-op). A per-key submission
+          still fully replaces that key's value as the save endpoint
+          requires. Mismatch on any locked key returns ``RESOURCE_LOCKED``
+          with the offending keys in the response's top-level
+          ``mismatched_keys`` (``create_error_response`` flattens the
+          ``context`` dict onto the response root).
+        - ``dry_run=True`` skips the hash check entirely for both forms;
+          the per-key form is therefore silently accepted on dry runs even
+          if its keys would mismatch the current state.
         - A local shape check runs before every write; malformed payloads
           are rejected with a ``shape_errors`` list.
         - After a successful write, the tool calls ``energy/validate`` and
@@ -635,13 +638,12 @@ class EnergyTools:
 
         ``config_hash`` accepts two forms. A ``str`` locks against the
         full prefs blob (the original optimistic-locking contract). A
-        ``dict[str, str]`` keyed by top-level keys locks each submitted
-        key individually — set-equality is enforced between the
-        ``config``-present top-level keys and the dict's top-level keys
-        (unknown keys silently dropped on both sides). Per-key mismatch
-        surfaces every offending key in the response's top-level
-        ``mismatched_keys`` (``create_error_response`` flattens the
-        ``context`` dict onto the response root).
+        ``dict[_PrefsKey, str]`` keyed by top-level keys locks each
+        submitted key individually — set-equality between ``config`` and
+        dict keys is enforced, and unknown keys on either side are
+        rejected (``VALIDATION_FAILED``) so an empty submission cannot
+        coincide as a no-op success. See the tool docstring for the full
+        agent-facing contract.
 
         ``current_prefs`` is an optional caller-supplied snapshot. When
         provided, the internal re-read is skipped — the convenience-mode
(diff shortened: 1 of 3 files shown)


kingpanther13 (changes requested):
## Critical (must fix before merge)

**Silent no-op "success" when typo'd canonical keys appear in both `config` AND `config_hash`** — `tools_energy.py:706-711, 779-784`. Trace:

1. `_shape_check` skips unknown keys (`:153-155`), passes.
2. Per-key branch filters both `submitted_keys` and `hashed_keys` through `_PREFS_TOP_LEVEL_KEYS` → both become `∅` → set-equality holds, no error.
3. `mismatched_keys = []` over empty set → no `RESOURCE_LOCKED`.
4. Save loop builds `{"type": "energy/save_prefs"}` with no top-level keys → HA Core accepts as no-op.
5. Returns `{"success": True, "message": "Energy prefs updated."}` with a fresh hash.

The agent's intent is silently discarded. The same coincidence trap exists for the legacy `str` form (less likely to coincide, since the full-blob hash usually mismatches), but the dict form makes it cheap to hit. This directly contradicts the docstring's "fail closed" claim at `:411-414`.

Two valid fixes — pick one:

- Guard at the dict branch: reject any key in `config` or `config_hash` outside `_PREFS_TOP_LEVEL_KEYS` with `VALIDATION_FAILED`. Belt-and-braces: also reject when `submitted_keys` is empty.
- **Better:** tighten the type to `dict[Literal["energy_sources", "device_consumption", "device_consumption_water"], str]`. Pydantic emits this as a JSON-Schema `enum` on property names — typos are rejected at the FastMCP boundary, the silent-drop loops at `:707, :710` become dead code, and the docstring's "unknown keys silently dropped" carve-out can be removed.

## Important (should fix)

- **Multi-key stale mismatch never asserted as a list** — `test_per_key_mismatch_lists_offending_keys` (`test_tools_energy.py:737-768`) only proves *one* offending key surfaces. A regression that replaces the list comprehension at `tools_energy.py:738-743` with `next(...)` would pass tests. Add a test with two stale keys; assert the full sorted list.
- **`dry_run=True` + dict-form hash is undocumented and untested** — `tools_energy.py:474-475` short-circuits before the hash check, so the dict form is silently accepted on dry runs. Document in the docstring at `:285-303` and pin with one test.
- **Convenience-mode "always passes str" contract untested** — docstring at `:651` claims this; no test pins it. A regression that forwards `config_hash_per_key` from `_mutate_atomic` would still functionally pass. Cheapest pin: monkeypatch `_set_prefs` from one convenience-mode test and assert `isinstance(call.args[1], str)`.

## Suggestions (nice-to-have)

- **`Field(description=...)` is ~115 words** at `:264-281`, restating the docstring at `:405-414`. Trim to ~45 words and defer to the docstring.
- **"Context flattening" caveat appears twice** in docstrings (`:412-414`, `:642-644`) — keep one.
- **Promote `_PREFS_TOP_LEVEL_KEYS` to a `Literal` alias** (`_PrefsKey`) and reuse it for the tuple, the parameter, and `_compute_per_key_hashes`'s return type. Single source of truth, and pairs naturally with the type-tightening fix above.
- **No E2E test for the dict form.** `AGENTS.md` test-pyramid expects E2E for new wire shapes. One round-trip (get → partial set with one key + that key's hash) would be enough.

## Strengths

- Backward compat verified byte-identical: the `str` branch is the existing logic moved into `else:` at `:762-776`.
- `test_response_includes_fresh_per_key_hashes_after_write` (`test_tools_energy.py:854-891`) pins the response chaining contract end-to-end.
- `66740c1` flattening fixup is correct — `errors.py:256-258` does `response.update(context)`.
- Convenience modes verified to never reach the dict branch (`_mutate_atomic` at `:1198, 1207` reads `current["config_hash"]` which is always `str`).
- No version bumps — release automation will handle on merge.
- Pydantic union precedent at `tools_config_dashboards.py:756` confirmed.

Patch76:
## Implementation Summary

**Choices Made:**
- **Pydantic Union for `config_hash`**, not a separate parameter — single `str | dict[_PrefsKey, str] | None` matches the in-repo precedent at `tools_config_dashboards.py:756`. Avoids a mutually-exclusive-parameter dance.
- **Per-key hash wraps each value in a single-key dict before hashing** — `compute_config_hash({key: prefs[key]})` in `_compute_per_key_hashes`. Hash carries the key name, so an `energy_sources` hash cannot authorise a `device_consumption` write even on byte-identical (empty-list) values.
- **`config_hash_per_key` echoed on every response** (mode='get' and mode='set') — agents can chain partial writes without an intermediate get.
- **Convenience modes never reach the dict-form code path** — `_mutate_atomic` reads `current["config_hash"]` (always `str` from `_get_prefs`) and forwards that to `_set_prefs`. The dict path is reachable only from direct mode='set'.
- **`_PrefsKey = Literal[...]` alias** (KP13 Suggestion #3, adopted) — single source of truth for the canonical top-level keys; reused in `_PREFS_TOP_LEVEL_KEYS`'s tuple type, the `config_hash` parameter type, and `_compute_per_key_hashes`'s return type. Pydantic emits the `Literal` keys as a JSON-Schema enum, giving MCP-wire callers typo-rejection at the boundary.
- **Belt-and-braces runtime guards** (KP13 Critical, adopted) — even with the typed boundary, the dict-branch in `_set_prefs` rejects unknown keys and empty `submitted_keys` so unit-test callers (which bypass Pydantic) cannot trigger the silent no-op trap. `tools_energy.py:702-746`.
- **E2E round-trip for the dict form** (KP13 Suggestion #4, adopted) — `test_energy_prefs_per_key_config_hash_roundtrip`. Per AGENTS.md test-pyramid for new wire shapes.
- **Field description trimmed to ~38 words** (KP13 Suggestion #1, adopted) — defers to the tool docstring, removes restated content.
- **Single context-flattening caveat** (KP13 Suggestion #2, adopted) — kept the agent-facing one in the tool docstring; removed the duplicate from `_set_prefs`'s internal docstring.

**Problems Encountered:**
- **Silent no-op trap on typo'd canonical keys** (KP13 Critical) — `_shape_check`'s permissive-drop semantics combined with the dict-branch's `set & _PREFS_TOP_LEVEL_KEYS` filter let `submitted_keys=∅, hashed_keys=∅` slip through set-equality. HA Core then accepted the empty `save_prefs` payload as a no-op. Addressed in `2dfedd2` by explicit unknown-key rejection plus an empty-`submitted_keys` guard.
- **Test assertions read `err["context"]["mismatched_keys"]`** when `errors.py:258`'s `response.update(context)` flattens those onto the response root. Caught pre-push; addressed in `66740c1` along with the misleading `context.mismatched_keys` docstring references at the same paths.
- **Multi-key mismatch test only proved one key surfaces** (KP13 Important #1) — addressed in `2dfedd2` with `test_per_key_mismatch_lists_all_offending_keys_sorted` constructing two stale keys and asserting the sorted list.
- **Convenience-mode str-only contract was untested** (KP13 Important #3) — addressed in `2dfedd2` with three `monkeypatch`-spy tests (one per convenience mode) asserting `type(config_hash).__name__ == "str"`.
- **`dry_run=True` + dict-form was undocumented** (KP13 Important #2) — addressed in `2dfedd2` with a docstring bullet and one wiring test.
- **Mypy: `dict[_PrefsKey, str]` not assignable to internal `dict[str, str]`** — dict is invariant in its key type, so the narrower `_PrefsKey`-keyed dict is not a subtype of the `str`-keyed one. Caught by CI on `2dfedd2`; addressed in `e0d9359` by tightening `_set_prefs`'s internal signature to match the public param plus a `cast(_PrefsKey, key)` at the mismatch-loop subscript where mypy cannot narrow `str → _PrefsKey` from the runtime guard.

**Suggested Improvements:**
- None — all review items addressed inline; no follow-up issues filed.

Local pre-push validation: full ha-mcp unit suite 1467 passed, 1 skipped (`test_numpy_version_constraint` — numpy is a transitive optional dependency, not resolved in this venv), 36.72s on Python 3.13.13.

kingpanther13 (approved):
Verified `2dfedd2` against the prior review — every Critical, Important, and Suggestion item addressed substantively:

- Silent no-op trap closed three ways (`tools_energy.py:708-746`): unknown-key guard, empty-`submitted_keys` guard, strict set-equality on raw sets.
- Type tightened to `dict[_PrefsKey, str]` at the Pydantic surface; `_PrefsKey` Literal alias promoted as the single source of truth.
- Multi-key sorted mismatch test, `dry_run` + dict-form documented & pinned, convenience-mode `str`-hash contract pinned with monkeypatch spies for all three modes.
- Field description trimmed, duplicate flattening caveat removed, E2E round-trip added (and passing in CI).

Only outstanding blocker is the Mypy failure at `tools_energy.py:493` — `_set_prefs`'s parameter type still says `str | dict[str, str]` while the public surface now passes `str | dict[_PrefsKey, str]` (`dict` is invariant in its key type). One-line fix on the helper signature.

kingpanther13:
Disregard the latest comment about mypy, you alreadty fixed that. CLaude was being stupid. This looks good to merge now.

Patch76:
@kingpanther13 The mypy issue at `tools_energy.py:493` was addressed in `e0d9359` — the very commit your review is pinned to. `_set_prefs`'s param signature is now `str | dict[_PrefsKey, str]` and the mismatch-loop subscript carries `cast(_PrefsKey, key)` since mypy cannot narrow `str → _PrefsKey` from the runtime guard. Mypy CI on this commit is SUCCESS (job [74128540264](https://github.com/homeassistant-ai/ha-mcp/actions/runs/25285046855/job/74128540264)); listed under Problems Encountered in the IS comment. Looks like the review was drafted against the prior `2dfedd2` mypy-failing snapshot before CI on `e0d9359` reported.

Squash-merging once you've confirmed.

kingpanther13:
@Patch76 yes it's fine to merge now

PR #1101, opened 2026-05-03
fix(webhook-proxy): surface webhook registration failures instead of silently loading

### LINKED ISSUE #1020: Webhook Proxy: truncated secret_path from log parser causes silent webhook registration failure (HA returns 404)
## Summary

Two linked defects in the **Webhook Proxy add-on** (`ha_mcp_webhook_proxy` v1.0.1) and the **`mcp_proxy` custom component** (v1.0.0) cause `/api/webhook/<id>` to return HTTP 404 even when:

- Both Apps report `state: started`
- `mcp_proxy` config entry reports `state: loaded`
- `/config/.mcp_proxy_config.json` exists with valid JSON

Symptom: the Nabu Casa webhook URL returns HTTP 404 "Not Found" on both GET and POST. MCP client (claude.ai) cannot connect.

## Environment

- Home Assistant Core: **2026.4.3** on HAOS (qemux86-64)
- ha-mcp App: **v7.3.0** (server prints `ha-mcp, version 7.2.0`)
- Webhook Proxy App: **v1.0.1**
- `mcp_proxy` custom component: **v1.0.0**
- Install source: both Apps from the `homeassistant-ai/ha-mcp` App store repository
- Nabu Casa: active and confirmed routing

## Reproduction

1. Install **ha-mcp** App, start it. Logs print:
   ```
   🔐 MCP Server URL: http://<host-ip>:9583/private_XXXXXXXXXXXXXXXXXXXXXX
      Secret Path: /private_XXXXXXXXXXXXXXXXXXXXXX
   ```
   (22-character secret token after `private_`)
2. Install **Webhook Proxy** App, start it. Restart HA when prompted.
3. After HA restart, the proxy fetches ha-mcp logs via `GET /addons/<slug>/logs` through the Supervisor API and parses them with `re.search(r"(/private_\S+)", logs)`.
4. **Match result is truncated to ~7 chars of the 22-char token.** Example logs:
   ```
   [INFO] Secret path from 81f33d0f_ha_mcp logs (/addons/.../logs): /private_ZZZZZZZ
   [INFO] Auto-discovered MCP server: http://127.0.0.1:9583/private_ZZZZZZZ
   ```
5. `/config/.mcp_proxy_config.json` written with truncated `target_url`:
   ```json
   {"target_url": "http://127.0.0.1:9583/private_ZZZZZZZ",
    "webhook_id": "mcp_[REDACTED]"}
   ```
6. The `mcp_proxy` integration's `async_setup_entry` runs, returns `True`, entry shows `state: loaded`.
7. `POST /api/webhook/<webhook_id>` on HA directly (not via Nabu Casa) returns **HTTP 404 "Not Found"**.

Reproduced across 3 fresh App restarts — **consistent truncation**, not transient.

### Defect 1 — `_discover_secret_path()` parser truncation

`homeassistant-addon-webhook-proxy/start.py`:

```python
match = re.search(r"(/private_\S+)", logs)
```

`\S+` is greedy to whitespace, but the Supervisor `/addons/<slug>/logs` endpoint apparently returns output with embedded control characters, chunk boundaries, or non-printable bytes that terminate the `\S` match early. Reproducible — always truncates to the same length.

### Defect 2 — silent webhook registration failure while `state: loaded`

With a truncated `target_url`, `async_setup_entry` in `custom_components/mcp_proxy/__init__.py` proceeds past the `if not target_url or not webhook_id` guard (truncated value is still truthy) to `async_register()`. From the outside:

- `GET /api/config/config_entries/entry` returns `state: "loaded"` for the `mcp_proxy` entry
- `POST /api/webhook/<webhook_id>` returns **HTTP 404** from HA core
- `home-assistant.log` contains **no** `MCP Proxy: webhook endpoint = ...` line despite the integration emitting `_LOGGER.info(...)` at that point — suggesting either the log emission is suppressed or a raised exception is swallowed before reaching `async_register`

## Fix verified

Manually rewriting `/config/.mcp_proxy_config.json` with the full 22-char secret path, then reloading the config entry via `POST /api/config/config_entries/entry/<id>/reload`, immediately unblocks both:

- Direct LAN POST to `/api/webhook/<webhook_id>` returns **HTTP 200** with a valid MCP `initialize` response identifying `ha-mcp, version 7.2.0`
- Nabu Casa webhook URL returns **HTTP 200** end-to-end with the same MCP response

This confirms **Defect 1 is the root cause**; **Defect 2 is the silent-failure bug that masked it** — without Defect 2, the truncated URL would have produced a visible error message in the integration, not a 404.

## Workaround

Set the `mcp_server_url` App option to the full URL (including the full secret path). `_discover_secret_path()` checks App options before parsing logs — this bypasses the parser.

## Requested fixes

1. **Parser hardening** — anchor/sanity-check the secret path extraction. Options:
   - Add a length assertion (expected secret is ~32 chars base64url-like), log WARNING and retry if shorter
   - Stream logs via a different endpoint / with different Accept headers that don't include control bytes
   - Have ha-mcp expose the secret path via a Supervisor-discoverable service/metadata endpoint instead of log scraping
2. **Registration failure surfacing** — in `async_setup_entry`, wrap `async_register()` in `try/except`, emit `_LOGGER.error` with the failure reason, and set config entry state to `error` with `reason` populated so the UI and API both reveal the problem instead of silently claiming `loaded`.

## Additional note

Separately, the Webhook Proxy App's logs include the line `The use of 'addons' is deprecated, please use 'apps' instead!` — consistent with HA renaming add-ons → Apps in 2026.2. The proxy's own log text still says "addon" in several places; worth a sweep in a follow-up.

kingpanther13:
## What does this PR do?

Closes Defect 2 of #1020 — the **mcp_proxy** custom integration would happily report `state: loaded` while a corrupted `target_url` caused every webhook request to return 404. The integration now refuses to load with a populated `reason` so the failure is visible in the UI and config-entries API.

Two changes in `homeassistant-addon-webhook-proxy/mcp_proxy/__init__.py`:

1. **Validate `target_url` shape before registering.** The path must match `/private_<token>` with `token >= 16 chars` (real ha-mcp generates a 22-char base64url token; the issue's truncation bug yielded ~7). On failure, raise `ConfigEntryError(reason)`.
2. **Wrap `async_register()` in `try/except`.** On failure, close the aiohttp session and re-raise as `ConfigEntryError` with the original error chained. `hass.data[DOMAIN]` is now assigned only after a successful register so a failed setup doesn't leave half-state.

The empty-string guard for `target_url`/`webhook_id` was also upgraded from `return False` to a `ConfigEntryError` with a populated reason for the same visibility reason.

> Defect 1 (the parser truncation itself) is being addressed in #952 by persisting the secret path to addon options. Defect 2 — the silent-failure bug that masked it — is what this PR fixes, per the maintainer's note on the issue.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing
- [x] All automated tests pass (`uv run pytest tests/addon/test_webhook_proxy.py` — 49 passed; 33 existing + 16 new)
- [x] Code follows style guidelines (`uv run ruff check` clean; `uv run ast-grep scan` clean)

**New test coverage** (`tests/addon/test_webhook_proxy.py`):

- `TestTargetUrlValidation` (10 cases): real 22-char token, 16-char minimum, https, the literal issue reproducer `/private_ZZZZZZZ`, missing path, missing host, non-http scheme, empty string, invalid chars in token, root path.
- `TestSetupEntrySurfaceFailures` (6 cases): truncated URL → `ConfigEntryError` raised + `async_register` not called + no leaked `hass.data`; missing `target_url` / missing `webhook_id` paths; `async_register` raising → session closed + error chained from original cause; happy path; no-config-file path returns `True` for fresh installs.

The new tests stub `homeassistant.*` and `aiohttp` via `sys.modules` so the integration imports under the existing pytest harness without pulling HA Core into dev deps — same approach the existing `_import_start()` helper uses.

## Checklist
- [x] I have updated documentation if needed (no user-facing docs change needed; the fix is internal error-handling behavior)

### DIFF
--- a/homeassistant-addon-webhook-proxy/config.yaml
+++ b/homeassistant-addon-webhook-proxy/config.yaml
@@ -1,6 +1,6 @@
 name: "Nabu Casa / Webhook Proxy for HA MCP"
 description: "Remote access proxy via Nabu Casa or any reverse proxy (Cloudflare, DuckDNS, nginx)"
-version: "1.0.1"
+version: "1.0.2"
 slug: "ha_mcp_webhook_proxy"
 url: "https://github.com/homeassistant-ai/ha-mcp"
 arch:
--- a/homeassistant-addon-webhook-proxy/mcp_proxy/__init__.py
+++ b/homeassistant-addon-webhook-proxy/mcp_proxy/__init__.py
@@ -33,15 +33,18 @@ DOMAIN = "mcp_proxy"
 CONFIG_FILE = Path("/config/.mcp_proxy_config.json")
 
 # ha-mcp generates a 22-char base64url token after `/private_`. We accept >=16
-# as a sanity floor — a truncated/corrupted config that yields a shorter token
-# is the failure mode this validator exists to catch.
+# as a sanity floor — a truncated/corrupted ha-mcp config yields a shorter
+# token, which is the failure mode this length check exists to catch.
 _SECRET_PATH_RE = re.compile(r"^/private_[A-Za-z0-9_-]{16,}$")
 
 
 def _validate_target_url(target_url: str) -> tuple[bool, str]:
-    """Check that target_url has the shape we expect.
+    """Check that target_url is a well-formed http(s) URL.
 
-    Returns (is_valid, reason).
+    When the path starts with `/private_` we additionally enforce the
+    ha-mcp secret-path shape so a truncated token (the issue we're guarding
+    against) is rejected. Other paths are accepted as-is — users with a
+    custom MCP server pointed at a different path are not constrained.
     """
     parsed = urlparse(target_url)
 
@@ -51,10 +54,10 @@ def _validate_target_url(target_url: str) -> tuple[bool, str]:
         return False, "URL is missing host"
     if parsed.params or parsed.query or parsed.fragment:
         return False, "URL must not contain query, fragment, or path parameters"
-    if not _SECRET_PATH_RE.match(parsed.path):
-        # Don't echo parsed.path — it contains the secret token.
+    if parsed.path.startswith("/private_") and not _SECRET_PATH_RE.match(parsed.path):
+        # Don't echo parsed.path — it contains the (truncated) secret token.
         return False, (
-            "path does not look like a valid secret path "
+            "secret path is too short or malformed "
             "(expected /private_<token> with token of at least 16 characters)"
         )
     return True, ""
--- a/homeassistant-addon-webhook-proxy/mcp_proxy/__init__.py
+++ b/homeassistant-addon-webhook-proxy/mcp_proxy/__init__.py
@@ -12,7 +12,9 @@ addon creates the config entry automatically via the HA API.
 
 import json
 import logging
+import re
 from pathlib import Path
+from urllib.parse import urlparse
 
 import aiohttp
 from aiohttp import web
@@ -22,12 +24,42 @@ from homeassistant.components.webhook import (
 )
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.core import HomeAssistant
+from homeassistant.exceptions import ConfigEntryError
 from homeassistant.helpers.typing import ConfigType
 
 _LOGGER = logging.getLogger(__name__)
 
 DOMAIN = "mcp_proxy"
 CONFIG_FILE = Path("/config/.mcp_proxy_config.json")
+
+# Real ha-mcp generates a 22-char base64url token after `/private_`. A shorter
+# token indicates a truncated/corrupted config (issue #1020 defect 1).
+_SECRET_PATH_RE = re.compile(r"^/private_[A-Za-z0-9_-]{16,}$")
+
+
+def _validate_target_url(target_url: str) -> tuple[bool, str]:
+    """Check that target_url has the shape we expect.
+
+    Returns (is_valid, reason). The URL must be http(s)://host with a path
+    matching `/private_<token>` where the token is at least 16 characters.
+    Catches the issue #1020 truncation, where the parsed secret_path was
+    ~7 characters instead of the expected 22.
+    """
+    try:
+        parsed = urlparse(target_url)
+    except ValueError as err:
+        return False, f"unparseable URL ({err})"
+
+    if parsed.scheme not in ("http", "https"):
+        return False, f"scheme must be http or https, got {parsed.scheme!r}"
+    if not parsed.netloc:
+        return False, "URL is missing host"
+    if not _SECRET_PATH_RE.match(parsed.path):
+        return False, (
+            f"path {parsed.path!r} does not look like a valid secret path "
+            "(expected /private_<token> with token of at least 16 characters)"
+        )
+    return True, ""
 
 
 async def async_setup(hass: HomeAssistant, config: ConfigType) -> bool:
@@ -65,7 +97,10 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
 
     if not target_url or not webhook_id:
         _LOGGER.error("MCP Proxy: Invalid config - missing target_url or webhook_id")
-        return False
+        raise ConfigEntryError(
+            "Missing target_url or webhook_id in /config/.mcp_proxy_config.json. "
+            "Restart the Webhook Proxy addon to regenerate it."
+        )
 
     # Mask sensitive values in logs to avoid leaking secrets
     if "/private_" in target_url:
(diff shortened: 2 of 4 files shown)


Patch76 (approved):
Solid Defect-2 fix. The narrowed `/private_*`-only length check (commit 3) keeps the validator from constraining custom MCP servers while still catching the truncated-token reproducer from #1020.

Things specifically worth keeping in the test suite:

- `test_truncated_url_does_not_log_full_token` — caplog regression guard against the masking ever silently breaking again.
- The parametrized `RuntimeError/ValueError/KeyError` on `async_register` documents *why* the broad `except Exception` is intentional — the test prevents future "narrow this catch" cleanups from removing the safety net.
- `_read_config` now distinguishes "no file" (returns None) from "corrupted file" (raises). That was a real silent-failure path previously, not just defensive coding.

`mock_session.assert_not_called()` on the validation and missing-field paths is the right move too — keeps a future "early connect probing" refactor from silently leaking sessions on those paths.

Patch76 (changes requested):
Reverting to changes-requested. The Implementation Summary is still missing on this PR — per the workflow it's required before marking Ready and before merge. Please add it as a comment with the standard sections (Choices Made / Problems Encountered / Suggested Improvements). Will switch back to approve once it's posted. Technical review above stands.

kingpanther13:
## Implementation Summary

**Choices Made:**
- **Scope:** Defect 2 of #1020 only — the silent-failure / "loaded with broken target_url" behavior. Defect 1 (the parser truncation that produced the corrupted `target_url` in the first place) is being addressed in #952 by persisting the secret path to addon options. Split per the maintainer's note on the issue.
- **`ConfigEntryError` over `return False`:** surfaces the failure `reason` in the HA UI and config-entries API. `return False` looked like a generic load failure with no diagnostics; `ConfigEntryError("…validator reason…")` makes the cause visible.
- **Validator shape:** regex `/private_<token>` with `token >= 16` chars. Real ha-mcp tokens are 22 chars (base64url); 16 is a sanity floor that catches the issue's 7-char truncation reproducer (`/private_ZZZZZZZ`) without false-positiving. **Commit 3 (a34d4dd) narrowed the length enforcement to `/private_*` paths only** so custom MCP servers with other path shapes aren't constrained.
- **`_read_config` propagates `OSError`/`JSONDecodeError`:** `async_setup_entry` now distinguishes "no file (fresh install)" from "broken file (raise `ConfigEntryError`)" rather than silently collapsing both into "treat as fresh install" — that collapse was a real silent-failure path, not just defensive cleanup.
- **Validator also rejects URLs with query strings, fragments, path parameters:** `urlparse` strips these into other fields, so `/private_<token>?x=y` was validating clean but the addon would 404 on the forwarded request.
- **Validator failure reasons no longer echo `parsed.path` verbatim:** was leaking the raw secret token into ERROR-level logs even when the caller's `masked_target` was correct. `test_truncated_url_does_not_log_full_token` is the regression guard.
- **Test stubs via `sys.modules` for `homeassistant.*` and `aiohttp`:** keeps HA Core out of dev deps, matches the existing `_import_start()` helper pattern.
- **Version alignment:** bumped both the addon (`config.yaml`: 1.0.1 → 1.0.2) and the bundled `mcp_proxy` integration (`manifest.json`: 1.0.0 → 1.0.2). They always ship together but had diverged because an earlier addon-only bump didn't touch the integration.

**Problems Encountered:**
- First-pass validator enforced `>=16` length on every path, which would have rejected legitimate custom MCP server URLs. Caught in review and narrowed to `/private_*` paths only (commit 3).
- Comment-vs-regex mismatch flagged in review: comment said "22-char token" but regex enforced `>=16`. Reworded to "real tokens are 22 chars; we accept >=16 as a sanity floor" to match what the code does.
- Dead `try/except ValueError` around `urlparse` — `urlparse` doesn't raise on any string input. Removed.
- `async_unload_entry` was untested. Boy-scout added two cases pinning that unload-after-failed-setup is a no-op (no `async_unregister`, no double-close) and unload-after-success unregisters and closes the session.
- The parametrized `RuntimeError`/`ValueError`/`KeyError` test on `async_register` documents *why* the broad `except Exception` is intentional — prevents a future "narrow this catch" cleanup from removing the safety net.

**Suggested Improvements:**
- None outstanding. Defect 1 follow-up is tracked in #952; this PR is complete on its own.

PR #1104, opened 2026-05-03
feat: add HA_VERIFY_SSL toggle to disable TLS verification

### LINKED ISSUE #1018: [FEATURE] Support mTLS connections

A way to either:

1. Specify a certificate file to use when connecting to the Home Assistant instance
2. Disable SSL validation

### Why do you need this?

My Home Assistant instance doesn't allow http connections. It's available via https://homeassistant.local:8123 or a domain managed by Cloudflare with mTLS enabled.

I can't seem to find any way to establish a connection with the mcp server. I've tried:

1. Going straight to https://homeassistant.local:8123. This fails because the SSL certificate doesn't match.
2. Tried adding the necessary certificates to my Windows certificate store and connecting to the public domain. This fails as well, apparently it isn't using the certificates
3. Disabling SSL validation by setting `"PYTHONHTTPSVERIFY": "0"` in the `claude_desktop_config.json` MCP server entry and going to the local address (Claude suggested this). This fails with an SSL validation error again.

### Additional context

I read the [FAQ](https://homeassistant-ai.github.io/ha-mcp/faq/ )entry for self-signed certificates, but that doesn't apply to me since I'm not using Docker (HA runs in a VM), and I'm not using a self-signed certificate.

kingpanther13:
## What does this PR do?

Closes #1018.

Adds a single `HA_VERIFY_SSL` boolean to disable TLS certificate verification on the REST + WebSocket clients. Some users (e.g. the reporter on #1018) connect to HA over HTTPS through a hostname that doesn't match the cert (`https://homeassistant.local:8123` with a Cloudflare-issued public cert, self-signed certs, etc.). `PYTHONHTTPSVERIFY=0` is not honored by `httpx` or `websockets`, so the only working fix is to plumb `verify=` and an `ssl.SSLContext` through the clients.

The toggle is exposed three ways so every install path covers it:

- **HA OS add-on** — new `verify_ssl` option (stable + dev `config.yaml` schemas, with translations + DOCS.md entries), wired through `homeassistant-addon/start.py` to the env var, sitting next to `enable_skills_as_tools` / `enable_tool_search` like the issue requested.
- **Docker / pip / pypi** — `HA_VERIFY_SSL=true|false` env var on `Settings`, documented in `.env.example`.
- **Programmatic** — `HomeAssistantClient(..., verify_ssl=False)` and `HomeAssistantWebSocketClient(..., verify_ssl=False)` for callers that construct clients directly.

When the toggle is off, both clients log a WARNING on every connect so the weakened state is visible in logs. `wss://` URLs get an `SSLContext` with `check_hostname=False` and `verify_mode=CERT_NONE`; plain `ws://` (Supervisor proxy) keeps getting `ssl=None` so we don't accidentally force TLS where there isn't any.

Default is `True` — nobody's existing setup changes unless they opt in.

## Type of change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

New unit tests in `tests/src/unit/test_ssl_verification.py` cover:

- `HA_VERIFY_SSL` env-var parsing (truthy/falsey/default)
- `httpx.AsyncClient` receives the right `verify=` value (default, explicit, env-driven)
- WARNING is logged when the toggle is off
- WebSocket `wss://` builds an `SSLContext` with the expected `verify_mode` / `check_hostname` for both states
- Plain `ws://` URLs do **not** get an SSL context attached
- `HomeAssistantWebSocketClient` falls back to `Settings.verify_ssl` when no kwarg is passed

19 new tests pass; the existing `test_rest_client_*`, `test_websocket_client`, and the broader unit suite are unaffected (1473 passing on my box; the 3 pre-existing `test_resources.py` failures are an uninitialized `skills-vendor/` submodule on this checkout, unrelated).

`uv run mypy src/ha_mcp/{config,client/rest_client,client/websocket_client}.py` is clean.

## Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -13,9 +13,8 @@ HA_TIMEOUT=30
 HA_MAX_RETRIES=3
 
 # Verify the Home Assistant server's TLS certificate. Set to "false" to
-# accept self-signed certificates or hostname mismatches (e.g. when
-# connecting to https://homeassistant.local:8123 from a LAN). Disabling
-# this weakens transport security — only use it on networks you trust.
+# accept self-signed certs or hostname mismatches (e.g. https://homeassistant.local:8123).
+# Trusted networks only.
 # HA_VERIFY_SSL=true
 
 # Tool Configuration
--- a/.env.example
+++ b/.env.example
@@ -12,6 +12,12 @@ HOMEASSISTANT_TOKEN=your_long_lived_access_token_here
 HA_TIMEOUT=30
 HA_MAX_RETRIES=3
 
+# Verify the Home Assistant server's TLS certificate. Set to "false" to
+# accept self-signed certificates or hostname mismatches (e.g. when
+# connecting to https://homeassistant.local:8123 from a LAN). Disabling
+# this weakens transport security — only use it on networks you trust.
+# HA_VERIFY_SSL=true
+
 # Tool Configuration
 FUZZY_THRESHOLD=60
 ENTITY_SEARCH_LIMIT=20
--- a/homeassistant-addon-dev/DOCS.md
+++ b/homeassistant-addon-dev/DOCS.md
@@ -23,6 +23,7 @@ The dev add-on uses the same configuration as the stable version. See the main a
 | `tool_search_max_results` | Max results from `ha_search_tools` (range 2-10) | `5` |
 | `disabled_tools` | Comma-separated list of tool names to disable (seed value; web UI is primary) | empty |
 | `pinned_tools` | Comma-separated list of tool names to pin when tool search is enabled (seed value; web UI is primary) | empty |
+| `verify_ssl` | Verify the HA server's TLS certificate. Disable for self-signed certs or hostname mismatches. Weakens security — leave on unless needed. | `true` |
 
 Beta options are hidden under "Show unused optional configuration options" in the add-on Configuration tab. See [beta.md](https://github.com/homeassistant-ai/ha-mcp/blob/master/docs/beta.md) for details.
 
--- a/homeassistant-addon-dev/config.yaml
+++ b/homeassistant-addon-dev/config.yaml
@@ -27,6 +27,7 @@ options:
   tool_search_max_results: 5
   disabled_tools: ""
   pinned_tools: ""
+  verify_ssl: true
 schema:
   backup_hint: list(strong|normal|weak|auto)
   secret_path: str?
@@ -39,5 +40,6 @@ schema:
   tool_search_max_results: int(2,10)?
   disabled_tools: str?
   pinned_tools: str?
+  verify_ssl: bool?
 ports:
   9583/tcp: 9583
--- a/homeassistant-addon-dev/translations/en.yaml
+++ b/homeassistant-addon-dev/translations/en.yaml
@@ -75,3 +75,13 @@ configuration:
       For a visual interface, click "Open Web UI" on the addon info page. This field
       seeds the initial config when the web UI hasn't been used yet.
       Requires restart.
+  verify_ssl:
+    name: Verify TLS certificate
+    description: >-
+      Verify the Home Assistant server's TLS certificate. The add-on
+      connects via the Supervisor proxy, so this normally has no effect
+      and should stay enabled. Disable only if you've reconfigured the
+      add-on to talk to HA over a public HTTPS hostname with a self-signed
+      certificate or hostname mismatch. Disabling weakens transport
+      security — leave on unless you know you need it. Requires restart
+      to take effect.
--- a/homeassistant-addon/DOCS.md
+++ b/homeassistant-addon/DOCS.md
@@ -233,11 +233,11 @@ Custom secret path override. **Leave empty for auto-generation** (recommended).
 
 Verify the Home Assistant server's TLS certificate.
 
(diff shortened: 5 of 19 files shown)


kingpanther13:
## Review-driven follow-up (commit bdab7b6)

Ran the four-agent pr-review-toolkit pass on Opus and implemented every actionable finding.

**Code:**
- Forward `verify_ssl` through `helpers.get_connected_ws_client` to all 5 callers (backup, addons, history, system, traces). Without this, `HomeAssistantClient(verify_ssl=False)` from a programmatic caller silently lost the toggle on any tool that opened its own WebSocket.
- Classify `ssl.SSLError` (and chain-wrapped variants in httpx) on REST and WebSocket connection failures. When verification is on and a TLS failure occurs, the error message now mentions `HA_VERIFY_SSL=false` as a workaround instead of bubbling up a generic "Failed to connect."
- Gate the WebSocket "TLS verification disabled" warning behind a per-client `_warned_verify_disabled` flag so HA reconnect storms don't spam logs.
- Catch unexpected exceptions from `get_global_settings()` in the WebSocket constructor; fall back to `verify_ssl=True` with a logged reason so a malformed env var elsewhere can't silently flip TLS off.
- `start.py`: extracted `resolve_bool_option` helper used for `verify_ssl`, and log the resolved value alongside `backup_hint` so users diagnosing TLS failures can confirm what was applied.
- `config.py`: added `_reset_global_settings()` test seam.

**Comments / docs:**
- Dropped `(True)` / `which is True` parentheticals from constructor docstrings — Field default is the source of truth.
- Rewrote the `wss://` block comment to describe what it actually does (Supervisor proxy uses `ws://` and gets `ssl=None`) instead of the half-true "would force TLS" claim.
- Replaced inaccurate "Cloudflare with mTLS" example in `homeassistant-addon/DOCS.md` with hostname-mismatch wording (`verify_ssl=False` does not address mTLS).
- Switched DOCS warning to `**Note:**` style for consistency with neighboring sections.
- Added a one-line note that OAuth users inherit the server-wide `verify_ssl` setting (no per-user override).
- Tightened `.env.example` and the `Settings` field comment to the surrounding terse voice.

**Tests** (now 33 passing in `test_ssl_verification.py`, 6 new in `test_addon_startup.py`):
- `TestResolveBoolOption` covers missing/explicit/string/int/None values for the new helper.
- `TestPoolFactoryHonorsEnv` asserts the `(url, token)`-positional factory call still picks up `HA_VERIFY_SSL` from env.
- `TestHelperPropagatesVerifySsl` asserts `helpers.get_connected_ws_client` forwards the flag.
- OAuth-path settings-fallback test on the REST client.
- `wss://`-with-path test (proxy setups) confirming the SSL context still attaches.
- "Warning emitted only once across reconnects" test.
- Constructor-safe-default test for the case where settings load raises.
- SSL-error-surfaces-actionable-hint test for the REST client.
- Negative log assertion (no warning when `verify_ssl=True`).
- Replaced wording-based assertions with level-based checks to reduce brittleness.
- Switched to the new `_reset_global_settings()` helper.

**Deliberately deferred** (out of scope for #1018):
- Custom CA bundle support (httpx `verify=` accepts paths; websockets via `SSLContext.load_verify_locations`). Worth a separate feature request — covers the corporate-private-CA case `verify_ssl=False` doesn't address.
- Pre-existing bare `Exception("Authentication failed: ...")` in `websocket_client.py` that should be `HomeAssistantAuthError`. Logged for follow-up.
- Splitting the `start.py` `try/except` per-key — broader refactor outside this PR's blast radius. `verify_ssl` now uses a tested helper, which is the high-value half.

CI on the new commit is green across all 13 real checks (Validate Hotfix Origin and dependabot are `skipping`, as expected).

Patch76 (approved):
The `verify_ssl` plumbing is complete end-to-end: REST + WebSocket constructors, helper-propagation through all 5 ws-callsites (`backup`, `addons`, `history`, `system`, `traces`), env-var fallback, and the `wss://`-only SSLContext branch that leaves the Supervisor `ws://` proxy on `ssl=None`. Default stays `True` everywhere — no behaviour change for existing installs.

Nice details: `resolve_bool_option` rejects string/int values and falls back to the secure default; `_warned_verify_disabled` gates the per-connect WARNING so HA reconnect storms don't flood logs; `_is_ssl_error` walks the `__cause__`/`__context__` chain to catch httpx's wrapped `ssl.SSLError`. The `HA_VERIFY_SSL=false` hint surfaced on TLS failures is exactly the right user-facing affordance.

Test coverage is unusually solid — 429 lines across `test_ssl_verification.py` + `TestResolveBoolOption`, including the wss-with-path proxy case, `ws://` non-attachment, the warning-only-once invariant, and the safe-default-when-settings-load-fails branch.

Translations only exist in `en.yaml` for both addons — no missed locales. The deferred items in the PR body (custom CA file, bare `Exception` in WS auth, splitting `start.py` per-key) are correctly out of scope.

**One scope note before merge:** #1018's title is "Support mTLS connections" and it asks for either cert-file specification (mTLS) or disable-SSL-validation. This PR delivers the second only — the first is on your deferred list as a separate feature request. With \`Closes #1018\` in the body, the squash auto-closes the issue while the mTLS half stays unbuilt. Either switch to \`Refs #1018\` / \`Partially addresses #1018\` and file a follow-up for mTLS, or keep as-is if you consider the toggle "resolved enough" for that issue — your call.

Approving.

PR #1094, opened 2026-05-02
fix(site): resolve client order collisions, anchor OpenCode shape

Patch76:
## What does this PR do?

Implements both follow-up cleanups from the PR #1080 review (`https://github.com/homeassistant-ai/ha-mcp/pull/1080#pullrequestreview-4213401158`).

**1. Client display-order collisions (`site/src/content/clients/*.md`).** Four pairs shared the same `order:` value, so display order fell back to alphabetical tiebreak. Renumbered to a lossless 1–19 sequence so every client has a unique, explicit value. The new sequence preserves today's display order 1:1 — alphabetical tiebreak is codified as the canonical ordering. Concrete delta on the four pairs:

| Client | Old | New |
|---|---|---|
| `windsurf.md` | 5 | 6 |
| `cline.md` | 6 | 7 |
| `copilot-cli.md` | 6 | 8 |
| `claude-ai.md` | 8 | 10 |
| `gemini-cli.md` | 8 | 11 |
| `open-webui.md` | 14 | 17 |
| `opencode.md` | 14 | 18 |

(Plus mechanical bumps to subsequent files to absorb the shifts: `chatgpt 7→9`, `jetbrains 9→12`, `zed 10→13`, `continue 11→14`, `raycast 12→15`, `codex 13→16`, `antigravity 15→19`.)

**2. OpenCode `uvx-stdio` shape — cross-reference anchors.** The shape lives in three places that already agree (`type, command, enabled, environment`); per the review this was flagged as a tracking item, not a fix. Added pointer comments at all three sites so a future contributor who edits one place sees the other two:

- `site/src/content/clients/opencode.md` — HTML comment above the JSON example
- `site/src/pages/setup.astro` — extended the existing inline comment in the `isOpenCode` branch (also documents the deliberate Docker-branch asymmetry: env vars inlined as `-e KEY=VAL` in the command array, top-level `environment` omitted)
- `site/src/data/clients.ts` — comment above the legacy mirror entry's `config` template string

The third review point (`accuracy` field not consumed in `setup.astro`) is intentionally not addressed — the reviewer scoped it explicitly to `setup.astro` and characterised it as a content-collection display value, which matches reality (the field is required by the Zod schema in `site/src/content.config.ts`).

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing
[...]

PR #1105, opened 2026-05-03
docs: address #1094 review nits on OpenCode mirror comments

Patch76:
## What does this PR do?

Two minor wording follow-ups to PR #1094, addressing optional nits from KP13's review (https://github.com/homeassistant-ai/ha-mcp/pull/1094#pullrequestreview-4216474820, marked "neither blocks merge"):

- Drop the hard-coded `"all three"` / `"Keep all three aligned"` phrasing in the OpenCode mirror cross-reference comments at `clients.ts:397`, `opencode.md:29`, and `setup.astro:1338`. The mirror peers are listed by name in each comment; the count adds noise without value.
- Add the Docker-asymmetry explanatory note to `clients.ts` and `opencode.md`, mirroring the existing note at `setup.astro:1339-1340`. Without it, a reader on either of those two files could misread the missing top-level `environment` key in the Docker variant as a bug.

No behaviour change; pure comments in `.ts` / `.md` / `.astro`.

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing
- [ ] I have tested these changes with a LLM agent — N/A, comment-only changes
- [ ] All automated tests pass (`uv run pytest`) — N/A, no Python touched
- [ ] Code follows style guidelines (`uv run ruff check`) — N/A, no Python touched

## Checklist
- [x] I have updated documentation if needed — these are the documentation updates

### DIFF
--- a/site/src/content/clients/opencode.md
+++ b/site/src/content/clients/opencode.md
@@ -26,7 +26,9 @@ OpenCode merges configs from all locations (project overrides global). See [conf
 <!-- The uvx-stdio shape below (type, command, enabled, environment) is mirrored in: -->
 <!--   - site/src/pages/setup.astro (wizard `isOpenCode` branch) -->
 <!--   - site/src/data/clients.ts (legacy mirror entry) -->
-<!-- Keep all three aligned when editing. -->
+<!-- Keep these aligned when editing. -->
+<!-- Note: the Docker variant in setup.astro intentionally omits the top-level -->
+<!-- `environment` key (env vars are inlined as `-e KEY=VAL` in the command array). -->
 
 ```json
 {
--- a/site/src/data/clients.ts
+++ b/site/src/data/clients.ts
@@ -395,7 +395,9 @@ HOMEASSISTANT_TOKEN = "your_long_lived_token"`,
     configLocation: '~/.config/opencode/opencode.json (global) or opencode.json (project)',
     description: 'AI coding agent by Anomaly with native MCP support over stdio and streamable HTTP.',
     // uvx-stdio shape (type, command, enabled, environment) is mirrored in
-    // site/src/content/clients/opencode.md and site/src/pages/setup.astro — keep all three aligned.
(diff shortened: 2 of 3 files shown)


Patch76:
## Implementation Summary

**Choices Made:**
- **Self-approved Mini-PR over tracking-issue or do-nothing** — KP13's review on #1094 ([4216474820](https://github.com/homeassistant-ai/ha-mcp/pull/1094#pullrequestreview-4216474820)) flagged two optional wording nits, both marked "neither blocks merge". Picked the Mini-PR path because the edits are 10 LOC of pure comments in 3 files, the Docker-asymmetry-note nit has real future-reader value (saves a "is this a bug?" pause on `clients.ts` / `opencode.md`), and self-merge fits the maintainers'-guide bar (chore/docs <300 LOC).
- **Drop hardcoded count, keep peer-name list** — replaced `"all three"` / `"Keep all three aligned"` at `clients.ts:397`, `opencode.md:29`, `setup.astro:1338` with `"keep aligned"` / `"Keep these aligned when editing."` (silent on count). The mirror peers are listed by name immediately above each phrase, so the count was redundant.
- **Docker-asymmetry note added to two of three mirror sites** — `setup.astro:1339-1340` already carried the note (added in #1094); replicated to `clients.ts` and `opencode.md`. Future syncing burden is real but minor (3 short comments).

**Problems Encountered:**
- None during implementation.

**Suggested Improvements:**
- None — both KP13 nits addressed inline; no follow-up issues filed.

kingpanther13 (approved):
This is actually a perfect example of when Claude was being lazy about requesting changes, and I was being lazy and missed that it did that... This is the type of thing that should really have been addressed within the same PR.

This isn't criticism, just an observation, I should have caught Claude before allowing it to word it the way it did and before approving the last PR. Imo too many small PRs really just create noise. It is not your fault though! Again I'm just making an observation.

Issue #1121, opened 2026-05-04
[FEATURE] Change ha_get_skill_home_assistant_best_practices to a shorter name


JonLaliberte:
### What would you like to see?

Could the tool name for `ha_get_skill_home_assistant_best_practices` be changed to be 40 chars or less?

I'm using this via Cloudflare's MCP portal and it complains about it:
> One or more tool names exceed 40 characters. Long tool names may cause issues when combined with server and portal names.

Maybe: `ha_get_skill_best_practices` ?

### Why do you need this?

Cloudflare MCP portal

kingpanther13:
Will look into this one, I might be refactoring the skills system a bit as well, so this can be part of that.

kingpanther13:
I'm going to close this issue but it's going to be a part of #1134, so it will get done when I get to #1134

Issue #1116, opened 2026-05-04
[BUG] ha_get_logs(source="supervisor") returns 403 for every slug on add-on installs


kingpanther13:
## 💬 What Happened?

`ha_get_logs(source="supervisor", slug=...)` returns 403 Forbidden for every add-on slug on a HA OS add-on install. The error gets mapped to `AUTH_INVALID_TOKEN` with a misleading "verify add-on slug is correct" suggestion, even when the slug is a known-running add-on. PR #951's commit message says this was verified working at the time, so this is a regression — I haven't bisected what changed in HA Core to break it.

## Reproduction

Tested across two MCP transport pathways (Nabu Casa → webhook proxy, and Cloudflared) and across two ha-mcp versions (stable v7.4.1, near-master fork build). Identical 403 in all combinations:

| Slug | Result |
|---|---|
| `81f33d0f_ha_mcp` (running) | 403 |
| `core_mosquitto` (running) | 403 |
| `core_matter_server` (running) | 403 |
| `5c53de3b_esphome` (running, just successfully accessed via `ha_manage_addon` proxy) | 403 |
| `self` (documented shorthand) | 403 |

Captured via httpx in ha-mcp's own log (`ha_report_issue` startup logs):

```
httpx: GET http://supervisor/core/api/hassio/addons/81f33d0f_ha_mcp/logs "HTTP/1.1 403 Forbidden"
```

## Environment

- ha-mcp Version: 7.4.1
- Installation Method: addon
- Platform: Linux 6.12.77-haos (x86_64)
- Python Version: 3.13.12
- Home Assistant Version: 2026.4.4

## Root cause (in ha-mcp's call chain)

`HomeAssistantClient.get_addon_logs()` in `src/ha_mcp/client/rest_client.py:436-457` (added in #951) hits `/api/hassio/addons/{slug}/logs` against `HOMEASSISTANT_URL=http://supervisor/core`, authenticating with the addon's `SUPERVISOR_TOKEN`. That HA-Core-proxy route now returns 403 for this token+path combination on this HA OS install. I haven't dug into HA Core to confirm exactly which component inside HA Core is doing the rejection.

By contrast, `tools_bug_report.py:186` calls `http://supervisor/addons/self/logs` directly with the same `SUPERVISOR_TOKEN` and works fine in the same environment, because that path bypasses HA Core entirely and hits Supervisor directly.

## Suggested fix

[...]

Issue #1099, opened 2026-05-03
[Docs] further Clarify Cloudflare setup in docs.md


kingpanther13:
### Discussed in https://github.com/homeassistant-ai/ha-mcp/discussions/1076

<div type='discussions-op-text'>

<sup>Originally posted by **beercity** April 29, 2026</sup>
Because I need to mainly use Claude in a browser, and I do not have a public IP at home, I found that using the HA MCP add-on was a better option than using the locally installed HA MCP tool. 

The issue I ran into was that my Cloudflare Tunnel was making it hard to get the MCP exposed over the public internet. This is the [Cloudflare Tunnel add-on I am using](https://github.com/homeassistant-apps/app-cloudflared)

The instructions said that in these cases, you should configure the Cloudflare tunnel with a sub-domain, for example `ha-mcp.domain.com`, that points to http://localhost:9583 which did not work for me.

What I found worked was getting the hostname of the HA MCP Server Add-on, which is available in the Add-on Information screen. An example of what this will look like is `b37de126-ha-mcp`

So you would configure the Cloudflare Tunnel HA add-on with `http://b37de126-ha-mcp:9583`

And then you would configure a Claude connector with `https://ha-mcp.example.com/[your_private_key_here]`

No port number, Cloudflare will do that translation.

I hope maybe this helps someone else, or maybe this can be used to enhance the documentation. </div>

Issue #1095, opened 2026-05-03
[FEATURE] Capture agent/model identity and config toggle state in ha_report_issue


kingpanther13:
## What would you like to see?

Extend `ha_report_issue` so the generated report templates auto-fill two new sections:

**1. Agent / model identity (self-reported by the LLM)**

The tool's instructions should ask the calling LLM to fill in:
- LLM provider and exact model identifier (e.g. `claude-opus-4-7`, `claude-sonnet-4-6`, `gpt-5-codex`, `gemini-3-pro`)
- MCP client / host application (Claude Code, Claude Desktop, Cursor, Continue, Copilot CLI, Gemini CLI, Codex, etc.)
- MCP transport in use (stdio / HTTP / SSE) — this one can be auto-detected, no self-report needed

Obviously this relies on the LLM accurately self-reporting, which is best-effort (Claude models are generally reliable, others are mixed). The template should make clear these fields are self-reported and ask the user to sanity-check them before submitting.

**2. ha-mcp config toggle state (read directly from config)**

Add the toggles that materially change the tool surface the agent sees, pulled from the running `Settings` object:
- `ENABLE_SKILLS` (skills as resources)
- `ENABLE_SKILLS_AS_TOOLS` (skills as tools)
- `ENABLE_TOOL_SEARCH` (tool search transform)
- `TOOL_SEARCH_MAX_RESULTS`
- `ENABLE_YAML_CONFIG_EDITING` (and any other safety/feature toggles worth surfacing)

These don't need self-reporting — `tools_bug_report.py` can read them off the config object the same way the rest of the addon does.

## Why do you need this?

A large chunk of triage on this repo is figuring out *which* agent and *which* tool surface produced the behavior being reported. Right now:

- `runtime_bug_template` has no model/client field at all.
- `agent_behavior_template` has a single freeform `**AI Client:** (Claude Desktop / Claude Code / Other)` line that most reports leave blank or vague.
- Neither template captures whether `enable_tool_search` or `enable_skills_as_tools` was on — and the available tools the agent sees differs dramatically between those configurations. The same complaint ("agent picked the wrong tool") can be a totally different bug depending on which toggles are active.

Capturing both up front means:
- Maintainers can tell immediately whether a "weird tool selection" report is the tool_search transform doing its job vs. an actual misbehavior.
- Skills-related issues can be filtered by whether skills were exposed as resources, as tools, or both.
[...]

PR #1122, opened 2026-05-04
fix(ci): harden gemini-triage so failures stop spamming user issues

kingpanther13:
## What does this PR do?

Stops the issue-triage bot from posting **❌ Evaluation failed** as a comment on every newly opened or commented issue (e.g. #1121, #1116, #1099, #1095, #1085, #1072, #1057, …) and hardens the workflow so future CLI / LLM / API breakage can't reach user issues the same way.

### Immediate fix
The bundled Gemini CLI inside `google-github-actions/run-gemini-cli@v0` was bumped to a version that refuses to run in untrusted directories:

```
Approval mode overridden to "default" because the current folder is not trusted.
Gemini CLI is not running in a trusted directory. To proceed, either use `--skip-trust`,
set the `GEMINI_CLI_TRUST_WORKSPACE=true` environment variable, or trust this directory
in interactive mode.
```

Set `GEMINI_CLI_TRUST_WORKSPACE: 'true'` on both `run-gemini-cli` steps — the headless escape hatch the CLI's own error message recommends ([trusted-folders docs](https://geminicli.com/docs/cli/trusted-folders/#headless-and-automated-environments)). `actions/checkout` is the only thing populating the workspace, so trusting it is in scope.

Example failed run before the fix: https://github.com/homeassistant-ai/ha-mcp/actions/runs/25345416494

### Hardening — the underlying bug
The triage workflow's `Update comment and labels` step ran with `if: always()` and read `steps.evaluate.outputs.error`, dumping raw CLI stderr onto the user's issue while reporting the Actions run green. Whatever Google shipped next inside `@v0` would have hit users the same way. Restructured so:

- **SHA-pinned** `run-gemini-cli@v0` → `f77273f4c914e4bf38440cf36a0369cb64a37489` (v0.1.22). Floating tag is what let the trust-workspace bump land silently in the first place.
- **`continue-on-error: true`** on both gemini-cli steps so the workflow controls how failures surface instead of letting the job die mid-flight.
- **Comment-update gates on `steps.<id>.outcome === 'failure'`** instead of presence of `outputs.error`. CLI step failure → generic *"Triage bot temporarily unavailable. Your issue will be triaged manually — no action needed on your part"* notice. No raw stderr on user-facing issues, ever.
- **`if: !cancelled()` on the comment-update step** so concurrency `cancel-in-progress` (a normal occurrence — every user follow-up comment cancels the prior triage run) doesn't paint cancelled runs red or overwrite live placeholders. Trade-off documented inline: job timeout also surfaces as `cancelled()`, so a real timeout exits green with the placeholder still on the issue; the next user comment retriggers triage and orphan-cleanup picks up the stale placeholder.
- **Trailing fail-job step** exits 1 only on actual `failure` outcomes (not `cancelled`), so the Actions UI shows red X for real failures and stays clean on supersedes.
- **`updateComment` and `addLabels` wrapped in try/catch** with `core.setFailed`. The `addLabels` failure message names the exact follow-up: *"Issue will re-trigger triage on next comment until 'triaged' is applied manually."*
- **Idempotent placeholder.** `Post initial "working" comment` lists existing comments, finds bot-authored placeholders matching the prefix (using `login === 'github-actions[bot]'` so other bots can't be hijacked), reuses the most recent and deletes older orphans. Defensive re-list **after** the create/reuse catches stragglers from concurrent runs that lost the cancel race. `listComments` failure → `setFailed` and bail (don't create duplicates the user can never tell apart).
- **Strict `COMPLETE` sentinel match.** Missing-info gate uses `evaluateResult.trim() === 'COMPLETE'` (not `.includes()`) so a future prompt rewrite that drifts the sentinel can't silently route wrong (`INCOMPLETE` / `MARKED_COMPLETE` / `IS_COMPLETE` no longer collide). The YAML-side `contains()` gate on `gemini_analysis` stays liberal on purpose: false-positive analysis is fine (user gets a real analysis attempt), false-negative isn't.
- **Missing-info path validation.** Refuses to dress up LLM refusals (`Sorry,`, `Unfortunately, I cannot…`, `As an AI…`, `Error:`, `safety policy`, `violates guidelines`, etc.) as *Information Needed*. First-person refusal verbs (`I cannot` / `I'm unable` / `We cannot`) are anchored to start AND required to pair with a help-action object (`help`/`assist`/`comply`/`fulfill`/etc.) so legitimate prose like *"I cannot find the version in the issue, please provide"* doesn't false-positive into the generic notice. Length-capped at 4000 chars to block runaway monologues. Off-shape prose (no list markers) gets posted verbatim with a `core.warning` so prompt drift surfaces in run logs without faking infra failure.
- **`core.warning`** when the LLM omits the `TRIAGE_LABELS` hidden comment, so a prompt-format regression is visible in run logs instead of issues silently labeling `triaged`-only with no classification.
- **Removed dead `comments_json` step output** that was set on the issue-data step and passed via env to gemini_analysis but never referenced in the prompt — eliminates an unused injection surface.

### Out of scope (intentional)
- **Model alias.** `gemini-3-flash-preview` is kept as-is; verified via Google's docs the alias is currently active and there's no GA counterpart yet (`gemini-3-pro-preview` was deprecated 2026-03-09 and replaced by `gemini-3.1-pro-preview`, but Flash hasn't followed yet). Added a `# Preview alias — review when gemini-3-flash GAs` comment so the next maintainer to read it knows to revisit.
- **TRIAGE_LABELS allowlist expansion** (e.g. `security`, `performance`). The current allowlist (`bug`/`enhancement`/`question`/`documentation`/`duplicate`/`invalid`) matches exactly what the prompt instructs Gemini to emit; expanding the allowlist without also updating the prompt does nothing. That's a feature change, not a fix.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

CI YAML only — local pytest/ruff aren't applicable, and the existing PR Validation Pipeline (Ruff / Mypy / AST / uv.lock / Unit / E2E / Docker) all pass on this branch. The refusal regex was additionally exercised against a sample matrix of 11 realistic Gemini refusals and 7 legitimate missing-info phrasings — 0 misses, 0 false positives. Real end-to-end validation happens on a triage run after merge:

- **Healthy `gemini` step** → normal analysis comment posted to the placeholder + green Actions run.
- **Broken `gemini` step** (e.g. quota, model deprecation, future trust-folder-style change) → generic *"temporarily unavailable"* comment + red X in Actions; user is told it'll be triaged manually.
- **Cancel-in-progress** (user posts a follow-up comment mid-triage) → cancelled run silently dies, replacement run picks up the orphan placeholder and posts the real analysis to it.
- **LLM refusal** (`"I cannot help with that"` / `"As an AI, I'm unable to comply"` / `"This violates my guidelines"`) → generic notice, not dressed as a polite info request.
- **LLM prose answer** without bullet markers → posted verbatim under *Information Needed* header with a warning logged to the run.
- **`addLabels` API failure** → red Actions run with a setFailed message naming the exact follow-up (apply `triaged` manually).

Happy to `workflow_dispatch` the triage against an existing issue from the fork before merging if useful.

## Checklist
- [x] I have updated documentation if needed _(no docs touched; CI-only change)_

### DIFF
--- a/.github/workflows/gemini-triage.yml
+++ b/.github/workflows/gemini-triage.yml
@@ -150,21 +150,32 @@ jobs:
             const PLACEHOLDER_PREFIX = '🤖 **Issue Triage Bot is analyzing this issue...**';
             const PLACEHOLDER_BODY = `${PLACEHOLDER_PREFIX}\n\nThis may take a minute. I\'m:\n- Fetching the full issue thread\n- Searching for related issues\n- Researching relevant code and documentation\n\nI\'ll update this comment when complete.`;
 
+            // Match strictly on `github-actions[bot]` (not just user.type === 'Bot')
+            // so a Dependabot/Renovate/etc. comment that happens to share the
+            // prefix can never be hijacked.
+            const isOurOrphan = (c) =>
+              c.user && c.user.login === 'github-actions[bot]' &&
+              c.body && c.body.startsWith(PLACEHOLDER_PREFIX);
+
             // Reuse / clean up orphan placeholders from prior cancelled runs.
             // The `Update comment` step is gated on `!cancelled()`, so a run
             // that gets cancelled (concurrency cancel-in-progress, job
             // timeout) never touches its own placeholder again — without
             // this cleanup, the placeholder stays as orphan text on the
             // issue forever and every retriggering comment piles up a new one.
-            const existing = await github.rest.issues.listComments({
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              issue_number: issueNumber,
-              per_page: 100,
-            });
-            const orphans = existing.data.filter(c =>
-              c.user && c.user.type === 'Bot' && c.body && c.body.startsWith(PLACEHOLDER_PREFIX)
-            );
+            let existing;
+            try {
+              existing = await github.rest.issues.listComments({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: issueNumber,
+                per_page: 100,
+              });
+            } catch (err) {
+              core.warning(`listComments failed during orphan check (${err.message}); proceeding to create a fresh placeholder.`);
+              existing = { data: [] };
+            }
+            const orphans = existing.data.filter(isOurOrphan);
 
             let commentId;
             if (orphans.length > 0) {
@@ -191,6 +202,36 @@ jobs:
               });
               commentId = created.data.id;
             }
+
+            // Defensive re-list AFTER our action: if a concurrent run lost
+            // the cancel race and was mid-createComment when we listed
+            // earlier, its orphan landed after our snapshot. Find and delete
+            // any other matching placeholders so the issue is left with
+            // exactly one (ours).
+            try {
+              const recheck = await github.rest.issues.listComments({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: issueNumber,
+                per_page: 100,
+              });
+              const stragglers = recheck.data.filter(c => isOurOrphan(c) && c.id !== commentId);
+              for (const stale of stragglers) {
+                try {
+                  await github.rest.issues.deleteComment({
+                    owner: context.repo.owner,
+                    repo: context.repo.repo,
+                    comment_id: stale.id,
+                  });
+                  core.info(`Cleaned up straggler placeholder ${stale.id} from a concurrent run.`);
+                } catch (err) {
+                  core.warning(`Failed to delete straggler placeholder ${stale.id}: ${err.message}`);
+                }
+              }
+            } catch (err) {
+              core.warning(`Defensive re-list for straggler cleanup failed: ${err.message}`);
+            }
+
             core.setOutput('comment_id', commentId);
             return commentId;
 
@@ -338,6 +379,8 @@ jobs:
           GEMINI_CLI_TRUST_WORKSPACE: 'true'
         with:
           gemini_api_key: '${{ secrets.GEMINI_API_KEY }}'
+          # Preview alias — review when gemini-3-flash GAs (or when a
+          # gemini-3.1-flash-preview lands, mirroring the 3-pro → 3.1-pro move).
           gemini_model: 'gemini-3-flash-preview'
           settings: |
             {
@@ -577,8 +620,15 @@ jobs:
       # superseded run (user posts a follow-up comment that retriggers triage)
       # would otherwise land here and overwrite the in-flight placeholder with
       # a "bot unavailable" notice — that's noise, not a failure. The
-      # replacement run will post the real analysis. Job timeout also surfaces
-      # as cancellation and we accept losing the placeholder in that rare case.
+      # replacement run will post the real analysis.
+      #
+      # Trade-off: GitHub Actions also surfaces job timeout (timeout-minutes:
+      # 12 above) as `cancelled()`, so a real timeout exits the run green
+      # with the placeholder still on the issue. The next user comment will
+      # retrigger triage and the orphan-cleanup in `Post initial` will pick
+      # up the stale placeholder. Acceptable: the user-visible artifact
+      # eventually gets repaired and timeouts are rare relative to
+      # superseded runs (which are common per issue).
       - name: Update comment and labels
         if: ${{ !cancelled() && steps.initial_comment.outputs.comment_id }}
         uses: actions/github-script@v9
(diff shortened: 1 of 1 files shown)


kingpanther13:
## Implementation Summary

**Choices Made:**

- **Bundled the cascade hardening into this PR** rather than landing the 2-line trust flag alone and filing follow-ups. The `if: always()` + dump-stderr-to-issue pattern is what *let* a transitive CLI bump turn into spam on every issue; without rewiring it, the next breaking change inside `@v0` ships the same UX bug to users. One PR, atomic.
- **SHA-pinned `run-gemini-cli@v0`** to `f77273f` (v0.1.22). Project convention is floating tags everywhere else; this is a deliberate deviation for an action that controls comment content on every issue. The floating tag is what let the bug land silently — pinning is the structural fix.
- **`continue-on-error: true` + `outcome === 'failure'`** instead of `conclusion`-based gating. With `continue-on-error`, `conclusion` is masked to `success`, so `outcome` is the only signal that catches the real CLI exit.
- **`!cancelled()` instead of `always()`** on the comment-update step. `concurrency.cancel-in-progress: true` is set, so every user follow-up comment cancels the prior triage. Treating `cancelled` as terminal would have painted those (normal) runs red and overwritten the in-flight placeholder with a "bot down" notice. Trade-off documented inline: job timeout also surfaces as `cancelled()`, so a real timeout exits green with the placeholder on the issue; the next user comment retriggers triage and the orphan-cleanup picks up the stale placeholder.
- **Idempotent placeholder.** Without orphan reuse, the cancel-then-cancel pattern accumulates stuck "🤖 analyzing…" comments on the issue with no path to clean them up. Now: list → reuse most recent orphan → delete older orphans → defensive re-list after action to catch stragglers from a concurrent run that lost the cancel race.
- **`login === 'github-actions[bot]'` for orphan match, `type === 'Bot'` for `should_run` skip.** Different invariants — the orphan path *deletes* comments and must not hijack other bots' (Dependabot, Renovate, etc.) comments; the should_run path just wants to ignore retriggers from any bot. Both checks documented inline.
- **Refusal-pattern divert before posting prose.** The shape check used to be a hard reject on anything that wasn't a bullet list, which dressed valid prose answers as "infra failure." Now: refusals → generic notice; runaway monologues (>4000 chars) → generic notice; everything else → posted verbatim under *Information Needed* with a `core.warning` if it didn't match list shape. After review iteration, refusal patterns were tightened to require the first-person verb to be paired with a help-action object (`help`/`assist`/`comply`/etc.) so legitimate prose like *"I cannot find the version in the issue body, please provide"* doesn't false-positive into the generic notice. Verified against a sample matrix of 11 realistic refusals (all match) and 7 legitimate missing-info phrasings (all pass): 0 misses, 0 false positives.
- **Strict `evaluateResult.trim() === 'COMPLETE'`** for the missing-info gate (instead of `.includes()`). Defends against a future prompt rewrite that drifts the sentinel — `INCOMPLETE` / `MARKED_COMPLETE` / `IS_COMPLETE` would no longer collide. The YAML-side `contains()` gate on `gemini_analysis` stays liberal on purpose: false-positive analysis is fine (user gets a real analysis attempt on a sparse issue), false-negative analysis isn't.
- **API calls wrapped in try/catch with `core.setFailed`.** `updateComment` failure → red Actions run (so a maintainer notices the placeholder is stuck). `addLabels` failure → red run **and** the message names the exact follow-up: *"Issue will re-trigger triage on next comment until 'triaged' is applied manually."* (Without the `triaged` label, `should_run` re-fires every reply and pile up placeholders.)
- **Removed dead `comments_json` step output.** Was set on the issue-data step and passed via env to gemini_analysis but never referenced in the prompt; deleting it eliminates an injection surface for free.

**Problems Encountered:**

This branch went through 6 rounds of multi-agent review (code-reviewer + silent-failure-hunter + idiot-checker, all on Opus) before being marked ready. The reviews caught several confidently-shipped bugs across iterations, every one of which was fixed in subsequent commits:

- **Cancelled-as-failure regression.** Mid-PR I expanded the failure-detection set to include `cancelled` (intending to cover step timeouts), which would have painted every concurrency-cancelled run red and overwritten live placeholders. Reverted: only `failure` is terminal; `cancelled` is normal and silenced via `!cancelled()` on the relevant steps.
- **`listComments` fallback regression.** The first orphan-cleanup version fell back to "create a fresh placeholder" if listing failed — strictly worse than the prior code, since a transient list error would leave the issue with both an orphan and a brand-new placeholder. Fixed: `listComments` failure → `setFailed` and bail (no comment_id, downstream skips, red run).
- **Shape-check too aggressive.** Earlier rounds rejected valid prose answers as "infra unavailable." Walked through Gemini refusal/error phrasings and added an explicit refusal regex so the shape check is now warning-only for non-refusal off-shape output.
- **Refusal regex too narrow / too broad.** First version was anchored to `^i` only, missing "Unfortunately, I cannot…" / "As an AI…". Broadened. Then the broader version was unanchored on `\bi (cannot|...)\b` which false-positived on legitimate prose like *"I cannot find the version, please provide"*. Final version anchors to `^` AND requires a help-action object (`help`/`assist`/`comply`/etc.) — verified zero false-positives and zero misses on a matrix of realistic Gemini outputs.
- **`tooShort` floor was numerology** (initially 20, then 8 — both pulled from nowhere and rejected legitimate one-liners like `- HA version`). Dropped entirely; the refusal regex catches what the floor was meant to catch, properly.

By the final review pass: code-reviewer returned *"ship it"* with no critical/important findings, silent-failure-hunter and idiot-checker each returned a single merge-blocker (the unanchored refusal regex) which was fixed in the final commit. Every actionable review finding across the full review history is either resolved in code or has explicit traced reasoning recorded in the file (e.g. the `INCOMPLETE` substring "dual-path firing" concern was traced through and confirmed to produce correct UX — `geminiSummary` taking precedence in the JS cascade is the right design, not a bug).

**Suggested Improvements (out of scope for this PR):**

- **Track `gemini-3-flash` GA / `gemini-3.1-flash-preview` arrival.** The model alias is preview-only with no GA counterpart yet (Pro went `3-preview` → `3.1-preview` in March 2026; Flash hasn't followed). Inline `# Preview alias — review when gemini-3-flash GAs` comment now flags this for whoever reads next.
- **Possible TRIAGE_LABELS allowlist expansion** (e.g. `security`, `performance`). Currently the allowlist matches exactly what the prompt instructs Gemini to emit; broadening it requires updating the prompt at the same time. A separate change if you want those classifications.
- **`looksLikeRefusal` regex is heuristic.** It catches the common Gemini refusal phrasings I could think of plus what reviewers surfaced, with a help-action pairing requirement to avoid false positives. If a future model emits a novel refusal style that doesn't pair with help/assist/comply/etc., it'll slip through and post as "Information Needed." Worth revisiting if telemetry shows it happening.

Patch76:
A few non-blocking follow-up notes flagged for visibility — none change the merge decision; the underlying bug ("evaluate stderr leaks onto issues via `if: always()` + `outputs.error`") is closed at every traceable path.

**1. PR-body / code mismatch on length-cap ordering** (`gemini-triage.yml:735-743`)

The body says *"Length-capped at 4000 chars to block runaway monologues"* but the actual code computes `looksLikeRefusal` (regex) before `tooLong` (length cap):

```js
const looksLikeRefusal = REFUSAL_PATTERNS.some(re => re.test(trimmed));
const tooLong = trimmed.length > 4000;
// ...
if (looksLikeRefusal || tooLong) { /* GENERIC_FAILURE_BODY */ }
```

Both branches divert to the same `GENERIC_FAILURE_BODY`, so observable behavior is identical to swapping the two — but the order is the inverse of what the body claims. Cosmetic.

**2. Three Gemini-style refusal phrasings slip through the regex** (`gemini-triage.yml:720-734`)

Reading the actual `REFUSAL_PATTERNS` array — these phrasings would not match any of the 9 patterns:

- `"I am afraid I cannot help with this"` — leads with `I am afraid`, not anchored by `^i (cannot|can't|won't) (help|...)` or `^i am (unable|not able)`
- `"Unable to help with this issue"` — no `^I'm` / `^I am` prefix to anchor
- `"I can't provide assistance"` — `provide` requires `that|this` follow-up in the help-action group; `\bcannot assist\b` doesn't match `can't provide assistance` either

None are dominant Gemini idioms. A future tightening could add `/^i'?m afraid (i )?cannot/i` and `/^unable to (help|assist|comply)/i` to the array. Gap, not blocker.

**3. YAML-level `contains('COMPLETE')` gate stays liberal** (`gemini-triage.yml:410`)

```yaml
if: contains(steps.evaluate.outputs.summary, 'COMPLETE') || ...
```

The strict-`COMPLETE` migration in `:687` (`evaluateResult.trim() !== 'COMPLETE'`) is the JS-side fix this PR's about — but the YAML expression-level gate is still substring-fuzzy: an evaluate output of `INCOMPLETE` would still trigger `gemini_analysis` here. Pre-existing, not introduced in this PR; the body explicitly notes this is on purpose ("false-positive analysis is fine, false-negative isn't"). Just flagging — a future tightening to `== 'COMPLETE'` would lock both sides symmetrically.

Patch76 (approved):
Approve. The bug the title is about — `if: always()` + `outputs.error` leaking CLI stderr onto user issues — is closed at every traceable path: CLI step failure routes to `GENERIC_FAILURE_BODY`, refusal-shaped CLI output routes to `GENERIC_FAILURE_BODY`, length-capped output routes to `GENERIC_FAILURE_BODY`, job cancellation skips the comment-update via `!cancelled()`, API call failures go through `core.setFailed` with the named follow-up, and concurrent-run straggler placeholders are deleted by the defensive re-list.

SHA-pinning `run-gemini-cli` to `f77273f4c91...` (v0.1.22) on both gemini-cli steps removes the floating-tag risk that triggered this. CI green on `2412d79`.

Posted three non-blocking follow-up notes in a separate comment for visibility — none affect the merge decision.

PR #1124, opened 2026-05-05
feat: emit progress via FastMCP Context in long-running tools

drseanwing:
## Why

Several long-running ha-mcp tools (`ha_deep_search`, `ha_get_history`,
`ha_get_automation_traces`, `ha_bulk_control`, `ha_hacs_search`) currently
return only after the entire operation completes. Clients (Claude Code, etc.)
have no way to render progress or surface intermediate state, even though
`ha_deep_search` for example may run a 30-second per-automation REST fallback
loop on a busy instance.

FastMCP 3.x already supports a `Context` parameter that, when declared by a
tool, is injected automatically and hidden from the LLM-visible schema. Tools
can then call `await ctx.report_progress(...)` and `await ctx.info(...)` to
emit MCP progress and log notifications back to the client.

## What

Adds an optional `ctx: Context | None = None` parameter to five long-running
tools (and the service-layer methods they delegate to):

- `ha_deep_search` — `tools_search.py` + `smart_search.py::deep_search`. One
  progress event per requested `search_type`, plus a leading event for the
  initial `get_states()` fetch.
- `ha_get_history` — `tools_history.py`. Events at WebSocket connect, recorder
  query dispatch, and completion.
- `ha_get_automation_traces` — `tools_traces.py`. Events at connect, fetch,
  and final formatting; an extra event marks the diagnostics-gather branch
  when no traces exist.
- `ha_bulk_control` — `tools_service.py` + `device_control.py::bulk_device_control`.
  Per-operation progress in sequential mode; framing-only events in parallel
  mode (asyncio.gather makes per-task progress tracking expensive).
- `ha_hacs_search` — `tools_hacs.py`. Events at HACS-availability check,
  list fetch, and filter-and-score completion.

In each case the existing `info(...)` event provides a high-level start
message and `report_progress(progress, total, message)` calls advance through
the phases.

## Compat

- `ctx` is **optional** with a default of `None`. Existing callers (and the
  test suite) that omit the parameter are unchanged.
- FastMCP 3.2.4 (the pinned version) auto-hides `Context`-typed parameters
  from the LLM-visible JSON schema via `transform_context_annotations` +
  `without_injected_parameters` in `fastmcp/tools/function_parsing.py`.
  Verified by inspecting the installed package; matches the existing
  precedent in `src/ha_mcp/transforms/categorized_search.py`.
- Return shapes are unchanged. `Context` is a pure side-channel.

## Tests

New: `tests/src/unit/test_context_injection.py` — 11 tests covering each of
the 5 tools with both the legacy (`ctx=None`) path and a fake AsyncMock
`Context`. Asserts `report_progress` / `info` are awaited at the expected
boundaries without changing return shape.

Full unit suite: **1713 passed, 1 skipped** (`numpy` unavailable on this
runner; same skip on `upstream/master`). `ruff check`, `mypy src/
homeassistant-addon/ scripts/`, and `mypy homeassistant-addon-webhook-proxy/`
all clean.

## Out of scope

- Short / synchronous tools (e.g. `ha_get_state`) — no useful progress to emit.
- Per-task progress in `ha_bulk_control(parallel=True)` — would need wrapping
  each `asyncio.gather` task in an individual reporter; deferred.
- Any output-shape changes — `Context` is intentionally side-channel only.


Patch76:
RfR?

kingpanther13 (approved):
LGTM. Verified FastMCP 3.2.4 auto-hides Context from the LLM schema (`function_parsing.py:169-172`), every emit is gated on `ctx is not None`, return shapes unchanged, `except ToolError: raise` guards preserved across all 5 tools.

Will follow up in a separate PR with a safe-emit wrapper (so transport errors on a progress notification can't convert a successful operation into a `ToolError`), expanded branch coverage (`deep_search` non-helper types, `ha_get_history` statistics, `ha_get_automation_traces` run_id/diagnostics branches), and `ctx:` Args-block docstring lines on the 4 tools that are missing them.

PR #1135, opened 2026-05-05
feat: expose device_class + options on ha_set_entity / ha_get_entity (Show As)

### LINKED ISSUE #1114: [FEATURE] support changing Device Class "Show As"

I needed to change 20 devices today and wanted Claude to do it. Claude told me... "Show As" is stored in the entity registry as options.binary_sensor.device_class. The ha_set_entity tool doesn't expose this. 

Hoping you can add this functionality. thx for the great work... using this has saved me many, many hours!

### Why do you need this?

I often find that some Integrations are 'lazy' and set generic device_classes... like "Opening" when it is a door or window.

### Additional context

_No response_

kingpanther13:
## What does this PR do?

Closes #1114.

Adds the registry-options write surface that's been missing from `ha_set_entity`, plus the matching read fields on `ha_get_entity`:

- **`device_class`** on `ha_set_entity` — the top-level field HA's UI Show As dropdown writes. Empty string clears the override and falls back to the integration's `original_device_class`. Applies instantly, no reload needed.
- **`options`** on `ha_set_entity` — per-domain registry options dict, e.g. `{"sensor": {"display_precision": 2}}`. Multi-domain dicts are split into one `config/entity_registry/update` call per domain, because HA's WS schema requires `options_domain` + `options` to be sent paired (`vol.Inclusive` group `entity_option`) with `options` carrying ONE domain's sub-dict per call.
- `ha_get_entity` now returns `device_class`, `original_device_class`, and `options` so agents can read state before deciding whether to mutate.

### Why two parameters, not one

Live verification on a deployed HA instance (template `binary_sensor`) confirmed the UI's "Show As" dropdown writes the **top-level** `device_class` field — *not* `options.binary_sensor.device_class`, which exists in the registry but is unused by the frontend. Routing Show As through `options` would silently no-op in the UI. The original issue's request was specifically "Show As", so it gets the dedicated parameter; the generic `options` parameter still covers `display_precision`, `unit_of_measurement`, `forecast_type`, etc.

## Type of change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing
- [x] I have tested these changes with a LLM agent (manual roundtrip on the live deployment)

Unit (`tests/src/unit/test_tools_entities.py`):
- `TestHaSetEntityShowAs` — top-level `device_class` field, empty-string clear, `options` single-domain pairing, multi-domain split, JSON-string parsing, validation error on bad shape
- `TestHaGetEntityRegistryOptions` — response includes the three new fields

E2E (`tests/src/e2e/workflows/entities/test_show_as.py`):
- Round-trip `device_class` set → get → clear on a template binary_sensor
- Round-trip `options.sensor.display_precision` on a template sensor

## Checklist

### DIFF
--- a/src/ha_mcp/tools/tools_entities.py
+++ b/src/ha_mcp/tools/tools_entities.py
@@ -38,6 +38,9 @@ def _format_entity_entry(entry: dict[str, Any]) -> dict[str, Any]:
         "aliases": entry.get("aliases", []),
         "labels": entry.get("labels", []),
         "categories": entry.get("categories", {}),
+        "device_class": entry.get("device_class"),
+        "original_device_class": entry.get("original_device_class"),
+        "options": entry.get("options", {}),
     }
 
 
@@ -75,6 +78,8 @@ def register_entity_tools(mcp: Any, client: Any, **kwargs: Any) -> None:
         parsed_expose_to: dict[str, bool] | None,
         new_entity_id: str | None = None,
         new_device_name: str | None = None,
+        device_class: str | None = None,
+        parsed_options: dict[str, dict[str, Any]] | None = None,
     ) -> dict[str, Any]:
         """Update a single entity. Returns the response dict."""
         # For add/remove operations, we need to fetch current labels first
@@ -119,6 +124,14 @@ def register_entity_tools(mcp: Any, client: Any, **kwargs: Any) -> None:
         if icon is not None:
             message["icon"] = icon if icon else None
             updates_made.append(f"icon='{icon}'" if icon else "icon cleared")
+
+        if device_class is not None:
+            message["device_class"] = device_class if device_class else None
+            updates_made.append(
+                f"device_class='{device_class}'"
+                if device_class
+                else "device_class cleared"
+            )
 
         if enabled is not None:
             try:
@@ -203,13 +216,17 @@ def register_entity_tools(mcp: Any, client: Any, **kwargs: Any) -> None:
         if new_device_name is not None:
             updates_made.append(f"device_name -> {new_device_name}")
 
+        if parsed_options:
+            for _domain, _sub in parsed_options.items():
+                updates_made.append(f"options[{_domain}]={_sub}")
+
         if not updates_made:
             raise_tool_error(
                 create_error_response(
                     ErrorCode.VALIDATION_INVALID_PARAMETER,
                     "No updates specified",
                     suggestions=[
-                        "Provide at least one of: area_id, name, icon, enabled, hidden, aliases, categories, labels, expose_to, new_entity_id, or new_device_name"
+                        "Provide at least one of: area_id, name, icon, device_class, enabled, hidden, aliases, categories, labels, options, expose_to, new_entity_id, or new_device_name"
                     ],
                 )
             )
@@ -241,15 +258,19 @@ def register_entity_tools(mcp: Any, client: Any, **kwargs: Any) -> None:
                     "Verify the entity_id exists using ha_search_entities()",
                 ]
                 if new_entity_id is not None:
-                    suggestions.extend([
-                        "Check that the new entity_id doesn't already exist",
-                        "Ensure the entity has a unique_id (some legacy entities cannot be renamed)",
-                    ])
+                    suggestions.extend(
+                        [
+                            "Check that the new entity_id doesn't already exist",
+                            "Ensure the entity has a unique_id (some legacy entities cannot be renamed)",
+                        ]
+                    )
                 else:
-                    suggestions.extend([
-                        "Check that area_id exists if specified",
-                        "Some entities may not support all update options",
-                    ])
+                    suggestions.extend(
+                        [
+                            "Check that area_id exists if specified",
+                            "Some entities may not support all update options",
+                        ]
+                    )
                 raise_tool_error(
                     create_error_response(
                         ErrorCode.SERVICE_CALL_FAILED,
@@ -264,6 +285,41 @@ def register_entity_tools(mcp: Any, client: Any, **kwargs: Any) -> None:
             # If entity was renamed, update entity_id for subsequent operations
             if new_entity_id:
                 entity_id = new_entity_id
+
+        # Per-domain options updates: HA's `config/entity_registry/update` WS schema
+        # treats `options_domain` + `options` as a `vol.Inclusive("entity_option")`
+        # group — they MUST be sent paired, and `options` carries only ONE domain's
+        # sub-dict per call. So an agent-supplied {domain: {...}, ...} is split into
+        # one WS call per domain.
+        if parsed_options:
+            for opts_domain, opts_sub in parsed_options.items():
+                opts_msg: dict[str, Any] = {
+                    "type": "config/entity_registry/update",
+                    "entity_id": entity_id,
+                    "options_domain": opts_domain,
+                    "options": opts_sub,
+                }
+                opts_result = await client.send_websocket_message(opts_msg)
+                if not opts_result.get("success"):
+                    error = opts_result.get("error", {})
+                    error_msg = (
+                        error.get("message", str(error))
+                        if isinstance(error, dict)
+                        else str(error)
+                    )
+                    raise_tool_error(
+                        create_error_response(
+                            ErrorCode.SERVICE_CALL_FAILED,
+                            f"Failed to update options for domain '{opts_domain}': {error_msg}",
+                            context={
+                                "entity_id": entity_id,
+                                "options_domain": opts_domain,
+                            },
+                        )
+                    )
+                entity_entry = opts_result.get("result", {}).get(
+                    "entity_entry", entity_entry
+                )
 
         # Handle new_device_name — rename the associated device
         # Normalize empty string to None (no-op, don't clear device name)
@@ -281,12 +337,16 @@ def register_entity_tools(mcp: Any, client: Any, **kwargs: Any) -> None:
                 if get_result.get("success"):
                     entity_entry = get_result.get("result", {})
                 else:
-                    logger.warning(f"Entity registry lookup failed for {entity_id}: {get_result.get('error')}")
+                    logger.warning(
+                        f"Entity registry lookup failed for {entity_id}: {get_result.get('error')}"
+                    )
                     device_rename_result = {
                         "warning": "Entity registry lookup failed — could not determine device. Retry may succeed.",
                     }
 
-            device_id = entity_entry.get("device_id") if not device_rename_result else None
+            device_id = (
+                entity_entry.get("device_id") if not device_rename_result else None
+            )
             if not device_id:
                 device_rename_result = {
                     "warning": "Entity has no associated device — device rename skipped",
@@ -351,12 +411,16 @@ def register_entity_tools(mcp: Any, client: Any, **kwargs: Any) -> None:
                     if has_registry_updates:
                         context["partial"] = True
                         context["entity_entry"] = _format_entity_entry(entity_entry)
-                    raise_tool_error(create_error_response(
-                        ErrorCode.SERVICE_CALL_FAILED,
-                        f"Exposure failed: {error_msg}",
-                        context=context,
-                        suggestions=["Check Home Assistant connection and entity availability"],
-                    ))
+                    raise_tool_error(
+                        create_error_response(
+                            ErrorCode.SERVICE_CALL_FAILED,
+                            f"Exposure failed: {error_msg}",
+                            context=context,
+                            suggestions=[
+                                "Check Home Assistant connection and entity availability"
+                            ],
+                        )
+                    )
 
                 # Track successful exposures
                 for a in assistants:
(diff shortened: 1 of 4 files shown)


kingpanther13:
Thanks for the review. Addressed below.

**1. Partial-success error context for the per-domain options loop (lines 294-322)** — Agreed and already addressed in 1505098d (pushed before this review landed). The implementation goes a bit further than the suggestion: in addition to `partial: True` and `entity_entry`, the error response now also carries `options_succeeded` (which domains were already written) and `updates_applied` (the list of fields that landed before the failure). The corresponding unit test is `test_options_partial_failure_surfaces_partial_state` in f2a03d43.

**2 & 3. Double-backticks for code spans in the new `options` Field description and in the SHOW AS / DEVICE CLASS docstring section** — Holding off on this one. I checked `.gemini/styleguide.md` for the cited rule and didn't find a backticks convention there. A grep across `src/ha_mcp/tools/` shows zero `Field(description=...)` strings using single OR double backticks — all of them (including pre-existing params on this same `ha_set_entity` tool: `area_id`, `name`, `icon`, `aliases`, `categories`, etc.) use prose with single-quoted code values, e.g. `"Use empty string '' to remove custom icon"` and `"e.g., 'mdi:thermometer'"`. Adopting double-backticks for just the new params would break that file-level consistency. The double-backtick precedent in `helpers.py` / `reference_validator.py` is in regular Sphinx-style module docstrings, not in `Field(description=...)` strings exposed to MCP clients via JSON Schema — different contexts. Happy to revisit if there's an explicit decision to migrate the whole file (or all tool descriptions) to double-backticks.

kingpanther13:
## Implementation Summary

**Choices Made:**
- **Two parameters (`device_class` + `options`), not one.** Live verification on a deployed HA instance showed the UI's "Show As" dropdown writes the **top-level `device_class`** field on the registry entry, NOT `options.binary_sensor.device_class` (which exists but is unused by the frontend for that purpose). Routing Show As through `options` would silently no-op in the UI — exactly the failure mode #1114 describes. The dedicated `device_class` parameter solves Show As; `options` covers the rest (sensor `display_precision`, weather `forecast_type`, etc.).
- **Per-domain split inside `_update_single_entity`.** HA's `config/entity_registry/update` WS schema treats `options_domain` + `options` as a `vol.Inclusive` group with `options` carrying ONE domain's sub-dict per call, so an agent-supplied `{domain: {...}, ...}` is split into one WS call per domain. Empty top-level dict and non-dict sub-values are rejected at validation with messages naming the offending key/type.
- **Mirrored the established `expose_to` partial-failure pattern.** Per-domain options loop tracks `options_succeeded`, appends `updates_made` only after each WS success, and on partial failure surfaces `partial=True`, `options_succeeded`, `updates_applied`, and an `entity_entry` snapshot in the error context. Boy-scout: also extended `expose_to`'s `partial` flag to include `options_succeeded`, `device_rename_result.success`, and the in-progress `succeeded` batch dict (was previously only `has_registry_updates`).
- **`_extract_ws_error()` helper applied to every WS-error site in the module** (5 callsites). Logs a warning and returns a stable placeholder when HA returns an empty error envelope, so the user-facing message never degrades to literal `"{}"`.
- **Two e2e test classes.** `TestShowAs` round-trips on template helpers (helpers have an alternative path via `ha_config_set_helper`); `TestShowAsOnIntegrationEntity` round-trips on a YAML template sensor (owned by the template integration, no config entry, NOT reachable via `ha_config_set_helper`) — gives literal coverage for the original-issue case.

**Problems Encountered:**
- **Issue bot's auto-generated diff was wrong.** It proposed a single `options` parameter sent as a bare WS message; HA's schema rejects that with `invalid_format: some but not all values in the same group of inclusion 'entity_option'`. Verified live; reshaped the implementation to split per-domain.
- **`sensor.demo_temperature` referenced by several existing tests didn't actually exist in CI.** The bare `demo:` line in `tests/initial_test_state/configuration.yaml` is a no-op in modern HA — the demo integration is config-flow-only since it was migrated. Existing tests in `test_helper_crud.py` and `test_config_entry_flow.py` "passed" because HA helper configs (min_max etc.) accept unknown entity IDs without validating they exist (the helper got created with `unknown` runtime state but the tests only verified the helper itself). Boy-scout fix: declared `sensor.demo_temperature` and `sensor.demo_outside_temperature` as YAML template sensors. Names match what existing tests already reference, so those tests now actually exercise real entities for free.
- **State-based YAML template entities require `unique_id` to register in the entity registry.** Without it they exist in HA's state machine but `ha_get_entity` (which queries `config/entity_registry/get`) returns "Entity not found". Plus HA 2026.4 stricter validation expects `state_class` alongside the `temperature` device_class. Both fixed in `876a8ba1`.

**Suggested Improvements:**
- The "tests reference non-existent entities" issue surfaced here may exist elsewhere in the suite. A grep audit comparing entity-ID strings in test files against entities the testcontainer actually creates could surface other silently-passing tests. Not done as part of this PR; flagging for follow-up.

src/ha_mcp/tools/tools_entities.py
```diff
@@ -265,6 +292,53 @@ async def _update_single_entity(
...
+                        create_error_response(
+                            ErrorCode.SERVICE_CALL_FAILED,
+                            f"{msg_prefix} domain '{opts_domain}': {err_msg}",
+                            context={
+                                "entity_id": entity_id,
+                                "options_domain": opts_domain,
+                                "partial": partial,
+                                "options_succeeded": options_succeeded,
+                                "updates_applied": list(updates_made),
+                                "entity_entry": _format_entity_entry(entity_entry),
```
Patch76: The per-domain options failure path injects `entity_entry` unconditionally:

```python
context={
    "entity_id": entity_id,
    "options_domain": opts_domain,
    "partial": partial,
    "options_succeeded": options_succeeded,
    "updates_applied": list(updates_made),
    "entity_entry": _format_entity_entry(entity_entry),
},
```

The `expose_to` failure path further down (lines 426-445) gates the same field — and `partial`, `options_succeeded`, `device_rename_succeeded` — behind `if prior_mutation:`, where `prior_mutation = has_registry_updates or bool(options_succeeded) or bool(succeeded) or bool(device_rename_result and device_rename_result.get("success"))`. On a clean first-call failure here (no prior registry update, no prior options domain succeeded), `entity_entry={}` and `_format_entity_entry({})` returns a stub of all-`None` / empty-list / empty-dict fields — indistinguishable from "entity exists but has nothing set" rather than "no snapshot was taken".

**Suggested**: mirror the `expose_to` shape:

```python
context: dict[str, Any] = {
    "entity_id": entity_id,
    "options_domain": opts_domain,
    "partial": partial,
    "options_succeeded": options_succeeded,
    "updates_applied": list(updates_made),
}
if partial:
    context["entity_entry"] = _format_entity_entry(entity_entry)
```

`test_options_first_call_failure_is_not_partial` (`tests/src/unit/test_tools_entities.py:1648`) currently doesn't assert on `entity_entry`, so the change wouldn't break it — but extending the test to assert absence on the not-partial path would lock the asymmetry from drifting back.

src/ha_mcp/tools/tools_entities.py
```diff
@@ -120,6 +137,14 @@ async def _update_single_entity(
             message["icon"] = icon if icon else None
             updates_made.append(f"icon='{icon}'" if icon else "icon cleared")
 
+        if device_class is not None:
+            message["device_class"] = device_class if device_class else None
```
Patch76: `message["device_class"] = device_class if device_class else None` uses Python truthiness, so `""` correctly clears (matches the docstring's "pass empty string to clear") but `"   "` (whitespace-only) is truthy and reaches HA literally. HA then rejects with whatever WS-validator error pops out, which surfaces as a generic registry-update failure rather than a clear validation error.

Two options: `.strip()` the value before the check (friendlier UX, mirrors `new_device_name` normalization 25 lines below) or document explicitly that whitespace is treated as a literal value. The `name` and `icon` paths a few lines above carry the same shape, but they're load-bearing only on rare paths.

tests/src/e2e/workflows/entities/test_show_as.py
```diff
@@ -0,0 +1,291 @@
...
+                "ha_set_entity",
+                {"entity_id": self.TARGET, "device_class": restore_dc},
+            )
+            # Restore the sensor sub-options to their pre-test value so any
+            # demo-platform default we overwrote is put back intact.
+            await mcp_client.call_tool(
+                "ha_set_entity",
+                {
+                    "entity_id": self.TARGET,
+                    "options": {"sensor": dict(original_sensor_options)},
```
Patch76: When `pre_data["entity_entry"]["options"]` lacks a `sensor` sub-key, `original_sensor_options` defaults to `{}` (line ~232 capture), and the restore call then emits `options={"sensor": {}}`. The validator at `tools_entities.py:917 / 926-930` accepts that — top-level dict is non-empty and the inner value is a dict — but no other test exercises whether HA's WS-side validator accepts an empty inner dict for `options_domain` + `options` payloads. CI passes so it works in practice today; either skip the restore call when the captured value is empty (cleaner, no round-trip on a no-op) or add a one-line comment so the next reader doesn't grep this case.

src/ha_mcp/tools/tools_entities.py
```diff
@@ -265,6 +292,53 @@ async def _update_single_entity(
...
+                                "partial": partial,
+                                "options_succeeded": options_succeeded,
+                                "updates_applied": list(updates_made),
+                                "entity_entry": _format_entity_entry(entity_entry),
+                            },
+                        )
+                    )
+                entity_entry = opts_result.get("result", {}).get(
+                    "entity_entry", entity_entry
+                )
```
Patch76, line 352:
```python
entity_entry = opts_result.get("result", {}).get(
    "entity_entry", entity_entry
)
```

Last-call-wins reassignment is the intended semantic — HA returns the cumulative entry on each per-domain call, so the final loop iteration carries the full state — but it's not obvious from the local code, and `test_options_multi_domain_splits_into_separate_calls` only asserts the final entry, not the reassignment per-call. A one-line comment here ("each WS round-trip returns the cumulative `entity_entry`; keep the last one") would save the next reader the trace.

src/ha_mcp/tools/tools_entities.py
```diff
@@ -451,6 +552,34 @@ async def ha_set_entity(
...
+        options: Annotated[
+            str | dict[str, dict[str, Any]] | None,
+            Field(
+                description=(
+                    "Per-domain entity registry options (e.g. sensor 'display_precision', "
+                    "weather 'forecast_type'). Pass a dict mapping domain to a sub-dict, "
+                    'e.g. {"sensor": {"display_precision": 2}}. JSON-string form also accepted. '
+                    "Multiple domains are sent as separate registry updates. "
+                    "For 'Show As' use the dedicated `device_class` parameter — that is "
+                    "what the HA UI Show As dropdown writes. Single entity only."
```
Patch76: The `ha_get_entity` docstring at lines 1131-1133 explicitly warns that voice-assistant exposure is stored in `options.<assistant>` but should be managed via `ha_set_entity(expose_to=...)`, not the options dict. The set-side `options` Field description here does not mention this collision — a user calling `ha_set_entity(options={"google_assistant": {"should_expose": True}})` would silently bypass the dedicated `expose_to` machinery. Appending one sentence here would close the gap:

> "Note: voice-assistant exposure is stored under `options.<assistant>.should_expose` but must be managed via the dedicated `expose_to` parameter."

Patch76 (commented):
Re-review of HEAD `876a8ba1`. Implementation is solid: the two-parameter design isolates Show As from generic options correctly, partial-failure context on the per-domain split is structured (`partial` / `options_succeeded` / `updates_applied` / `entity_entry`), `_extract_ws_error` is applied at every WS-error site in the module, and the `sensor.demo_temperature` template-sensor fixture turns several previously-silent helper-source tests into real assertions.

One consistency nit between the per-domain options failure path and the `expose_to` failure path, plus four small docstring / edge-case items. Inline.

src/ha_mcp/tools/tools_entities.py
kingpanther13: ✅ Fixed in 8ae02e42. Mirrored the expose_to shape — context dict built incrementally, `entity_entry` only injected when `partial` is True. Also extended `test_options_first_call_failure_is_not_partial` with `assert "entity_entry" not in body` to lock the symmetry.

src/ha_mcp/tools/tools_entities.py
kingpanther13: ✅ Fixed in 8ae02e42 — went with the `.strip() or None` route to match `new_device_name`. The `name` and `icon` paths above still use the truthiness shape; happy to align those in a follow-up if you want, but kept this PR focused on the device_class addition.

tests/src/e2e/workflows/entities/test_show_as.py
kingpanther13: ✅ Fixed in 8ae02e42. Guarded the restore call with `if original_sensor_options:` — cleaner than the no-op round-trip and removes the dependency on HA's WS validator continuing to accept the empty inner dict.

src/ha_mcp/tools/tools_entities.py
kingpanther13: ✅ Fixed in 8ae02e42. Added the one-liner: "HA returns the cumulative entity_entry on each per-domain call, so last-call-wins reassignment leaves the final loop iteration carrying the full state."

src/ha_mcp/tools/tools_entities.py
kingpanther13: ✅ Fixed in 8ae02e42. Added: "Voice-assistant exposure is stored under `options.<assistant>.should_expose` but must be managed via the dedicated `expose_to` parameter, not this options dict."

kingpanther13 (commented):
✅ Addressed all 5 review findings in 8ae02e42:

- **Finding 1** (entity_entry asymmetry): Per-domain options failure now gates `entity_entry` on `partial`, mirroring the expose_to path. Extended `test_options_first_call_failure_is_not_partial` to assert `entity_entry` is absent on a clean failure so the asymmetry can't drift back.
- **Finding 2** (whitespace device_class): `device_class.strip() or None` so accidental whitespace hits the documented clear sentinel instead of HA's validator.
- **Finding 3** (empty options restore): Guarded the test restore with `if original_sensor_options:` so the no-op `options={"sensor": {}}` round-trip is skipped.
- **Finding 4** (last-call-wins comment): Added the one-line comment on the per-domain `entity_entry` reassignment.
- **Finding 5** (options/expose_to collision): Added the expose_to warning sentence to the `ha_set_entity` `options` Field description.

Thanks for the consistency pass.

Patch76 (approved):
Round-3 fixes verified at HEAD `8ae02e42`: all 5 round-2 findings addressed surgically — incremental `options_failure_context` build with gated `entity_entry` injection (`tools_entities.py:332–344`), `device_class` `.strip() or None` parity with `new_device_name` (`tools_entities.py:139–151`), e2e `if original_sensor_options:` restore guard (`test_show_as.py:287`), last-call-wins comment on the `entity_entry` reassignment (`tools_entities.py:352–354`), voice-assistant Field-description warning consistent with the read-side docstring (`tools_entities.py:591–595` / `1148–1150`). Lockdown assertion `assert "entity_entry" not in body` lands on the clean-fail unit test.

Issue #1134, opened 2026-05-05
[FEATURE] Combine the 3 skill-related tools (list_resources, read_resource, ha_get_skill_home_assistant_best_practices) into 1


kingpanther13:
### What would you like to see?

Replace these three tools with a single polymorphic skill tool reading directly from `src/ha_mcp/resources/skills-vendor/skills/`:

- `list_resources` (generated by the `ResourcesAsTools` transform)
- `read_resource` (generated by the `ResourcesAsTools` transform)
- `ha_get_skill_home_assistant_best_practices` (registered by `_register_skill_guidance_tools`)

Tool name TBD during PR work — `ha_skill` is just `ha_<noun>` and breaks the `ha_<verb>_<noun>` convention used elsewhere in the repo.

Tool shape (progressive disclosure preserved across three tiers):

- no args → list bundled skills with frontmatter descriptions
- `skill` arg only → list reference files for that skill
- `skill + file` args → return file content

Tool description carries SKILL.md frontmatter trigger conditions for all bundled skills, preserving the claude.ai-discoverability behavior added in #732.

Code changes:

- Add the new tool (~50 lines of plain disk I/O)
- Drop `add_transform(ResourcesAsTools(self.mcp))` (`server.py:611`)
- Drop `_register_skill_guidance_tools` and `_collect_skill_ref_files` (`server.py:624-714`)
- **Keep** `SkillsDirectoryProvider` registered — resource-capable clients still get native `skill://` URI access; tool-only clients use the new tool
- Update tool-search pinning (`server.py:520-524`) to pin the new tool (replaces the old `_skill_tool_names` extension)

### Why do you need this?

Three tools to deliver one capability (skill content with progressive disclosure) is more catalog surface than the work justifies. The single-tool shape:

- **Reduces idle context** — three tool-catalog entries (~450 tokens, dominated by the per-skill tool's trigger-condition description) collapse to one (~350 tokens; trigger-condition framing dedupes across modes).
- **Preserves progressive disclosure** — three tiers map cleanly to the three arg combinations. Each call only does the I/O for the requested tier; reference files stay on disk until explicitly fetched. Same on-demand model as today.
- **Single description** carrying all skill trigger conditions, instead of three descriptions where the per-skill one duplicates content reachable elsewhere.
[...]

PR #1136, opened 2026-05-05
refactor: always-on skills; rename list/read resource tools with ha_ prefix

### LINKED ISSUE #1133: [FEATURE] Remove ENABLE_SKILLS / ENABLE_SKILLS_AS_TOOLS toggles, surface skill tools in web settings UI

Drop the `ENABLE_SKILLS` and `ENABLE_SKILLS_AS_TOOLS` toggles entirely (env vars, addon `config.yaml` schema entries, Supervisor UI labels, `start.py` env-var setup, and migration markers). Always register `SkillsDirectoryProvider`, the `ResourcesAsTools` transform, and the per-skill guidance tool(s).

Then extend the web settings UI (introduced via #798) so the two `ResourcesAsTools`-generated tools (`list_resources` and `read_resource`) appear there. Today `_get_tool_metadata` reads `server.mcp.local_provider._list_tools()` (`settings_ui.py:185`), which does not include transform-generated tools — the two skill-as-tools entries are invisible in the UI even when enabled. Users who want them off should disable them per-tool from the web UI like any other tool.

The per-skill guidance tool (`ha_get_skill_home_assistant_best_practices`) already surfaces in the UI because it registers via `self.mcp.tool(...)` directly into `local_provider`. No fix needed there.

Affected surfaces (all install methods):

- `src/ha_mcp/config.py` — Pydantic fields + `_skills_dependency` validator
- `src/ha_mcp/server.py:556-622` — drop the conditional gates in `_register_skills`
- `src/ha_mcp/settings_ui.py` — extend `_get_tool_metadata` to include transform-generated tools (likely mirroring the `FEATURE_GATED_TOOLS` stub-injection pattern at L208-226)
- `homeassistant-addon/config.yaml` and `homeassistant-addon-dev/config.yaml` schema entries
- `homeassistant-addon/translations/en.yaml` and `-dev` variant — Supervisor UI toggle labels
- `homeassistant-addon/start.py` — env-var setup + `.skills_as_tools_default_migration_v1` marker logic
- README + `homeassistant-addon/DOCS.md` toggle tables
- `tests/addon/test_skills_config.py` and adjacent

### Why do you need this?

The original justifications for the toggles do not survive scrutiny:

1. **"Avoid conflict with locally-installed skills"** — skills served via MCP and skills installed locally (e.g., `~/.claude/skills/`) live in separate namespaces and never collide. Two skills with the same name from different sources are independent entries to the LLM.
2. **"Don't burn extra context"** — skills are not auto-injected. `list_resources`/`read_resource`/`ha_get_skill_*` use progressive disclosure: list/file-listing returns metadata only; content loads on demand. Disabled skills cost nothing if unused; enabled skills cost only what's actually consulted.

Real friction the toggles create:

- Multiple disable mechanisms (env var, addon Supervisor toggle, migration marker) where one (per-tool visibility config from #798) would suffice
- Two of the three skill-related tools never appear in the web settings UI even when enabled — users have no per-tool way to disable them
- `_skills_dependency` validator silently auto-enables `ENABLE_SKILLS` when `ENABLE_SKILLS_AS_TOOLS=true` (because the latter wraps the former); collapsing to one always-on path removes the implicit dependency

This is the consolidation #798 contemplated ("I'm unsure if I should keep the toggles for enabling skills as resources and enabling skills as tools the same or if they should be part of this new system"). Now that #798 has shipped, retiring these as separate toggles and treating skill tools like any other tool in the visibility UI is the consistent move.

### Additional context

If a follow-up issue to combine the three skill tools into one lands first or simultaneously, the `list_resources`/`read_resource` rename to the `ha_*` prefix becomes throwaway work (those tools go away). Worth deciding ordering: if the combine issue lands soon, skip the rename in this PR; if it is deferred, include the rename, which costs an ~80-line `ResourcesAsTools` subclass since FastMCP hardcodes the names at `resources_as_tools.py:63-69, 76-79`.

Migration note: existing users with `ENABLE_SKILLS_AS_TOOLS=false` lose the env-var path. Per-tool visibility config offers equivalent opt-out via the web UI — should be documented in DOCS.md and README.

kingpanther13:
## What does this PR do?

Closes #1133. Removes the `ENABLE_SKILLS` and `ENABLE_SKILLS_AS_TOOLS` toggles entirely (env vars, addon `config.yaml` schema, supervisor labels, `start.py` env setup, the one-time migration marker) and renames the `ResourcesAsTools`-generated pair to `ha_list_resources` / `ha_read_resource` so they follow `ha_<verb>_<noun>` like every other tool. The web settings UI (#960) is now the single mechanism for hiding either tool — extended `_get_tool_metadata` with stub injection because FastMCP's `local_provider` doesn't see transform-appended tools.

### Why retire the toggles

The original justifications don't survive scrutiny:

- **"Avoid conflict with locally-installed skills"** — skills served via `skill://` URIs and skills installed under `~/.claude/skills/` live in separate namespaces and never collide. Same name from two sources is just two independent entries to the LLM. Searched #679 / #806 / #960 — this rationale was never actually documented as a motivation; it appears to be folk knowledge.
- **"Don't burn extra context"** — confirmed in #679's PR body: resources are not auto-injected, `list_resources` returns ~200 tokens of metadata, and the actual ~70KB of skill content only loads on `resources/read`. Disabled skills cost nothing if they're not consulted.

What actually motivated the toggles: a generic opt-out for #679, then flipped to default-true in #806 to preserve documentation access for tool-only clients after four doc tools were converted to skill references. With #960 shipped, every tool has its own per-tool toggle — the global gates duplicate that mechanism and have the additional bug that two of the three skill-related tools (`list_resources`, `read_resource`) never showed up in the UI even when enabled, because `_get_tool_metadata` reads from `local_provider._list_tools()` which doesn't include transform-generated tools.

### Migration

Existing users with `ENABLE_SKILLS_AS_TOOLS=false` lose the env-var path. Equivalent opt-out is available per-tool via the web settings UI, called out in README and `homeassistant-addon-dev/DOCS.md`. The same outcome is achievable via the alternative, so this is not a breaking change per `AGENTS.md`'s definition.

### Addendum

The real motivation for the toggles initially was that I didn't want to add too many new tools at a time that we were trying to consolidate tools. Now that we have user configurable tools and that we have significantly consolidated a lot of tools this really isn't necessary anymore; furthermore we have a growing number of things that require the skills tools, which has lead to friction in other areas. We did make #948 to make them default to on, but then I realized there really is no sense in turning them off in the first place. If someone truly still wants them off they can just simply go into #960 to turn them off, but it'll significantly reduce the capability of their MCP (same as turning the toggles off now).

## Type of change
- [x] ✨ New feature
- [x] 🔧 Maintenance/refactor

## Testing

## Checklist


src/ha_mcp/server.py
```diff
@@ -16,18 +16,68 @@
...
+    ) -> Tool | None:
+        if name == self.LIST_TOOL_NAME:
+            return self._make_list_resources_tool().model_copy(
+                update={"name": self.LIST_TOOL_NAME}
+            )
+        if name == self.READ_TOOL_NAME:
+            return self._make_read_resource_tool().model_copy(
+                update={"name": self.READ_TOOL_NAME}
+            )
+        return await call_next(name, version=version)
```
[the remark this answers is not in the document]
kingpanther13: 📝 Not addressing — verified against fastmcp 3.2.4 source. The `Transform` base class (`fastmcp/server/transforms/__init__.py:69-220`) defines `list_tools`/`get_tool`/etc. but no `call_tool` method. Tool dispatch in `FastMCP.call_tool` (`server/server.py:1180-1245`) calls `await self.get_tool(name)` to walk the transform chain (where this subclass's rename branches fire), then `await tool._run(arguments)`. `FunctionTool.run` (`tools/function_tool.py:240`) invokes `self.fn` (the wrapped function pointer); renaming via `model_copy(update={"name": ...})` doesn't touch `fn`.

This is the same pattern fastmcp's own `Namespace` transform uses (`transforms/namespace.py:97-113`) — `model_copy` rename in both `list_tools` and `get_tool`, no `call_tool` override. If the pattern were broken, `Namespace` would be too.

Confirmed end-to-end by the new `test_ha_list_resources_invocation` and `test_ha_read_resource_invocation` e2e tests in 50ab72be, which actually call `mcp_client.call_tool("ha_list_resources", {})` through the full dispatch path.

src/ha_mcp/server.py
```diff
@@ -16,18 +16,68 @@
...
 from mcp.types import Icon
 
 from .config import _PACKAGE_VERSION, get_global_settings
 from .tools.enhanced import EnhancedToolsMixin
 from .transforms import DEFAULT_PINNED_TOOLS
 
 if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+    from fastmcp.server.transforms import GetToolNext
```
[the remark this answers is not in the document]
kingpanther13: 📝 Not addressing — `CallToolNext` is not exported by `fastmcp.server.transforms`. The module's `__all__` lists `Namespace, PromptsAsTools, ResourcesAsTools, ToolTransform, Transform, VersionFilter, VersionSpec, Visibility, is_enabled` (verified against fastmcp 3.2.4). The `Transform` protocol declares `list_tools(tools)` and `get_tool(name, call_next, *, version)` only — no `call_tool` override surface (see thread above). The current `GetToolNext` import is correct and sufficient.

kingpanther13:
✅ Addressed pr-review-toolkit feedback in 50ab72be. CI green.

**Implemented:**
- `read_resource` → `ha_read_resource` in skill-guidance docstring, tool description, handler `how_to_use`, and the s13 UAT story.
- `HaResourcesAsTools.list_tools` rewritten to scan-by-name over the full upstream sequence (no more `result[-2:]` slicing). Warns if the matched count != 2 so a future fastmcp regression is loud at boot rather than silent.
- Class constants (`HaResourcesAsTools.LIST_TOOL_NAME` / `READ_TOOL_NAME`) are the single source of truth for the pinned list and search-tool description; the `TRANSFORM_GENERATED_TOOLS` ↔ class-constants invariant is now covered by a dedicated unit test.
- Upgrade-fragility note added to the `HaResourcesAsTools` docstring flagging the dependency on fastmcp's `_make_*_tool` private factories.
- New `tests/src/unit/test_ha_resources_as_tools.py`: rename happy-path for `list_tools`/`get_tool`, fall-through for unprefixed and unrelated names, and a drift-warning test that monkey-patches the base class to drop one of the appended tools.
- New e2e tests `test_ha_list_resources_invocation` / `test_ha_read_resource_invocation` actually invoke the renamed tools via `mcp_client.call_tool(...)` to verify dispatch routing through the full FastMCP path (`call_tool` → `get_tool` → `tool._run`).
- `best_practice_checker` docstring tightened to drop the implication that a canonical alternative prefix still exists in the module.

**Not addressing (out of scope for this PR):**
- Class rename `HaResourcesAsTools` → `RenamedResourcesAsTools` — judgment call; the docstring already explains the intent, and this matches the FastMCP convention (`Namespace`, `PromptsAsTools`, `ResourcesAsTools`).
- TypedDict refactor for `TRANSFORM_GENERATED_TOOLS` / `FEATURE_GATED_TOOLS` shared shape — would touch the existing feature-gated dict and is unrelated to this PR's scope.
- Boot-time skill-system summary log line — separate observability improvement; skill registration failures are still logged.
- Stale `/data/.skills_as_tools_default_migration_v1` marker cleanup on existing installs — harmless litter; can be filed as a separate hygiene PR if it bothers anyone.

**Gemini bot comments:** both are false positives based on a non-existent fastmcp API (`Transform.call_tool` / `CallToolNext`). Verified against fastmcp 3.2.4 source — the `Transform` base defines `list_tools` / `get_tool` only; tool dispatch in `FastMCP.call_tool` (`server.py:1180-1245`) goes through `get_tool` → `tool._run` → `FunctionTool.run` → `self.fn` (function pointer, name-independent). Same `model_copy(update={"name": ...})` pattern is used by fastmcp's own `Namespace` transform. The new e2e invocation tests prove the rename works end-to-end. Replies posted on both threads and threads resolved.

kingpanther13:
✅ Update: address remaining review items in 4c413712. CI green.

Folded in three follow-ups previously called "out of scope":

**Marker cleanup (`homeassistant-addon/start.py`):** new `cleanup_stale_migration_marker` runs once at boot to `unlink(missing_ok=True)` the orphaned `/data/.skills_as_tools_default_migration_v1` file left over from the previous version's migration helper. Best-effort — a stale dotfile is harmless if removal fails.

**Skill registration summary log (`server._register_skills`):** four phases now track per-phase status (`provider`, `transform`, `guidance_tools`) and emit one summary line via `_log_skill_registration_summary` — `info` when both provider and transform succeeded, `warning` otherwise. Without the toggle, every install hits this code path on every boot, so a single line operators can grep for beats reconstructing state from scattered `logger.exception` calls. Per-phase exception logs remain for stack traces.

**`ToolStub` TypedDict refactor (`settings_ui.py`):** both `TRANSFORM_GENERATED_TOOLS` and `FEATURE_GATED_TOOLS` now have a shared `ToolStub` shape with `NotRequired` `disabled_by` / `readOnlyHint` / `destructiveHint`. Annotation values converted from `"true"` strings to actual `bool`. New `_render_stub` helper collapses the previously-duplicated stub-build loops into one function, so a misspelled key fails type-checking instead of silently producing an entry with the wrong shape. Behavioral output is unchanged; existing unit tests still pass.

Everything from the original feedback batch (commit 50ab72be) plus these three is now addressed. The only remaining "not addressed" item is the class rename `HaResourcesAsTools` → `RenamedResourcesAsTools` (judgment call — the current name matches FastMCP's `Namespace`/`PromptsAsTools`/`ResourcesAsTools` convention).

homeassistant-addon-dev/DOCS.md
```diff
@@ -14,8 +14,6 @@ The dev add-on uses the same configuration as the stable version. See the main a
 |--------|-------------|---------|
 | `backup_hint` | Backup strength preference | `normal` |
 | `secret_path` | Custom secret path (optional) | auto-generated |
```
Patch76, line 16: README.md adds a sentence at line 230 telling users to disable tools via the web settings UI, but this DOCS.md (the file Supervisor renders for the dev addon's "Documentation" tab) just removes the two `enable_skills*` rows with no replacement prose. Existing users who had `enable_skills_as_tools: false` saved will see Supervisor strip the field on update, both tools become enabled, and there's nothing in this doc that explains why or how to restore the opt-out.

Suggested: one line under the options table — "Removed in 7.4.x: `enable_skills` and `enable_skills_as_tools`. To hide either tool, use the web settings UI."

src/ha_mcp/server.py
```diff
@@ -590,36 +667,58 @@ def _register_skills(self) -> None:
...
-
-        try:
-            self.mcp.add_transform(ResourcesAsTools(self.mcp))
-            logger.info("Skills also exposed as tools (ResourcesAsTools)")
+            status["transform"] = "ok"
         except Exception:
             logger.exception(
                 "Failed to expose skills as tools (resources still available)"
             )
+            status["transform"] = "failed"
```
Patch76, line 695: When `HaResourcesAsTools(self.mcp)` raises, the failure surfaces only as the `logger.warning` summary at line 721 plus a `logger.exception` for the stack trace. Resource-capable clients still see skills (Phase 2 succeeded), but tool-only clients (claude.ai and similar) lose the renamed pair from their catalog with no MCP-protocol-level error or in-instructions hint — the degradation is operator-visible in logs but invisible at the protocol layer.

Worth documenting in `_register_skills`'s docstring (lines 624-636) that this failure mode degrades silently for tool-only clients, so an operator hitting the warning understands the user-facing effect without tracing through three independent log lines.

src/ha_mcp/server.py
```diff
@@ -16,18 +16,92 @@
...
+            renamed.append(tool.model_copy(update={"name": new_name}))
+            matches += 1
+        if matches != len(self._RENAMES):
+            logger.warning(
+                "HaResourcesAsTools: expected to rename %d tools (%s) but "
+                "matched %d in the upstream tool list — fastmcp's "
+                "ResourcesAsTools contract may have changed",
+                len(self._RENAMES),
+                ", ".join(self._RENAMES),
+                matches,
```
Patch76, line 84: `if matches != len(self._RENAMES): logger.warning(...)` is permissive in both directions: `matches > 0 and matches < len(self._RENAMES)` is plausibly a partial fastmcp upgrade transition (one tool renamed, one not yet), but `matches == 0` is unambiguously a regression — the upstream `_make_*` factories are gone or moved, no rename happened, and the unprefixed `list_resources` / `read_resource` ship in the catalog. Operators who don't watch boot logs won't notice the leak.

Consider a hard fail on `matches == 0` (raise so registration aborts loudly) and keep the warning for the partial-mismatch case. The fastmcp version pin in `pyproject.toml` is the real safety net for routine upgrades, but defensive-by-default for the catastrophic path is cheap.

src/ha_mcp/server.py
```diff
@@ -590,36 +667,58 @@ def _register_skills(self) -> None:
...
         # server instructions (e.g., claude.ai). The tool description contains
         # the trigger conditions so the AI sees them in the tool listing.
         # Names stored for pinning in search transforms (always-visible).
         self._register_skill_guidance_tools(skills_dir)
+        status["guidance_tools"] = len(self._skill_tool_names)
+
+        self._log_skill_registration_summary(status)
+
+    @staticmethod
+    def _log_skill_registration_summary(status: dict[str, str | int]) -> None:
```
Patch76, line 707: This helper plus `cleanup_stale_migration_marker` (`homeassistant-addon/start.py:185`) are both new in `4c41371` and neither is exercised in the test suite — `git grep` returns zero hits for either name across `tests/`. Both are best-effort and small, but:

- `cleanup_stale_migration_marker` runs on every boot; a defect (e.g., `unlink` raising on a permission-restricted `/data` mount despite the `OSError` catch) only manifests in production.
- `_log_skill_registration_summary` has four meaningful states (`(skipped, skipped)`, `(failed, *)`, `(ok, failed)`, `(ok, ok)`) and the warning-vs-info branch matters for operators grepping logs.

A small unit test per function would lock the behaviour without much cost.

src/ha_mcp/server.py
```diff
@@ -16,18 +16,92 @@
...
+    Upgrade fragility: depends on FastMCP's ``_make_list_resources_tool`` /
+    ``_make_read_resource_tool`` private factories and on the names
+    ``list_resources`` / ``read_resource`` produced by them. A FastMCP
+    upgrade that renames either factory or either tool name will require
+    a matching update here. ``list_tools`` logs a warning if the rename
+    fails to match exactly two tools so the regression is loud at boot.
+    """
+
+    LIST_TOOL_NAME = "ha_list_resources"
+    READ_TOOL_NAME = "ha_read_resource"
```
Patch76, line 56: `READ_TOOL_NAME = "ha_read_resource"` follows fastmcp's underlying `read_resource` verb, which is intuitive for resource-style URI access — but `.gemini/styleguide.md:64-76` enumerates the allowed verbs as `get / list / search / set / delete / call / manage` and explicitly says "Flag MEDIUM severity if tools don't follow this pattern." Either add `read` to the styleguide list with a one-line rationale ("read for resource-style URI access") or rename to `ha_get_resource`. Otherwise expect Gemini to re-flag the deviation on every adjacent PR that touches the catalog.

src/ha_mcp/server.py
```diff
@@ -590,36 +667,58 @@ def _register_skills(self) -> None:
...
+        """
+        provider = status.get("provider")
+        transform = status.get("transform")
+        guidance = status.get("guidance_tools", 0)
+
+        message = (
+            "Skill system summary: provider=%s, transform=%s, guidance_tools=%d"
+        )
+        args = (provider, transform, guidance)
+        if provider == "ok" and transform == "ok":
```
Patch76: The summary gates info-vs-warning purely on `provider == "ok" and transform == "ok"`, ignoring `guidance_tools` count. When the skills directory is empty (no `SKILL.md` files vendored, or the bundle was stripped at build time), `_register_skill_guidance_tools` runs to completion with `_skill_tool_names == []` and `status["guidance_tools"] == 0` — the summary still emits `info` because both prior phases succeeded, even though the user-facing effect is "skill system is technically up but exposes nothing".

Tightening to `if provider == "ok" and transform == "ok" and guidance > 0:` pushes the empty-directory case into the warning branch where operators expect to grep.

tests/src/unit/test_ha_resources_as_tools.py
```diff
@@ -0,0 +1,110 @@
...
+
+from ha_mcp.server import HaResourcesAsTools
+from ha_mcp.settings_ui import TRANSFORM_GENERATED_TOOLS
+
+
+def test_transform_generated_tool_names_match_class_constants():
+    """The settings UI stub keys must match the class constants so a future
+    rename of either tool can't drift the two sides apart."""
+    expected = {HaResourcesAsTools.LIST_TOOL_NAME, HaResourcesAsTools.READ_TOOL_NAME}
+    assert set(TRANSFORM_GENERATED_TOOLS) >= expected, (
```
Patch76: `assert set(TRANSFORM_GENERATED_TOOLS) >= expected` lets `TRANSFORM_GENERATED_TOOLS` carry extra keys without a corresponding rename — exactly the drift the test is meant to catch. If a future PR adds a stub for a tool that isn't actually transform-renamed, this test still passes.

Either tighten to `==` (if `TRANSFORM_GENERATED_TOOLS` should always equal exactly the renamed set) or add a complementary assertion that every key in `TRANSFORM_GENERATED_TOOLS` corresponds to a rename target — `set(TRANSFORM_GENERATED_TOOLS) <= set(HaResourcesAsTools._RENAMES.values())` would catch the drift in the other direction.

Patch76 (commented):
Re-review of HEAD `4c413712`. Solid refactor — three-commit iteration shows good quality progression (base feat → harden rename → typing + marker cleanup). Breaking-change call holds under `AGENTS.md:613-622` (functionality preserved via the per-tool web settings UI), and the squash-subject `feat:` correctly reflects the user-visible footprint that `refactor:` (patch bump per `pyproject.toml [tool.semantic_release]`) wouldn't capture.

Four mediums plus three nits inline. The current `dirty` merge state is a single-file, single-hunk conflict against master's `5de8679` on `homeassistant-addon-dev/DOCS.md` — the `enable_tool_search` row got a Sonnet/Opus warning while this PR drops the two `enable_skills*` rows above it. Mechanical to resolve.

Note: the README migration sentence at line 230 covers the public site, so `homeassistant-addon-dev/DOCS.md` is the only place that still reads silently for users with `enable_skills_as_tools: false` saved.

homeassistant-addon-dev/DOCS.md
kingpanther13: ✅ Fixed in fd39c73c. Added the migration line under the options table — kept the "Removed in 7.4.x" framing and pointed at the Tool Settings Web UI section right below it. Verified the line lands between the table and the Beta-options paragraph.

src/ha_mcp/server.py
kingpanther13: ✅ Fixed in 2a3c69c4. Added a paragraph to `_register_skills`'s docstring spelling out that Phase 3 transform failure degrades silently for tool-only clients (claude.ai etc.) — only the warning summary signals it at the protocol layer.

src/ha_mcp/server.py
kingpanther13: 📝 Not adding the hard-fail. Re-traced the leak path:

`matches == 0` means `super().list_tools()` returned NO tools whose names equal `list_resources` / `read_resource`. For the unprefixed names to ship in the catalog, those exact strings would have to appear in super's output — which would mean `matches > 0`, not `matches == 0`.

The actual leak scenario you're guarding against is `matches == 1`: one factory still produces the original name (gets renamed), one was renamed upstream (doesn't appear in `_RENAMES.keys()` so doesn't ship under the unprefixed name either). That's already caught by the existing `matches != len(self._RENAMES)` warning.

`matches == 0` corresponds to "factories renamed/removed wholesale" — the unprefixed names cannot leak from this code path because super's output didn't contain them. Catalog gets either nothing extra or whatever fastmcp's new factory names are. The warning still fires, which is the right signal level for that case.

The fastmcp version pin in `pyproject.toml` is the routine-upgrade safety net; the existing warning covers the genuinely-leaky path. Hard-fail would just abort registration loudly without preventing a leak that this branch can't actually cause.

src/ha_mcp/server.py
kingpanther13: ✅ Fixed in 2a3c69c4. Added `TestLogSkillRegistrationSummary` (six tests covering all four meaningful states plus the new guidance>0 gate and the missing-key fallback) in `tests/src/unit/test_skills_instructions.py`, and `TestCleanupStaleMigrationMarker` (three tests: removal / no-op / OSError-swallow on permission-restricted /data) in `tests/addon/test_addon_startup.py`. The OSError test is the production-only case you flagged.

src/ha_mcp/server.py
kingpanther13: 📝 Deferring to #1134. That issue replaces all three skill-related tools (`ha_list_resources`, `ha_read_resource`, `ha_get_skill_home_assistant_best_practices`) with a single polymorphic skill tool, which makes the `read` verb question moot. Renaming to `ha_get_resource` (or extending the styleguide) on this PR would be churn for a name that's about to disappear.

src/ha_mcp/server.py
kingpanther13: ✅ Fixed in 2a3c69c4. `_log_skill_registration_summary` now requires `guidance > 0` for the info branch — the empty-skills-dir / all-SKILL.md-fail-to-parse case now lands in the warning branch where operators grep. New unit test `test_logs_warning_when_guidance_zero_despite_ok_phases` locks the behaviour. Updated docstring to explain the gate.

tests/src/unit/test_ha_resources_as_tools.py
kingpanther13: ✅ Fixed in 2a3c69c4. Tightened to `==` against `set(HaResourcesAsTools._RENAMES.values())` so the assertion catches drift in both directions — orphan stub or missing stub both fail. Used `_RENAMES.values()` directly rather than the `LIST_TOOL_NAME`/`READ_TOOL_NAME` constants so the source of truth is unambiguous.

kingpanther13:
## Implementation Summary (Patch76 review pass)

Thanks for the re-review @Patch76 — addressed in 2a3c69c4 + 2e60bde0, plus the rebase that pulled in master's Sonnet/Opus warning on `enable_tool_search`.

**Choices Made:**
- **DOCS.md migration line** — added under the options table with a "(see below)" tie-in to the Tool Settings Web UI section already present in DOCS.md.
- **Summary gating** — added `guidance > 0` to the info-vs-warning gate in `_log_skill_registration_summary`. Catches the "shipped but exposes nothing" case where both prior phases succeed but no SKILL.md is reachable.
- **Test assertion** — tightened `TRANSFORM_GENERATED_TOOLS` drift check to `==` against `_RENAMES.values()` directly (source of truth), catches both orphan stubs and missing stubs.
- **Docstring** — `_register_skills` now explicitly notes the silent-degradation failure mode for tool-only clients.
- **New unit tests** — `TestLogSkillRegistrationSummary` (six tests, all four meaningful states + new gate + missing-key fallback) and `TestCleanupStaleMigrationMarker` (removal / no-op / OSError-swallow on permission-restricted /data).

**Push-backs / defers:**
- **Hard-fail on \`matches == 0\`** — declined. \`matches == 0\` means \`super().list_tools()\` returned no tools whose names match \`_RENAMES.keys()\`, so the unprefixed-leak scenario this would guard against can't actually occur from that branch. The genuine leak path (\`matches == 1\`, partial upstream rename) is already covered by the existing warning. Full trace in the inline thread.
- **\`READ_TOOL_NAME\` vs \`.gemini/styleguide.md\`** — deferred to #1134. That issue replaces all three skill-related tools with a single polymorphic skill tool, so the \`read\` verb question disappears.

**Problems Encountered:**
- PR was DIRTY against master after #1140 landed the Sonnet/Opus warning on \`enable_tool_search\`. Rebased onto master and resolved the single-hunk conflict in \`homeassistant-addon-dev/DOCS.md\` (kept master's enriched description, dropped the two \`enable_skills*\` rows, added the migration line).
- Initial new test \`test_missing_guidance_key_treated_as_zero\` failed CI because \`LogRecord.message\` is set by caplog to the already-formatted string; applying \`% args\` over-formats. Fixed in 2e60bde0 by using \`getMessage()\`.

**Suggested Improvements:**
- None additional from this pass. #1134 will subsume the renamed tools entirely.

Patch76 (approved):
Round-3 fixes verified at HEAD `2e60bde0`: 5/7 round-2 findings code-fixed (DOCS migration line, `_register_skills` asymmetric-degradation paragraph, `TestCleanupStaleMigrationMarker` + `TestLogSkillRegistrationSummary` coverage, `guidance > 0` summary gate, `==`-tightened drift assertion). The two declines (M3 `matches==0` hard-fail, N1 `READ_TOOL_NAME` styleguide deviation deferred to #1134) hold against scrutiny.

NIT — `server.py:561`: parenthetical reads `ResourcesAsTools (list_resources/read_resource)` (FastMCP base-class names); the actual pin extension three lines below uses `HaResourcesAsTools.LIST_TOOL_NAME` / `READ_TOOL_NAME`. Trivial single-line cleanup.

PR #1138, opened 2026-05-05
fix: survive read-only filesystems at startup

### LINKED ISSUE #1125: Updating to :latest (dev) is failing
":latest" image does not start.

```
> docker logs ha-mcp
Traceback (most recent call last):
  File "/app/.venv/bin/ha-mcp-web", line 10, in <module>
    sys.exit(main_web())
             ~~~~~~~~^^
  File "/app/src/ha_mcp/__main__.py", line 758, in main_web
    _run_http_server("http", default_port=8086)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/src/ha_mcp/__main__.py", line 739, in _run_http_server
    register_browser_landing(_get_mcp(), path)
                             ~~~~~~~~^^
  File "/app/src/ha_mcp/__main__.py", line 323, in _get_mcp
    _server = _create_server()
  File "/app/src/ha_mcp/__main__.py", line 309, in _create_server
    return HomeAssistantSmartMCPServer()
  File "/app/src/ha_mcp/server.py", line 93, in __init__
    self._initialize_server()
    ~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/app/src/ha_mcp/server.py", line 151, in _initialize_server
    self._apply_settings_visibility()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/app/src/ha_mcp/server.py", line 335, in _apply_settings_visibility
    config = load_tool_config(self.settings)
  File "/app/src/ha_mcp/settings_ui.py", line 125, in load_tool_config
    path = _get_config_path()
  File "/app/src/ha_mcp/settings_ui.py", line 119, in _get_config_path
    home_dir.mkdir(parents=True, exist_ok=True)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/pathlib/_local.py", line 722, in mkdir
    os.mkdir(self, mode)
    ~~~~~~~~^^^^^^^^^^^^
OSError: [Errno 30] Read-only file system: '/.ha-mcp'
Traceback (most recent call last):
  File "/app/.venv/bin/ha-mcp-web", line 10, in <module>
    sys.exit(main_web())
             ~~~~~~~~^^
  File "/app/src/ha_mcp/__main__.py", line 758, in main_web
    _run_http_server("http", default_port=8086)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/src/ha_mcp/__main__.py", line 739, in _run_http_server
    register_browser_landing(_get_mcp(), path)
                             ~~~~~~~~^^
  File "/app/src/ha_mcp/__main__.py", line 323, in _get_mcp
    _server = _create_server()
  File "/app/src/ha_mcp/__main__.py", line 309, in _create_server
    return HomeAssistantSmartMCPServer()
  File "/app/src/ha_mcp/server.py", line 93, in __init__
    self._initialize_server()
    ~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/app/src/ha_mcp/server.py", line 151, in _initialize_server
    self._apply_settings_visibility()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/app/src/ha_mcp/server.py", line 335, in _apply_settings_visibility
    config = load_tool_config(self.settings)
  File "/app/src/ha_mcp/settings_ui.py", line 125, in load_tool_config
    path = _get_config_path()
  File "/app/src/ha_mcp/settings_ui.py", line 119, in _get_config_path
    home_dir.mkdir(parents=True, exist_ok=True)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/pathlib/_local.py", line 722, in mkdir
    os.mkdir(self, mode)
    ~~~~~~~~^^^^^^^^^^^^
OSError: [Errno 30] Read-only file system: '/.ha-mcp'
Traceback (most recent call last):
  File "/app/.venv/bin/ha-mcp-web", line 10, in <module>
    sys.exit(main_web())
             ~~~~~~~~^^
  File "/app/src/ha_mcp/__main__.py", line 758, in main_web
    _run_http_server("http", default_port=8086)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/src/ha_mcp/__main__.py", line 739, in _run_http_server
    register_browser_landing(_get_mcp(), path)
                             ~~~~~~~~^^
  File "/app/src/ha_mcp/__main__.py", line 323, in _get_mcp
    _server = _create_server()
  File "/app/src/ha_mcp/__main__.py", line 309, in _create_server
    return HomeAssistantSmartMCPServer()
  File "/app/src/ha_mcp/server.py", line 93, in __init__
    self._initialize_server()
    ~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/app/src/ha_mcp/server.py", line 151, in _initialize_server
    self._apply_settings_visibility()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/app/src/ha_mcp/server.py", line 335, in _apply_settings_visibility
    config = load_tool_config(self.settings)
  File "/app/src/ha_mcp/settings_ui.py", line 125, in load_tool_config
    path = _get_config_path()
  File "/app/src/ha_mcp/settings_ui.py", line 119, in _get_config_path
    home_dir.mkdir(parents=True, exist_ok=True)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/pathlib/_local.py", line 722, in mkdir
    os.mkdir(self, mode)
    ~~~~~~~~^^^^^^^^^^^^
OSError: [Errno 30] Read-only file system: '/.ha-mcp'
Traceback (most recent call last):
  File "/app/.venv/bin/ha-mcp-web", line 10, in <module>
    sys.exit(main_web())
             ~~~~~~~~^^
  File "/app/src/ha_mcp/__main__.py", line 758, in main_web
    _run_http_server("http", default_port=8086)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/src/ha_mcp/__main__.py", line 739, in _run_http_server
    register_browser_landing(_get_mcp(), path)
                             ~~~~~~~~^^
  File "/app/src/ha_mcp/__main__.py", line 323, in _get_mcp
    _server = _create_server()
  File "/app/src/ha_mcp/__main__.py", line 309, in _create_server
    return HomeAssistantSmartMCPServer()
  File "/app/src/ha_mcp/server.py", line 93, in __init__
    self._initialize_server()
    ~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/app/src/ha_mcp/server.py", line 151, in _initialize_server
    self._apply_settings_visibility()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/app/src/ha_mcp/server.py", line 335, in _apply_settings_visibility
    config = load_tool_config(self.settings)
  File "/app/src/ha_mcp/settings_ui.py", line 125, in load_tool_config
    path = _get_config_path()
  File "/app/src/ha_mcp/settings_ui.py", line 119, in _get_config_path
    home_dir.mkdir(parents=True, exist_ok=True)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/pathlib/_local.py", line 722, in mkdir
    os.mkdir(self, mode)
    ~~~~~~~~^^^^^^^^^^^^
OSError: [Errno 30] Read-only file system: '/.ha-mcp'

```

Switching back to :stable works.

kingpanther13:
## What does this PR do?

Fixes #1125 — `ghcr.io/homeassistant-ai/ha-mcp:latest` crashes at startup
under hardened Docker setups (`read_only: true`, `user: 1000:1000`,
`tmpfs: /tmp`):

```
OSError: [Errno 30] Read-only file system: '/.ha-mcp'
  in _get_config_path() → home_dir.mkdir(parents=True, exist_ok=True)
```

Two contributing factors:

1. The Dockerfile didn't set `ENV HOME`, so under `USER mcpuser` Docker
   left `HOME=/` ([moby/moby#2968](https://github.com/moby/moby/issues/2968)).
   `Path.home()` resolved to `/`, ha-mcp tried to mkdir `/.ha-mcp`, and
   `read_only: true` made that fatal. Even on writable filesystems this
   silently polluted the container's filesystem root with a `/.ha-mcp/`
   directory.
2. `_get_config_path()` (added in #960) unconditionally `mkdir`s the
   home dir on every server start, with no fallback.

### Coordinated changes

**Shared resolver** (`src/ha_mcp/utils/data_paths.py`, new):
single source of truth for "where does ha-mcp write its persistent files".

| # | Source                       | Used for                                |
|---|------------------------------|-----------------------------------------|
| 1 | `HA_MCP_CONFIG_DIR` env var  | Hardened Docker bind-mount override     |
| 2 | `/data` (when `SUPERVISOR_TOKEN` set) | Home Assistant add-on persistent dir |
| 3 | `~/.ha-mcp`                  | Standard install                        |
| 4 | `<tempdir>/ha-mcp`           | Last-resort fallback (server still runs, no persistence) |

The resolved dir is memoized at module level so the fallback warning
emits **once at startup** instead of on every save/load HTTP request.

**`settings_ui`**:
- `_get_config_path` becomes a thin `get_data_dir() / "tool_config.json"` wrapper.
- `load_tool_config` replaces the `path.exists()` + `read_text()` two-step with a single `try: read_text() except OSError:`. `Path.exists()` only swallows `ENOENT/ENOTDIR/EBADF/ELOOP`, so `EACCES` (e.g. `HA_MCP_CONFIG_DIR` pointing at an existing 0700 dir owned by another UID) would re-introduce the same class of crash. Confirmed empirically: `errno=13` propagates from `pathlib.Path.exists()`.
- `save_tool_config` now returns `bool`, and the `_save_tools` HTTP route returns 500 when the write fails. The JS already handles `!resp.ok` with "Save failed!" — it was never given the chance.

**`utils/usage_logger`**: switch the default log path from `Path.home() / ".ha-mcp" / "logs"` to `get_data_dir() / "logs"`. Logs now follow the same precedence as the tool config — `HA_MCP_CONFIG_DIR` works for them too, and add-on mode lands in `/data/logs/` (the supervisor's persistent location, was previously ephemeral `~/.ha-mcp/logs/`).

**`Dockerfile`**:
- `ENV HOME=/home/mcpuser` so `Path.home()` resolves correctly under the default user.
- `chmod 0755 /home/mcpuser` because the base image's `/etc/login.defs` sets `HOME_MODE=0700`; that's fine when the container runs as mcpuser, but un-traversable when users override `--user UID:GID` (the issue reporter does so) — anything that stats a path under HOME then raises `PermissionError`. Mode 0755 keeps write access restricted to mcpuser.

### Tests

| File | Class | What it pins |
|---|---|---|
| `test_data_paths.py` (new) | `TestPriorityOrder` | `HA_MCP_CONFIG_DIR > /data > ~/.ha-mcp` |
| | `TestFallbacks` | Issue #1125 regression; `HA_MCP_CONFIG_DIR` mkdir-fail chains to tmpdir |
| | `TestMemoization` | Single warning per startup; resolver re-runs when cache cleared |
| `test_settings_ui.py` | `TestConfigPath` | Wrapper sanity; `load_tool_config` doesn't crash on EACCES (same-class regression guard) |
| | `TestSaveToolConfig` (new) | bool return contract |
| | `TestSaveToolsValidation` | 500 surfaced when save fails |
| `test_usage_logger.py` | `TestUsageLoggerDefaults` | Honors `HA_MCP_CONFIG_DIR` |

Full unit suite: **1684 passed, 1 skipped** (numpy not installed, unrelated). `ruff check` clean. `mypy src/` clean.

### Local Docker verification

Built locally and confirmed three real scenarios:

| Scenario                                                          | Before | After |
|-------------------------------------------------------------------|--------|-------|
| Reporter's compose: `--read-only --user 1000:1000 --tmpfs /tmp`   | crash  | runs (config falls back to `/tmp/ha-mcp/`)  |
| Default user (no overrides)                                       | "works" but creates `/.ha-mcp` 🚨 | works (creates `~/.ha-mcp`) |
| `HA_MCP_CONFIG_DIR=/data/ha-mcp` with bind-mount + read-only      | n/a    | runs, settings + logs persist to bind-mount |

In all "after" cases the FastMCP banner displays and uvicorn binds `http://0.0.0.0:8086`. The original stack trace from #1125 no longer reproduces.

### Release note

Users who ran `:latest` between #960 (2026-05-02) and this fix and had a writable `/` (default Docker, no `read_only: true`) will have had their settings UI tool config silently written to `/.ha-mcp/tool_config.json` at the filesystem root, because Docker leaves `HOME=/` under `USER mcpuser` without an explicit `ENV HOME` (moby/moby#2968). After this fix the resolver writes to `~/.ha-mcp` (or `HA_MCP_CONFIG_DIR` / `/data` when applicable) and no code path reads from `/.ha-mcp/` — any per-tool enable/disable/pin choices made on broken `:latest` need to be recreated in the settings UI after upgrading. (Users who hit the actual #1125 crash had nothing persisted, so nothing to migrate.)

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing
- [x] All automated tests pass (`uv run pytest`) — 1684 passed, 1 skipped
- [x] Code follows style guidelines (`uv run ruff check`) — and `mypy src/` clean

## Checklist
- [x] I have updated documentation if needed — `HA_MCP_CONFIG_DIR` is documented in the resolver's docstring; happy to surface it in `README.md` or the setup wizard if you'd like.

### DIFF
--- a/src/ha_mcp/settings_ui.py
+++ b/src/ha_mcp/settings_ui.py
@@ -21,6 +21,7 @@ from starlette.responses import HTMLResponse, JSONResponse
 
 from .errors import ErrorCode, create_error_response
 from .transforms import DEFAULT_PINNED_TOOLS
+from .utils.data_paths import get_data_dir
 
 if TYPE_CHECKING:
     from fastmcp import FastMCP
@@ -112,23 +113,37 @@ def _is_addon() -> bool:
 
 
 def _get_config_path() -> Path:
-    """Return the path to the tool config JSON file."""
-    if _is_addon():
-        return Path("/data") / "tool_config.json"
-    home_dir = Path.home() / ".ha-mcp"
-    home_dir.mkdir(parents=True, exist_ok=True)
-    return home_dir / "tool_config.json"
+    """Return the path to the tool config JSON file.
+
+    Delegates directory resolution to :func:`utils.data_paths.get_data_dir`,
+    which handles ``HA_MCP_CONFIG_DIR`` override, add-on ``/data``,
+    home-dir, and tmpdir fallback (memoized).
+    """
+    return get_data_dir() / "tool_config.json"
 
 
 def load_tool_config(settings: Settings | None = None) -> dict[str, Any]:
     """Load persisted tool config, seeding from env vars if no file exists."""
     path = _get_config_path()
-    if path.exists():
+    # ``Path.exists()`` only swallows ``ENOENT/ENOTDIR/EBADF/ELOOP``; an
+    # ``EACCES`` (e.g. ``HA_MCP_CONFIG_DIR`` pointing at a dir that exists
+    # but isn't readable by the runtime UID) propagates. Read directly and
+    # treat ``FileNotFoundError`` as "no config yet"; log other ``OSError``s.
+    try:
+        raw = path.read_text()
+    except FileNotFoundError:
+        raw = None
+    except OSError:
+        logger.warning("Cannot read tool config at %s", path)
+        raw = None
+
+    if raw is not None:
         try:
-            result: dict[str, Any] = json.loads(path.read_text())
+            result: dict[str, Any] = json.loads(raw)
+        except json.JSONDecodeError:
+            logger.warning("Tool config at %s is not valid JSON; ignoring.", path)
+        else:
             return result
-        except (OSError, json.JSONDecodeError):
-            logger.warning("Failed to read tool config from %s", path)
 
     if settings is None:
         return {}
@@ -156,14 +171,23 @@ def load_tool_config(settings: Settings | None = None) -> dict[str, Any]:
     return {}
 
 
-def save_tool_config(config: dict[str, Any]) -> None:
-    """Persist tool config to disk."""
+def save_tool_config(config: dict[str, Any]) -> bool:
+    """Persist tool config to disk.
+
+    Returns True on success, False on failure (read-only filesystem,
+    permission denied, etc.). Caller is responsible for surfacing the
+    failure to the user — the HTTP route at ``_save_tools`` returns 500
+    so the UI's ``saveConfig`` shows "Save failed!" instead of the
+    misleading "Saved — restart required".
+    """
     path = _get_config_path()
     try:
         path.write_text(json.dumps(config, indent=2))
-        logger.info("Saved tool config to %s", path)
     except OSError:
         logger.exception("Failed to save tool config to %s", path)
+        return False
+    logger.info("Saved tool config to %s", path)
+    return True
 
 
 async def _get_tool_metadata(server: HomeAssistantSmartMCPServer) -> list[dict[str, Any]]:
@@ -760,7 +784,18 @@ def register_settings_routes(
 
         config = load_tool_config()
         config["tools"] = states
-        save_tool_config(config)
+        if not save_tool_config(config):
+            return JSONResponse(
+                create_error_response(
+                    ErrorCode.INTERNAL_ERROR,
+                    "Failed to persist tool config to disk",
+                    suggestions=[
+                        "Set HA_MCP_CONFIG_DIR to a writable path (read-only filesystem?)",
+                        "Check the server logs for the underlying OSError",
+                    ],
+                ),
+                status_code=500,
+            )
 
         disabled_count = sum(1 for s in states.values() if s == "disabled")
         pinned_count = sum(1 for s in states.values() if s == "pinned")
--- a/src/ha_mcp/settings_ui.py
+++ b/src/ha_mcp/settings_ui.py
@@ -19,6 +19,7 @@ import httpx
 from starlette.requests import Request
 from starlette.responses import HTMLResponse, JSONResponse
 
+from ._version import is_running_in_addon
 from .errors import ErrorCode, create_error_response
 from .transforms import DEFAULT_PINNED_TOOLS
 from .utils.data_paths import get_data_dir
@@ -100,18 +101,6 @@ FEATURE_GATED_TOOLS: dict[str, dict[str, str]] = {
 }
 
 
-def _is_addon() -> bool:
-    """Return True when running inside the Home Assistant add-on container.
-
-    Mirrors the existing convention in this module (and ``__main__.py``)
-    of treating ``SUPERVISOR_TOKEN`` as the add-on detector. Using the env
-    var is more reliable than checking for ``/data`` because some Docker
-    setups (and macOS dev environments) have a ``/data`` directory that
-    isn't the add-on data dir.
-    """
-    return bool(os.environ.get("SUPERVISOR_TOKEN"))
-
-
 def _get_config_path() -> Path:
     """Return the path to the tool config JSON file.
 
@@ -134,7 +123,7 @@ def load_tool_config(settings: Settings | None = None) -> dict[str, Any]:
     except FileNotFoundError:
         raw = None
     except OSError:
-        logger.warning("Cannot read tool config at %s", path)
+        logger.warning("Cannot read tool config at %s", path, exc_info=True)
         raw = None
 
     if raw is not None:
@@ -858,11 +847,11 @@ def register_settings_routes(
 
     async def _settings_info(_: Request) -> JSONResponse:
         return JSONResponse({
-            "is_addon": _is_addon(),
+            "is_addon": is_running_in_addon(),
         })
 
     secret_prefix = secret_path.rstrip("/") if secret_path else ""
-    is_addon = _is_addon()
+    is_addon = is_running_in_addon()
 
     if not is_addon and not secret_prefix:
         logger.warning(
--- a/src/ha_mcp/utils/data_paths.py
+++ b/src/ha_mcp/utils/data_paths.py
@@ -40,16 +40,24 @@ def get_data_dir() -> Path:
        server start; users wanting persistence should set
        ``HA_MCP_CONFIG_DIR`` to a writable path.
 
-    Memoized so the fallback warning emits once at startup rather than on
-    every save/load HTTP request. Tests reset via
-    ``get_data_dir.cache_clear()``.
+    Memoized so the fallback warning typically emits once at startup
+    rather than on every save/load HTTP request. ``lru_cache`` serializes
+    its internal dict but does not serialize the wrapped call when the
+    cache is empty, so two threads racing on first access (e.g.
+    ``UsageLogger.__init__`` from a worker thread plus a settings UI HTTP
+    handler) may each run ``_resolve_data_dir`` once and emit the warning
+    twice. The mkdir calls are idempotent, so this is cosmetic.
+    Tests reset via ``get_data_dir.cache_clear()``.
     """
     return _resolve_data_dir()
 
 
 def _resolve_data_dir() -> Path:
     """Resolve the data directory (uncached); see ``get_data_dir`` for priority."""
-    config_dir_env = os.environ.get("HA_MCP_CONFIG_DIR")
+    # ``.strip()``: ``HA_MCP_CONFIG_DIR="   "`` is truthy and ``Path("   ")``
+    # resolves cwd-relative, which would mkdir a literal whitespace-named
+    # directory next to whatever cwd happens to be at startup.
+    config_dir_env = os.environ.get("HA_MCP_CONFIG_DIR", "").strip()
     preferred: Path | None = None
     if config_dir_env:
         custom_dir = Path(config_dir_env)
@@ -68,7 +76,26 @@ def _resolve_data_dir() -> Path:
             return custom_dir
 
     if is_running_in_addon():
-        return Path("/data")
+        addon_dir = Path("/data")
+        try:
+            addon_dir.mkdir(parents=True, exist_ok=True)
+        except OSError as e:
+            logger.warning(
+                "/data is not writable in add-on mode (%s: %s); "
+                "falling back. Set HA_MCP_CONFIG_DIR to override.",
+                type(e).__name__,
+                e,
+            )
+            if preferred is None:
+                preferred = addon_dir
+        else:
+            # Honor an explicit HA_MCP_CONFIG_DIR override even in add-on
+            # mode: if the user set it and its mkdir failed (preferred is
+            # not None), fall through to the tmpdir fallback rather than
+            # silently writing to /data — they chose the override
+            # deliberately.
+            if preferred is None:
+                return addon_dir
 
     if preferred is None:
         home_dir = Path.home() / ".ha-mcp"
@@ -85,8 +112,10 @@ def _resolve_data_dir() -> Path:
     except OSError as e:
         # Even the tmpdir is unwritable. Return the path anyway: callers
         # that wrap writes in try/except OSError can degrade gracefully
-        # (no persistence, but the server still starts).
-        logger.warning(
+        # (no persistence, but the server still starts). ``error`` rather
+        # than ``warning`` because persistence is silently disabled — the
+        # supervisor log viewer surfaces errors more prominently.
+        logger.error(
             "Cannot write ha-mcp data to %s or fallback %s (%s: %s); "
             "persistence is disabled. "
             "Set HA_MCP_CONFIG_DIR to a writable path for persistence.",
--- a/src/ha_mcp/utils/data_paths.py
+++ b/src/ha_mcp/utils/data_paths.py
@@ -13,35 +13,32 @@ import os
 import tempfile
 from pathlib import Path
 
+from .._version import is_running_in_addon
+
 logger = logging.getLogger(__name__)
-
-
-def _is_addon() -> bool:
-    """Return True when running inside the Home Assistant add-on container.
-
-    Mirrors the convention in ``settings_ui.py`` of treating
-    ``SUPERVISOR_TOKEN`` as the add-on detector — more reliable than
-    checking for ``/data`` because some Docker setups have a ``/data``
-    directory that isn't the supervisor data dir.
-    """
-    return bool(os.environ.get("SUPERVISOR_TOKEN"))
 
 
 @functools.lru_cache(maxsize=1)
 def get_data_dir() -> Path:
     """Return a writable directory for ha-mcp persistent data (memoized).
 
-    Priority:
+    Resolution order:
 
     1. ``HA_MCP_CONFIG_DIR`` env var — explicit override, e.g. for hardened
        Docker setups bind-mounting a writable volume into a
        ``read_only: true`` container.
-    2. ``/data`` — Home Assistant add-on (writable supervisor data dir).
-    3. ``~/.ha-mcp`` — standard.
-    4. ``<tempdir>/ha-mcp`` — last-resort fallback when (1) and (3) fail
-       (read-only filesystem, or ``HOME`` unset so ``Path.home()`` resolves
-       to ``/``). Loses persistence across restarts but lets the server
-       start; users wanting persistence should set ``HA_MCP_CONFIG_DIR``.
+    2. ``/data`` — Home Assistant add-on (``SUPERVISOR_TOKEN`` set; writable
+       supervisor data dir).
+    3. ``~/.ha-mcp`` — standard install. Skipped when ``HA_MCP_CONFIG_DIR``
+       was set but failed: an explicit override means "use this exact
+       location", and silently writing to ``$HOME`` instead would surprise
+       users who chose the override deliberately.
+    4. ``<tempdir>/ha-mcp`` — last-resort fallback when the previously
+       chosen step fails (read-only filesystem; ``HOME`` unset so
+       ``Path.home()`` resolves to ``/``; or ``HA_MCP_CONFIG_DIR`` set but
+       its mkdir raises). Loses persistence across restarts but lets the
+       server start; users wanting persistence should set
+       ``HA_MCP_CONFIG_DIR`` to a writable path.
 
     Memoized so the fallback warning emits once at startup rather than on
     every save/load HTTP request. Tests reset via
@@ -70,7 +67,7 @@ def _resolve_data_dir() -> Path:
         else:
             return custom_dir
 
-    if _is_addon():
+    if is_running_in_addon():
         return Path("/data")
 
     if preferred is None:
@@ -86,10 +83,9 @@ def _resolve_data_dir() -> Path:
     try:
         fallback.mkdir(parents=True, exist_ok=True)
     except OSError as e:
-        # Even the tmpdir is unwritable. Return the path anyway so the
-        # caller's own try/except (save_tool_config wraps writes in
-        # try/except OSError; usage_logger disables itself) can degrade
-        # gracefully.
+        # Even the tmpdir is unwritable. Return the path anyway: callers
+        # that wrap writes in try/except OSError can degrade gracefully
+        # (no persistence, but the server still starts).
         logger.warning(
             "Cannot write ha-mcp data to %s or fallback %s (%s: %s); "
             "persistence is disabled. "
--- a/src/ha_mcp/utils/data_paths.py
+++ b/src/ha_mcp/utils/data_paths.py
@@ -1,0 +1,110 @@
+"""Resolve a writable directory for ha-mcp persistent data.
+
+Single source of truth for "where does ha-mcp write its files?" — used
+by both ``settings_ui`` (tool config) and ``usage_logger`` (rolling
+JSONL).
+"""
+
+from __future__ import annotations
+
+import functools
+import logging
+import os
+import tempfile
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+
+def _is_addon() -> bool:
+    """Return True when running inside the Home Assistant add-on container.
+
+    Mirrors the convention in ``settings_ui.py`` of treating
+    ``SUPERVISOR_TOKEN`` as the add-on detector — more reliable than
+    checking for ``/data`` because some Docker setups have a ``/data``
+    directory that isn't the supervisor data dir.
+    """
+    return bool(os.environ.get("SUPERVISOR_TOKEN"))
+
+
+@functools.lru_cache(maxsize=1)
+def get_data_dir() -> Path:
+    """Return a writable directory for ha-mcp persistent data (memoized).
+
+    Priority:
+
+    1. ``HA_MCP_CONFIG_DIR`` env var — explicit override, e.g. for hardened
+       Docker setups bind-mounting a writable volume into a
+       ``read_only: true`` container.
+    2. ``/data`` — Home Assistant add-on (writable supervisor data dir).
+    3. ``~/.ha-mcp`` — standard.
+    4. ``<tempdir>/ha-mcp`` — last-resort fallback when (1) and (3) fail
+       (read-only filesystem, or ``HOME`` unset so ``Path.home()`` resolves
+       to ``/``). Loses persistence across restarts but lets the server
+       start; users wanting persistence should set ``HA_MCP_CONFIG_DIR``.
+
+    Memoized so the fallback warning emits once at startup rather than on
+    every save/load HTTP request. Tests reset via
+    ``get_data_dir.cache_clear()``.
+    """
+    return _resolve_data_dir()
+
+
+def _resolve_data_dir() -> Path:
+    """Resolve the data directory (uncached); see ``get_data_dir`` for priority."""
+    config_dir_env = os.environ.get("HA_MCP_CONFIG_DIR")
+    preferred: Path | None = None
+    if config_dir_env:
+        custom_dir = Path(config_dir_env)
+        try:
+            custom_dir.mkdir(parents=True, exist_ok=True)
+        except OSError as e:
+            logger.warning(
+                "HA_MCP_CONFIG_DIR=%s could not be prepared (%s: %s); "
+                "falling back to a tmpdir.",
+                custom_dir,
+                type(e).__name__,
+                e,
+            )
+            preferred = custom_dir
+        else:
+            return custom_dir
+
+    if _is_addon():
+        return Path("/data")
+
+    if preferred is None:
+        home_dir = Path.home() / ".ha-mcp"
+        try:
+            home_dir.mkdir(parents=True, exist_ok=True)
+        except OSError:
+            preferred = home_dir
+        else:
+            return home_dir
+
+    fallback = Path(tempfile.gettempdir()) / "ha-mcp"
+    try:
+        fallback.mkdir(parents=True, exist_ok=True)
+    except OSError as e:
+        # Even the tmpdir is unwritable. Return the path anyway so the
+        # caller's own try/except (save_tool_config wraps writes in
+        # try/except OSError; usage_logger disables itself) can degrade
+        # gracefully.
+        logger.warning(
+            "Cannot write ha-mcp data to %s or fallback %s (%s: %s); "
+            "persistence is disabled. "
+            "Set HA_MCP_CONFIG_DIR to a writable path for persistence.",
+            preferred,
+            fallback,
+            type(e).__name__,
+            e,
+        )
+    else:
+        logger.warning(
+            "Cannot write ha-mcp data to %s (read-only filesystem or HOME unset). "
+            "Falling back to %s — data will NOT persist across restarts. "
+            "Set HA_MCP_CONFIG_DIR to a writable path for persistence.",
+            preferred,
+            fallback,
+        )
+    return fallback
--- a/tests/src/unit/test_data_paths.py
+++ b/tests/src/unit/test_data_paths.py
@@ -115,6 +115,41 @@ class TestFallbacks:
         assert result.is_dir()
         assert not broken_target.exists()
 
+    def test_returns_unwritable_tmpdir_when_everything_fails(
+        self, monkeypatch, tmp_path, caplog
+    ):
+        """Last-resort branch: env mkdir fails, home mkdir fails, tmpdir mkdir
+        fails. The resolver returns the tmpdir path anyway so callers (which
+        wrap their own writes in ``try/except OSError``) can degrade
+        gracefully rather than crashing the server. Warning fires.
+        """
+        monkeypatch.delenv("SUPERVISOR_TOKEN", raising=False)
+        monkeypatch.delenv("HA_MCP_CONFIG_DIR", raising=False)
+        readonly_home = tmp_path / "ro-home"
+        readonly_home.mkdir()
+        monkeypatch.setattr(Path, "home", lambda: readonly_home)
+        fake_tmp = tmp_path / "ro-tmp"
+        fake_tmp.mkdir()
+        monkeypatch.setattr(
+            "ha_mcp.utils.data_paths.tempfile.gettempdir", lambda: str(fake_tmp)
+        )
+        original_mkdir = Path.mkdir
+
+        def fake_mkdir(self: Path, *args, **kwargs):
+            if self in (readonly_home / ".ha-mcp", fake_tmp / "ha-mcp"):
+                raise OSError(30, "Read-only file system")
+            return original_mkdir(self, *args, **kwargs)
+
+        monkeypatch.setattr(Path, "mkdir", fake_mkdir)
+
+        with caplog.at_level(logging.WARNING, logger="ha_mcp.utils.data_paths"):
+            result = get_data_dir()
+
+        assert result == fake_tmp / "ha-mcp"
+        assert any(
+            "persistence is disabled" in r.getMessage() for r in caplog.records
+        )
+
 
 class TestMemoization:
     """The resolver must memoize so warnings emit once at startup."""
--- a/tests/src/unit/test_data_paths.py
+++ b/tests/src/unit/test_data_paths.py
@@ -1,0 +1,172 @@
+"""Unit tests for the shared data-directory resolver.
+
+The resolver is the single source of truth for where ha-mcp writes its
+persistent files (tool config, usage logs). These tests pin its priority
+order and the fallback behavior added for issue #1125.
+"""
+
+from __future__ import annotations
+
+import logging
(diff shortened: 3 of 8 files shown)


src/ha_mcp/settings_ui.py
```diff
@@ -112,12 +113,62 @@ def _is_addon() -> bool:
...
+                fallback,
+            )
+        logger.warning(
+            "Cannot write tool config to %s (read-only filesystem or HOME unset). "
+            "Falling back to %s — settings will NOT persist across restarts. "
+            "Set HA_MCP_CONFIG_DIR to a writable path for persistence.",
+            home_dir,
+            fallback,
+        )
+        return fallback / "tool_config.json"
```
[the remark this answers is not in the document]
kingpanther13: ✅ Addressed in 5024a43 (force-push). The double-warning concern was real — verified empirically that when both home and tmpdir mkdir fail, the original code emitted two warnings, and the second misleadingly said "Falling back to %s" even though the fallback also failed. The new `_resolve_config_path` emits exactly one warning per resolution, with the message branching on whether the fallback mkdir succeeded. The result is also memoized at module level (`_CONFIG_PATH_CACHE`), so the warning fires once at startup rather than on every save/load HTTP request — `_get_config_path` is hit by the GET /api/settings/tools route, the POST handler (twice), and `load_tool_config` at server init.

I went a different direction on the "operation" prefix suggestion — included `type(e).__name__` and `e` in the warnings instead, so operators can distinguish EROFS (read-only fs), EACCES (perm), ENOSPC (disk full), etc., since the remediation differs for each.

kingpanther13 (commented):
## Review pass complete (force-pushed `5024a43`)

Ran the PR review toolkit (`code-reviewer`, `pr-test-analyzer`, `comment-analyzer`, `silent-failure-hunter`) plus addressed Gemini's inline suggestion. Findings I confirmed empirically and fixed:

| Finding | Source | Status |
|---|---|---|
| `load_tool_config()` `path.exists()` propagates `PermissionError` (EACCES not in `pathlib._IGNORED_ERRNOS`) — same class as the original #1125 bug | silent-failure-hunter | **Fixed**: replaced `exists()`+`read_text()` two-step with single `try: read_text() except OSError:` |
| `HA_MCP_CONFIG_DIR` set-but-unwritable returned a broken path; every save then OSError'd in `save_tool_config` | code-reviewer + pr-test-analyzer + silent-failure-hunter | **Fixed**: `HA_MCP_CONFIG_DIR` mkdir failure now chains into the tmpdir fallback (consistent with home-dir branch) |
| Warning emitted on every `_get_config_path()` call (called from each settings UI HTTP request) → log spam | silent-failure-hunter | **Fixed**: result memoized in module-level `_CONFIG_PATH_CACHE`; warning emits once at startup |
| Double warning when home AND tmpdir both fail; second one misleadingly says "Falling back to ..." | Gemini + silent-failure-hunter | **Fixed**: single warning per resolution, message branches on fallback mkdir success |
| ~50% of original diff was unrelated `ruff format` reflows on untouched code | code-reviewer | **Reverted**: restored `settings_ui.py` from `origin/master` and re-applied only the targeted fix; diff is now focused (87 added / 11 removed in `settings_ui.py`, vs. 152/38 before) |
| Dockerfile chmod comment framing about `useradd default 0700` was image-specific; comment-analyzer also flagged "Path.home() raises PermissionError" being technically inaccurate (traversal raises, not Path.home() itself) | code-reviewer + comment-analyzer | **Reworded**: comment now references `/etc/login.defs HOME_MODE` and "stats a path under HOME" rather than `Path.home()` |

### Confirmed but deliberately deferred (out of PR scope)

- `usage_logger.py:120-127` already uses `try/except OSError` but its `logger.exception` on save failures will now write to a same-class fallback path (i.e. `/home/mcpuser/.ha-mcp/logs/`) that's unwritable for `--user UID:GID` users; should adopt the same memoized-resolution + tempdir-fallback pattern in a follow-up.
- `save_tool_config` returns `None` on OSError and the HTTP handler reports `success: True` to the browser — this is a UX bug (user thinks save worked), but predates this PR and has its own scope.
- Test edge case (both home AND tmpdir unwritable) — defensive-only, low priority; left as a possible follow-up.

### Tests

5 unit tests in `TestConfigPath` (all in `tests/src/unit/test_settings_ui.py`):

- `test_ha_mcp_config_dir_overrides_supervisor_token` — env var precedence
- `test_falls_back_to_tmpdir_when_home_unwritable` — issue #1125 regression
- `test_ha_mcp_config_dir_unwritable_chains_to_tmpdir` — new fallback chaining
- `test_load_tool_config_does_not_crash_on_unreadable_config_dir` — same-class follow-up regression
- `test_warning_emitted_only_once_via_module_cache` — log-spam guard

Full unit suite: 1679 passed, 1 skipped (numpy unrelated). `ruff check src/ tests/ homeassistant-addon/ homeassistant-addon-webhook-proxy/ scripts/` clean. `mypy src/` clean.

Re-verified the three Docker scenarios (reporter's compose, default user, `HA_MCP_CONFIG_DIR` bind-mount) — all start cleanly under `:fix-1125` build.

tests/src/unit/test_settings_ui.py
```diff
@@ -137,20 +137,192 @@ def test_empty_config_no_disable(self):
         mcp.disable.assert_not_called()
 
 
+@pytest.fixture(autouse=True)
+def _reset_config_path_cache():
+    """Clear the module-level resolved-path cache between tests."""
+    import ha_mcp.settings_ui as su
```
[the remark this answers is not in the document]
kingpanther13: ✅ Fixed in 8f8efc0. Replaced `import ha_mcp.settings_ui as su; su._CONFIG_PATH_CACHE = None` with `monkeypatch.setattr("ha_mcp.utils.data_paths._DATA_DIR_CACHE", None)` (no `import ha_mcp.settings_ui as su` anywhere now). Bonus: monkeypatch auto-restores on teardown, so the cache state is properly isolated between tests instead of relying on the fixture's own cleanup. The cache itself moved out of `settings_ui` into the new shared `utils/data_paths` module — the directory resolver is now used by both `settings_ui` (tool config) and `usage_logger` (logs).

kingpanther13 (commented):
## Follow-ups addressed (force-pushed `8f8efc0`)

Pulled in the items I'd previously marked deferred:

| Concern | Source | Fix |
|---|---|---|
| `usage_logger.py` had the same root cause but wasn't patched (would silently fall back to `Path.home()/...` and ignore `HA_MCP_CONFIG_DIR`) | code-reviewer issue #1 | Extracted shared resolver to new `utils/data_paths.py`; both `settings_ui` and `usage_logger` now go through `get_data_dir()`. Add-on mode now lands logs in `/data/logs/` (the supervisor's persistent dir) instead of ephemeral `~/.ha-mcp/logs/`. |
| `save_tool_config` swallowed `OSError` and `_save_tools` returned `{success: true}` while the user's change was lost | silent-failure-hunter issue #4 | `save_tool_config` now returns `bool`; `_save_tools` returns 500 with a structured error body when it fails. The JS at `saveConfig()` already shows "Save failed!" on `!resp.ok` — it just wasn't being given the chance. |
| Mixed `import` and `from … import` in test fixture | github-code-quality bot | Replaced `import ha_mcp.settings_ui as su` + manual reset with `monkeypatch.setattr("ha_mcp.utils.data_paths._DATA_DIR_CACHE", None)` — also gives proper between-test isolation via monkeypatch's auto-restore. Thread resolved. |
| PR title was vague | user feedback | Renamed: `fix: tolerate read-only filesystem when locating tool config` → `fix: survive read-only filesystems at startup`. |

**New tests** (9 total across new + existing files):
- `test_data_paths.py`: priority order, both fallback branches, memoization (single-warning guard), uncached resolver re-runs.
- `test_settings_ui.py`: `save_tool_config` bool contract (success + OSError → False); `_save_tools` returns 500 when save fails.
- `test_usage_logger.py`: default path uses shared resolver; `HA_MCP_CONFIG_DIR` redirects logs.

Full unit suite: **1684 passed, 1 skipped**. `ruff check src/ tests/ ...` clean. `mypy src/` clean. Re-verified the three Docker scenarios under the new build.

src/ha_mcp/settings_ui.py
```diff
@@ -760,7 +783,19 @@ async def _save_tools(request: Request) -> JSONResponse:
...
         config["tools"] = states
-        save_tool_config(config)
+        if not save_tool_config(config):
+            return JSONResponse(
+                create_error_response(
+                    ErrorCode.INTERNAL_ERROR,
+                    "Failed to persist tool config to disk",
+                    suggestions=[
+                        "Set HA_MCP_CONFIG_DIR to a writable path "
+                        "(common with read_only Docker setups)",
```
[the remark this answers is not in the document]
kingpanther13: ✅ Fixed in 3a53371. The two-string adjacency was an unintentional smell — the bot's preferred fix (add a comma) would have changed semantics (made it two suggestions instead of one with a parenthetical), so I collapsed to a single string instead: `"Set HA_MCP_CONFIG_DIR to a writable path (read-only filesystem?)"`. Same intent, no ambiguity.

While I was at it I scanned my new code with stricter ruff selectors and cleaned up two more bot-flag-worthy patterns (which the project's local ruff config doesn't enable but a stricter checker like CodeQL would catch):
- `PLW0603`: replaced `global _DATA_DIR_CACHE` with `functools.lru_cache(maxsize=1)` on `get_data_dir()`. Tests now reset via `get_data_dir.cache_clear()` instead of poking a module attribute.
- `TRY300`: moved the success-path `return`s out of the `try:` blocks into `else:` clauses (or after the `try` in `save_tool_config`'s case).

kingpanther13:
## Implementation Summary

**Choices Made:**
- Extracted a shared resolver to a new `utils/data_paths.py` rather than duplicating fallback logic in `settings_ui` and `usage_logger`. Both modules now route through `get_data_dir()`, so `HA_MCP_CONFIG_DIR` and add-on `/data` mode work identically for tool config and logs.
- 4-tier precedence: `HA_MCP_CONFIG_DIR` env → `/data` (when `SUPERVISOR_TOKEN` is set) → `~/.ha-mcp` → `<tempdir>/ha-mcp` last-resort. The server keeps running even when nothing is writable; only persistence is lost.
- Memoized result in a module-level `_DATA_DIR_CACHE` so the fallback warning emits **once at startup** instead of on every settings-UI HTTP request.
- Replaced `path.exists() + read_text()` in `load_tool_config` with a single `try: read_text() except OSError:`. `Path.exists()` only swallows `ENOENT/ENOTDIR/EBADF/ELOOP`; `EACCES` propagates and would re-introduce the same class of crash as #1125.
- `save_tool_config` now returns `bool`; `_save_tools` returns 500 with a structured error body on failure. The JS at `saveConfig()` already had a `!resp.ok → "Save failed!"` branch — it just wasn't being given the chance.
- `Dockerfile`: explicit `ENV HOME=/home/mcpuser` (otherwise Docker leaves `HOME=/` under non-root users — moby/moby#2968) and `chmod 0755 /home/mcpuser` so users overriding `--user UID:GID` can still stat paths under HOME without `PermissionError`. Mode 0755 keeps writes restricted to mcpuser.

**Problems Encountered:**
- Initial fix introduced a same-class regression: `load_tool_config()`'s `path.exists()` would still propagate `PermissionError` on an unreadable config dir. Caught by `silent-failure-hunter` and pinned with `test_load_tool_config_does_not_crash_on_unreadable_config_dir`.
- `HA_MCP_CONFIG_DIR` set-but-unwritable returned a broken path; every save then `OSError`'d. Chained the env-var branch into the same tmpdir fallback as the home-dir branch (`test_ha_mcp_config_dir_unwritable_chains_to_tmpdir`).
- `usage_logger.py` had the same root cause but wasn't part of the original fix — it would silently fall back to `Path.home()/.ha-mcp/logs/` and ignore `HA_MCP_CONFIG_DIR`. Initially deferred, then pulled in once the shared resolver existed.
- Original diff accidentally bundled ~50% unrelated `ruff format` reflows on untouched code. Reverted `settings_ui.py` to `origin/master` and re-applied only the targeted fix; settings_ui diff dropped from 152/38 to 87/11.
- Comment rot: Dockerfile comment cited `useradd` defaults (image-specific) and said `Path.home()` raises `PermissionError` (technically the traversal raises, not `Path.home()` itself). Reworded to reference `/etc/login.defs HOME_MODE` and "stats a path under HOME". Caught by `comment-analyzer`.
- A test fixture mixed `import ha_mcp.settings_ui as su` + manual cache reset. Replaced with `monkeypatch.setattr("ha_mcp.utils.data_paths._DATA_DIR_CACHE", None)` so test isolation comes from monkeypatch's auto-restore rather than ad-hoc teardown.

**Verification:**
- Full unit suite: 1684 passed, 1 skipped (numpy unrelated). `ruff check` clean. `mypy src/` clean. All PR Validation Pipeline checks green.
- Local Docker: reporter's `--read-only --user 1000:1000 --tmpfs /tmp` compose, default-user, and `HA_MCP_CONFIG_DIR` bind-mount scenarios all start cleanly under `:fix-1125`. Original #1125 stack trace no longer reproduces.

src/ha_mcp/utils/data_paths.py
```diff
@@ -0,0 +1,106 @@
...
+                custom_dir,
+                type(e).__name__,
+                e,
+            )
+            preferred = custom_dir
+        else:
+            return custom_dir
+
+    if is_running_in_addon():
+        return Path("/data")
```
Patch76: The addon branch returns `Path("/data")` without verifying writability — every other branch in `_resolve_data_dir()` wraps the chosen path in `try: mkdir(parents=True, exist_ok=True) except OSError:` and chains forward (env at 56-67, home at 73-79, fallback at 82-95). The addon branch does not, and crucially it also returns *before* checking `preferred`.

Two consequences:

1. **Residual same-class bug**: a user who adopts `read_only: true` on the supervisor container, or the supervisor in a degraded state where `/data` is mounted read-only, gets `Path("/data")` returned and the very first `path.write_text(...)` in `save_tool_config` raises `OSError` — exactly the crash class #1125 set out to fix, just one floor up.
2. **Silent override discard**: when `HA_MCP_CONFIG_DIR` is set, its mkdir fails (lines 56-67), and `SUPERVISOR_TOKEN` is also set, `preferred` carries the user's intended override — but this branch returns `Path("/data")` and silently ignores it. That contradicts the resolver's documented intent (the docstring says explicit overrides win; the home branch at line 73 does honor `preferred`).

Suggested fix:

```python
if is_running_in_addon():
    addon_dir = Path("/data")
    try:
        addon_dir.mkdir(parents=True, exist_ok=True)
    except OSError as e:
        logger.warning(
            "/data is not writable in add-on mode (%s: %s); falling back. "
            "Set HA_MCP_CONFIG_DIR to override.",
            type(e).__name__, e,
        )
        if preferred is None:
            preferred = addon_dir
    else:
        if preferred is None:
            return addon_dir
        # preferred set → env-var override failed mkdir; fall through to honor user intent
```

Plus a regression test — currently `TestPriorityOrder` and `TestFallbacks` don't cover the `env-fail + SUPERVISOR_TOKEN-set` interaction at all.

src/ha_mcp/utils/data_paths.py
```diff
@@ -0,0 +1,106 @@
...
+    ``get_data_dir.cache_clear()``.
+    """
+    return _resolve_data_dir()
+
+
+def _resolve_data_dir() -> Path:
+    """Resolve the data directory (uncached); see ``get_data_dir`` for priority."""
+    config_dir_env = os.environ.get("HA_MCP_CONFIG_DIR")
+    preferred: Path | None = None
+    if config_dir_env:
```
Patch76, line 62: `config_dir_env = os.environ.get("HA_MCP_CONFIG_DIR")` then `if config_dir_env:` is pure truthiness with no whitespace handling. Three cases worth pinning:

- `HA_MCP_CONFIG_DIR=""` — `bool("")` is False, branch skipped. Fine.
- `HA_MCP_CONFIG_DIR="   "` (three spaces, easy to introduce via a `.env` file with trailing space, or YAML quoting) — truthy. `Path("   ")` resolves cwd-relative (verified locally: `Path("   ").absolute()` returns a path ending in `/   `), and mkdir creates a literal three-space-named directory under whatever `cwd` happens to be at startup. The user thinks their override is being honored; it isn't, in any meaningful sense.
- `HA_MCP_CONFIG_DIR="."` — truthy, resolves to `cwd`, mkdir succeeds (already exists), config files land next to wherever the working directory points. For the Docker container that's `/app` (where the venv lives). Not catastrophic but not what anyone meant.

Smallest fix:

```python
config_dir_env = os.environ.get("HA_MCP_CONFIG_DIR", "").strip()
preferred: Path | None = None
if config_dir_env:
    custom_dir = Path(config_dir_env)
    ...
```

`"."` is the user's choice to make so I'd leave that alone, but `""` and whitespace-only both need to skip the branch.

src/ha_mcp/utils/data_paths.py
```diff
@@ -0,0 +1,106 @@
...
+import os
+import tempfile
+from pathlib import Path
+
+from .._version import is_running_in_addon
+
+logger = logging.getLogger(__name__)
+
+
+@functools.lru_cache(maxsize=1)
```
Patch76, line 21: `@functools.lru_cache(maxsize=1)` makes the cache itself thread-safe — CPython's `lru_cache` serializes reads and writes to the internal dict — but the inner `_resolve_data_dir()` call is not serialized when the cache is empty. Two simultaneous startup paths (e.g. `UsageLogger.__init__` from a worker thread plus `_get_tool_metadata` from the HTTP path) hitting an empty cache will both call `_resolve_data_dir`, both call `mkdir(exist_ok=True)` (idempotent — fine), and both log the fallback warning if it fires.

So the docstring's "Memoized so the fallback warning emits once at startup" claim holds in practice for single-threaded startup but not for concurrent first-access. Either weaken the docstring (`Memoized; warning typically emits once but may emit twice under concurrent first-access`) or wrap `_resolve_data_dir` in a module-level lock for the stronger guarantee. Low practical impact either way; only the docstring's framing is wrong.

src/ha_mcp/utils/data_paths.py
```diff
@@ -0,0 +1,106 @@
...
+        # Even the tmpdir is unwritable. Return the path anyway: callers
+        # that wrap writes in try/except OSError can degrade gracefully
+        # (no persistence, but the server still starts).
+        logger.warning(
+            "Cannot write ha-mcp data to %s or fallback %s (%s: %s); "
+            "persistence is disabled. "
+            "Set HA_MCP_CONFIG_DIR to a writable path for persistence.",
+            preferred,
+            fallback,
+            type(e).__name__,
```
Patch76, line 124: When this branch fires, persistence is silently disabled — the user's settings UI changes evaporate on every restart and their logs evaporate on every restart. That's a degraded operational state, not a "configuration warning".

`logger.error(...)` would match the practical impact and would be more visible in the supervisor log viewer, where `WARNING` lines blur into noise. The `try/except` in `save_tool_config` keeps the server running — that's the case for not raising — but the log level should match the operational reality of "persistence is silently disabled".

tests/src/unit/test_data_paths.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def test_home_path_when_no_supervisor_token(self, monkeypatch, tmp_path):
+        monkeypatch.delenv("SUPERVISOR_TOKEN", raising=False)
+        monkeypatch.delenv("HA_MCP_CONFIG_DIR", raising=False)
+        monkeypatch.setattr(Path, "home", lambda: tmp_path)
+        result = get_data_dir()
+        assert result == tmp_path / ".ha-mcp"
+        assert result.is_dir()
+
+    def test_ha_mcp_config_dir_overrides_supervisor_token(self, monkeypatch, tmp_path):
```
Patch76, line 52: This test pins the success case for `HA_MCP_CONFIG_DIR > /data`. The failure-case interaction — env mkdir fails + `SUPERVISOR_TOKEN` is set — is not exercised anywhere in the suite. That's exactly the path where the addon-mode branch silently discards `preferred` (see the `data_paths.py:71` finding), so any regression of a fix there would slip through CI.

A `test_ha_mcp_config_dir_unwritable_in_addon_mode_chains_to_data` mirroring this structure but with `monkeypatch.setattr(Path, "mkdir", lambda *a, **kw: (_ for _ in ()).throw(PermissionError(13, "forced")))` (or a real unwritable parent dir) to force the env-var mkdir to fail would lock the behavior.

Patch76 (commented):
Re-review of HEAD `d66c651a`. The fix closes #1125. `ENV HOME`, the EACCES handling in `load_tool_config`, and the shared `data_paths.get_data_dir()` resolver are all correct. Test coverage — including the real `chmod 0o000` regression guard at `tests/src/unit/test_settings_ui.py:198` — is solid.

Three sticking points before squash:

- One residual same-class bug: the addon-mode branch at `data_paths.py:70-71` returns `Path("/data")` without verifying writability and silently overrides an explicit `HA_MCP_CONFIG_DIR` if its mkdir failed.
- One env-var edge case: `HA_MCP_CONFIG_DIR="   "` (whitespace) is truthy, mkdirs a three-space-named directory under `cwd`. `Path("   ").absolute()` empirically resolves cwd-relative.
- The PR title — auto-merge is queued, so the squash subject inherits `fix: fix read-only filesystems /:latest at startup ( fixes #1125)` verbatim into `master` history the moment the next approval lands. `git log upstream/master --grep="^fix: fix "` returns nothing else; this would be the first double-verb fix subject, alongside the `/:latest` typo and the leading space inside the parens. Worth editing before approving.

For existing `:latest` users who hit the original bug, data was previously written to `/.ha-mcp/` (filesystem root, from `HOME=/`); after this fix no code path reads from there. Worth a one-line release note.

Findings inline.

src/ha_mcp/utils/data_paths.py
kingpanther13: ✅ Fixed in befcd9a7. Addon branch now mkdir's `/data` first; on success returns it only when `preferred is None` (so a failed `HA_MCP_CONFIG_DIR` falls through to tmpdir instead of being silently discarded), on failure sets `preferred = /data` and continues into the home/tmpdir chain. Two new tests pin both interactions: `test_falls_back_when_addon_data_dir_unwritable` (read-only `/data`) and `test_ha_mcp_config_dir_unwritable_in_addon_mode_chains_to_tmpdir` (silent-override-discard regression).

src/ha_mcp/utils/data_paths.py
kingpanther13: ✅ Fixed in befcd9a7. `os.environ.get("HA_MCP_CONFIG_DIR", "").strip()` so `""` and whitespace-only both skip the branch. Pinned by `test_whitespace_only_ha_mcp_config_dir_is_ignored`. Left `"."` alone per your note — that's the user's choice to make.

src/ha_mcp/utils/data_paths.py
kingpanther13: ✅ Fixed in befcd9a7. Weakened the docstring: now flags that `lru_cache` doesn't serialize the wrapped call when the cache is empty, so concurrent first-access can emit the warning twice (idempotent mkdirs, cosmetic only). Skipped the lock — the impact doesn't justify it.

src/ha_mcp/utils/data_paths.py
kingpanther13: ✅ Fixed in befcd9a7. Upgraded to `logger.error`. `test_returns_unwritable_tmpdir_when_everything_fails` now asserts `levelno == logging.ERROR` so a future refactor can't quietly downgrade it.

tests/src/unit/test_data_paths.py
kingpanther13: ✅ Fixed in befcd9a7. Added `test_ha_mcp_config_dir_unwritable_in_addon_mode_chains_to_tmpdir` mirroring `test_ha_mcp_config_dir_overrides_supervisor_token` but with the env-var mkdir forced to fail and `/data` mkdir stubbed to succeed — locks the silent-override-discard fix.

kingpanther13 (commented):
## Patch76's review addressed in befcd9a7

All five inline findings + the title.

| Finding | Fix |
|---|---|
| Addon branch returns `Path("/data")` without writability check; silently discards `HA_MCP_CONFIG_DIR` override on env-var mkdir failure | `mkdir`-then-check; on success returns `/data` only when `preferred is None` (failed override falls through to tmpdir); on failure sets `preferred = /data` and continues into the home/tmpdir chain |
| `HA_MCP_CONFIG_DIR="   "` is truthy → mkdir's a literal whitespace-named dir under cwd | `os.environ.get("HA_MCP_CONFIG_DIR", "").strip()` |
| Docstring "warning emits once" overclaims under concurrent first-access | Weakened: now flags `lru_cache` doesn't serialize the wrapped call when empty; idempotent mkdir, cosmetic-only double warning |
| Last-resort branch logged at WARNING when persistence is silently disabled | Upgraded to `logger.error` |
| No test for env-fail + `SUPERVISOR_TOKEN`-set interaction | Added `test_ha_mcp_config_dir_unwritable_in_addon_mode_chains_to_tmpdir` (silent-override-discard regression) and `test_falls_back_when_addon_data_dir_unwritable` (residual same-class bug) |
| PR title double-verb / `/:latest` typo / stray space | Renamed to `fix: survive read-only filesystems at startup` |

Existing `test_addon_path_when_supervisor_token_set` updated to stub `Path.mkdir` for `/data` (the new writability check would otherwise hit the real path on dev/CI machines). `test_returns_unwritable_tmpdir_when_everything_fails` now asserts `levelno == logging.ERROR`.

> For existing `:latest` users who hit the original bug, data was previously written to `/.ha-mcp/` (filesystem root, from `HOME=/`); after this fix no code path reads from there. Worth a one-line release note.

Good call — that's a stranded-data note worth surfacing, and we already have the conventional-commit body to drive a release note. I'll add it to the PR description rather than the commit (the description is what the addon CHANGELOG sync uses).

src/ha_mcp/utils/data_paths.py
Patch76: Verified — addon branch now mkdirs `/data` at line 81 and the `if preferred is None:` gate at line 97 honors an explicit `HA_MCP_CONFIG_DIR` override even when its mkdir failed. Trace lands at the tmpdir branch.

src/ha_mcp/utils/data_paths.py
Patch76: Verified — line 61 has `os.environ.get("HA_MCP_CONFIG_DIR", "").strip()`. Regression pinned by `test_whitespace_only_ha_mcp_config_dir_is_ignored` at `tests/src/unit/test_data_paths.py:66` with the literal-whitespace-dir guard.

src/ha_mcp/utils/data_paths.py
Patch76: Verified — docstring at lines 43-50 names the concrete racing pair (`UsageLogger.__init__` from a worker thread plus a settings UI HTTP handler) and acknowledges the cosmetic double-warning is idempotent.

src/ha_mcp/utils/data_paths.py
Patch76: Verified — line 118 uses `logger.error`, scope correct (only on tmpdir-fallback failure, the persistence-genuinely-disabled state). `assert all(r.levelno == logging.ERROR ...)` pins it at `tests/src/unit/test_data_paths.py:207`.

tests/src/unit/test_data_paths.py
Patch76: Verified — `test_ha_mcp_config_dir_unwritable_in_addon_mode_chains_to_tmpdir` at `tests/src/unit/test_data_paths.py:141` exercises exactly the silent-discard regression path (SUPERVISOR_TOKEN set, broken `HA_MCP_CONFIG_DIR`, fakes `/data` mkdir as success, asserts result is tmpdir).

Patch76 (commented):
Round 2: all 5 inline findings verified fixed in `befcd9a`. CI green (17 pass / 2 skipping for hotfix-origin and dependabot). Per-thread verifications posted as replies above; threads already resolved.

Two unrelated observations for before-approval:

**High — `auto_merge.commit_title` is cached at enable-time and ships at squash time:**

```
.title:                    "fix: survive read-only filesystems at startup"           (clean)
.auto_merge.commit_title:  "fix: fix read-only filesystems /:latest at startup ( fixes #1125) (#1138)"  (cached)
```

The PR title rename did not refresh the auto-merge cache. [GitHub Community Discussion #16271](https://github.com/orgs/community/discussions/16271) reproduces that the cached value is what lands on master at squash time, not the live `.title`. Either disable + re-enable auto-merge to refresh, or use `gh pr merge 1138 --squash --subject "..."` for manual merge. Otherwise the squash commit on master will carry `fix: fix`, `/:latest`, and the leading space inside the parens.

**Low — cosmetic double-warning** on the env-fail + addon-set + `/data`-unwritable trace (`src/ha_mcp/utils/data_paths.py:82-87`): user sees both `HA_MCP_CONFIG_DIR=... could not be prepared` and `/data is not writable in add-on mode` warnings, even though the env-fail short-circuit already determined the outcome (the override would have won). Cosmetic only — no functional issue, and a guard would complicate the flow disproportionately.

Patch76 (approved):
Verified `befcd9a` against the Round-2 review above — all 5 inline fixes land cleanly, CI green on required checks. Auto-merge disabled to prevent the cached typo squash-subject from shipping; merge timing and subject are yours to control.

PR #1147, opened 2026-05-06
feat: add JMESPath response-filter middleware

tomwilkie:
## What does this PR do?

Adds an optional `_jmespath` parameter to every MCP tool via server-level middleware. Agents can provide a [JMESPath](https://jmespath.org/) expression and the server filters the response **server-side** before returning it, reducing context-window token usage by 80–90% on large HA payloads — without any per-tool changes.    

  **Example:** 

Before:

```bash
> fastmcp call http://localhost:8086/mcp ha_config_list_areas
{
  "success": true,
  "count": 16,
  "areas": [
    {
      "aliases": [],
      "area_id": "basement",
      "floor_id": "basement",
      ...
> fastmcp call http://localhost:8086/mcp ha_config_list_areas | wc
     215     411    5519
```

After:
```bash
> fastmcp call http://localhost:8086/mcp ha_config_list_areas '_jmespath={success: success, areas:areas[*].{id: area_id, name: name}}' 
{
  "success": true,
  "areas": [
    {
      "id": "basement",
      "name": "Basement"
    },
...
> fastmcp call http://localhost:8086/mcp ha_config_list_areas '_jmespath={success: success, areas:areas[*].{id: area_id, name: name}}' | wc
     69     117    1153
```

## Changes

  - `src/ha_mcp/middleware/response_filter.py` — JMESPathFilterMiddleware with two hooks:
    - `on_list_tools`: injects `_jmespath` into every tool's advertised JSONSchema so agents discover it automatically. Uses shallow copy + parameters-only deepcopy to avoid "cannot pickle '_thread.RLock' object" (FastMCP Tool objects contain threading locks internally)
    - `on_call_tool`: strips `_jmespath` from arguments before the underlying tool sees them, then applies the expression to `structured_content`
  - `src/ha_mcp/server.py` — one-line middleware registration
  - `pyproject.toml` — adds `jmespath==1.0.1` (pure Python, no native extensions; important for the HA add-on and Docker images); removes the stale `jq` entry from mypy overrides (was never imported)

##  Behaviour on failure

  Invalid JMESPath expressions raise `ToolError` with `VALIDATION_INVALID_PARAMETER`. Non-matching expressions (result is `None`) return `{"result": null}`. Envelope keys (`success`, `partial`, `warning`, etc.) are always preserved in the filtered response.

##  Test plan

  - `_jmespath` present in `inputSchema.properties` for all tools
  - Projection `{entity_id: data.entity_id, state: data.state}` returns only those two keys
  - Invalid expression `!!not valid!!` raises `ToolError` with `VALIDATION_INVALID_PARAMETER`
  - `_jmespath` is stripped from tool args before delegation (proves middleware strips it before FastMCP schema validation)
  - Regression: `on_list_tools` does not raise when Tool objects contain `threading.RLock` (unit test with `_ToolStub`)
  - Regression: `on_list_tools` does not mutate the original tool's parameters
  - Full E2E suite passes

## Type of change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

## Checklist

### DIFF
--- a/homeassistant-addon-dev/config.yaml
+++ b/homeassistant-addon-dev/config.yaml
@@ -26,6 +26,7 @@ options:
   enable_tool_search: false
   enable_yaml_config_editing: false
   enable_code_mode: false
+  enable_jmespath_filter: false
   tool_search_max_results: 5
   disabled_tools: ""
   pinned_tools: ""
@@ -36,6 +37,7 @@ schema:
   enable_tool_search: bool?
   enable_yaml_config_editing: bool?
   enable_code_mode: bool?
+  enable_jmespath_filter: bool?
   enable_filesystem_tools: bool?
   enable_custom_component_integration: bool?
   tool_search_max_results: int(2,10)?
--- a/homeassistant-addon/config.yaml
+++ b/homeassistant-addon/config.yaml
@@ -32,11 +32,13 @@ image: "ghcr.io/homeassistant-ai/ha-mcp-addon-{arch}"
 options:
   backup_hint: "normal"
   enable_tool_search: false
+  enable_jmespath_filter: false
   verify_ssl: true
 schema:
   backup_hint: list(strong|normal|weak|auto)
   secret_path: str?
   enable_tool_search: bool?
+  enable_jmespath_filter: bool?
   verify_ssl: bool?
   advanced_debug_logging: bool?
 # Add-on exposes HTTP port for MCP communication (fixed internal port)
--- a/homeassistant-addon/start.py
+++ b/homeassistant-addon/start.py
@@ -214,6 +214,7 @@ def main() -> int:
     custom_secret_path = ""  # default
     enable_tool_search = False  # default
     enable_yaml_config_editing = False  # default
+    enable_jmespath_filter = False  # default
     enable_filesystem_tools = False  # default
     enable_custom_component_integration = False  # default
     enable_code_mode = False  # default
@@ -233,6 +234,8 @@ def main() -> int:
             enable_tool_search = raw_tool_search if isinstance(raw_tool_search, bool) else False
             raw_yaml_config = config.get("enable_yaml_config_editing", False)
             enable_yaml_config_editing = raw_yaml_config if isinstance(raw_yaml_config, bool) else False
+            raw_jmespath = config.get("enable_jmespath_filter", False)
+            enable_jmespath_filter = raw_jmespath if isinstance(raw_jmespath, bool) else False
             raw_filesystem_tools = config.get("enable_filesystem_tools", False)
             enable_filesystem_tools = raw_filesystem_tools if isinstance(raw_filesystem_tools, bool) else False
             raw_custom_component = config.get("enable_custom_component_integration", False)
@@ -275,6 +278,7 @@ def main() -> int:
     os.environ["BACKUP_HINT"] = backup_hint
     os.environ["ENABLE_TOOL_SEARCH"] = str(enable_tool_search).lower()
     os.environ["ENABLE_YAML_CONFIG_EDITING"] = str(enable_yaml_config_editing).lower()
+    os.environ["ENABLE_JMESPATH_FILTER"] = str(enable_jmespath_filter).lower()
     os.environ["HAMCP_ENABLE_FILESYSTEM_TOOLS"] = str(enable_filesystem_tools).lower()
     os.environ["HAMCP_ENABLE_CUSTOM_COMPONENT_INTEGRATION"] = str(enable_custom_component_integration).lower()
     os.environ["ENABLE_CODE_MODE"] = str(enable_code_mode).lower()
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,6 +26,7 @@ classifiers = [
 dependencies = [
     "fastmcp==3.2.4",
     "httpx[socks]==0.28.1",
+    "jmespath==1.0.1",
     "pydantic==2.13.3",
     "python-dotenv==1.2.2",
     "truststore==0.10.4",
@@ -75,7 +76,8 @@ explicit_package_bases = true
 [[tool.mypy.overrides]]
 module = [
     "fastmcp.*",
-    "jq",
+    "jmespath",
+    "jmespath.*",
     "pydantic_monty",
     "pydantic_monty.*",
 ]
--- a/src/ha_mcp/config.py
+++ b/src/ha_mcp/config.py
@@ -102,6 +102,11 @@ class Settings(BaseSettings):
     # files. Disabled by default; only for YAML-only features with no UI/API path.
     enable_yaml_config_editing: bool = Field(False, alias="ENABLE_YAML_CONFIG_EDITING")
 
+    # JMESPath response-filter middleware — injects an optional _jmespath parameter
+    # into every tool, allowing agents to filter responses server-side before they are
+    # returned. Reduces token usage for large HA payloads. Disabled by default.
+    enable_jmespath_filter: bool = Field(False, alias="ENABLE_JMESPATH_FILTER")
+
     # Seed values for tool visibility (comma-separated tool names).
     # Used as initial config when no tool_config.json exists.
     # The web settings UI (/settings) is the primary interface for managing these.
--- a/src/ha_mcp/middleware/__init__.py
+++ b/src/ha_mcp/middleware/__init__.py
@@ -1,0 +1,3 @@
+from .response_filter import JMESPathFilterMiddleware
+
+__all__ = ["JMESPathFilterMiddleware"]
--- a/src/ha_mcp/middleware/response_filter.py
+++ b/src/ha_mcp/middleware/response_filter.py
@@ -85,7 +85,7 @@ def _apply_jmespath(result: ToolResult, expr: str) -> ToolResult:
     except jmespath.exceptions.JMESPathError as exc:
         warning: dict[str, Any] = dict(data) if isinstance(data, dict) else {"result": data}
         warning["_jmespath_warning"] = str(exc)
-        return ToolResult(content=warning, structured_content=warning)
+        return ToolResult(content=[mt.TextContent(type="text", text=json.dumps(warning))], structured_content=warning)
 
     if filtered is None:
         filtered_dict: dict[str, Any] = {}
@@ -94,4 +94,4 @@ def _apply_jmespath(result: ToolResult, expr: str) -> ToolResult:
     else:
         filtered_dict = {"result": filtered}
 
-    return ToolResult(content=filtered_dict, structured_content=filtered_dict)
+    return ToolResult(content=[mt.TextContent(type="text", text=json.dumps(filtered_dict))], structured_content=filtered_dict)
--- a/src/ha_mcp/middleware/response_filter.py
+++ b/src/ha_mcp/middleware/response_filter.py
@@ -53,9 +53,7 @@ class JMESPathFilterMiddleware(Middleware):
         jmespath_expr: str | None = args.get(_PARAM_NAME)
 
         if jmespath_expr is not None:
-            clean_args = {k: v for k, v in args.items() if k != _PARAM_NAME}
-            new_message = context.message.model_copy(update={"arguments": clean_args})
-            context = context.copy(message=new_message)
+            context.message.arguments = {k: v for k, v in args.items() if k != _PARAM_NAME}
 
         result = await call_next(context)
 
--- a/src/ha_mcp/middleware/response_filter.py
+++ b/src/ha_mcp/middleware/response_filter.py
@@ -21,7 +21,7 @@ _PARAM_NAME = "_jmespath"
 _PARAM_SCHEMA: dict[str, Any] = {
     "type": "string",
     "description": (
-        "Optional JMESPath filter applied server-side before returning (see jmespath.org). "
+        "Optional JMESPath filter for results - see jmespath.org"
         "Examples: 'state', '{id: entity_id, state: state}', 'entities[?domain==`light`]'."
     ),
 }
@@ -42,10 +42,17 @@ class JMESPathFilterMiddleware(Middleware):
         call_next: CallNext[mt.ListToolsRequest, Sequence[Tool]],
     ) -> Sequence[Tool]:
         tools = await call_next(context)
-        tools_copy = copy.deepcopy(list(tools))
-        for tool in tools_copy:
-            tool.parameters.setdefault("properties", {})[_PARAM_NAME] = _PARAM_SCHEMA
-        return tools_copy
+        result = []
+        for tool in tools:
+            # Shallow-copy the Tool to avoid mutating the registry.
+            # Full deepcopy raises "cannot pickle '_thread.RLock' object" because
+            # FastMCP Tool objects hold threading locks internally; only the
+            # parameters dict (plain JSON Schema) needs to be copied.
+            tool_copy = copy.copy(tool)
+            tool_copy.parameters = copy.deepcopy(tool.parameters)
+            tool_copy.parameters.setdefault("properties", {})[_PARAM_NAME] = _PARAM_SCHEMA
+            result.append(tool_copy)
+        return result
 
     async def on_call_tool(
         self,
--- a/src/ha_mcp/middleware/response_filter.py
+++ b/src/ha_mcp/middleware/response_filter.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import copy
 import json
 import logging
 from collections.abc import Sequence
@@ -8,8 +9,11 @@ from typing import Any
 import jmespath
 import jmespath.exceptions
 import mcp.types as mt
+from fastmcp.exceptions import ToolError
 from fastmcp.server.middleware import CallNext, Middleware, MiddlewareContext
 from fastmcp.tools.base import Tool, ToolResult
+
+from ha_mcp.errors import ErrorCode, create_error_response
 
 logger = logging.getLogger(__name__)
 
@@ -17,14 +21,12 @@ _PARAM_NAME = "_jmespath"
 _PARAM_SCHEMA: dict[str, Any] = {
     "type": "string",
     "description": (
-        "Optional JMESPath expression applied server-side to filter this tool's "
-        "response before it is returned, reducing token usage. "
-        "Examples: 'state' — single field; "
-        "'{id: entity_id, state: state}' — projection; "
-        "'entities[?domain==`light`].entity_id' — filter array. "
-        "On expression error the full response is returned with a '_jmespath_warning' key."
+        "Optional JMESPath filter applied server-side before returning (see jmespath.org). "
+        "Examples: 'state', '{id: entity_id, state: state}', 'entities[?domain==`light`]'."
     ),
 }
+
+_ENVELOPE_KEYS = ("success", "partial", "warning", "error", "count", "message")
 
 
 class JMESPathFilterMiddleware(Middleware):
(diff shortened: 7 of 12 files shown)


kingpanther13:
This is very interesting. While it seems that it will reduce context usage overall my concern is it will increase idle context usage since we have to add this to every tool docstring.  This also is a pretty major change overall. I at the very least would want it to have an opt-in toggle in the config settings for the addon, and env vars on the non-addon side so it's fully optional to use, and it should be default off.  I haven't reviewed code yet so just bringing that up in case you haven't already included those things. 

For a feature this big I'm going to request the owner @julienld to give a final thumbs up on. I'll still do my normal code review process on my side, but I want his blessings before merging this.

kingpanther13 (changes requested):
Thanks for the PR — the use case is real and the JMESPath choice for filter-only (vs. transform) is fine. Three independently-verifiable contract violations need to be fixed before this lands, and there's a meta-ask about scoping.

### Blockers

**1. Envelope drop violates the documented degraded-response contract.**
`src/ha_mcp/middleware/response_filter.py:86,99` runs JMESPath against the entire response and returns the search result as the new payload. AGENTS.md (lines 606–611) documents the degraded shape `{success, partial, warning, ...}`. A reasonable expression like `data.entity_id` collapses the response to `{"result": "..."}`, silently dropping `success`/`partial`/`warning`. Three current tool families produce this shape (`tools_registry.py`, `tools_search.py`, `tools_service.py`), so it isn't theoretical.
Fix: filter `data` only and re-attach the envelope, or merge `success`/`partial`/`warning` back onto the filtered result.

**2. `on_list_tools` mutates live `Tool.parameters`.**
`response_filter.py:42-45` calls `tool.parameters.setdefault(...)` on the registered Tool. FastMCP's `LocalProvider` returns the registry-stored objects and `Tool.parameters` is a shared `dict[str, Any]`, so the first `tools/list` permanently injects `_jmespath` into every registered Tool, leaking into anything else that introspects the registry (other middleware, transforms, future caching). Idempotent today, fragile.
Fix: build a copy of the parameters dict and return new Tool instances.

**3. JMESPath errors should raise `ToolError`, not embed a warning key.**
`response_filter.py:85-90` swallows `JMESPathError` into a `_jmespath_warning` sentinel. AGENTS.md lines 564–611 require tool-level failures to `raise ToolError` (sets `isError=true` per MCP spec). A malformed caller-supplied expression is exactly that kind of error, and agents have no guarantee they surface a sentinel key inside an arbitrary response shape.
Fix: raise `ToolError(json.dumps(create_error_response(ErrorCode.VALIDATION_INVALID_PARAMETER, ...)))` following `errors.py`/`helpers.py`.

### Required: opt-in/opt-out toggle

This adds an unconditional `_jmespath` parameter to every tool's schema (~341 chars × 92 tools ≈ 31 KB per `tools/list`) and routes every call through filter logic. AGENTS.md treats tool-description budget and progressive disclosure as priorities. Please add a setting in `config.py` (e.g. `enable_jmespath_filter`, env-toggleable) so the middleware can be disabled cleanly. With a toggle this becomes a feature users who want it can turn on, rather than a server-wide schema mutation.

### While you're in there

- `response_filter.py:72-83` — JSON-parse fallback silently no-ops; `logger` is imported and never used. At minimum `logger.warning(...)` so this branch is observable.
- `response_filter.py:92-93` — an expression returning `None` collapses to `{}`, indistinguishable from a successful empty match. `{"result": None}` or a "no match" warning would let callers tell them apart.
- `tests/src/e2e/basic/test_jmespath_filter.py` — please add coverage for (a) degraded responses preserving `partial`/`warning` after a sub-field projection, (b) expression returning `None`, (c) `on_list_tools` mutation idempotency.
- `_apply_jmespath` is a pure function — `tests/src/unit/` exists and would be the right home for fast tests of the filter behaviors above.
- `_PARAM_SCHEMA` description (~44 words) could be tightened to ~10 words + a JMESPath link, since it ships on every tool.

### One more ask

Even after the above is addressed, this is a server-wide architectural addition touching every tool's request and response path. I'd like @julienld to weigh in with a thumbs-up before merge — not a code review, just a sign-off on introducing an MCP-level filter layer. If the toggle above lands as opt-in (or with a clean opt-out), this becomes much less load-bearing and the second-maintainer ask is more courtesy than gating.

tomwilkie:
Thank you for the comments @kingpanther13! That all looks reasonable, I'll work through them this afternoon / evening.

tomwilkie (commented):
(Sorry, this was claude responding, let me review everything first!)

✅ Addressed all feedback from kingpanther13 in commit 20ab39f.

**Blockers:**

1. **Envelope drop** — `_apply_jmespath` now merges `success`, `partial`, `warning`, `error`, `count`, and `message` from the original response back onto the filtered result. Dict projections are returned directly (not wrapped); scalars/lists are wrapped in `{"result": ...}`.

2. **`on_list_tools` registry mutation** — replaced the in-place `setdefault` with `copy.deepcopy(list(tools))` so the middleware operates on cloned objects and returns those, leaving the FastMCP registry untouched.

3. **`ToolError` on bad expressions** — `_jmespath_warning` sentinel removed. Invalid expressions now raise `ToolError(VALIDATION_INVALID_PARAMETER)` with the expression in context and a link to jmespath.org. Updated the e2e test accordingly.

**Opt-in toggle:**

Added `ENABLE_JMESPATH_FILTER` env var (default `false`) to `config.py`, `server.py`, both addon `config.yaml` files, and `start.py`. The middleware is no longer registered unconditionally — it only loads when the flag is set. E2e conftest enables it alongside the other feature flags.

**While you're in there:**

- Added `logger.warning(...)` on the non-JSON text fallback branch.
- `None` result now returns `{"result": None}` instead of `{}`.
- `_PARAM_SCHEMA` description trimmed from ~44 words to ~10 words + `jmespath.org` link.
- Added e2e tests for: None result, envelope preservation after projection, `list_tools` idempotency.
- Added `tests/src/unit/test_response_filter.py` with 11 pure unit tests covering all `_apply_jmespath` behaviors (all pass, no Docker needed).

tomwilkie:
I think thats addressed all the feedback - I still need to add documentation, but before I do let me know what you think?

kingpanther13 (commented):
Sorry my Claude responded before I had a chance to review too lol. Sorry for poor wording , but it's pretty accurate. This is it's response:

Thanks for the fast turnaround — blockers all show up in the code (`response_filter.py:110-113` for envelope merge, `:51-53` for the copy, `:95-101` for `ToolError`), and the toggle is wired through `config.py` / addon configs / `start.py` / e2e conftest. Nice unit-test layer in `tests/src/unit/test_response_filter.py`.

A few items before I'd lift the "changes requested":

**CI is red on the envelope test that's supposed to prove the fix.**
[`test_jmespath_projection_preserves_envelope`](https://github.com/homeassistant-ai/ha-mcp/actions/runs/25506898074/job/74866887987) asserts `"success" in data` after `_jmespath: "data.state"` and gets back `{"result": "above_horizon"}` only. The unit test passes because it constructs `_make_result({"success": True, "data": ...})` directly; the live `ha_get_state` response apparently doesn't have `success` at the top level where `_apply_jmespath` looks, so the merge loop at `response_filter.py:110-113` finds nothing to copy. Worth tracing where the envelope keys actually live in the structured response.

**Other CI / merge cleanup**
- Ruff lint fails on `tests/src/unit/test_response_filter.py:3` (import sort) — `uv run ruff check --fix` clears it.
- Conflicts against current `master` in `pyproject.toml` and `homeassistant-addon-dev/config.yaml` — please rebase.

**Minor — `_PARAM_SCHEMA` description** (`response_filter.py:23-26`)
The implicit-concat strings render as `"...see jmespath.orgExamples: '{areas:areas[*].{id: area_id, name: name}'."` — missing space, and the example has unbalanced braces (opens 2, closes 1). Suggest `"Optional JMESPath filter — see https://jmespath.org. Example: '{areas: areas[*].{id: area_id, name: name}}'"`.

**Before merge — BAT round on my end**
Once CI is green and the rebase is done, I want to run thorough bot-acceptance testing before this lands. Specifically checking that (a) it actually saves tokens in realistic agent loops, not just on the example payloads, (b) `_jmespath` showing up on every tool doesn't degrade tool selection or push agents into weird routing, and (c) filter expressions don't accidentally strip information agents need (e.g. attribute keys that aren't the obvious projection target). The toggle helps a lot here — I can A/B with it on/off.

kingpanther13 (changes requested):
Round-2 cleanups are clean in isolation (description fix balanced, ruff is a no-op import sort, conflicts gone), but the envelope-preservation fix from round-1 has a hole the round-2 test switch papers over instead of closing.

### Blocker (carried over from round-1 — not yet fixed)

`add_timezone_metadata` (`src/ha_mcp/tools/util_helpers.py:352-375`) wraps responses as `{"data": <inner>, "metadata": {...}}`. For tools using this helper, `success`/`partial`/`warning` live one level deep inside `data["data"]`, so the envelope-merge loop at `response_filter.py:110-113` (which only inspects top-level keys) never matches. The round-1 fix silently does nothing for those tools.

The commit message on `c5cccf54` acknowledges the structural problem: *"`ha_get_state` wraps its response via `add_timezone_metadata` ({data: ..., metadata: ...}), so success is never at the top level where the envelope merge looks."* The chosen action was to switch the test to a tool that doesn't use the wrapper, not to fix the merge.

**Affected surface (~12 tools), grepped from wrap sites:**
- `ha_get_state` (single + batch via `_fetch_state` at `tools_search.py:1034`)
- `ha_search_entities` (`tools_search.py:310,353,367,412,524`)
- `ha_get_history` / statistics (`tools_history.py:569,743`)
- `ha_get_logs` (`tools_utility.py:433`)
- `ha_install_mcp_tools` (`tools_mcp_component.py:230,290`)
- `ha_hacs_search` / `repository_info` / `add_repository` / `download` (`tools_hacs.py:256,354,488,604`)

Concrete failure: `tools_search.py:519-524` sets `result["partial"] = True` and `result["warning"]` immediately before the `add_timezone_metadata` wrap. An agent calling `ha_search_entities` with `_jmespath="data.results"` receives `{"result": [...]}` with no envelope keys re-attached — exactly the silent failure round-1 was meant to prevent.

**Two ways to fix:**
1. Make `_apply_jmespath` wrapper-aware: when `data` has the `{"data": ..., "metadata": ...}` shape, merge envelope keys from `data["data"]`.
2. Promote `success`/`partial`/`warning` outside the wrapper in `add_timezone_metadata` (touches more code, gives a consistent invariant project-wide).

### Test gaps

- **No E2E for the wrapped path.** `test_jmespath_projection_preserves_envelope` was deliberately moved to `ha_config_list_areas`. Add a companion test against `ha_get_state` (or any wrapped tool) asserting `success` survives a sub-field projection — with current code, that test would fail and would expose the blocker.
- **Unit tests don't reproduce the wrapped shape.** `_make_result()` at `tests/src/unit/test_response_filter.py:16-20` only builds flat `{success: True, data: ...}` shapes. Add a fixture constructing `{"data": {"success": True, "partial": True, "warning": "..."}, "metadata": {...}}` and pin the contract.
- **No `partial=True` exercised end-to-end.** Easiest entry point is the partial-listing branch at `tools_search.py:128`. Once the wrapper fix is in, one E2E hitting that path with a projection locks it in.

### Positive

- `response_filter.py:23-26` description renders clean.
- Ruff fix is pure import sort.
- RLock-pickling protection (round-1) still solid: `test_on_list_tools_idempotent` + `TestOnListToolsDeepCopy` lock the regression.

### Verdict

Re-iterating CHANGES_REQUESTED. Round-1 blocker isn't fixed, just hidden. Once the merge handles the wrapped shape and a wrapped-tool envelope test is added, I'll do the BAT round we discussed.

kingpanther13 (changes requested):
I tested this against my live HA with the toggle on and ran sonnet sub-agents through realistic read-only scenarios (lights summary, temperature per room, HA overview). Laying out what I found objectively, then the changes I'd need to merge it as-is, then a cleaner alternative if the refactor is too much.

## What works

Round-3's fixes apply cleanly — wrap-detection-via-test-switch passed, lint/mypy/builds green, conflicts resolved. Filtered-call savings on tools that don't go through `add_timezone_metadata` are real and large:

| Tool | Wrapped via `add_timezone_metadata`? | Baseline | Filtered (correct path) | Reduction | Envelope keys preserved? |
|---|---|---|---|---|---|
| `ha_config_list_areas` | no | ~5,500 | ~1,400 | 75% | yes |
| `ha_get_overview` (light filter) | no | ~2,300 | ~145 | 94% | yes |
| `ha_get_state` (bulk 7 entities) | yes | ~52,000 | ~440 | 99% | **no** |
| `ha_get_history` (1 entity) | yes | ~700 | ~65 | 91% | **no** |
| `ha_get_history` (3 entities) | yes | ~2,000 | ~180 | 91% | **no** |
| `ha_search_entities` (5 lights) | yes | ~1,400 | ~280 | 80% | **no** |

## What didn't work in real agent flows

Sonnet sub-agents attempted `_jmespath` on read-only tasks 3 of 3 runs and guessed the wrong inner path on the first call every time. Examples:
- `results[*]` instead of `data.results[*]` for `ha_search_entities`
- `data.results` instead of `data.states.*` for `ha_get_state` (whose inner shape is a dict keyed by `entity_id`, not a list)

Each got `{"result": null}` back — indistinguishable from a successful empty match — and retried without the filter, costing the round-trip. Each agent independently flagged the silent-null behavior unprompted (*"a footgun worth noting"*, *"silent null-out makes mismatches hard to debug"*, *"JMESPath applied pre-wrap... failed silently with null"*). Even with full schema knowledge in front of me, my own first-attempt path on `ha_get_state` was wrong.

Inner shapes vary across tools (`data.results`, `data.states` (dict), `data.entities`, `data.areas`, plain top-level for unwrapped tools) and the per-tool schema doesn't tell the agent what to write. The headline 99% reduction on bulk `ha_get_state` is exactly where the path-guess failure mode is worst.

## What still needs to land

These were either carried from prior rounds or surfaced by live testing:

**1. Wrap-aware envelope merge** (`response_filter.py:110-113`). The current loop only inspects top-level keys, so for the ~12 tools wrapped via `add_timezone_metadata` (`tools_search.py:310,353,367,412,524,932`, `tools_history.py:569,743`, `tools_hacs.py:256,354,488,604`, `tools_mcp_component.py:230,290`, `tools_utility.py:433`), envelope keys silently drop. Discriminator I'd want:
```python
def _is_timezone_wrapped(d):
    return (
        isinstance(d, dict)
        and {"data", "metadata"} <= d.keys()
        and isinstance(d.get("metadata"), dict)
        and "home_assistant_timezone" in d["metadata"]
    )
```
When wrapped: unwrap, apply JMESPath to `d["data"]`, restore `metadata` and inner envelope keys.

**2. Silent-null → `ToolError` on no-match** (`response_filter.py:103-108`). JMESPath returns `None` for both "key missing" and "key present with null value" — a naive raise breaks the second. Sentinel pattern handles both:
```python
_SENTINEL = "__HAMCP_NO_MATCH__"
filtered = jmespath.compile(f"({expr}) || `\"{_SENTINEL}\"`").search(data)
if filtered == _SENTINEL:
        ErrorCode.VALIDATION_INVALID_PARAMETER,
        "JMESPath expression matched no path",
        context={"expression": expr},
        suggestions=["Check the path against the tool's response schema"],
    ))
```
Use `raise_tool_error()` from `helpers.py` per AGENTS.md "Error Handling," not raw `ToolError(json.dumps(...))` like at `response_filter.py:96-101`.

**3. Per-tool opt-in via `annotations`**, replacing inject-on-every-tool. `response_filter.py:53` does `setdefault("properties", {})[_PARAM_NAME] = _PARAM_SCHEMA` with no collision check — any tool that ever defines its own `_jmespath` param gets silently overwritten. Scoping fixes that AND bounds the ~2.5K tokens of catalog overhead (`_jmespath` description × 92 tools on every `tools/list`). Note: `tool.annotations` can be `None`, so guard explicitly:
```python
def _is_filterable(tool):
    ann = getattr(tool, "annotations", None)
    if ann is None:
        return False
    extra = getattr(ann, "model_extra", None) or {}
    return bool(extra.get("jmespathFilterable", False))
```

**4. Don't merge envelope keys on dict projections** (`response_filter.py:110-113`). When filter result is a dict, don't splice `_ENVELOPE_KEYS` (`success`, `partial`, `warning`, `error`, `count`, `message`) into it — caller chose that exact shape, and many HA entity attribute dicts have their own `message`/`count` keys with different semantics. Only merge envelope keys when filter result is scalar or None-wrapped.

**5. Drop the `note` field from `add_timezone_metadata`** (`util_helpers.py:352-375`). The English sentence "All timestamps are in UTC. Home Assistant timezone is ..." ships on every wrapped response, ~80 bytes of constant token waste regardless of whether `_jmespath` is even on. Free win.

**6. Initial opt-in list verified from source.** BAT inference is not source-truth — please confirm `ha_get_overview` wrap status from `tools_overview.py` (or wherever it's defined) before adding it.

**7. Tests rewritten.** `test_none_result_explicit` and `test_structured_content_none_match_explicit` currently assert silent-null behavior items 1–4 replace. They need to assert `ToolError` on missing-path while still passing on legit null-leaf values. Also add: wrapped-tool envelope preservation after sub-field projection, leaf-null pass-through, no-match raises, opted-out tool absence of `_jmespath` in schema, dict projection doesn't inject envelope keys.

## Honest read after live testing

The marginal benefit of the middleware over native per-tool `fields=` parameters is small. The 75-94% wins are on small list tools where ~4KB savings don't move the needle in real agent budgets. The 99% headline on bulk `ha_get_state` is precisely where path-guessing fails worst — and that case is better served by a native `compact: bool` or split `fields=`/`attribute_keys=` parameter on the tool itself. The middleware ends up being a second response-shaping path the codebase has to maintain alongside whatever native work eventually happens.

That said, your diagnostic work is what proved the token problem is real and quantifiable, and you've already absorbed the wrap-bug's full shape. I'm willing to keep working on this if you are.

## Two ways forward — your call

**Option 1 — Refactor + prove benefit.** Implement items 1–7. Then re-run agent-flow testing (any approach you'd like, BAT-style or otherwise) and demonstrate objectively that with the fixes in place, real agent flows actually realize the savings — i.e. agents don't trip over the silent-null/path-guess failure mode 3-of-3-runs the way they do today. If the post-refactor data shows real benefit, we merge.

**Option 2 — Close this PR; redirect to native `fields=`.** I file a parent issue for native per-tool `fields=` plumbing on the read-heavy tools (`ha_get_state` split into `fields=` + `attribute_keys=` to handle domain-variant attributes; `ha_get_history`, `ha_search_entities`, `ha_config_list_areas`, `ha_get_overview`, `ha_list_services` with simpler `fields=`). Each tool ships in its own small PR. You'd be welcome to claim any tool from the list — your BAT work translates directly to the per-tool implementations.

Both are real options. Option 1 honors what you've built; Option 2 lands a cleaner long-term architecture sooner. If you go Option 1, please also confirm `ha_get_overview`'s wrap status from source before locking the opt-in list.

Tell me which way you want to go and I'll set it up.

tomwilkie:
Thank you for the very thorough review! I had to step away halfway through the last set of changes - I should have put the PR back in to draft mode, sorry. 

My motivation here was using Claude to bring some consistency to my device and entity names. I've got ~300 and ~2.5k entities and I was burning my entire Claude window's token budget on a single pass on a single room...

I started optimising this by going device-by-device in sub agents, and then noticed the lists calls were burning tons of tokens (hence this PR). 

If you're open to per-tool 'fields=' parameters, I think that's probably a better approach. I can put up a PR for list devices and entities. 

Cheers!

kingpanther13:
Thanks @tomwilkie for the work on this and for the patient iteration through the review rounds. The middleware approach surfaced the response-shape problems clearly enough that we could see the right design — even if it ended up not being the design that ships.

After live testing on a real HA instance, the conclusion was that per-tool typed parameters (`fields=` / `attribute_keys=`) are a better fit than a server-wide JMESPath layer. Reasoning is in the most recent review on this PR; short version is the path-guess failure mode (3/3 sonnet runs missed the inner shape) plus the wrap-envelope handling complexity made the middleware less usable in real agent flows than per-tool params would be.

Filed the parent issue: #1199 — prioritized list of tools, the per-tool param shape, and an open invitation for you to claim any tool as your first conversion. No pressure either way; the list stays open if you'd rather pass.

Issue #1152, opened 2026-05-06
[FEATURE] Replace wait_for_entity_registered REST polling with WebSocket event subscriptions


kingpanther13:
## 💡 Suggest a Feature

Replace the REST-polling in `wait_for_entity_registered` (and its sibling `wait_for_entity_removed`, `wait_for_state_change`) with WebSocket event subscriptions for `entity_registry_updated` / `state_changed` / `config_entry_changed`.

## What would you like to see?

Today, after a write tool (`ha_config_set_helper`, `ha_config_set_automation`, `ha_config_set_script`, etc.) finishes its WS write, the tool calls `wait_for_entity_registered(client, entity_id)` to confirm the entity is queryable before returning. That helper, in `src/ha_mcp/tools/util_helpers.py`, polls the REST `/api/states/<entity_id>` endpoint at 200-500ms intervals up to a deadline.

Switch this to an event-based wait: open a WebSocket subscription to `entity_registry_updated` (and `state_changed` for state-dependent waits), filter for the target `entity_id`, and resolve as soon as HA pushes the event. Same for the other `wait_for_*` helpers.

## Why do you need this?

**Concrete failure mode in `ha_config_set_helper`:** on a slow HA instance (Raspberry Pi with many integrations, large recorder backlog), the polling can time out before the entity becomes queryable, surfacing a `helper_data["warning"]` like:

> *"Helper created but input_number.foo not yet queryable. It may take a moment to become available."*

The create actually succeeded — HA persisted the storage entry and started entity registration. The warning fires because the REST `get_state` poll didn't see the entity hydrated within the deadline. The agent then sees `success: true` with a warning and may decide to retry, query, or report the operation as partially failed.

WS event subscription would fix this:
- **No deadline-based polling.** The event arrives the instant HA registers the entity. No false timeouts.
- **No race window.** Currently the entity may be in the entity registry but its `state` not yet populated. Subscribing to `state_changed` (or both events) eliminates the window between registry-add and state-hydration.
- **Faster.** Removes 200-500ms median latency per write tool.

## Additional context

- **Source files**:
  - `src/ha_mcp/tools/util_helpers.py:wait_for_entity_registered` (and `wait_for_entity_removed`, `wait_for_state_change`)
  - `src/ha_mcp/client/websocket_client.py` (already supports subscriptions for the listener path)
[...]

Issue #1150, opened 2026-05-06
[BUG] ha_config_set_helper: 19 bugs found in audit (silent drops, destructive UPDATE, broken multi-step flows, phantom-ID acceptance, etc.)


kingpanther13:
## 💬 What Happened?

While stress-testing `ha_config_set_helper` for the issue #1074 fix (PR #1143), I ran 30 parallel agents against the test instance and noticed lots of false-success patterns — agents got `success: true` from the tool, but the helpers in HA didn't actually have the requested config. Digging in turned up not just the original `name`-required bug but a wide ecosystem of silent-drop, destructive-update, and discriminator issues across all 27 helper types.

This is filed as a meta-issue covering the full audit. **A single PR is in flight to fix all of them.** Filing this so progress isn't lost mid-implementation.

Related: #1074 (original symptom), #1143 (partial fix), #1149 (schema-awareness feature request that complements these fixes).

### 19 confirmed bugs

Source citations are line numbers in `src/ha_mcp/tools/tools_config_helpers.py` unless noted.

| # | Bug | Severity | Source |
|---|---|---|---|
| 1 | `input_number.initial` silently dropped on CREATE — never written to message | HIGH | `:909-936` |
| 1b | Same, on UPDATE — and **always wipes existing initial since it's never in the update message** | HIGH | `:1464-1480` |
| 2 | `min`/`max` shorthand kwargs rejected (must be `min_value`/`max_value`) | MEDIUM | pydantic schema |
| 3 | `name`-required runtime error reachable; #1143 docstring patched discoverability but root cause is schema-awareness | LOW | `:867-872` |
| 4a | `input_select.initial` silently dropped if not in `options` (no warning) | MEDIUM | `:906-907` |
| 4b | **Wrong-type params silently ignored** — passing `min_value` to input_boolean, `options` to input_number, `latitude` to person, etc. is silently no-op'd. Affects every type. ~70-90% of typed signature params silent-drop for any single helper-type call. | MEDIUM | every `elif helper_type == X` branch |
| 5 | Truthy `if param:` checks drop falsy-but-valid values: `initial=""` for input_text, `duration=""` for timer | LOW | `:945`, `:984`, `:1004` |
| 6 | Out-of-domain enum values silently coerced: `mode="default"` for input_number → silently becomes "slider"; `mode="slider"` for input_text → silently becomes "text" | MEDIUM | `:935`, `:943`, `:1479`, `:1487` |
[...]

Issue #1149, opened 2026-05-06
[FEATURE] Surface helper-creation schema inside ha_config_set_helper instead of a separate ha_get_helper_schema tool


kingpanther13:
### What would you like to see?

`ha_get_helper_schema` is currently a separate read-only tool that returns the form schema for **flow-type** helpers (template, group, utility_meter, etc.). It's only useful in a single moment — when an LLM is about to call `ha_config_set_helper` and doesn't know what fields go in the `config` dict for the chosen helper type. After that one moment it has no purpose.

Two consolidations to consider, possibly together:

1. **Merge schema-discovery into `ha_config_set_helper`.** Either via an explicit `dry_run=true` / `return_schema=true` flag that returns the field list without creating, or by auto-attaching the relevant schema to validation errors (e.g. when `config` is missing required fields, or when the simple-type call is missing a required parameter, the error response includes the schema inline). The model sees the schema *at the exact moment* it's about to make the call, not via a tool it has to remember to use.

2. **Extend schema coverage to simple types.** Right now `ha_get_helper_schema` rejects simple types (`input_number`, `input_text`, `counter`, etc.) with a `literal_error` — only the 15 flow types are supported. Simple types have their own per-type required/optional fields too (e.g. `options` is required for `input_select`, `latitude`+`longitude` for `zone`, `tag_id` for `tag`). The LLM has to derive these from the docstring rather than getting an authoritative schema.

If both happen, `ha_get_helper_schema` can be removed entirely (per [AGENTS.md guidance](https://github.com/homeassistant-ai/ha-mcp/blob/master/AGENTS.md) on tool consolidation: *"With 92+ tools, this project exceeds the 10-20 tool threshold where tool selection accuracy degrades"*).

### Why do you need this?

Two concrete cases where schema-aware-at-the-call-site would have prevented bugs:

[...]

PR #1154, opened 2026-05-06
fix(helpers): validation framework + control-flow + remaining audit bugs (2/3)

kingpanther13:
## What does this PR do?

The second of three stacked PRs closing #1150. **Stacked on PR 1** (`fix-helper-critical`). Adds the pre-validation framework + remaining bug fixes that tighten the tool's input handling and surface clearer errors instead of silent drops.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [x] 🔧 Maintenance/refactor (validation framework added)
- [x] 🧪 Tests only (7 new unit-test files added)
- [ ] 💥 Breaking change

## Testing
- [x] I have tested these changes with a LLM agent (full no-context test runs across haiku/sonnet/opus, against the stacked branch tip)
- [x] All automated tests pass (`uv run pytest`) — 341 helper-related unit tests, 0 regressions
- [x] Code follows style guidelines (`uv run ruff check` clean, `uv run mypy src/` clean)

## Bugs fixed in this PR (16 of 19 from #1150)

This PR contains both **validation framework** bugs AND **control-flow** bugs because their source changes share the same function bodies and can't be cleanly atomically separated without significant refactoring risk on a 19-bug fix scope. Both groups are listed below.

### Validation framework

| # | Bug | Fix |
|---|---|---|
| 4a | `input_select.initial` silently dropped if not in `options` | Now raises `VALIDATION_INVALID_PARAMETER` with the offending value and the available options |
| **4b** | **Wrong-type params silently ignored** | `_TYPE_TYPED_PARAMS` allowlist + `_validate_applicable_params` rejects inapplicable params with a clear "applicable parameters: …" message |
| 7c | `icon` silently stripped for `person`/`tag` | Folded into Bug 4b — now rejected with hint to use `picture` (person) or that tags don't support icons |
| 10 | `input_button` silently accepts and drops everything | Folded into Bug 4b allowlist (`input_button` accepts only `name`/`icon`) |
| 13 | Range/step validation gaps (`min==max`, `step=0`, `step > range`) | `_validate_numeric_range` pre-validates with clear messages |
| 14 | Flow types silently drop simple-helper params | Folded into Bug 4b — flow types accept only `config` + cross-cutting params |
| **16** | **Phantom `area_id`/`labels`/`category` silently accepted** | `_validate_registry_ids` validates against HA's registries before any registry-update WS call. Distinguishes lookup-failure (fail-open) from legitimately-empty registry (reject phantom) via an `(ok, items)` tuple |
| 17 | HA-side validations bubble through cryptically | `_validate_input_select_options` rejects duplicates; `_validate_schedule_days` rejects overlapping/malformed time ranges. `_validate_numeric_range` also catches input_text length 0-255 |

### Control-flow + UX

| # | Bug | Fix |
|---|---|---|
| 2 | shorthand `min`/`max`/`unit` kwargs rejected | Pydantic `validation_alias=AliasChoices(...)` on the relevant Fields. FastMCP routes args through pydantic TypeAdapter which honors the alias; the generated MCP schema still surfaces only canonical names |
| 3 | name-required runtime error reachable | Improved error messages and the explicit-action discriminator (Bug 11) make the intent clearer up front |
| 5 | Truthy `if param:` drops (`initial=""` for input_text, etc.) | Replaced with `if param is not None:` for `initial`, `duration`, etc. |
| 6 | Out-of-domain `mode` silently coerced | Now rejects with the allowed values for the chosen helper_type. Applied to `input_number` (`box`/`slider`) and `input_text` (`text`/`password`) on both CREATE and UPDATE paths |
| 7a | `schedule` no-day-of-week creates always-off | Now rejects with hint to pass at least one day-of-week with at least one time range |
| 7b | `zone` without lat/long → bubbled HA error | Tool-side preflight rejects with a friendly error naming both fields |
| **11** | **Action discriminator (create vs update) was implicit** | New `action: Literal["create","update"] | None` parameter. When omitted, behavior is unchanged (back-compat: implicit from `helper_id` presence). When explicit, contradictions with `helper_id` are rejected up front. ENTITY_NOT_FOUND from misspelled helper_id now also suggests "did you mean to create?" |
| **12** | **Silent name collision creates `_2` duplicates** | `_check_name_collision` queries `{type}/list` (or `config_entries/get` for flow types) before create; rejects with the existing helper_id in the suggestion if the slug already exists |
| 15 | Cryptic `"API error: 400 - Bad Request"` from flow types | Parses HA's voluptuous error body for field-level details; attaches the helper_type's `data_schema` when the body is unstructured |

## Out of scope (deferred to PR 3)

- **Performance simplifications**: parallel `asyncio.gather` for the registry-id lookups, per-entity registry-update loop, and within-entity update concurrency. These are pure perf wins from the simplify-skill review and are isolated in PR 3.
- **DRY refactors**: `_validate_mode` helper extracting the duplicated mode validation; `_id_set` helper extracting the registry-list parsing; narrative-comment cleanup. Mechanical de-dup, also in PR 3.
- **Bot review-comment fixes**: 4 inline-bot threads from the original PR (implicit string concat, unused vars/imports). PR 3.

This PR also defers:
- Issue #1149 (schema-awareness consolidation) — separate feature work
- Issue #1152 (WS event subscriptions to replace polling) — separate refactor

## Future improvements

- Extract a `_TYPE_FIELD_SPEC` table to unify the per-helper-type CREATE+UPDATE elif chains (~150 line dedup) — flagged as HIGH-severity by the simplify-skill review but each type has quirks (string-to-bool for input_boolean, int casting for counter, has_date/has_time defaulting, etc.) that don't fit cleanly. Worth a focused follow-up PR after this lands.

## Checklist

## Tests added in this PR

- `test_helper_param_rejection.py` (39 tests, Bug 4b/7c/10/14)
- `test_helper_input_validation.py` (24 tests, Bug 13/17)
- `test_helper_registry_id_validation.py` (5 tests, Bug 16)
- `test_action_and_collision.py` (14 tests, Bug 12 plus rename-pattern protection)
- `test_action_discriminator.py` (14 tests, Bug 11 explicit action)
- `test_helper_kwarg_aliases.py` (7 tests, Bug 2)
- `test_flow_error_parsing.py` (6 tests, Bug 15)

Plus mock-response queue updates to existing `test_helper_update_persistence.py` for the new validation lookups.

Refs #1150.

### DIFF
--- a/src/ha_mcp/tools/tools_config_entry_flow.py
+++ b/src/ha_mcp/tools/tools_config_entry_flow.py
@@ -19,6 +19,7 @@ from fastmcp.exceptions import ToolError
 from fastmcp.tools import tool
 from pydantic import Field
 
+from ..client.rest_client import HomeAssistantAPIError
 from ..errors import ErrorCode, create_error_response
 from .helpers import (
     exception_to_structured_error,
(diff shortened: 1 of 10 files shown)


kingpanther13:
Closing — the split was 4047/153 instead of the requested ~2000/~2000. Re-doing the split properly.

Issue #995, opened 2026-04-16
[FEATURE] Add scene configuration tools (ha_config_get/set/remove_scene)


Patch76:
### What would you like to see?

Scene CRUD tools analogous to the existing script and automation families:

- `ha_config_get_scene` — retrieve a single scene or list all scenes
- `ha_config_set_scene` — create or update a scene
- `ha_config_remove_scene` — delete a scene

The HA REST API already exposes `GET/POST/DELETE /api/config/scene/config/{scene_id}` for individual scene operations. Listing all scenes reads from `scenes.yaml` via the same config view layer that scripts and automations use.

### Why do you need this?

Scenes are a first-class HA concept alongside automations and scripts, but ha-mcp has no way to inspect or manage scene configuration. An agent can activate a scene via `ha_call_service(scene.turn_on)` but cannot read, create, or modify the underlying config — making scene management entirely opaque.

The implementation path is well-established: `tools_config_scripts.py` is the direct reference pattern, and the REST endpoints are stable public API.

### Additional context

ha-mcp already handles the parallel families:
- Automations: `ha_config_{get,set,remove}_automation`
- Scripts: `ha_config_{get,set,remove}_script`

Scenes are the missing third member of this group. The HA Core source (`homeassistant/components/config/scene.py`) confirms the config view is registered and stable.

kingpanther13:
Hmm good catch I didn't realize we didn't have coverage for scenes, probably because I don't personally use them. I'm doubting that we can merge this into existing tools cleanly but it'd be good to look around to see if it may fit elsewhere... Otherwise these might genuinely be good new tools.

Patch76:
For reference I checked script + automation the same way — both also return 404 on the bare `config/{domain}/config` path. There is no REST listing for any of the three. So scenes are fully aligned with the established pattern, not a special case.

## Correction to the issue-bot analysis

The bot suggested making `scene_id` optional so `ha_config_get_scene()` can list all scenes, and extending the same to scripts/automations. Two problems:

1. There is no REST endpoint to list scene configs (verified above). Listing would require reading `scenes.yaml` directly or enumerating `scene.*` entities and fetching each — a scope jump well beyond this issue.
2. `ha_config_get_script` and `ha_config_get_automation` do **not** currently have optional identifiers; `script_id` / `identifier` are required. So the "consistency" argument for making `scene_id` optional actually cuts the other way.

[...]

PR #1160, opened 2026-05-07
fix(resources): reject HA-config YAML in dashboard resource content

### LINKED ISSUE #1072: [BUG] Entities created without entity ID
### 💬 What Happened?

I asked the agent to create a scene for a light entity. Depending of the model, it sometimes ended in creating a _group_ (gpt, haiku, minimax) rather than a scene (opus, sonnet) but ok i'm still experimenting on it.

but the point is that every entities ha-mcp creates it doesn't have the unique entity_id. Thus, it's "locked", not editable and impossible to delete. In facts, as usual with such kind of content, they disappears after a restart.

I'm probably issuing the wrong prompt ("create a scenery for desk led strip with warm colors"), nonetheless this is the behaviour i get sistematically.

I'm using VS Code as host, but Claude Desktop gave the same results.

```markdown
User: create a scene for LED Desk Strip to be used as night light.
```

Patch76:
## What does this PR do?

Closes #1072 (Path 2). `ha_config_set_dashboard_resource(content=...)` is for Lovelace JavaScript or CSS, but smaller LLMs sometimes pick this tool to "create a scene/automation/helper" and pass HA-config YAML as `content`. The misroute lands as a Lovelace module containing YAML, creating orphaned, unreachable HA entities — exactly theclue's reproduction in the issue body.

This PR adds an early content-shape check in `_set_inline_resource`:

- New module-level `_HA_CONFIG_YAML_MARKERS` mapping 20 top-level YAML keys (`scene:`, `automation:`, `script:`, `group:`, `input_boolean:`, …) to suggested replacement tools, or `None` where no direct tool exists yet (scenes point at #995).
- New helper `_detect_ha_config_yaml(content)` returns the matching domain when the first non-empty line opens a known HA-config block. Tight matching: exact `<key>:`, or `<key>:` followed by whitespace / EOL / YAML markers (`|`, `>`, `-`, `[`, `{`, `#`). Non-YAML separators after the colon (e.g. CSS `automation:hover`) don't match.
- Reject inline in `_set_inline_resource` after the empty-content and `resource_type='js'` guards. The error message names the marker, points at issue #1072, and surfaces the right replacement tool (or #995 for scenes) via `suggestions`. No WebSocket call is issued.

### Scope

This PR addresses **Path 2** of #1072 only. **Path 1** (Sonnet's `ha_call_service(domain="scene", service="create", ...)` falling into HA's documented transient-scene service) is the broader fix — tracked in **#995** (scene configuration tools, not yet shipped). The two paths together close the issue; this is the smaller, tracked follow-up KP13 explicitly invited in his 2026-05-06 reproduction comment ("tighten `ha_config_set_dashboard_resource` to warn or reject when `content` doesn't look like JS/CSS").

### Boy-Scout

`tools_resources.py` was already format-drifted on master (`ruff format --check` reported `Would reformat` against `60ba1f2` without any of my changes). The diff includes the format pass alongside the substantive change — the format-only lines are mostly `raise_tool_error(create_error_response(...))` block-style adjustments. New logic is ~70 LOC; the rest of `+303` is pre-existing format drift KP13 will see in the diff regardless.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

(Not breaking — pure additive validation. Any caller passing legit JS/CSS keeps working; only the documented misroute pattern is now rejected with a clear pointer to the right tool.)

## Testing
- [x] I have tested these changes with a LLM agent _(reproduced theclue's exact `scene:` payload locally; rejection now fires with the #995 pointer)_

Local: 67/67 tests in `test_tools_resources.py` pass; ruff lint + format clean. Broader unit suite has one unrelated environmental failure (`test_resources.py::test_skills_vendor_directory_exists`) that's a worktree-specific submodule-not-populated symptom, not present on master / CI.

## Future improvements

- **#995 (scene config tools)** is the real fix for Sonnet's path. When that lands, `scene.create` workarounds become unnecessary.
- The marker list could be widened to other less-common HA YAML keys (`zone:`, `device_tracker:`, `weather:`, …) if the misroute pattern keeps surfacing for them. The current 20 cover the documented `theclue` reproduction plus the most-misrouted helper types.
- E2E test coverage: an integration test that posts the `scene:` payload and asserts the structured error would lock the contract end-to-end. Could land as a follow-up PR alongside #995's E2E setup.

## Checklist
- [x] I have updated documentation if needed _(no doc-only changes; the rejected-content error message is self-documenting and points at the right tool)_


Patch76:
## Implementation Summary

**Choices Made:**
- **Reject early at validation, not at WebSocket-call time** — the misroute is a tool-selection mistake by the LLM, not a backend error. Rejecting at the inline-content guard (`tools_resources.py`, immediately after the empty-content and `js`-type guards) keeps the failure mode close to the cause and avoids any side effect on Lovelace state. No WebSocket call is issued for a YAML payload.
- **Marker-map keyed by domain → suggested tool** — `_HA_CONFIG_YAML_MARKERS` doubles as the detection set and the suggestion source, so adding a new marker is a single-line dict edit rather than touching the helper, the error builder, and a separate "right tool" map.
- **Match precision: exact `<key>:` or `<key>:` followed by `[ \t|>\-[{#]`** — narrow enough to skip CSS pseudo-classes (`automation:hover`) and JS/CSS strings/comments containing the marker word, broad enough to catch all real-world YAML shapes from theclue's reproduction (block-form, inline value, block-string `|`, flow-sequence `[…]`, comment `#…`, leading whitespace).
- **Don't match `homeassistant`/`sensor`/`binary_sensor`/etc. with a direct tool** — for these, suggesting a non-existent tool would mislead. Setting them to `None` in the map produces a generic "this tool only stores Lovelace JS/CSS resources" message + an honest "no direct tool exists; configure via UI/YAML packages" note.
- **Scene path explicitly cites #995** — when `_HA_CONFIG_YAML_MARKERS["scene"]` is hit, the error message includes the #995 pointer ("Scene configuration tools are tracked in #995; until they ship, scenes can only be created via the HA UI"). Closes the loop on the broader fix path KP13's reproduction comment outlined.
- **Boy-Scout reformat included** — `ruff format --check` reported `Would reformat: tools_resources.py` against `60ba1f2` (master) before my changes. The format pass is included rather than left as drift; per AGENTS.md `Boy Scout Rule` for trivial fixes in the same PR.

**Problems Encountered:**
- **`error_data["error"]["context"]` doesn't exist** — initial test assertions assumed `create_error_response`'s `context=` argument lived under `error.context`. Per `errors.py:256-258` it's flattened to the top level (`response.update(context)`). Updated the test to assert `error_data["detected_marker"]` and `error_data["resource_type"]` at top level. Caught by the first pytest run — fixed before push.
- **One unrelated worktree test failure** — `test_resources.py::test_skills_vendor_directory_exists` fails in the worktree because the `skills-vendor/skills` submodule directory isn't populated there (passes on the main checkout). Not caused by this PR; CI gets a fresh clone with full submodule init, so it'll pass on the PR validation pipeline.

**Suggested Improvements:**
- **#995 (scene config tools)** is the broader fix for Path 1 of #1072 (Sonnet's `ha_call_service(scene, create, ...)` workaround). When that lands, the `scene` row in `_HA_CONFIG_YAML_MARKERS` can be flipped from `None` to `ha_config_set_scene` (or whatever the final tool name lands as).
- **Widen the marker list** if the misroute pattern surfaces for other YAML keys (`zone:`, `device_tracker:`, `weather:`, …) — current 20 covers the documented case + most-confused helper types.
- **E2E coverage** — an integration test posting the `scene:` payload through the MCP boundary and asserting the structured-error response would lock the contract end-to-end. Worth a follow-up PR alongside #995's E2E setup.

Patch76:
Self-review before maintainer review.

The PR body and commit message say the new test "reproduces theclue's **exact** `scene:` payload locally". `test_inline_rejects_scene_yaml_with_995_pointer` actually uses a reduced version: `name` is truncated to `Luce Notturna` (theclue's #1072 paste was `Luce Notturna Tavolo`), and `icon: mdi:weather-night` / `brightness: 30` / `color_temp: 400` are dropped.

The detection logic doesn't depend on those fields — first-line `scene:` matches regardless — so the test's reject behaviour is sound. But the "exact" framing is overstated; the test literal is shape-equivalent, not byte-equivalent.

Two ways to reconcile when this lands:

1. Replace the test literal with theclue's full payload verbatim.
2. Rephrase the PR body / commit message to "reproduced the shape of theclue's `scene:` payload" or "matches theclue's reproduction pattern".

Flagging now so the discrepancy doesn't get lost between bot-wave and merge.

---

**Edit — BAT findings against live HA via OpenRouter (gemini-2.5-flash, gpt-4o-mini):**

1. Natural-language theclue prompt, gemini-2.5-flash, tool-search on — model didn't pick `ha_config_set_dashboard_resource`; gave up after failing to find scene-creation tools. Path-2 not exercised.

2. Natural-language theclue prompt, gpt-4o-mini, tool-search on — model reproduced **Path 1** (the broader misroute tracked in #995): created an automation with `scene.create` as a service action. Path-2 not touched. Side finding for #995, out of scope here.

3. Direct-stimulus call to `ha_config_set_dashboard_resource(content="scene:\n  - id: ...", resource_type="module")`, tool-search off, gpt-4o-mini — **rejection fires live as designed**: `code: VALIDATION_INVALID_PARAMETER`, message names `scene:` and cites #1072, suggestions include the #995 pointer plus the JS/CSS-only hint, `detected_marker: "scene:"`, `resource_type: "module"`. No WebSocket call issued (rejection is pre-`_upsert_resource`).

Unit-test-asserted behaviour reproduces under real MCP wire-up. Natural-language reproduction of Path-2 specifically is hard to surface — small models either give up or take Path 1 — so direct-stimulus is the cleanest validation path, and the unit tests already cover the marker matrix.

Patch76:
## Implementation Summary

**Choices Made:**
- **Detection placement** — heuristic in `_set_inline_resource` after the empty-content and `resource_type='js'` guards, before the size-cap check. Rejection fires as early as possible, before any base64-encoding work; no WebSocket call is issued.
- **8-byte separator allowlist** (` `, `\t`, `|`, `>`, `-`, `[`, `{`, `#`) — covers YAML's lexical surface (block-string `|`/`>`, sequence `-`, flow `[`/`{`, comment `#`, whitespace) without false-positiving on CSS selectors with alpha after colon (e.g. `automation:hover`).
- **20-marker dictionary instead of regex** — explicit dict mapping top-level HA-config keys to replacement tools (`None` where no direct tool exists yet; `scene` calling out #995 specifically). Per-key replacement-tool hint produces a more actionable error than a generic "looks like YAML".
- **Boy-Scout: ruff-format pass on `tools_resources.py`** — file was already format-drifted on master pre-PR (`ruff format --check` against `60ba1f2` reports drift independent of these changes). Folded into the substantive PR per AGENTS.md Boy-Scout rule. New logic is ~70 LOC; the rest of the diff is pre-existing reformat.

**Problems Encountered:**
- None blocking. Pre-existing format drift on master made the diff visually larger; documented in the PR body.

**Suggested Improvements (post-merge):**
- **Test-literal byte-exactness** — see the self-review comment above. Test uses a structurally-equivalent but reduced version of theclue's payload (`name` truncated, `icon`/`brightness`/`color_temp` dropped). Detection logic is unaffected, but the "exact" framing in the PR body and commit message is overstated. Either copy theclue's full payload verbatim or rephrase to "shape of theclue's payload".
- **Marker list expansion** — could widen to `zone:`, `device_tracker:`, `weather:` etc. if the misroute pattern surfaces for them.
- **E2E test** — current coverage is unit-only; an E2E posting the `scene:` payload and asserting the structured error would lock the contract end-to-end. Reasonable follow-up alongside #995's E2E setup.

**BAT validation** — see the BAT-findings edit on the self-review comment above; rejection fires live as designed under direct-stimulus against the RBO instance via gpt-4o-mini.

kingpanther13 (changes requested):
**1. Tighten `_detect_ha_config_yaml` to skip BOM, leading-comment, and `---` doc-start lines.**

- `"\ufeffscene:\n…"` slips through — `content.strip()` doesn't strip `\ufeff`.
- `"# header\nscene:\n…"` slips through — the first `\n`-split line is the comment, not `scene:`.
- `"---\nautomation:\n…"` slips through for the same reason.

One preprocessing step covers all three:

```python
stripped = content.lstrip("\ufeff")
first_line = ""
for line in stripped.splitlines():
    bare = line.strip()
    if not bare or bare == "---" or bare.startswith("#"):
        continue
    first_line = line.rstrip()
    break
if not first_line:
    return None
```

**2. Add cases to `test_real_world_yaml_shapes` covering BOM, leading-comment, and `---` doc-start** so the helper-level coverage matches the new behavior.

**3. Add an update-path rejection test.**

`_set_inline_resource` runs for both create and update via `resource_id`. All three reject tests pass `resource_id=None`. If a refactor moves the YAML check below `_upsert_resource`, a misroute could overwrite an existing resource and the suite stays green.

Mirror `test_inline_rejects_scene_yaml_with_995_pointer` with `resource_id="existing-id"` and the same `mock_client.send_websocket_message.assert_not_called()` assertion.

Patch76:
Addressed in 73c6328:

- ✅ **Ask 1** — `_detect_ha_config_yaml` skips leading BOM (`\ufeff`), blank lines, full-line `#` comments, and `---` doc-start before reading the first content line. Single preprocessing block following your sketch; the kept line is set to `line.strip()` (rather than `rstrip()`) so leading whitespace is also dropped — incidentally tightens the existing `"\n\n  scene:\n…"` case. Match loop after the preprocessing is unchanged.
- ✅ **Ask 2** — three new cases in `test_real_world_yaml_shapes` (BOM `\ufeff`, leading `#` comment, `---` doc-start) plus one combined-decoration case.
- ✅ **Ask 3** — new `test_inline_update_rejects_scene_yaml_with_995_pointer` mirrors the scene-YAML reject under `resource_id="existing-resource-id"`, locking the YAML check above `_upsert_resource`.

72/72 unit tests pass locally; ruff + mypy clean.

kingpanther13 (approved):
All three asks addressed:

1. BOM / blank-line / `#` comment / `---` doc-start preprocessing in `_detect_ha_config_yaml` — using `bare` (`strip()`) for the kept line is a sensible tightening.
2. `test_real_world_yaml_shapes` covers BOM, leading-comment, `---`, and combined-decoration cases.
3. `test_inline_update_rejects_scene_yaml_with_995_pointer` locks the YAML check above `_upsert_resource`.

Thanks.

PR #1178, opened 2026-05-08
fix: return newest automation traces, add offset+order pagination (#1177)

### LINKED ISSUE #1177: [BUG] ha_get_automation_traces returns oldest traces instead of newest when total > limit
## 📋 Bug Description

`ha_get_automation_traces` (list mode) returns the oldest N traces instead of the newest N when total stored traces exceed the requested `limit`. This makes it impossible to debug recent automation runs on automations with `stored_traces` set higher than 50.

## 🔄 Steps to Reproduce

1. Create an automation that runs frequently (e.g. every 15 minutes via time_pattern)
2. Set `stored_traces: 200` on that automation
3. Wait until the trace buffer fills (50+ hours for a 15-min automation)
4. Call `ha_get_automation_traces(automation_id, limit=50)`

**Expected:** The 50 most recent traces are returned (e.g. from the last ~12 hours for a 15-min automation).

**Actual:** The 50 oldest traces are returned (e.g. from 50+ hours ago). The response includes `total_available: 200` but the tool only ever surfaces the first 50 chronologically. There is no `offset` parameter, so the most recent traces are unreachable through the tool when `total_available > limit`.

## 🔍 Root Cause

In `src/ha_mcp/tools/tools_traces.py`, function `_format_trace_list`:

```python
for trace in traces[:limit]:
```

Home Assistant's `trace/list` WebSocket command returns traces in chronological order (oldest first). Slicing `[:limit]` therefore returns the oldest N. To return the most recent N, the slice should be `[-limit:]`.

## 💡 Suggested Fix

Change `traces[:limit]` to `traces[-limit:]` in `_format_trace_list`.

Optionally also consider:
- Reversing the result so newest appears first in the response (matches user expectation)
- Adding an `offset` parameter to `ha_get_automation_traces` so older traces remain reachable via pagination when needed

## 🔧 Environment

- **ha-mcp Version:** 7.4.0
- **Platform:** Windows 11 (AMD64)
- **Python Version:** 3.13.13
- **Home Assistant Version:** 2026.4.4
- **Connection Status:** Connected

## 💡 Additional Context

This bug was discovered while debugging a heating-control automation with `stored_traces: 200`. The user reported behaviour from the previous overnight run, but the tool consistently returned traces from 36+ hours earlier even with `limit=50`, making it impossible to investigate the actual incident through the AI agent.

Workaround: reduce `stored_traces` on the automation to ≤ the tool's max limit (50) so the buffer never grows beyond what `[:limit]` returns.

```markdown
today is 8th may roughly 6pm. you need to go back to 7th may 20:00 onwards

18:27
Claude responded: You're right.
You're right. I had the trace tool with limit which appears to return oldest 50, but I see the tool definition more clearly now — limit is max 50, returns most recent. Let me try again with limit 50 and look for May 7-8 traces in the response.


Get Automation Traces
Get Automation Traces
Same 50 traces every call. The tool returns the OLDEST 50 of the 200 stored, with no offset/pagination available in the parameters. I cannot get to the May 7-8 traces through this tool.

I'm not making excuses — I genuinely cannot retrieve the May 7-8 traces. The traces exist in HA (200 stored) but the tool only exposes the oldest 50. There's no offset parameter in the ha_get_automation_traces tool definition. I've checked.
```

kingpanther13:
## What does this PR do?

Closes #1177.

`ha_get_automation_traces` (list mode) was sliced as `traces[:limit]` against HA's `trace/list`, which returns traces oldest-first. With `stored_traces > limit` the user only ever saw the *oldest* N traces — recent runs were unreachable through the tool.

- Returns newest traces by default (most common debugging case).
- New `offset` parameter pages through stored traces beyond `limit`.
- New `order` parameter (`newest` | `oldest`) restores chronological access when needed.
- Response now echoes `offset`, `order`, and adds `has_more` so an agent can paginate.

The reporter's case (200 stored, `limit=50`, only oldest visible) is now solved by either the default newest-first behavior or `offset` for paging deeper.

## Type of change
- [x] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

Behavior of the default `limit=N` call changes (now newest-first instead of oldest-first), but the previous behavior was unusable when `total_available > limit`, so this is a fix rather than a breaking change. Pre-existing oldest-first behavior is recoverable via `order=\"oldest\"`.

## Testing
- [ ] Code follows style guidelines (\`uv run ruff check\`)

Reproduced live against an automation with 5 stored traces — calling with `limit=2` returned the oldest 2 (run timestamps confirmed). New unit tests cover newest/oldest ordering, offset paging in both directions, and the `has_more` flag at boundaries.

## Future improvements

None for this PR. If trace volume / pagination ergonomics become a concern, a `since` / `until` timestamp filter would be a natural next step but is out of scope here.

## Checklist

Tool docstring and field descriptions updated to mention `offset` / `order` / `has_more`.

### DIFF
--- a/src/ha_mcp/tools/tools_traces.py
+++ b/src/ha_mcp/tools/tools_traces.py
@@ -63,27 +63,12 @@ class TraceTools:
         limit: Annotated[
             int,
             Field(
-                description="Maximum number of traces to return when listing (default: 10, max: 50). Combined with `offset` and `order` to page through stored traces.",
+                description="Maximum number of traces to return when listing (default: 10, max: 50).",
                 default=10,
                 ge=1,
                 le=50,
             ),
         ] = 10,
-        offset: Annotated[
-            int,
-            Field(
-                description="Number of traces to skip from the start of the requested order. Use with `limit` to page through stored traces when `total_available > limit`.",
-                default=0,
-                ge=0,
-            ),
-        ] = 0,
-        order: Annotated[
-            Literal["newest", "oldest"],
-            Field(
-                description="Order traces are returned in. 'newest' (default) returns most-recent first; 'oldest' returns chronological-first.",
-                default="newest",
-            ),
-        ] = "newest",
         deduplicate: Annotated[
             bool,
             Field(
@@ -109,6 +94,21 @@ class TraceTools:
                 default=None,
             ),
         ] = None,
+        offset: Annotated[
+            int,
+            Field(
+                description="Number of traces to skip from the start of the requested order. Use with `limit` to page through stored traces when `total_available > limit`.",
+                default=0,
+                ge=0,
+            ),
+        ] = 0,
+        order: Annotated[
+            Literal["newest", "oldest"],
+            Field(
+                description="Order traces are returned in. 'newest' (default) returns most-recent first; 'oldest' returns chronological-first.",
+                default="newest",
+            ),
+        ] = "newest",
         ctx: Context | None = None,
     ) -> dict[str, Any]:
         """
@@ -507,7 +507,7 @@ def _format_trace_list(
     """
     # HA's trace/list returns traces oldest-first. Pick a window from the end
     # for newest-first, or from the start for oldest-first, with offset for
-    # pagination through stored traces beyond `limit` (issue #1177).
+    # pagination through stored traces beyond `limit`.
     if order == "newest":
         end = len(traces) - offset
         start = max(end - limit, 0)
--- a/src/ha_mcp/tools/tools_traces.py
+++ b/src/ha_mcp/tools/tools_traces.py
@@ -7,7 +7,7 @@ to help debug automation and script issues.
 
 import json
 import logging
-from typing import Annotated, Any
+from typing import Annotated, Any, Literal
 
 from fastmcp import Context
 from fastmcp.exceptions import ToolError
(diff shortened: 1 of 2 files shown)


kingpanther13:
## Implementation Summary

**Choices Made:**
- Default ordering changed to newest-first rather than just adding `offset` and keeping oldest-first. The bug-driving symptom is that recent traces are unreachable — making the common case (debug a recent run) work without parameters was worth a behavior flip. Old behavior remains accessible via `order="oldest"`.
- Added `order: Literal["newest", "oldest"]` rather than a boolean — clearer at call sites and leaves room to extend.
- Placed new `offset`/`order` params after `sections` (before `ctx`) so existing positional callers of earlier params still bind correctly. FastMCP clients pass kwargs over JSON, but the convention is worth keeping.
- Response echoes `offset`, `order`, and adds `has_more` so an LLM caller has explicit pagination signals without having to compute them from `total_available`.

**Problems Encountered:**
- None worth noting. Reproduced live against a 5-trace automation (`limit=2` returned the oldest two — confirmed via timestamps), confirmed `_format_trace_list` is the only sliced site, fixed in one place.

**Suggested Improvements:**
- None.

Patch76 (approved):
[Round 1 — review on `9bbf91b`]

Window logic in `_format_trace_list` is correct across the boundary cases I traced — `offset == total` returns empty + `has_more=False` (caught by the `if end > 0` guard), `offset > total` falls into the same guard, and `limit > available` returns all without overshoot. The three-tier slice (`end = len-offset`, `start = max(end-limit, 0)`, then `reversed`) handles the newest path cleanly.

`TestFormatTraceList`'s 9 new cases hit the right edges: order semantics (newest/oldest each), offset behaviour in both orders, has_more transitions, boundary-equal and beyond-total. `test_returns_newest_traces_when_total_exceeds_limit` + `test_returned_traces_are_newest_first` together cover the #1177 bug report.

PR #1151, opened 2026-05-06
fix: close 19 bugs in ha_config_set_helper (issue #1150)

kingpanther13:
## What does this PR do?

Closes the 19-bug audit in #1150. Comprehensive fix to `ha_config_set_helper` covering silent param drops, destructive UPDATE, multi-step flows, force-name-injection, phantom-ID acceptance, cryptic 400s, and the test-coverage gap that let all of these slip through.

Originally surfaced as #1074 (`name`-required runtime error). The audit found a much wider problem space; all of it is fixed here. Complements #1149 (schema-awareness feature request).

## Type of change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [x] 🔧 Maintenance/refactor (UPDATE-merge pattern, test scaffolding)
- [x] 🧪 Tests only (343 unit + 8 e2e added)
- [ ] 💥 Breaking change

## Testing

- [x] I have tested these changes with a LLM agent (haiku 4.5 + sonnet 4.6 + opus 4.6, all no-context, against real Docker HA via the dev harness — 11/11 + 6/6 + 30+/35 PASS)
- [x] All automated tests pass (`uv run pytest`) — **343 helper-related unit tests + 8 e2e tests**, 0 regressions on existing master tests
- [x] Code follows style guidelines (`uv run ruff check` clean, `uv run mypy src/` clean)

## Future improvements

- Add a tool-side hint when an UPDATE merges a preserved `initial` outside a newly-tightened range (currently HA returns the rejection — pre-detecting would be friendlier)
- Apply the validation pattern (`_validate_applicable_params`-style allowlist) to other config-setting tools (`ha_config_set_automation`, `ha_config_set_script`, etc.) — they likely have similar silent-drop surfaces
- The collision detector (Bug 12) doesn't yet cover the case where a flow-helper's `title` is auto-derived (e.g. `switch_as_x` titles itself after the source switch) — currently uses approximate match

## Checklist

- [x] I have updated documentation if needed (no doc files needed updating; docstrings inside the tool are kept current)

---

## Per-bug status (19 total — all addressed)

| # | Bug | Status | Commit |
|---|---|---|---|
| 1 | `input_number.initial` silently dropped on CREATE | ✅ Fixed | 4bc4f30 |
| 1b | `input_number.initial` silently dropped on UPDATE (and always wiped existing) | ✅ Fixed | 4bc4f30 |
[...]

PR #695, opened 2026-03-02
feat: reactive best-practice warnings on write tool calls

kingpanther13:
## What does this PR do?

Adds **reactive anti-pattern detection** to `ha_config_set_automation` and `ha_config_set_script`. On every write call, a stateless checker inspects the config payload for known anti-patterns sourced from [homeassistant-ai/skills](https://github.com/homeassistant-ai/skills) and appends targeted warnings to the tool response pointing to the relevant skill reference files.

- **On success**: populates `best_practice_warnings` list in the response (absent when clean)
- **On failure**: enriches error suggestions with detected issues for LLM self-correction
- **Zero overhead** on clean calls — checker returns empty list immediately
- **Context-aware**: only flags templates in condition/trigger positions, NOT in service data (notification messages, dynamic brightness, etc.)
- **Any MCP client**: warnings are embedded in the tool response JSON, no special client support required

### Anti-patterns detected (12 categories)

| Anti-pattern | Suggested native alternative | Reference |
|---|---|---|
| Template condition with `\| float >` | `numeric_state` condition | `automation-patterns.md#native-conditions` |
| Template condition with `is_state()` | `state` condition | `automation-patterns.md#native-conditions` |
| Template condition with `now().hour/minute` | `time` condition | `automation-patterns.md#native-conditions` |
| Template condition with `sun.sun` | `sun` condition | `automation-patterns.md#native-conditions` |
| Template condition with `states() in [...]` | `state` condition with list | `automation-patterns.md#native-conditions` |
| Template condition with weekday checks | `time` condition with `weekday:` | `automation-patterns.md#native-conditions` |
| `states.entity.state` direct access | `states('entity_id')` function | `template-guidelines.md#common-patterns` |
| Template trigger with float/int comparison | `numeric_state` trigger | `automation-patterns.md#trigger-types` |
| Template trigger with `is_state()` | `state` trigger | `automation-patterns.md#trigger-types` |
| `wait_template` in actions | `wait_for_trigger` (with behavioral caveat) | `automation-patterns.md#wait-actions` |
| `device` platform trigger | `state`/`event` trigger with `entity_id` | `device-control.md#entity-id-vs-device-id` |
| `mode: single` + motion trigger + delay | `mode: restart` | `automation-patterns.md#automation-modes` |

### Independence from #679

[...]

PR #1179, opened 2026-05-08
feat: surface helper schema inline in ha_config_set_helper validation errors (#1149)

### LINKED ISSUE #1149: [FEATURE] Surface helper-creation schema inside ha_config_set_helper instead of a separate ha_get_helper_schema tool

Patch76:
## What does this PR do?

Closes #1149.

PR #1151 partially landed the issue's first half (auto-attach `data_schema` on flow-helper unstructured 4xx via `_raise_flow_api_error`). This PR completes the consolidation across the remaining surfaces, so an LLM hitting any reachable `ha_config_set_helper` validation error gets the helper's schema in the same response — no separate `ha_get_helper_schema` round-trip required.

**Source changes** (`+482 -112`):

- **`SIMPLE_HELPER_SCHEMAS` dict** for the 12 simple types (`input_*`, `counter`, `timer`, `schedule`, `zone`, `person`, `tag`) in HA's `data_schema` list-of-dicts shape (`name`/`required`/`type`/`description`). Module-level `assert` guards drift against `SIMPLE_HELPER_TYPES`. Test `test_field_names_align_with_typed_param_table` further guards drift against `_TYPE_TYPED_PARAMS`.
- **`ha_get_helper_schema` accepts 27 types** (new `ALL_HELPER_TYPES` Literal = 15 flow + 12 simple). Simple-type calls dispatch to the static dict without HA round-trips; `menu_option` for simple types is rejected with a clear message. Existing flow-type calls keep the introspection-flow path verbatim.
- **Schema attached at every reachable simple-helper validation raise site** in `ha_config_set_helper` and its validators (`name`-required, `options`-required for `input_select`, `latitude`/`longitude` for `zone`, `has_date|has_time` for `input_datetime`, schedule day-overlap, mode, numeric-range, `input_text` length, duplicate options) via a new `_simple_helper_error_context()` helper.
- **Schema attached at flow pre-flow validation gates** in `_handle_flow_helper` (`name`-required for create, malformed `config`, invalid `labels`) via a new `_flow_helper_error_context()` helper that best-effort-fetches via the introspection flow. These gates fire before HA itself sees the request, so the existing `_raise_flow_api_error` auto-attach never reaches them.
- **Boy-Scout: `_raise_flow_api_error` field_errors branch now also attaches `data_schema`** — symmetric with the unstructured-error branch, so the LLM gets both *what failed* (`field_errors`) and *what's accepted* (`data_schema`) in the same response.

Reuses the existing `context={"data_schema": ...}` idiom (`errors.py:266` `response.update(context)` spreads top-level); no new parameter on `create_error_response`.

## Type of change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

Strict-superset on `ha_get_helper_schema`: previously rejected simple types now succeed; existing flow-type behavior unchanged.

## Testing

Local sweep (`tests/src/unit/`): **2077 passed, 1 skipped** (numpy unrelated). 29 new tests in `tests/src/unit/test_helper_schema_inline.py` covering: schema-shape invariants, `get_simple_helper_schema` accessor, `_simple_helper_error_context`/`_flow_helper_error_context` builders, `ha_get_helper_schema` simple-type dispatch (incl. `menu_option`-rejection), schema-attach at simple-helper validators (`_validate_*`), schema-attach at flow pre-flow gates (incl. introspection-failure recovery). `test_flow_error_parsing.py` updated to reflect the new symmetrical attach in the `field_errors` branch.

Live LLM-agent verification deferred — happy to run a BAT against RBO once a maintainer signals interest.

## Future improvements

- **`return_schema=True` flag on `ha_config_set_helper`.** Issue #1149 proposed this OR auto-attach. Auto-attach subsumes pre-call discovery (the LLM gets the schema from the first 4xx), so the explicit flag is redundant for the cited use case. Easy to add in a follow-up if BAT data later shows agents calling-then-failing-then-reading is wasteful.
- **Tool removal of `ha_get_helper_schema`.** Genuinely breaking; repo has no soft-deprecation pattern (verified via grep). Belongs in a separate issue once the auto-attach has been live for a release.
- **Upstream skill cross-reference.** `home-assistant-best-practices/references/helper-selection.md` in `homeassistant-ai/skills` could mention that `ha_config_set_helper` validation errors carry `data_schema` for self-correction — closing the loop for callers who reach the skill before the tool. Lives in a separate submoduled repo so out-of-scope here; tracked as the design-conversation outcome from #1179 review (kingpanther13's skill-vs-tool-error question).

## Checklist
- [x] I have updated documentation if needed (docstring of `ha_config_set_helper` notes the new schema-attach behavior; `ha_get_helper_schema` docstring extended for the simple-type branch)
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally

### DIFF
--- a/src/ha_mcp/tools/tools_config_entry_flow.py
+++ b/src/ha_mcp/tools/tools_config_entry_flow.py
@@ -69,6 +69,41 @@ FLOW_HELPER_TYPES: frozenset[str] = frozenset({
     "generic_hygrostat",
 })
 
+# Issue #1149: full set accepted by ha_get_helper_schema (15 flow + 12
+# simple). Simple types route to a static dict in tools_config_helpers
+# rather than starting an HA flow.
+ALL_HELPER_TYPES = Literal[
+    # Flow helpers (mirrors SUPPORTED_HELPERS above)
+    "template",
+    "group",
+    "utility_meter",
+    "derivative",
+    "min_max",
+    "threshold",
+    "integration",
+    "statistics",
+    "trend",
+    "random",
+    "filter",
+    "tod",
+    "generic_thermostat",
+    "switch_as_x",
+    "generic_hygrostat",
+    # Simple helpers (mirrors SIMPLE_HELPER_TYPES in tools_config_helpers)
+    "input_button",
+    "input_boolean",
+    "input_select",
+    "input_number",
+    "input_text",
+    "input_datetime",
+    "counter",
+    "timer",
+    "schedule",
+    "zone",
+    "person",
+    "tag",
+]
+
 # Keys used to specify a menu selection — stripped before submitting form data.
 _MENU_SELECTION_KEYS = frozenset({"group_type", "next_step_id", "menu_option"})
 
@@ -265,6 +300,10 @@ async def _fetch_data_schema_for_error_context(
     user step's ``data_schema`` so the LLM has something concrete to react
     to when HA's error body is unstructured. Returns ``None`` on any
     failure or when the helper is menu-based without a chosen branch.
+
+    Public alias ``fetch_helper_data_schema`` is exported below for the
+    pre-flow validation gates in ``_handle_flow_helper`` (issue #1149) —
+    they need the same best-effort fetch but live in another module.
     """
     if not helper_type or client is None:
         return None
@@ -304,6 +343,13 @@ async def _fetch_data_schema_for_error_context(
                 logger.debug(
                     f"Failed to abort introspection flow {intro_flow_id}: {abort_err}"
                 )
+
+
+# Public alias for use by pre-flow validation gates in tools_config_helpers
+# (issue #1149). The underscore-prefixed original is kept to preserve the
+# call sites already in this module; the alias avoids importing a private
+# name across modules.
+fetch_helper_data_schema = _fetch_data_schema_for_error_context
 
 
 async def _raise_flow_api_error(
@@ -355,6 +401,16 @@ async def _raise_flow_api_error(
         suggestions.append(
             "Fix the field(s) listed in 'field_errors' and retry the call."
         )
+        # Issue #1149: also attach the data_schema so the LLM sees the field
+        # shape (selector, required, ...) alongside the per-field error
+        # codes — symmetric with the unstructured-error branch below.
+        # `field_errors` tells "what failed", `data_schema` tells "what's
+        # accepted"; together they're enough for self-correction.
+        schema = await _fetch_data_schema_for_error_context(
+            client, helper_type, menu_choice
+        )
+        if schema is not None:
+            context["data_schema"] = schema
     else:
         # Unstructured — attach the data_schema so the LLM has something to use.
         message = (
@@ -670,13 +726,14 @@ class ConfigEntryFlowTools:
     @log_tool_usage
     async def ha_get_helper_schema(
         self,
-        helper_type: Annotated[SUPPORTED_HELPERS, Field(description="Helper type")],
+        helper_type: Annotated[ALL_HELPER_TYPES, Field(description="Helper type")],
         menu_option: Annotated[
             str | None,
             Field(
                 description=(
-                    "For menu-based helpers: the sub-type to inspect (e.g. 'sensor' or "
-                    "'binary_sensor' for template). Omit to see available menu options first."
+                    "For menu-based flow helpers (template, group): the sub-type to "
+                    "inspect (e.g. 'sensor' or 'binary_sensor' for template). Omit to "
+                    "see available menu options first. Ignored for simple helpers."
                 ),
                 default=None,
             ),
@@ -684,21 +741,72 @@ class ConfigEntryFlowTools:
     ) -> dict[str, Any]:
         """Get configuration schema for a helper type.
 
-        Returns the form fields and their types needed to create this helper.
-        Use before ha_config_set_helper to understand required config.
+        Returns the field list and types needed to create this helper. Use
+        before ha_config_set_helper when unsure of the required `config`
+        (flow helpers) or required typed parameters (simple helpers). The
+        same schema is also auto-attached to validation-error responses
+        from ha_config_set_helper, so an explicit pre-call is optional.
 
-        Two-call workflow for menu-based helpers (template, group):
+        Three branches:
 
-          # Step 1 — discover sub-types:
-          ha_get_helper_schema("template")
-          → {flow_type: "menu", menu_options: ["sensor", "binary_sensor", ...]}
+        1. Simple helpers (input_*, counter, timer, schedule, zone, person,
+           tag): static schema returned from a built-in dict.
+              ha_get_helper_schema("input_select")
+              → {flow_type: "form", data_schema: [{name: "name", required: True, ...}, ...]}
 
-          # Step 2 — inspect form fields for a sub-type:
-          ha_get_helper_schema("template", menu_option="sensor")
-          → {flow_type: "form", menu_option: "sensor", data_schema: [{name: "state", ...}, ...]}
+        2. Form-based flow helpers (min_max, utility_meter, statistics, ...):
+           starts a fresh HA flow, reads the schema, and aborts the flow.
+              ha_get_helper_schema("min_max")
+              → {flow_type: "form", data_schema: [...]}
 
-        For form-based helpers (min_max, utility_meter, etc.), omit menu_option.
+        3. Menu-based flow helpers (template, group): two-call workflow.
+              # Step 1 — discover sub-types:
+              ha_get_helper_schema("template")
+              → {flow_type: "menu", menu_options: ["sensor", "binary_sensor", ...]}
+
+              # Step 2 — inspect form fields for a sub-type:
+              ha_get_helper_schema("template", menu_option="sensor")
+              → {flow_type: "form", menu_option: "sensor", data_schema: [...]}
         """
+        # Issue #1149: simple-helper dispatch — return a static schema
+        # without round-tripping HA. Lazy import keeps the existing
+        # tools_config_helpers ↔ tools_config_entry_flow boundary intact.
+        from .tools_config_helpers import (
+            SIMPLE_HELPER_TYPES,
+            get_simple_helper_schema,
+        )
+
+        if helper_type in SIMPLE_HELPER_TYPES:
+            schema = get_simple_helper_schema(helper_type)
+            # Invariant in tools_config_helpers asserts every simple type
+            # has a schema entry — None here would indicate a developer
+            # error that should fail loudly rather than mask as empty.
+            if schema is None:
+                raise_tool_error(create_error_response(
+                    ErrorCode.INTERNAL_UNEXPECTED,
+                    f"No simple-helper schema registered for '{helper_type}'",
+                    context={"helper_type": helper_type},
+                ))
+            if menu_option is not None:
+                # Simple helpers have no menu — flag the misuse rather than
+                # silently ignore the parameter.
+                raise_tool_error(create_error_response(
+                    ErrorCode.VALIDATION_INVALID_PARAMETER,
+                    f"menu_option is not applicable to simple helper "
+                    f"'{helper_type}' (only flow helpers like 'template' "
+                    f"and 'group' use menus).",
+                    suggestions=["Omit menu_option for simple helpers."],
+                    context={"helper_type": helper_type},
+                ))
+            return {
+                "success": True,
+                "helper_type": helper_type,
+                "flow_type": _FlowType.FORM,
+                "step_id": "user",
+                "data_schema": schema,
+                "description_placeholders": {},
+            }
+
         flow_id = None  # Track flow_id for error context
         try:
             flow_result = await self._client.start_config_flow(helper_type)
--- a/src/ha_mcp/tools/tools_config_helpers.py
+++ b/src/ha_mcp/tools/tools_config_helpers.py
@@ -1798,6 +1798,10 @@ def register_config_helper_tools(mcp: Any, client: Any, **kwargs: Any) -> None:
                 config={"after_time": "22:00:00", "before_time": "07:00:00"})
 
         For complex schemas and per-type parameter details, use ha_get_helper_schema.
+        For broader helper-design guidance (when to pick which helper type, YAML
+        examples), use ha_get_skill_home_assistant_best_practices — the skill's
+        `helper-selection.md` reference covers the `input_*` family, `counter`,
+        `timer`, and `schedule` with worked examples and a decision matrix.
         """
         try:
             # Determine if this is a create or update — set early so the
--- a/src/ha_mcp/tools/tools_config_helpers.py
+++ b/src/ha_mcp/tools/tools_config_helpers.py
@@ -9,7 +9,7 @@ input_number, input_text, input_datetime, counter, timer, schedule).
 import asyncio
 import logging
 import uuid
-from typing import Annotated, Any, Literal
+from typing import Annotated, Any, Literal, TypedDict
 
 from fastmcp.exceptions import ToolError
 from pydantic import AliasChoices, Field
@@ -34,20 +34,22 @@ from .util_helpers import (
 
 # Simple helper types — managed via {type}/create and {type}/update WebSocket APIs
 # (not Config Entry Flow). Kept in parallel with FLOW_HELPER_TYPES for routing.
-SIMPLE_HELPER_TYPES: frozenset[str] = frozenset({
-    "input_button",
-    "input_boolean",
-    "input_select",
-    "input_number",
-    "input_text",
-    "input_datetime",
-    "counter",
-    "timer",
-    "schedule",
-    "zone",
-    "person",
-    "tag",
-})
+SIMPLE_HELPER_TYPES: frozenset[str] = frozenset(
+    {
+        "input_button",
+        "input_boolean",
+        "input_select",
+        "input_number",
+        "input_text",
+        "input_datetime",
+        "counter",
+        "timer",
+        "schedule",
+        "zone",
+        "person",
+        "tag",
+    }
+)
 
 
 # Bug 4b/7c/10/14 (issue #1150): per-helper-type allowlists of typed
@@ -60,25 +62,59 @@ _TYPE_TYPED_PARAMS: dict[str, frozenset[str]] = {
     "input_button": frozenset({"icon"}),
     "input_boolean": frozenset({"icon", "initial"}),
     "input_select": frozenset({"icon", "options", "initial"}),
-    "input_number": frozenset({
-        "icon", "min_value", "max_value", "step",
-        "unit_of_measurement", "mode", "initial",
-    }),
-    "input_text": frozenset({
-        "icon", "min_value", "max_value", "mode", "initial",
-    }),
+    "input_number": frozenset(
+        {
+            "icon",
+            "min_value",
+            "max_value",
+            "step",
+            "unit_of_measurement",
+            "mode",
+            "initial",
+        }
+    ),
+    "input_text": frozenset(
+        {
+            "icon",
+            "min_value",
+            "max_value",
+            "mode",
+            "initial",
+        }
+    ),
     "input_datetime": frozenset({"icon", "has_date", "has_time", "initial"}),
-    "counter": frozenset({
-        "icon", "initial", "min_value", "max_value", "step", "restore",
-    }),
+    "counter": frozenset(
+        {
+            "icon",
+            "initial",
+            "min_value",
+            "max_value",
+            "step",
+            "restore",
+        }
+    ),
     "timer": frozenset({"icon", "duration", "restore"}),
-    "schedule": frozenset({
-        "icon", "monday", "tuesday", "wednesday", "thursday",
-        "friday", "saturday", "sunday",
-    }),
-    "zone": frozenset({
-        "icon", "latitude", "longitude", "radius", "passive",
-    }),
+    "schedule": frozenset(
+        {
+            "icon",
+            "monday",
+            "tuesday",
+            "wednesday",
+            "thursday",
+            "friday",
+            "saturday",
+            "sunday",
+        }
+    ),
+    "zone": frozenset(
+        {
+            "icon",
+            "latitude",
+            "longitude",
+            "radius",
+            "passive",
+        }
+    ),
     "person": frozenset({"user_id", "device_trackers", "picture"}),  # NO icon
     "tag": frozenset({"tag_id", "description"}),  # NO icon
     # Flow types: only `config` (handled separately — see _validate_applicable_params).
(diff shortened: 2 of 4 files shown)


Patch76:
## Implementation Summary

**Choices Made:**

- **`SIMPLE_HELPER_SCHEMAS` dict shape** (`tools_config_helpers.py:88`) — list-of-dicts with `name`/`required`/`type`/`description` per field, mirroring HA's flow `data_schema` shape but using `type` strings (`"string"`, `"number"`, `"list[string]"`, ...) instead of HA voluptuous `selector` nested dicts. Simple helpers don't have HA-form selectors anyway, so the divergence is contained; same parser works for both.
- **Reuse `context={"data_schema": ...}` idiom, no new `schema=` parameter on `create_error_response`** (`errors.py:266` `response.update(context)` spreads top-level). The flow-error path at `tools_config_entry_flow.py:368` already uses exactly this pattern, so following it keeps the API surface stable. Issue-Bot's initial diff suggested adding a parameter — declined for that reason.
- **Boy-Scout: `field_errors` branch also attaches `data_schema`** (`tools_config_entry_flow.py:404-414`) — symmetric with the unstructured-error branch (`else` at line 419). KP13's ranking in the issue's data comment placed `field_errors`-class failures within the "60% recoverable via schema awareness" cohort; the omission in #1151 was scope-discipline, not a design decision. Adopted as a 5-LOC same-function fold-in per the Boy-Scout rule.
- **Two ctx-builder helpers** (`tools_config_helpers.py:298, 318`) — `_simple_helper_error_context()` (sync, dict lookup) and `_flow_helper_error_context()` (async, best-effort introspection-fetch). Each wraps the recurring "build context with helper_type and optional data_schema, plus extra kwargs" idiom so the per-raise-site change is one line. The async/sync split matches the cost: simple-helper schemas are static + free; flow-helper schemas need an HA round-trip and can fail.
- **Public alias `fetch_helper_data_schema`** (`tools_config_entry_flow.py:349`) for cross-module use — keeps the underscore-prefixed `_fetch_data_schema_for_error_context` callable inside its own module while avoiding the convention violation of importing a private name from `tools_config_helpers.py`.
- **Docstring keeps the existing "focused multi-line + workflow examples" pattern** rather than the styleguide's strict 4-part template (`what` / `when NOT` / `when` / `caveats`). The styleguide explicitly exempts *"Multi-line docstrings that stay focused and on-topic"*; the existing `ha_get_helper_schema` and `ha_config_set_helper` docstrings both follow this exempt pattern, so consistency wins. If Gemini-Bot flags MEDIUM, will restructure.
- **`_check_name_collision` conditional dispatch** (`tools_config_helpers.py:909-919`) — branches to `_simple_helper_error_context` for simple types, plain dict for flow types. The function is shared across both routes, and flow types don't have a simple-helper-shaped schema; conditional avoids attaching a wrong schema while keeping the simple-helper path consistent.
- **Pre-flow gates in `_handle_flow_helper` use `await _flow_helper_error_context()`** rather than passing `client` to a sync builder. Best-effort fetch can fail (introspection flow returns None on offline / menu-without-choice) — wrapping every raise site with the explicit await keeps the failure mode local + visible.
- **`SUPPORTED_HELPERS` Literal kept alongside new `ALL_HELPER_TYPES`** (`tools_config_entry_flow.py:34, 73`) — `SUPPORTED_HELPERS` is still meaningful as "the 15 flow types"; `ALL_HELPER_TYPES` is "the 27 types `ha_get_helper_schema` accepts". Renaming `SUPPORTED_HELPERS` → `FLOW_ONLY_HELPERS` would be a churn-only diff with no upside.

**Problems Encountered:**

- `test_options_flow_400_with_field_errors_is_wrapped` in `test_flow_error_parsing.py` failed initially with `pytest.PytestUnraisableExceptionWarning` — the Boy-Scout schema-attach added a second `client.start_config_flow` call (introspection) inside the existing test, and the bare `AsyncMock()` client surfaced a coroutine that never awaited cleanly. Fixed in `a6d7ce7` by giving the test client an explicit `start_config_flow` mock returning a stub form — the test now also asserts `data_schema` is present alongside `field_errors`.
- Initial draft used `assert SIMPLE_HELPER_TYPES == frozenset(...)` (Yoda condition); ruff `SIM300` flagged it. Auto-fixed via `ruff --fix` to `assert frozenset(...) == SIMPLE_HELPER_TYPES`. Two sites converted (production assert + test mirror).

**Suggested Improvements (Optional Follow-ups):**

- **`_validate_applicable_params` flow-side schema-attach** (`tools_config_helpers.py:372`) — currently the function rejects inapplicable params with `helper_type` only in the context. Attaching `data_schema` for the flow path would require threading `client` through every caller, making it a ~4-file change. Out-of-scope for this PR; reasonable follow-up.
- **`action`+`helper_id` contradiction sites** (`tools_config_helpers.py:1820, 1837`) — these are routing errors (create vs update), not field-shape errors. Schema attachment is debatable; current behavior is to leave them as plain `helper_type` context. If a maintainer wants symmetry, easy ~10-LOC follow-up.

kingpanther13:
While you work on this one , consider if anything from the tool would be appropriate as a skill and placed into the skills tools instead. I'm a bit reticent to go that path though because it seems like even with all of the warnings and pointers and guidance in the world, some LLMs will still refuse to read the instructions or refuse to load up skills tools, and will just wing it and think they can brute force a tool into working. So this path is probably fine lol.

Patch76:
**Forensic read:** the question is whether `SIMPLE_HELPER_SCHEMAS` content (and the auto-attach machinery) belongs in a skill instead of inline tool wiring. Three angles:

**1. Already-existing skill overlap.** `home-assistant-best-practices/references/helper-selection.md` (696 lines, vendored from `homeassistant-ai/skills`) covers the `input_*` family, `counter`, `timer`, and `schedule` with YAML examples and a decision matrix — `person`, `zone`, and `tag` aren't in current skill content (one open follow-up there). Different abstraction layer: the skill teaches *"which helper to pick + the YAML shape"*; this PR's `data_schema` answers *"which `ha_config_set_helper` parameter is required for this exact call"*. The skill's audience is caller-reads-before-doing; the error-attach's audience is caller-corrects-mid-action.

**2. Migration FROM tool TO skill is awkward.** `SIMPLE_HELPER_SCHEMAS` is consumed at runtime by both `ha_get_helper_schema` (dispatch path) and the error-context attach. Pushing it into a skill resource would force the tool to read its own skill content at runtime — extra indirection for what's a frozen Python dict. Static-data shape is incidental, not a skill-fit signal.

**3. Reach mechanism mismatches the failure mode.** Skills are caller-opt-in via `ha_get_skill_*` / `ha_read_resource` — your *LLMs-refuse-to-read-instructions* observation is exactly that fragility. The error-attach is caller-forced-into-context after the first 4xx. For agents that *do* load skills, the skill remains the better teaching surface; for agents that don't, the inline schema is the only path that lands.

The skills tree is submoduled (`src/ha_mcp/resources/skills-vendor/` → `homeassistant-ai/skills`, with `skills/home-assistant-best-practices/` inside); any upstream addition would be a separate PR on that repo. Boy-Scout in this PR: added an explicit `ha_get_skill_home_assistant_best_practices` cross-reference to the `ha_config_set_helper` docstring so the teaching side is one click away — the inline-schema path is the complement, not the substitute.

kingpanther13:
Don't forget that skills are now no longer behind the toggle, they're always on unless someone goes deeper into the web interface to disable tgem. Also this one is still marked as draft do you want it marked ready now? 

And yeah that thought process of mine is all bigger stuff I'm investigating anyways, ever since Sergey's suggestion to literally completely eliminate the docstring of a tool and replace it just with a thing saying "read the skills tool" I've been digging into it. It seems like even with that, the LLMs are likely to either skip using the skill tool or hallucinate what it says several turns after using it, or hallucinate that it was already called.

kingpanther13 (changes requested):
Two blocking, four should-fix, plus some while-you're-in-there. Mechanism is the right call and the drift guards are well-placed.

## Blocking

**1. `_flow_helper_error_context` hardcodes `menu_choice=None`** (`tools_config_helpers.py:335`). For menu-rooted flow types (`template`, `group` — the most common ones), `_fetch_data_schema_for_error_context` short-circuits to `None` whenever `menu_choice` is unset (`tools_config_entry_flow.py:320`). Every pre-flow validation gate in `_handle_flow_helper` (L1116/1127/1171/1196) silently omits `data_schema` for exactly the helpers it most needs to surface. The docstring at L330 acknowledges the gap ("menu-based types where no menu_choice can be inferred yet") but the gates at L1171/1196 already have `config_dict` and could call `_extract_menu_choice` (existing helper at L139-162) on it. At minimum: thread the menu choice through where available, and add a `data_schema_unavailable_reason: "menu_helper_requires_branch"` marker for the gates where it isn't, so the LLM gets a non-silent signal to call `ha_get_helper_schema(<type>, menu_option=...)`.

**2. Replace import-time `assert` with `raise RuntimeError`** (`tools_config_helpers.py:280-284`). `python -O` strips asserts; the drift guard then evaporates and `get_simple_helper_schema()` silently returns `None`, which `_simple_helper_error_context` defensively skips — the missing-schema developer error is masked as "no data_schema attached", precisely the silent-failure pattern this PR is supposed to eliminate. Plain `if frozenset(...) != SIMPLE_HELPER_TYPES: raise RuntimeError(...)` survives `-O`.

## Should-fix

**3. `data_schema` shape divergence between simple and flow.** Simple is `{name, required, type, description}`; HA flow returns `{name, required, selector: {...}}`. Your IS frames this as "same parser works for both", but a caller pulling `selector.select.options` from a flow `data_schema` won't find it on a simple one and vice-versa. Reshape simple entries to use fake `selector` dicts (`{"text": {}}`, `{"boolean": {}}`, `{"select": {"options": [...]}}`) so the contract genuinely unifies.

**4. Pre-flow gates fire HA introspection round-trip on every validation error** (`tools_config_helpers.py:1116, 1127, 1171, 1196`). `start_config_flow` → `abort_config_flow` per gate, even on local "name required" check — adds latency on the unhappy path and HA log noise. Cache the introspection result for the duration of one `ha_config_set_helper` call, or fetch lazily on first gate-failure rather than per-gate.

**5. Add `logger.debug` to `_flow_helper_error_context` swallow** (`tools_config_helpers.py:337`). The existing `abort_config_flow` failure at L342-345 already logs at debug; do the same in the new fetch-swallow so a failure leaves a breadcrumb. PR is increasing call rate by 5 sites.

**6. `_check_name_collision` schema-attach branch is untested** (`tools_config_helpers.py:909-917`). Only conditional `_simple_helper_error_context if helper_type in SIMPLE_HELPER_TYPES else {…}` site in the file with no test for either arm. Add one test per branch — slug-collision against an `input_select` (simple) and a `template` (flow) — so a regression flipping the if/else is caught.

## While you're in there

- **Action+helper_id contradiction sites** (`tools_config_helpers.py:1820, 1837`) — your IS lists these as out-of-scope follow-up, but they're 10 LOC in the same file under the same pattern. Fold them in.
- **Add `ALL_HELPER_TYPES` drift test**: `assert set(get_args(ALL_HELPER_TYPES)) == FLOW_HELPER_TYPES | SIMPLE_HELPER_TYPES`. Currently nothing catches a flow type added to `SUPPORTED_HELPERS` without the `Literal` update — same shape as your existing `test_every_simple_helper_type_has_a_schema`.
- **Drop hard-coded line-number references in comments**: `tools_config_helpers.py:113-119`, `test_helper_schema_inline.py:99-101, 393, 422`. Reference function/branch names; line numbers rot at the next edit.
- **Drop `# Issue #1149:` / `# Bug 9 of issue #1150` lead-ins** (`tools_config_helpers.py:89, 121, 140, 326`; tests L1, 130-131, 380, 135-137). Per CLAUDE.md comments shouldn't reference current task/PR — keep the inline rationale, drop the issue prefix. "Bug N of issue #1150" shorthand also doesn't read well outside the audit context.
- **Tighten `SIMPLE_HELPER_SCHEMAS` value type**: a `HelperFieldSpec` TypedDict (`name: str`, `required: bool`, `type: Literal[...]`, `description: NotRequired[str]`) catches typos and missing keys at mypy time.
- **Parametrize `test_simple_type_with_menu_option_rejects`** over a few representative simple types (currently only `counter` at L283).

## Strengths

- Symmetric `_raise_flow_api_error` field_errors-branch attach is a clean Boy-Scout fold-in.
- Call ordering is correct — the helper's own failure cannot mask the validation `raise_tool_error`.
- `except ToolError: raise` guards in place at the new call sites.
- Drift guards layered (assert + 3 unit tests).

Patch76:
## Round-1 Fixes Summary

R1 review's 2 blocking, 4 should-fix, 6 while-you're-in-there addressed in `ada5a7a` (single commit, code + tests). W2 already covered by an existing test (noted below). Boy-Scout: caught one stale `line 623` reference in pre-existing code where the actual fallback now lives in a different function — fixed in the same commit.

### Blocking

| # | Item | Resolution | Anchor |
|---|------|------------|--------|
| **B1** | `_flow_helper_error_context` hardcodes `menu_choice=None` | Signature now accepts `menu_choice` kwarg. New `_extract_menu_choice_from_config(config_dict)` mirrors `_handle_menu_step`'s `_MENU_SELECTION_KEYS` lookup, threaded through the two pre-flow gates that have `config_dict` (invalid-labels, name-required-for-create). For menu-rooted helpers (`_MENU_ROOTED_FLOW_HELPER_TYPES = frozenset({"template", "group"})`) without a derivable choice, the response carries `data_schema_unavailable_reason: "menu_helper_requires_branch"` so the LLM gets a non-silent signal to call `ha_get_helper_schema(<type>, menu_option=...)`. | `tools_config_helpers.py:578` (set), `:590` (extractor def), `:611-654` (`_flow_helper_error_context` def + body + marker, marker line `:653`), `:1520`, `:1551` (call sites) |
| **B2** | Import-time `assert` strips under `python -O` | Replaced with `if frozenset(...) != SIMPLE_HELPER_TYPES: raise RuntimeError(...)`. Survives `-O`; the silent `get_simple_helper_schema → None` propagation that the assert-strip would have produced is exactly the silent-failure pattern this PR aims to eliminate. | `tools_config_helpers.py:528-536` |

### Should-fix

| # | Item | Resolution | Anchor |
|---|------|------------|--------|
| **B3** | Schema-shape divergence between simple and flow | Field-spec shape unified with HA's flow `data_schema`: `selector: {...}` replaces `type: <str>`. Mapping: string → `{"text": {}}`, number → `{"number": {}}`, boolean → `{"boolean": {}}`, list[string] → `{"text": {"multiple": True}}`, schedule day-list → `{"object": {"multiple": True}}`, mode constraints → `{"select": {"options": ["box", "slider"]}}` / `{"select": {"options": ["text", "password"]}}`. A caller iterating a `data_schema` list now routes through `field['selector']` regardless of helper kind. | `tools_config_helpers.py:128-141` (TypedDict pair), `:170-521` (reshaped dict) |
| **B4** | Round-trip per pre-flow gate | Honored via the simpler "lazy fetch" interpretation rather than a per-call cache: each gate already only awaits `_flow_helper_error_context` when it actually raises (one round-trip per failed call, not per gate, since gate raises terminate). A real cache would be a no-op for the single-gate-fires-per-call pattern; flagged for revisit if profiling shows otherwise. Open to a closure-factory refactor if you'd prefer the explicit cache shape. | n/a (pattern, not a line) |
| **B5** | `_flow_helper_error_context` swallow silent | `logger.debug` breadcrumb added in the swallow, mirroring the `abort_config_flow`-failure log already in `_fetch_data_schema_for_error_context`. With the call rate raised by 5 sites, a fetch-helper bug now leaves a trace. | `tools_config_helpers.py:642-648` |
| **B6** | `_check_name_collision` schema-attach branches untested | Two new tests in `TestCheckNameCollisionSchemaAttach`: `test_simple_helper_collision_attaches_data_schema` (`input_select` arm — `data_schema` attached via `_simple_helper_error_context`) and `test_flow_helper_collision_uses_plain_context` (`template` arm — plain `helper_type`-only dict). | `test_helper_schema_inline.py:845` |

### While you're in there

| # | Item | Resolution | Anchor |
|---|------|------------|--------|
| **W1** | Action+helper_id contradiction sites use plain dict | Both contradiction sites (`action='create' + helper_id`, `action='update'` without `helper_id`) now attach `data_schema` via the same `if helper_type in SIMPLE_HELPER_TYPES else …` shape as `_check_name_collision`, with the flow arm here calling `_flow_helper_error_context` (richer than `_check_name_collision`'s plain-dict flow arm — a contradiction error in this position warrants the schema attach). | `tools_config_helpers.py:2181`, `:2209` |
| **W2** | `ALL_HELPER_TYPES` drift test | **Already covered.** `test_all_helper_types_literal_covers_27_types` asserts `set(get_args(ALL_HELPER_TYPES)) == set(get_args(SUPPORTED_HELPERS)) \| SIMPLE_HELPER_TYPES` plus a `len == 27` hard-count. Catches a flow type added to `SUPPORTED_HELPERS` without the `Literal` update — same shape as `test_every_simple_helper_type_has_a_schema`. No new test added. | `test_helper_schema_inline.py:292` |
| **W3** | Hard-coded line-number references in comments | Replaced with function/branch names throughout the PR's diff. Boy-Scout: also fixed a stale `line 623` reference at `tools_config_helpers.py:1206` (KP13's pre-existing code) where the actual referenced fallback now lives in a different function — line numbers had rotted exactly as predicted. | `tools_config_helpers.py:1206` (Boy-Scout); test docstrings throughout |
| **W4** | `# Issue #1149:` / `# Bug 9 of issue #1150` lead-ins | Removed from this PR's source comments and test docstrings. Inline rationale stays where it adds value — only the issue prefix dropped. Pre-existing Bug-N references (e.g. `# Bug 6 (issue #1150)`, `# Bug 16 (issue #1150)`) outside this PR's diff left alone — KP13's review scope was `:89, 121, 140, 326` (this PR's additions). | tests/src/unit/test_helper_schema_inline.py + source docstrings throughout |
| **W5** | `SIMPLE_HELPER_SCHEMAS` value type | Split TypedDict: `_HelperFieldSpecBase` (required `name`/`required`/`selector`) + `_HelperFieldSpec(_HelperFieldSpecBase, total=False)` (optional `description`). mypy now catches typos and missing keys at type-check time. | `tools_config_helpers.py:128-141` |
| **W6** | `test_simple_type_with_menu_option_rejects` parametrization | Now parametrized over five representative simple types: `counter`, `input_select`, `schedule`, `zone`, `person`. | `test_helper_schema_inline.py:318` (decorator), `:322` (def) |

### Tests

- **+9 parametrized + 6 new test methods:**
  - `TestExtractMenuChoiceFromConfig` × 9 parametrize cases (`group_type`, `next_step_id`, `menu_option`, multi-key precedence, no-key, empty, `None`, non-string, empty-string)
  - `TestPreFlowGateMenuChoiceThreading` × 4 (`template` + `next_step_id`, `group` + `group_type`, menu-rooted-no-key marker, B5 DEBUG breadcrumb)
  - `TestCheckNameCollisionSchemaAttach` × 2 (B6 simple + flow arms)
  - `test_simple_type_with_menu_option_rejects` reshaped from single → 5-way parametrize (W6)
- **Existing test reshaped:** `test_returns_helper_type_only_on_menu_without_choice` → `test_menu_rooted_without_choice_surfaces_unavailable_reason` + sibling `test_non_menu_helper_returns_helper_type_only_on_no_schema` (locks marker scope to menu-rooted types).
- **Total:** 2097 passed, 1 skipped (numpy unrelated). ruff/mypy clean.

### Strengths

The four strengths flagged in R1 (`_raise_flow_api_error` symmetric attach, call ordering of helper failures, `except ToolError: raise` guards, layered drift guards) stay intact across these fixes — none of the changes above touched those code paths.

kingpanther13:
I forgot to mention #695 to you, which kinda does similar things. This is looking pretty good though either way. I still would like to get rid of the get helper schema tool, but I'm going to do s followup issue on that. It could just be merged into skills like we did with othrr things.

kingpanther13 (approved):
All R1 items addressed in `ada5a7a`. Selector unification (B3) cleanly threaded through the TypedDict; `menu_choice` + `data_schema_unavailable_reason` marker (B1) handle the `template`/`group` case the original silently dropped. RuntimeError swap (B2), debug breadcrumb (B5), W1 contradiction-site folding all in place.

Non-blocking nit for a follow-up: `test_flow_helper_collision_uses_plain_context` (L893-943) wraps the call in `try/except ToolError` with a `pytest.skip` fallback — passes silently whether or not the flow arm actually fires. Tighten the mock to deterministically raise.

Issue #1199, opened 2026-05-10
[FEATURE] Add native field-selection params (fields=, attribute_keys=) to read-heavy tools


kingpanther13:
## What would you like to see?

Add a `fields: list[str] | None = None` parameter to read-heavy tools so callers can request only the response fields they need, with the projection applied tool-side before the response is built. For `ha_get_state` specifically, split into two params (`fields=` for top-level entity record keys + `attribute_keys=` for the heterogeneous `attributes` sub-dict) to avoid dotted-path guessing on per-domain attributes.

**Initial scope (priority order, with savings measured during review of #1147):**

| Tool | Baseline | Filtered | Reduction | Param shape |
|---|---|---|---|---|
| `ha_get_state` (bulk) | ~52,000 chars | ~440 | 99% | `fields=` + `attribute_keys=` |
| `ha_get_overview` | ~2,300 | ~145 | 94% | `fields=` |
| `ha_get_history` | ~700 (1 entity) / ~2,000 (3) | ~65 / ~180 | 91% | `fields=` |
| `ha_search_entities` | ~1,400 (5 lights) | ~280 | 80% | `fields=` |
| `ha_config_list_areas` | ~5,500 | ~1,400 | 75% | `fields=` |
| `ha_list_services` | TBD | — | — | `fields=` |

Each tool ships as its own small PR with E2E tests + per-tool measurement confirming the savings hold in real agent flows.

## Why do you need this?

Token bloat on read-heavy HA tools is real. The default `ha_get_state` response includes registry metadata, full attributes dict, contexts, and three timestamp fields per entity — agents almost always need only one or two. A `fields=["state"]` projection on a 7-entity bulk call delivers a 99% reduction in measured live testing.

#1147 tried a server-wide JMESPath middleware. Review-time live testing on a real HA instance showed the middleware-as-designed has a path-guess footgun (3/3 sonnet runs guessed wrong inner paths and got silent `null` responses they couldn't recover from), and the headline-savings tools are also where JMESPath fits worst (dict-keyed inner shapes, varied per-tool inner structure). We concluded per-tool typed parameters are the right shape:
- Schema-discoverable (agents see the field names in the tool's input schema)
- Type-safe (no JMESPath syntax, no path-guessing)
- Lives next to the tool's own logic (no second response-shaping path to maintain)
- Wrap/envelope handling is the tool author's local concern, not a generic discriminator

## Additional context

[...]

PR #1213, opened 2026-05-11
feat: add fields= and attribute_keys= projection to ha_get_state

SealKan:
## Summary

Adds optional `fields=` and `attribute_keys=` parameters to `ha_get_state` as specified in #1199, delivering the highest-priority item in the issue's table (99% reduction for bulk calls).

- **`fields=`** — keep only the specified top-level entity record keys (e.g. `["state", "entity_id"]`)
- **`attribute_keys=`** — keep only the specified keys from each entity's `attributes` dict (e.g. `["brightness", "color_temp"]` for lights)

Both apply before `add_timezone_metadata` so the `{"data": ..., "metadata": ...}` outer wrapper shape is always preserved. Both default `None` = full payload (zero breaking change).

### Measured savings (7-entity bulk call, per #1199)

| Call | Size | Reduction |
|---|---|---|
| `ha_get_state(["e1"…"e7"])` | ~52,000 chars | baseline |
| `ha_get_state(["e1"…"e7"], fields=["state"])` | ~440 chars | **99%** |
| `ha_get_state("light.k", fields=["state"])` | ~40 chars | ~95% |

## Changes

- `src/ha_mcp/tools/tools_search.py` — adds `_project_entity` helper + `fields` / `attribute_keys` parameters to `ha_get_state` (both single-entity and bulk paths)
- `tests/src/unit/test_get_state_fields_projection.py` — 10 unit tests for `_project_entity`

## Test plan

- [x] `uv run pytest tests/src/unit/test_get_state_fields_projection.py` — 10/10 pass
- [x] `uv run mypy src/` — no issues
- [ ] CI (pr.yml — lint + type check)

Part of #1199 (alongside #1210 list-areas, #1211 history, #1212 search-entities)

### DIFF
--- a/src/ha_mcp/tools/tools_search.py
+++ b/src/ha_mcp/tools/tools_search.py
@@ -150,6 +150,27 @@ async def _exact_match_search(
     }
 
 
+def _project_entity(
+    record: dict[str, Any],
+    fields: list[str] | None,
+    attribute_keys: list[str] | None,
+) -> dict[str, Any]:
+    """Apply optional field projection to a HA entity record.
+
+    ``fields`` filters which top-level keys to keep (e.g. ["state", "attributes"]).
+    ``attribute_keys`` further filters the ``attributes`` sub-dict.
+    Both default None = full payload (no-op).
+    """
+    if fields is not None:
+        keep = set(fields)
+        record = {k: v for k, v in record.items() if k in keep}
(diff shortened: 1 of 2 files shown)


SealKan:
Superseded by #1225 which consolidates all six fields= projection PRs into a single implementation.

PR #1241, opened 2026-05-11
test(addon): fix base64 padding-bit flake in token tamper tests

### LINKED ISSUE #1238: [BUG] test_token_with_tampered_signature_rejected has ~6% flake rate (base64 padding bits)
## Summary

\`tests/addon/test_webhook_proxy.py::test_token_with_tampered_signature_rejected\` (added in #1184) fails about 6% of runs. Triggered on the first CI attempt of #1229; passed cleanly on retry.

## Root cause (empirically confirmed)

The test tampers the **last base64-url character** of the HMAC-SHA256 signature:

\`\`\`python
def test_token_with_tampered_signature_rejected(self, provider):
    token = provider.issue_access_token()
    body, sig = token.rsplit(".", 1)
    tampered_sig = sig[:-1] + ("A" if sig[-1] != "A" else "B")
    assert provider.validate_access_token(f"{body}.{tampered_sig}") is False
\`\`\`

A 32-byte HMAC-SHA256 is **256 bits**. Base64-url encoded, that's **43 characters**: 42 chars × 6 bits = 252 bits, plus a 43rd char encoding only **4 meaningful bits** of HMAC data and 2 zero-padding bits that \`base64.urlsafe_b64decode\` discards.

The base64-url alphabet has groups of 4 chars that share the same first 4 bits — e.g. \`{A=000000, B=000001, C=000010, D=000011}\` all start with \`0000\`. When \`sig[-1] ∈ {A, B, C, D}\` the test tampers within that same group (\`A↔B\`, or \`C/D→A\`), and the decoded HMAC bytes are **byte-identical**. \`hmac.compare_digest\` returns True, validation correctly passes, and the test (which expects rejection) reports a false failure.

\`P(sig[-1] ∈ {A,B,C,D}) = 4/64 = 6.25%\`. Verified empirically over 1000 simulated tokens with \`signing_key=b\"\\x00\" * 32\`: 60 hit the byte-identical case (6.0%).

## Suggested fix

Tamper a **byte** in the decoded HMAC and re-encode, or pick the tampered char such that its first 4 bits differ from the original's:

\`\`\`python
def test_token_with_tampered_signature_rejected(self, provider):
    token = provider.issue_access_token()
    body, sig = token.rsplit(".", 1)
    raw = base64.urlsafe_b64decode(sig + "=" * (-len(sig) % 4))
    tampered = bytes([raw[0] ^ 0xFF]) + raw[1:]  # flip the first byte
    tampered_sig = base64.urlsafe_b64encode(tampered).rstrip(b"=").decode()
    assert provider.validate_access_token(f"{body}.{tampered_sig}") is False
\`\`\`

Same fix likely needed for \`test_token_with_tampered_payload_rejected\` — depending on where the body payload's base64 boundary lands, the same padding-bit collision can occur there too.

## Related

- Discovered while shipping #1229. The Renovate token fix itself doesn't touch this test; the flake just happened to land on the first CI attempt and was misread as caused by that PR.

SealKan:
Fixes #1238

## Root cause

The last character of a 43-char base64-url HMAC-SHA256 encoding only carries
4 meaningful bits (2 bits are zero-padding). Characters in the same 4-char
group (e.g. `A`/`B`/`C`/`D`) share identical first 4 bits, so flipping the
last character within that group produces a byte-identical decoded value.
`hmac.compare_digest` returns `True`, validation correctly passes, and the
test — which expects rejection — fails. Probability: 4/64 = 6.25%.

## Fix

Target the **first** base64-url character instead. The first character always
encodes 6 full data bits with no padding, so any change to it is guaranteed to
produce a different decoded byte.

No new imports needed — the fix is a one-character positional change in both
affected tests.

Applied to both `test_token_with_tampered_payload_rejected` and
`test_token_with_tampered_signature_rejected` as the same padding-bit
collision can occur in the payload encoding depending on JSON length.

### DIFF
--- a/tests/addon/test_webhook_proxy.py
+++ b/tests/addon/test_webhook_proxy.py
@@ -1224,13 +1224,17 @@ class TestOAuthProvider:
     def test_token_with_tampered_payload_rejected(self, provider):
         token = provider.issue_access_token()
         body, sig = token.rsplit(".", 1)
-        tampered_body = body[:-1] + ("A" if body[-1] != "A" else "B")
+        # Flip the first base64-url character (6 full data bits, no padding)
+        # to guarantee a different decoded byte regardless of key material.
+        tampered_body = ("A" if body[0] != "A" else "B") + body[1:]
         assert provider.validate_access_token(f"{tampered_body}.{sig}") is False
 
     def test_token_with_tampered_signature_rejected(self, provider):
         token = provider.issue_access_token()
         body, sig = token.rsplit(".", 1)
-        tampered_sig = sig[:-1] + ("A" if sig[-1] != "A" else "B")
+        # Flip the first base64-url character (6 full data bits, no padding)
+        # to guarantee a different decoded byte regardless of key material.
+        tampered_sig = ("A" if sig[0] != "A" else "B") + sig[1:]
         assert provider.validate_access_token(f"{body}.{tampered_sig}") is False
 
     def test_expired_token_rejected(self, tmp_path):


kingpanther13 (approved):
lgtm

Issue #1313, opened 2026-05-16
[FEATURE] Extend identifier validation to tools_config_{automations,scripts,dashboards} + scenes DRY consolidation


Patch76:
### What would you like to see?

Extend the `validate_identifier_not_empty()` helper introduced in PR #1312 to the three `tools_config_*` tool families whose CRUD entry points still fall through to Home Assistant on empty/whitespace identifiers:

- `tools_config_automations.py` — `ha_config_get_automation` / `ha_config_set_automation` (update branch when `identifier` is provided) / `ha_config_remove_automation`
- `tools_config_scripts.py` — `ha_config_get_script` / `ha_config_set_script` (update branch) / `ha_config_remove_script`
- `tools_config_dashboards.py` — `ha_config_get_dashboard` / `ha_config_set_dashboard` (update branch) / `ha_config_delete_dashboard`

### Why do you need this?

#1294 closed the destructive intent-loss class (`action = "update" if id else "create"` on the registry-metadata writes) and the whitespace gap on existing partial guards in `tools_config_helpers.py` + `tools_areas.py`. The three families listed here were deliberately deferred — @kingpanther13's signal on #1294 ("belt and suspenders, depends on the tool and how complicated and/or destructive it can be") reserves the validation cost for the destructive / silent-failure class, not blanket family-wide, so a separate scope check is warranted before extending the pattern there.

The case for extending:

- **Consistency** — once #1312 lands, scenes is the only `tools_config_*` member with the empty/whitespace guard; the four siblings differ from it. Cross-family asymmetry adds cognitive load when reasoning about validation behaviour.
- **Marginal UX win** — HA's `RESOURCE_NOT_FOUND` for `automation.` / `script.` / `dashboard ""` is informative enough today, so the urgency is lower than #1294 was. But a structured `VALIDATION_INVALID_PARAMETER` with the parameter name in `context` is still cleaner than a generic not-found.

The case against extending:

- **No destructive routing** — none of these tools have `action = "update" if id else "create"` falsy semantics. Empty/whitespace `identifier` produces a clean not-found, not a silent create.
- **Cost ≠ benefit** for the simple read-only `get_*` tools — they already return structured errors via HA's response path.

### Additional context

**Scope sizing if implemented (Approach b, shared helper):**

| File | Entry points to guard | Approximate LOC |
|---|---|---|
| `tools_config_automations.py` | get + set (update branch) + remove | ~30 LOC + 1 import |
[...]

PR #1312, opened 2026-05-16
feat: reject empty/whitespace identifiers on registry-metadata writes (closes #1294)

Patch76:
## What does this PR do?

Adopts tool-side empty/whitespace identifier rejection across the destructive class of tools per the maintainer signal on #1294 (kingpanther13: belt-and-suspenders, per-tool nuance). Three failure modes are closed:

1. **Destructive intent loss** — `action = "update" if id else "create"` previously routed an empty-string `label_id` / `category_id` / area `id` / `helper_id` / `item` silently to `create` (or `update`, in the `ha_set_todo_item` case) when the caller intended the other. Now surfaces a structured `VALIDATION_INVALID_PARAMETER` error naming the offending parameter instead.

2. **Whitespace gap on partial guards** — existing `if not name` / `if not helper_id` / `if x is None` checks on `ha_config_set_helper` (simple + flow create branches), `ha_set_area_or_floor` (area + floor create branches), and `ha_manage_energy_prefs` (`stat_consumption` on add_device/remove_device) were truthy-true for `"   "` because `bool(" ") is True`. Upgraded to also reject whitespace-only strings.

3. **Misleading delete/service-call errors** — destructive remove/delete/disable/update/install tools that passed identifier parameters straight to the backend would surface as misleading HA-side "not found" or service-call errors when the identifier was empty/whitespace. Each is now guarded with a pre-flight that names the offending parameter.

A new shared helper `validate_identifier_not_empty()` in `src/ha_mcp/tools/helpers.py` centralises the empty/whitespace pre-flight check. It returns the validated value (typed `str`) so call sites can rebind to narrow `str | None → str` for mypy without a duplicate inline check, and accepts a `message` override for context-specific phrasing. `None` stays rejected by the helper — callers permitting the documented "list-all" / "create-new" sentinel gate with `if value is not None:` first.

**Tier 1 — destructive intent-loss class (silent-failure):**
- `tools_labels.py:70, 191, 276` — `action = "update" if label_id else "create"` falsy-routing fix; guard added to `set_label`, `get_label`, `remove_label` for symmetry
- `tools_categories.py:83, 214, 312` — same falsy-routing pattern; guard added to all three CRUD entry points
- `tools_areas.py:473, 597` — `id == ""` guard upgraded to also reject whitespace-only `"   "`; same upgrade applied to `remove_area_or_floor`
[...]

PR #1320, opened 2026-05-17
refactor: migrate tools_config_scenes inline empty-id guards to shared helper (closes #1314)

### LINKED ISSUE #1314: [FEATURE] Migrate tools_config_scenes.py inline empty-id guards to shared validate_identifier_not_empty helper

Migrate the three inline empty/whitespace `scene_id` guards in `tools_config_scenes.py` (L225, L524, L946) to the shared `validate_identifier_not_empty()` helper introduced in PR #1312.

Each site currently inlines:

```python
if not scene_id or not scene_id.strip():
    raise_tool_error(
        create_error_response(
            ErrorCode.VALIDATION_INVALID_PARAMETER,
            "scene_id must not be empty",
            suggestions=[
                "Pass a non-empty scene identifier (e.g. 'movie_night')",
                "Use ha_search_entities(domain_filter='scene') "
                "to find existing scene_ids",
            ],
            context={"scene_id": scene_id},
        )
    )
```

Replace each with one helper call:

```python
validate_identifier_not_empty(
    scene_id,
    "scene_id",
    suggestions=[
        "Pass a non-empty scene identifier (e.g. 'movie_night')",
        "Use ha_search_entities(domain_filter='scene') to find existing scene_ids",
    ],
)
```

### Why do you need this?

Pure-refactor scope — zero behaviour change, zero bug-class fixed. The motivation is **single source of truth**:

- The shared helper has the canonical empty/whitespace policy (including unicode whitespace via `str.strip()` semantics, parameter-name in structured context, etc.). Future tightening (e.g. unicode-whitespace normalisation, stripped-but-meaningful detection) becomes a one-line diff instead of three inline diffs spread across `tools_config_scenes.py`.
- Once #1312 and #1313 land, scenes is the only `tools_config_*` member still inlining the pattern; the four siblings use the helper. Consistency.
- Removes ~30 LOC of repeated raise-blocks, replaces with ~9 LOC of helper calls.

### Additional context

**Scope sizing:**

| Change | Approximate LOC |
|---|---|
| `tools_config_scenes.py` — 3 inline-guard sites migrated | -30 / +12 (net -18) |
| Import update (`validate_identifier_not_empty` added to existing helpers import) | +1 |
| Existing scenes negative-input tests in `tests/src/unit/` continue to pass unchanged (the helper raises the same `VALIDATION_INVALID_PARAMETER` with the same `scene_id` in context) | 0 |
| **Total** | **~-17 LOC** |

Existing test coverage for empty/whitespace `scene_id` lives in `tests/src/unit/test_helper_input_validation.py` and adjacent files (originally landed alongside PR #1168 R6); behaviour-equivalent migration means no test changes needed beyond updating any string-match assertions that depend on the exact message wording (the helper produces `"scene_id must be a non-empty, non-whitespace string"` instead of `"scene_id must not be empty"`).

**Cross-refs:**
- #1294 — original policy decision
- #1312 — introduces `validate_identifier_not_empty()`, demonstrates the pattern
- #1313 — Tier 3 extension to `tools_config_{automations,scripts,dashboards}` (independent scope; that issue adds new guards, this one migrates existing ones)

Labelled `priority: low` — refactor, no user-visible improvement.

Patch76:
## What does this PR do?

Migrates the three inline `if not scene_id or not scene_id.strip():` empty-id guards in `src/ha_mcp/tools/tools_config_scenes.py` to the shared `validate_identifier_not_empty` helper introduced in #1312. Pure refactor — zero behaviour change for callers.

The 3 sites migrated (call-line anchors):
- `tools_config_scenes.py:228` — `ha_config_get_scene` pre-flight
- `tools_config_scenes.py:527` — `ha_config_set_scene` pre-flight
- `tools_config_scenes.py:949` — `ha_config_remove_scene` pre-flight

### Behaviour-preservation notes

- **Error code**: `VALIDATION_INVALID_PARAMETER` unchanged.
- **Error message**: `"scene_id must not be empty"` preserved at each site via `message=` override (the helper's default message would have been `"scene_id must be a non-empty, non-whitespace string"`; preserving the existing wording keeps the `TestEmptySceneIdGuard` contract intact without test changes).
- **Error context shape**: `context["scene_id"]` key preserved at each site; the helper additionally merges its own `parameter` + `value` keys, so the resulting `context` is a backward-compatible superset (`{"scene_id": value, "parameter": "scene_id", "value": value}`). Grep across `tests/` and `src/` shows zero consumers of `context["scene_id"]` on a `VALIDATION_INVALID_PARAMETER` response from scenes tools — extension is safe.
- **`None` and whitespace handling**: original `if not scene_id or not scene_id.strip()` rejected `None`, `""`, and any pure-whitespace string. The helper's `if value is not None and value.strip()` check produces the same accept/reject set.
- **Backend-call ordering**: each guard fires before the first client call (`get_scene_config` at L244, `upsert_scene_config` at L710/837, `resolve_scene_id` / `delete_scene_config` at L965/L973), same as the original inline guards.

### Tests

No test changes. The existing `TestEmptySceneIdGuard` class (`tests/src/unit/test_tools_config_scenes.py:1341`) — including its 4-value whitespace parametrization at L1380 — continues to pass exactly as before. That's the strongest signal of true behavioural equivalence: the migration was contract-preserving enough that no test update was needed.

- ruff clean on `tools_config_scenes.py`
- mypy clean on `tools_config_scenes.py`
- Full local unit suite: **2615 passed, 1 skipped** (the skip is `test_numpy_version_constraint.py` — by-design on Alpine musl).

### Net LOC reality

Issue body estimated −17 LOC for the migration; actual diff is **+42 / −41 = +1 LOC**. The three new "Migrated to the shared helper (#1314)" comment paragraphs offset most of the inline-guard removal. Worth surfacing so the diff doesn't read as a no-op cleanup — the value is in deduplicating the pattern across CRUD families, not in line count.

### Dependency on #1312

This branch is stacked on `feature/issue-1294-validate-identifier-family` because `validate_identifier_not_empty` doesn't exist on `master` yet — it lands with #1312. When #1312 merges, this PR will rebase cleanly onto `master` and the helper-import resolves naturally; the diff narrows to just the 3 site-migrations (no helper-add overlap because the helper is shared infrastructure not duplicated here).

Until #1312 merges, the diff against `master` will appear to include both PRs' contents. Reviewing this PR as a stack-on-top-of-#1312 (or after #1312 merges) gives the accurate scope.

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [x] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

## Future improvements

None planned for scenes specifically. The companion Tier-3 expansion — extending the helper to `tools_config_{automations,scripts,dashboards}` get/set/remove — is tracked under #1313 (also stacked on #1312).

## Checklist
- [x] I have updated documentation if needed — no doc updates required for a behaviour-preserving refactor; cross-references in `AGENTS.md`/skill content unchanged.

Closes #1314

### DIFF
--- a/src/ha_mcp/tools/tools_config_scenes.py
+++ b/src/ha_mcp/tools/tools_config_scenes.py
@@ -32,6 +32,7 @@ from .helpers import (
     log_tool_usage,
     raise_tool_error,
     register_tool_methods,
+    validate_identifier_not_empty,
 )
 from .reference_validator import validate_config_references
 from .util_helpers import (
@@ -221,20 +222,20 @@ class ConfigSceneTools:
             # Issue #1168 R6 blocker 16: empty ``scene_id`` previously
             # surfaced as ``RESOURCE_NOT_FOUND`` with a misleading
             # `entities`-related suggestion. Pre-flight here so the caller
-            # gets the actual problem.
-            if not scene_id or not scene_id.strip():
-                raise_tool_error(
-                    create_error_response(
-                        ErrorCode.VALIDATION_INVALID_PARAMETER,
-                        "scene_id must not be empty",
-                        suggestions=[
-                            "Pass a non-empty scene identifier (e.g. 'movie_night')",
-                            "Use ha_search_entities(domain_filter='scene') "
-                            "to find existing scene_ids",
-                        ],
-                        context={"scene_id": scene_id},
-                    )
-                )
+            # gets the actual problem. Migrated to the shared
+            # ``validate_identifier_not_empty`` helper (#1314) — message
+            # and ``context["scene_id"]`` key preserved for callers.
+            validate_identifier_not_empty(
+                scene_id,
+                "scene_id",
+                message="scene_id must not be empty",
+                suggestions=[
+                    "Pass a non-empty scene identifier (e.g. 'movie_night')",
+                    "Use ha_search_entities(domain_filter='scene') "
+                    "to find existing scene_ids",
+                ],
+                context={"scene_id": scene_id},
+            )
             # Issue #1168 R3 blockers 3 + 6: unwrap the rest-client envelope
             # so the response carries the scene body directly (no nested
             # `success`/`scene_id`/`config` chain), and use the storage key
@@ -520,19 +521,19 @@ class ConfigSceneTools:
             # Issue #1168 R6 blocker 16: empty ``scene_id`` pre-flight before
             # any config dispatch — keeps the error code/message aligned with
             # the actual problem rather than the misleading
-            # ``RESOURCE_NOT_FOUND`` from a downstream lookup.
-            if not scene_id or not scene_id.strip():
-                raise_tool_error(
-                    create_error_response(
-                        ErrorCode.VALIDATION_INVALID_PARAMETER,
-                        "scene_id must not be empty",
-                        suggestions=[
-                            "Pass a non-empty scene identifier (e.g. 'movie_night')",
-                            "For a fresh create, use a name-derived slug",
-                        ],
-                        context={"scene_id": scene_id},
-                    )
-                )
+            # ``RESOURCE_NOT_FOUND`` from a downstream lookup. Migrated to
+            # the shared ``validate_identifier_not_empty`` helper (#1314) —
+            # message and ``context["scene_id"]`` key preserved for callers.
+            validate_identifier_not_empty(
+                scene_id,
+                "scene_id",
+                message="scene_id must not be empty",
+                suggestions=[
+                    "Pass a non-empty scene identifier (e.g. 'movie_night')",
+                    "For a fresh create, use a name-derived slug",
+                ],
+                context={"scene_id": scene_id},
+            )
             # Validate mutual exclusivity of config and python_transform
             if config is not None and python_transform is not None:
                 raise_tool_error(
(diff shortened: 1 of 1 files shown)


Patch76:
## Implementation Summary

**Choices Made:**
- **Preserved the original error message via `message=` override** — the helper's default would have read `"scene_id must be a non-empty, non-whitespace string"`, which differs from the original `"scene_id must not be empty"`. The existing `TestEmptySceneIdGuard` class asserts the original wording verbatim; using `message=` keeps the test contract intact and avoids touching the test file at all. Listed as a Minor in the idiot-check pass.
- **Preserved the `context["scene_id"]` key** — the helper additionally merges its own `parameter` + `value` keys, so the resulting `context` is a backward-compatible superset (`{"scene_id": value, "parameter": "scene_id", "value": value}`). Grep across `tests/` and `src/` returned zero consumers of `context["scene_id"]` on a `VALIDATION_INVALID_PARAMETER` response from scenes tools, so dropping it would have been safe too — the explicit preservation is for forward-compat with any future caller that grep-discovers the key.
- **Side-effect helper call, no narrowing-rebind** — scene_id is `Annotated[str, ...]` in all three signatures (always non-None), so the helper's return value isn't needed for mypy narrowing. Matches the simpler call form, not the `name = validate_identifier_not_empty(name, …)` rebind pattern from `tools_areas.py` (which needed it for `str | None → str` narrowing).
- **Stacked on `feature/issue-1294-validate-identifier-family` (#1312)** — the helper doesn't exist on master yet. Branching off #1312 makes the diff appear large here (21 files / +1661 / −80) until #1312 merges; after that this PR rebases onto master and the diff narrows to **1 file / +42 / −41**.

**Problems Encountered:**
- **Submodule init missed on worktree creation** — full unit suite initially failed on `test_skills_vendor_directory_exists` because `src/ha_mcp/resources/skills-vendor/` was empty. Resolved with `git -C worktree/issue-1314 submodule update --init --recursive`. The resolver-skill's Phase 1 step lists the submodule init; missing it during this run was on me. Filing as a self-note for next worktree creation.

**Suggested Improvements:**
- **Tier-3 expansion (#1313)** — extending the helper to `tools_config_{automations,scripts,dashboards}` get/set/remove is in flight as a separate PR stacked on top of this one. That PR will continue to use the same `message=` / context-preservation conventions established here.

kingpanther13 (approved):
Clean mechanical migration. Verified:

- Behavior preserved — original `if not scene_id or not scene_id.strip()` and the helper's `if value is not None and value.strip()` accept/reject the same set (`None`, `""`, whitespace-only).
- Error code (`VALIDATION_INVALID_PARAMETER`) and message (`"scene_id must not be empty"`) byte-identical via `message=` override at all three sites.
- Context now `{scene_id, parameter, value}` — additive superset, no consumers depend on the exact shape.
- Existing `TestEmptySceneIdGuard` (`test_tools_config_scenes.py:1341`) plus helper-contract tests in `test_identifier_validation_family.py:41` cover the contract end-to-end; no new tests needed.

Approving — this is exactly the consolidation the helper was added for.

Issue #1318, opened 2026-05-17
[FEATURE] chore: ruff format repo-wide sweep + add ruff format --check to CI


Patch76:
### What would you like to see?

Two coupled changes to close the accumulated `ruff format` debt:

1. **A `chore: ruff format` sweep PR** applying `uv run ruff format` repo-wide across `src/` and `tests/`. Pure mechanical, no behaviour change.
2. **Add `ruff format --check` to `.github/workflows/pr.yml`** (alongside the existing `ruff check` step at `pr.yml:46`) so the debt does not re-accumulate.

### Why do you need this?

Three merged PRs in the last week explicitly documented pre-existing format-debt on touched files and deferred it as a separate sweep — none of those follow-ups was ever filed. Quoting the implementation summaries:

- **#1254** ([comment](https://github.com/homeassistant-ai/ha-mcp/pull/1254#issuecomment-4425752310)): *"pre-existing master had 140 lines of format-reflow debt across these 3 files, all at lines not touched by this PR."*
- **#1258** ([comment](https://github.com/homeassistant-ai/ha-mcp/pull/1258#issuecomment-4425800576)): *"the 681-line pre-existing format-debt on 4 of 5 touched files (separate `chore: ruff format tests/src/e2e/workflows` sweep candidate)."*
- **#1303** ([comment](https://github.com/homeassistant-ai/ha-mcp/pull/1303#issuecomment-4463596572)): *"e2e ruff-format debt on touched e2e files (master-state, CI enforces only `ruff check`)."*

Live state right now (verified on master `bad13aa`):

| Scope | Files needing reformat | Per `ruff format --check` |
|---|---:|---|
| `src/ha_mcp/` | 44 | production code |
| `tests/src/unit/` | 73 | unit tests |
| `tests/src/e2e/` | 47 | e2e tests |
| `tests/uat/` | 10 | UAT runner + stories |
| `tests/addon/` + `tests/test_env_manager.py` | 7 | addon-suite + env helper |
| **total** | **181** | ~11.6k diff lines |

[...]

PR #1258, opened 2026-05-11
test(e2e): use time.monotonic() in workflow polling loops

Patch76:
## What does this PR do?

Closes #1255.

Continuation of the `time.time()` → `time.monotonic()` sweep started in PR #1249 (e2e framework: `wait_helpers.py`, `assertions.py`, `test_network_errors.py`, `conftest.py`) and PR #1254 (UAT runner + `test_env_manager.py`). This PR addresses the 5 e2e workflow test files that roll their own polling loops instead of using `wait_helpers.py`.

`time.monotonic()` is NTP-immune and forward-only — the textbook-correct primitive for measuring elapsed time. Same motivation as #1234 / #1249 / #1250 / #1254. Risk is low: `time.monotonic()` has the same return type and call signature as `time.time()` for relative-duration math; only the absolute reference point changes.

**Scope: 15 mechanical substitutions across 5 files** — every callsite is duration math (`start_time = time.time(); while time.time() - start_time < timeout` or `elapsed = time.time() - start_time`):

- `tests/src/e2e/workflows/scenes/test_lifecycle.py` — 4 callsites
- `tests/src/e2e/workflows/scripts/test_lifecycle.py` — 5 callsites
- `tests/src/e2e/workflows/device_control/test_lights.py` — 2 callsites
- `tests/src/e2e/workflows/todo/test_lifecycle.py` — 2 callsites
- `tests/src/e2e/workflows/config/test_helper_crud.py` — 2 callsites

**Untouched (absolute-timing exceptions):** `int(time.time())` callsites in the same files (`todo/test_lifecycle.py:183, 411, 454, 455, 523`) are wall-clock-derived unique-ID generators and must stay absolute-timing — exactly the same class as the documented exceptions in PR #1249 (e.g. `conftest.py:136` recorder-timestamp metric). The substitution was applied with a placeholder-protect pattern (`int(time.time())` → placeholder → substitute → restore) to guarantee these stay untouched.

Issue #1255 enumerated 3 of these unique-ID callsites; live audit on the touched files found 6 in total (the 5 listed above plus `tests/src/e2e/workflows/convenience/test_backup.py:97`, which is in a different file and not touched by this PR). The expanded count is informational — all 6 are the same class and intentionally out of sweep scope.

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [x] 🧪 Tests only
- [ ] 💥 Breaking change

[...]

PR #1321, opened 2026-05-17
refactor: extend validate_identifier_not_empty to automations/scripts/dashboards CRUD (closes #1313)

### LINKED ISSUE #1313: [FEATURE] Extend identifier validation to tools_config_{automations,scripts,dashboards} + scenes DRY consolidation

Patch76:
## What does this PR do?

Extends the `validate_identifier_not_empty` helper (introduced in #1312) to the three remaining `tools_config_*` CRUD families: `automations`, `scripts`, `dashboards`. Explicit Tier-3 follow-up from the #1312 PR body — "a separate PR can extend the helper to `tools_config_{automations,scripts,dashboards}` get/set/remove if a maintainer signal endorses the blanket expansion. The helper is ready; only call-site additions would be needed." This is that separate PR.

### Sites guarded (7 new + 2 pre-existing)

#1312 already guarded `ha_config_remove_automation` (`tools_config_automations.py:970`) and `ha_config_remove_script` (`tools_config_scripts.py:624`); they're listed for completeness, not re-touched here.

| File | Tool | Anchor | Pattern |
|---|---|---|---|
| `tools_config_automations.py` | `ha_config_get_automation` | L296 | unconditional (identifier is required `str`) |
| `tools_config_automations.py` | `ha_config_set_automation` | L565 | conditional `if identifier is not None:` (None → create new; non-None → update — guard fires only on update path) |
| `tools_config_automations.py` | `ha_config_remove_automation` | L1002 | pre-existing from #1312 |
| `tools_config_scripts.py` | `ha_config_get_script` | L111 | unconditional |
| `tools_config_scripts.py` | `ha_config_set_script` | L434 | unconditional (script_id is required `str`) |
| `tools_config_scripts.py` | `ha_config_remove_script` | L652 | pre-existing from #1312 |
| `tools_config_dashboards.py` | `ha_config_get_dashboard` | L618 | conditional `if url_path is not None:` (placed after the `if list_only:` early-return so list-mode is unaffected; the documented "omit url_path → default dashboard" fallback remains reachable via `url_path=None` without `list_only`) |
| `tools_config_dashboards.py` | `ha_config_set_dashboard` | L1026 | unconditional, placed BEFORE the `"default"` alias so empty doesn't accidentally normalise |
| `tools_config_dashboards.py` | `ha_config_delete_dashboard` | L1534 | unconditional, placed BEFORE `_resolve_dashboard` so empty doesn't surface as a misleading "no dashboard found" |

Plus 1 import added to `tools_config_dashboards.py` (the other two files already imported the helper via #1312).

### Mypy narrowing

Per the issue body's caveat: only `set_automation` has a `str | None` identifier. The implementation uses the side-effect call form (`validate_identifier_not_empty(identifier, …)` not `identifier = validate_identifier_not_empty(…)`) because downstream code at the python_transform branch already gates on `if not identifier:` independently — no narrowing-rebind needed for type safety. `set_script` and `set_dashboard` both take `str` (always non-None) so unconditional side-effect calls fit cleanly.

### Tests

`tests/src/unit/test_identifier_validation_family.py` — 15 new test scenarios across 3 classes:

- **`TestAutomationsIdentifierValidation`** extended with `test_get_rejects_empty_identifier` (×2 params) + `test_set_rejects_empty_identifier_on_update` (×2 params)
- **`TestScriptsIdentifierValidation`** extended with `test_get_rejects_empty_script_id` (×2) + `test_set_rejects_empty_script_id` (×2)
- **`TestDashboardsIdentifierValidation`** (new) — closure-pattern register-and-capture, covers all three dashboard tools plus a `test_get_dashboard_list_only_skips_guard` control test that proves `list_only=True` does NOT trip the conditional guard (positive assertion: `send_websocket_message.call_count > 0` confirms the call reached `_fetch_dashboards_list`)

Each rejection test asserts:
- `VALIDATION_INVALID_PARAMETER` error code
- `'"parameter": "<name>"' in str(excinfo.value)` to discriminate the new guard from any pre-existing validation in the same function
- Underlying client method (`upsert_*_config` / `send_websocket_message` / etc.) NOT called

### Local

- ruff `check` clean on all 4 modified files
- mypy clean on the 3 modified production files
- Full unit suite: **2658 passed, 1 skipped** (the skip is `test_numpy_version_constraint.py` — by-design on Alpine musl); +15 new tests vs the master baseline

### Diff

Both parent PRs (#1312 + #1320) have merged. Rebased on master @ `264bfc2`. Diff: **4 files / +283 / −0** (3 production + 1 test).

### Out of scope (intentional)

- **Scenes inline-guard migration** — already handled in #1320 (the parent of this PR). Not duplicated here.
- **Dashboard-resource tools** (`tools_resources.py`) — already covered by #1312 (resource_id guard on `_upsert_resource` + `ha_config_delete_dashboard_resource`).
- **ruff format pre-existing debt** — all 4 modified files would be reformatted by `ruff format`, but the format-debt is inherited from master / base branch (verified by stash-and-recheck during idiot-check). New guard lines themselves are well-formatted. Out-of-scope per the established mechanical-PR scope-brake convention; tracked under #1318.

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [x] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

## Future improvements

None planned for these three families. The helper is now applied across all `tools_config_*` CRUD entry points that have an identifier-shaped parameter and pass it to a backend call.

## Checklist
- [x] I have updated documentation if needed — no doc updates required for additive validation guards; cross-references in `AGENTS.md` / skill content unchanged.

Closes #1313

### DIFF
--- a/src/ha_mcp/tools/tools_config_automations.py
+++ b/src/ha_mcp/tools/tools_config_automations.py
@@ -287,6 +287,21 @@ class AutomationConfigTools:
         For comprehensive automation documentation, use ha_get_skill_guide.
         """
         try:
+            # Empty/whitespace identifier would propagate to the internal
+            # config lookup and surface as a misleading
+            # ``RESOURCE_NOT_FOUND``. Structured ``VALIDATION_INVALID_PARAMETER``
+            # naming the parameter is cleaner — extension of the #1312
+            # validate_identifier_not_empty pattern to the automations
+            # family per #1313.
+            validate_identifier_not_empty(
+                identifier,
+                "identifier",
+                suggestions=[
+                    "Pass an automation entity_id (e.g. 'automation.morning_routine')",
+                    "Or pass the unique_id of an existing automation",
+                    "Use ha_search_entities(domain_filter='automation') to list automations",
+                ],
+            )
             normalized_config, config_hash = await self._get_automation_config_internal(identifier)
 
             # Resolve entity_id and fetch category from entity registry
@@ -539,6 +554,23 @@ class AutomationConfigTools:
         """
         bp_warnings: list[str] = []
         try:
+            # ``identifier`` is optional (omit → create new with generated
+            # unique_id; pass → update existing). When provided, reject
+            # empty/whitespace up-front so the caller gets a structured
+            # parameter error instead of a misleading ``RESOURCE_NOT_FOUND``
+            # from the downstream lookup. The ``not identifier`` check
+            # further down the python_transform branch still handles the
+            # explicit ``identifier is None`` case for that mode.
+            if identifier is not None:
+                validate_identifier_not_empty(
+                    identifier,
+                    "identifier",
+                    suggestions=[
+                        "Omit identifier to create a new automation",
+                        "Or pass a valid automation entity_id / unique_id to update",
+                    ],
+                    context={"action": "set"},
+                )
             # Validate mutual exclusivity of config and python_transform
             if config is not None and python_transform is not None:
                 raise_tool_error(
--- a/src/ha_mcp/tools/tools_config_dashboards.py
+++ b/src/ha_mcp/tools/tools_config_dashboards.py
@@ -26,6 +26,7 @@ from .helpers import (
     extract_tool_error_message,
     log_tool_usage,
     raise_tool_error,
+    validate_identifier_not_empty,
 )
 from .util_helpers import parse_json_param
 
@@ -605,6 +606,25 @@ def register_config_dashboard_tools(mcp: Any, client: Any, **kwargs: Any) -> Non
                     "count": len(dashboards),
                 }
 
+            # ``url_path`` is optional in this tool (omitted with
+            # ``list_only=True`` lists all dashboards — handled above; omitted
+            # without ``list_only`` falls back to the default dashboard via
+            # the resolver below). When provided, reject empty/whitespace
+            # up-front so the caller gets a structured parameter error
+            # instead of a misleading ``RESOURCE_NOT_FOUND``. Extension of
+            # the #1312 validate_identifier_not_empty pattern to the
+            # dashboards family per #1313.
+            if url_path is not None:
+                validate_identifier_not_empty(
+                    url_path,
+                    "url_path",
+                    suggestions=[
+                        "Pass a dashboard URL path (e.g. 'lovelace-home')",
+                        "Omit url_path and pass list_only=True to list dashboards",
+                        "Use 'default' to target the default dashboard",
+                    ],
+                )
+
             # Search mode — find cards, badges, or header cards
             if search_mode:
                 get_data: dict[str, Any] = {"type": "lovelace/config", "force": True}
@@ -996,6 +1016,22 @@ def register_config_dashboard_tools(mcp: Any, client: Any, **kwargs: Any) -> Non
         are also updated if explicitly provided alongside (or instead of) a config change.
         """
         try:
+            # ``url_path`` is required (always non-None). Reject empty/
+            # whitespace up-front so the caller gets a structured parameter
+            # error instead of a misleading downstream failure (the
+            # subsequent "default" alias, pre-resolver, and hyphen check
+            # all assume a usable string). Extension of the #1312
+            # validate_identifier_not_empty pattern to the dashboards
+            # family per #1313.
+            validate_identifier_not_empty(
+                url_path,
+                "url_path",
+                suggestions=[
+                    "Pass a dashboard URL path (e.g. 'my-dashboard')",
+                    "Use 'default' or 'lovelace' for the default dashboard",
+                ],
+                context={"action": "set"},
+            )
             # Handle "default" as alias for the default dashboard
             # (matches ha_config_get_dashboard behavior)
             if url_path == "default":
@@ -1490,6 +1526,20 @@ def register_config_dashboard_tools(mcp: Any, client: Any, **kwargs: Any) -> Non
         Note: The default dashboard cannot be deleted via this method.
         """
         try:
+            # ``url_path`` is required. Reject empty/whitespace up-front so
+            # the caller gets a structured parameter error instead of a
+            # misleading "no dashboard found" from the resolver below.
+            # Extension of the #1312 validate_identifier_not_empty pattern
+            # to the dashboards family per #1313.
+            validate_identifier_not_empty(
+                url_path,
+                "url_path",
+                suggestions=[
+                    "Pass a dashboard URL path or internal ID (e.g. 'my-dashboard')",
+                    "Use ha_config_get_dashboard(list_only=True) to list dashboards",
+                ],
+                context={"action": "delete"},
+            )
             resolved, _ = await _resolve_dashboard(client, url_path)
             if resolved is None:
                 raise_tool_error(
--- a/src/ha_mcp/tools/tools_config_scripts.py
+++ b/src/ha_mcp/tools/tools_config_scripts.py
@@ -103,6 +103,19 @@ class ConfigScriptTools:
         For detailed script configuration help, use ha_get_skill_guide.
         """
         try:
+            # Empty/whitespace script_id would propagate to
+            # ``get_script_config`` and surface as a misleading
+            # ``RESOURCE_NOT_FOUND``. Extension of the #1312
+            # validate_identifier_not_empty pattern to the scripts
+            # family per #1313.
+            validate_identifier_not_empty(
+                script_id,
+                "script_id",
+                suggestions=[
+                    "Pass a script identifier (e.g. 'morning_routine')",
+                    "Use ha_search_entities(domain_filter='script') to list scripts",
+                ],
+            )
             config_result = await self._client.get_script_config(script_id)
             # Extract actual script config body and compute hash before category injection
             actual_config = config_result.get("config", config_result)
@@ -412,6 +425,21 @@ class ConfigScriptTools:
         """
         bp_warnings: list[str] = []
         try:
+            # ``script_id`` is required (always non-None). Reject empty/
+            # whitespace up-front so the caller gets a structured parameter
+            # error instead of a misleading ``RESOURCE_NOT_FOUND`` from
+            # the downstream upsert/fetch. Extension of the #1312
+            # validate_identifier_not_empty pattern to the scripts family
+            # per #1313.
+            validate_identifier_not_empty(
+                script_id,
+                "script_id",
+                suggestions=[
+                    "Pass a script identifier (e.g. 'morning_routine')",
+                    "Use ha_search_entities(domain_filter='script') to list scripts",
+                ],
+                context={"action": "set"},
+            )
             # Validate mutual exclusivity of config and python_transform
             if config is not None and python_transform is not None:
                 raise_tool_error(
--- a/tests/src/unit/test_identifier_validation_family.py
+++ b/tests/src/unit/test_identifier_validation_family.py
@@ -567,8 +567,7 @@ class TestAutomationsIdentifierValidation:
         )
         tools._client.send_websocket_message.assert_not_called()
         # Also no REST GET on the automation config.
-        if hasattr(tools._client, "get_automation_config"):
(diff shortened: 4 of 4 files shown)


Patch76:
## Implementation Summary

**Choices Made:**
- **Side-effect helper call form for `ha_config_set_automation`** (vs `identifier = validate_identifier_not_empty(identifier, …)` reassign-for-narrowing) — `identifier` stays typed `str | None` after the guard. Defensible because the python_transform branch downstream at `tools_config_automations.py:591` has its own `if not identifier:` gate, and the config-update branches don't re-narrow. Could be tightened to capture the return value if a future maintainer prefers stricter narrowing; not needed for the current downstream shape.
- **Conditional vs unconditional guard placement matches each tool's param shape:**
  - `set_automation` (`identifier: str | None`) → `if identifier is not None:` so the documented create-new path (`identifier=None`) is unaffected
  - `get_dashboard` (`url_path: str | None` with `list_only` escape hatch) → `if url_path is not None:` and placed AFTER the `if list_only:` early-return, so list-mode never reaches the guard
  - All other 5 guards → unconditional side-effect call (required `str` params)
- **`message=` override deliberately NOT used** (different choice from sibling PR #1320) — these tools had no pre-existing test contract on a specific message string, so the helper's default `"<param> must be a non-empty, non-whitespace string"` is fine and reads cleanly. Tests assert `'"parameter": "<name>"'` for discrimination, not message text.
- **`context={"action": "set"}` / `context={"action": "delete"}` on the set/delete guards** — small extra context to discriminate the guard's structured error from the action it pre-empts, mirroring the convention from PR #1312 (Tier 1 entries like `tools_config_helpers.py:2412` that pass `context={"helper_type": ..., "action": action}`). The `get`-side guards omit this since there's only one action they can pre-empt.
- **Test class layout follows the existing `test_identifier_validation_family.py` convention** — class-based tools (`AutomationConfigTools`, `ConfigScriptTools`) get extension methods on their existing `Test*IdentifierValidation` classes; the closure-based `tools_config_dashboards.py` gets a new `TestDashboardsIdentifierValidation` class with the register-and-capture pattern from #1312's entities/registry tests.
- **`test_get_dashboard_list_only_skips_guard` control test** — positive-proof of the conditional-guard correctness. Asserts `send_websocket_message.call_count > 0` after `list_only=True` (proves `_fetch_dashboards_list` was reached, which means the guard didn't trip on `url_path=None`). Without this control, a regression that flipped the `if url_path is not None:` condition to unconditional would still pass the rejection tests but break legitimate list-mode usage.

**Problems Encountered:**
- **`ha_config_set_script` body location took an extra grep pass** — `tools_config_scripts.py` has 4 `try:` blocks at L105, L201, L427, L635. The L201 block is inside `_normalize_and_validate_set_script_config` (a private helper), not the public tool. Resolved via `awk '/^    async def ha_config_set_script/,/^    @/'` to scope the search to the tool's body. The right try-block is L427.
- **Submodule init was caught at worktree-creation time this run** — the resolver-skill Phase 1 step `git -C worktree/issue-1313 submodule update --init --recursive` ran cleanly. No mid-run failures on `test_skills_vendor_directory_exists` this time (vs the #1320 run where I missed the init initially).

**Suggested Improvements:**
- **`ruff format` pre-existing debt** across all 4 modified files — verified via stash-and-recheck during idiot-check that the format-debt is inherited from the base branch, not introduced here. Tracked under #1318; out-of-scope per the mechanical-PR scope-brake convention established in #1258 / #1303.
- **`set_automation` mypy-narrowing tightening** — could rewrite the guard as `identifier = validate_identifier_not_empty(identifier, …)` inside the `if identifier is not None:` block to narrow `str | None → str` for downstream. Not needed currently because the python_transform branch has its own gate, but worth doing if a future refactor removes that gate.

**Bot review:**
- Gemini Code Assist (review submitted `2026-05-17T11:10:59Z` on pre-rebase HEAD `0b220bb`) — no inline findings, no PR-level comments. Body: *"I have no feedback to provide as there were no review comments to assess."* Rebased on master @ `264bfc2` (HEAD now `281fe0d`) post #1320 merge; expect a fresh Gemini pass on Mark-Ready.

kingpanther13 (changes requested):
Pattern and placement look right — guards sit in the correct relation to the alias normalization, resolver, and list-mode early-return at every site, and all 7 try blocks have `except ToolError: raise` ahead of the broad catch so the helper's structured error propagates unmodified. Asking for four test changes before merge.

### Missing control tests (regression protection)

The family convention pairs every conditional-guard rejection with a positive `None`-path control. Two are missing:

1. **`test_set_automation_with_none_identifier_routes_to_create`** — mirror the existing pattern at `test_identifier_validation_family.py:158` (labels) / 202 (categories) / 279 (areas). Without it, a future tightening of `if identifier is not None:` to `if identifier is None or not identifier.strip():` would silently break the documented create-mode and no test in the file would catch it.

2. **`test_get_dashboard_with_none_url_path_routes_to_default`** — `tools_config_dashboards.py:569-571` documents "defaulting to the main dashboard if url_path is omitted". The `list_only` control proves one no-fire path; this proves the other. Same regression class as above.

### Test hygiene

3. **`test_get_dashboard_list_only_skips_guard` bare `except Exception: pass`** at line 700 — if `_fetch_dashboards_list` starts raising for an unrelated reason (including a stray `ToolError` from the guard regressing), the `call_count > 0` assertion still runs on a partially-completed body. Tighten to the specific exception classes the unrigged mock can produce (`AttributeError`, `KeyError`, `TypeError`) so a guard-regression `ToolError` surfaces instead of getting swallowed.

4. **Dead `hasattr` guards** at lines 570 / 621 / 638 — the mock is a bare `MagicMock` so `hasattr` is always True; the False branches never execute. Either drop the guards or attach `spec=AutomationConfigTools._client_type` (or equivalent) to the mock so the check becomes meaningful.

Production code is solid; once the controls land and re-review finds nothing else, this should be a straightforward approve.

Patch76:
Thanks for the detailed control-test pairing. All four addressed in `decff06`.

### 1. `test_set_automation_with_none_identifier_routes_to_create`

Added to `TestAutomationsIdentifierValidation`, mirroring the labels (`test_set_with_none_label_id_routes_to_create`, L158) / categories (L202) / areas (L279) pattern. Positive-proof: mocks `upsert_automation_config`, calls with `identifier` omitted + `wait=False`, asserts `call_args[0][1] is None`. A regression flipping `if identifier is not None:` to `if identifier is None or not identifier.strip():` now fails this test instead of silently breaking create-mode.

### 2. `test_get_dashboard_with_none_url_path_routes_to_default`

Added next to `test_get_dashboard_list_only_skips_guard`. Mocks `send_websocket_message`, calls `ha_config_get_dashboard()` (no args), asserts the get-mode WS payload contains `type: "lovelace/config"` but **no** `url_path` key — HA returns the default dashboard when the key is absent (`tools_config_dashboards.py` L734: `if url_path and url_path != "default":`). A regression flipping the L617 guard to unconditional, or normalising `None` to `""` before the L734 gate, now fails this assertion.

### 3. `test_get_dashboard_list_only_skips_guard` bare except

Tightened to `except (AttributeError, KeyError, TypeError)` — the unstructured-mock noise from list-mode touching missing keys on the bare `MagicMock` response is what the prior `except Exception` was actually catching. Any `ToolError` from a guard regression now propagates and the `call_count > 0` assertion can't pass by accident.

### 4. Dead `hasattr` guards (570 / 621 / 638)

Removed. Also swept the same-class sibling at the original L590 (`hasattr(tools._client, "upsert_automation_config")` in `test_set_rejects_empty_identifier_on_update`) — same `TestAutomationsIdentifierValidation` class, same dead-branch pattern. Now `grep -n "hasattr" tests/src/unit/test_identifier_validation_family.py` returns zero hits. I considered the `spec=` alternative; the direct `assert_not_called()` is simpler since every guarded path in this file has the attribute on the same `MagicMock`-shaped client and there's no other reason to spec the mock.

### Local

- `pytest tests/src/unit/test_identifier_validation_family.py` → 129 passed in 2.58s (+2 new)
- Ruff clean on the modified file
- HEAD: `decff06` (CI re-running)

kingpanther13 (approved):
All four addressed cleanly in `decff06c`.

- Both None-routing controls assert positively (`call_args[0][1] is None` for set_automation; `"url_path" not in sent` for get_dashboard) — exactly the regression class I had in mind.
- Tightened `except` is the right tuple; comment explains why bare-MagicMock list-mode body produces those classes.
- Dead `hasattr` sweep also caught the sibling at the original L590 on the rejection test that I missed — and the rejection test now rigs `upsert_automation_config = AsyncMock()` so the `assert_not_called()` isn't a no-op post-sweep. Nice catch.

LGTM.

Issue #1332, opened 2026-05-17
[FEATURE] Migrate result["warning"] singular → result["warnings"] list across tool modules


Patch76:
### What would you like to see?

Migrate the project's per-tool `result["warning"] = "..."` (singular string) sites to `result.setdefault("warnings", []).append("...")` (plural list), repo-wide. Today each site sets only one warning condition, so the overwrite-the-previous-warning risk is latent — but every future "add a second warning condition" PR has to either (a) migrate the singular key first, or (b) silently shadow the existing warning.

Concretely the singular sites on master (as of `bd9397f`):

| File | Sites |
|---|---|
| `src/ha_mcp/tools/tools_config_automations.py` | L687, L706, L708, L993, L995 |
| `src/ha_mcp/tools/tools_config_scenes.py` | L731, L740, L852, L861, L980, … (multiple) |
| `src/ha_mcp/tools/tools_config_scripts.py` | L541, L543, L641, L643 |
| `src/ha_mcp/tools/tools_groups.py` | L311, L313, L412, L414 |
| `src/ha_mcp/tools/tools_search.py` | L739 |
| `src/ha_mcp/tools/tools_system.py` | L377 (migrated in #1328's branch — drops out once #1328 merges) |
| `src/ha_mcp/tools/tools_integrations.py` | L805 |

The target shape is documented in `AGENTS.md` L641-649 (tool return contract) and is already canonical on the `ha_get_integration` / `ha_get_system_health` newer paths added in #1328.

Suggested approach: one sweep PR per tool-domain file (so the diffs stay small and reviewable), each one migrating the file's singular sites + flipping any in-tree test assertions that read `result["warning"]` → `result["warnings"][0]` or equivalent.

### Why do you need this?

Surfaced as a "Suggested Improvement" in #1328's Implementation Summary. The PR adopted the singular → plural migration locally on `tools_system.py:377` (Gemini round-3 GR3-1) as a sibling-bug sweep, but explicitly noted the rest of the repo's singular sites were out-of-scope for a diagnostics-feature PR.

The current state is correct *today* (no overwrites), but the next addition of a second warning condition to any of those tools is a footgun. Migrating proactively closes the class of bug rather than letting it land in a future PR's review cycle.

### Additional context

[...]

PR #1336, opened 2026-05-17
perf: parallelize ha_get_system_health optional sections via asyncio.gather

### LINKED ISSUE #1331: [FEATURE] Parallelize ha_get_system_health optional sections via asyncio.gather

Run the optional sections of `ha_get_system_health` in parallel via `asyncio.gather` (with `return_exceptions=True` for per-section error containment) instead of sequentially. The four optional sections all `await` independent backend round-trips:

- `await self._fetch_repairs(...)` (REST + WS, dismissed-filter pass)
- `await self._fetch_zha_network(...)` (WS)
- `await self._fetch_zwave_network(...)` (WS)
- `await fetch_integration_diagnostics(...)` (REST, up to 60 s on ZHA)

Currently `src/ha_mcp/tools/tools_system.py:ha_get_system_health` runs these in series (`tools_system.py:381-403` after #1328). When an agent calls e.g. `include="zha_network,diagnostics"` against ZHA + Hue, the wall-clock is `t_zha + t_diag` instead of `max(t_zha, t_diag)`. On the worst-case timeouts, that's 60 + 60 = 120 s vs. ~60 s.

Sketch (not for direct copy-paste — error attribution + ws_client lifecycle need care):

```python
async def _zha_or_none():
    if zha_full or zha_summary:
        return await self._fetch_zha_network(ws_client, full=zha_full)
    return None

# ...one wrapper per optional section, then:
repairs_r, zha_r, zwave_r, diag_r = await asyncio.gather(
    _repairs_or_none(), _zha_or_none(), _zwave_or_none(), _diag_or_none(),
    return_exceptions=True,
)
```

### Why do you need this?

The `ha_get_system_health` use-case is "triage by bundling everything in one round-trip" — the same agent invocation that benefits from `include=` aggregation also benefits from running those independent fetches concurrently. Today the win is left on the table.

`#1328`'s Implementation Summary listed this as a "Suggested Improvement" with the rationale "Left out so error attribution stays straightforward; refactor with deliberate per-section error containment fits a separate PR" — this is that separate PR's tracking issue.

### Additional context

Constraints / design points to settle in the PR:

- **Per-section error containment** — each `_fetch_*` already returns embeddable sub-dicts on failure (`_fetch_repairs`, `_fetch_zha_network`, `fetch_integration_diagnostics`). Need to confirm `_fetch_zwave_network` matches the same convention; if not, that's a precondition fix.
- **`ws_client` lifecycle** — the function reuses one `ws_client` for the WS-bound sections. Parallel access to a single client may need serialization at the client layer or a per-section client.
- **`return_exceptions=True` vs. inline `try/except` per wrapper** — the latter keeps the exception-to-error-string mapping symmetrical with the helper, the former centralises but loses per-section context. Decide explicitly in the PR.
- **Existing per-section unit tests** unchanged in behaviour (each test mocks one section); concurrency-specific tests would need `mock.gather`-style verification.
- **Cross-ref**: `#1328` adds `diagnostics_fields` + `diagnostics_truncate_at_bytes`. The parallelization landing on top of that should keep both new params threaded correctly into the wrapped helper call.

Patch76:
## What does this PR do?

Closes #1331. Replaces the sequential helper-fetch chain in `ha_get_system_health` with `asyncio.gather(*, return_exceptions=True)`, so independent backend calls for `repairs` / `zha_network` / `zwave_network` overlap on the wait side. Per-section error attribution preserved: a raising helper surfaces as `result["<section>"] = {"error": "..."}` while sibling sections still populate.

This PR sits on master at `bd9397f` (pre-#1328); the `diagnostics` section will need adding to the gather list once #1328 lands.

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

(Closest fit: `perf:` per the conventional-commits matrix — patch-bump, user-facing.)

## Testing

`ruff` + `mypy` clean on the touched source file; 2643 unit tests pass, 1 musl-skip. New `TestGetSystemHealthGather` class with 5 cases (all-three / subset / raising-helper / `zha_network_full` route / no-include baseline).

## Future improvements

- After #1328 merges, add the `diagnostics` section (+ `diagnostics_fields` / `diagnostics_truncate_at_bytes` params) to the gather list.
- Real-world wall-clock measurement against a live HA install with ZHA + Z-Wave + dismissed repairs to quantify the win.

## Checklist

### DIFF
--- a/src/ha_mcp/tools/tools_system.py
+++ b/src/ha_mcp/tools/tools_system.py
@@ -9,6 +9,7 @@ This module provides tools for Home Assistant system administration including:
 
 import asyncio
 import logging
+from collections.abc import Coroutine
 from typing import Any
 
 from fastmcp.exceptions import ToolError
@@ -392,7 +393,7 @@ class SystemTools:
             want_zha = zha_full or zha_summary
             want_zwave = "zwave_network" in includes
 
-            sections: list[tuple[str, Any]] = []
+            sections: list[tuple[str, Coroutine[Any, Any, dict[str, Any]]]] = []
             if want_repairs:
                 sections.append(
                     (
@@ -414,13 +415,39 @@ class SystemTools:
 
             if sections:
                 gathered = await asyncio.gather(
-                    *(coro for _, coro in sections), return_exceptions=True
+                    *[coro for _, coro in sections], return_exceptions=True
                 )
-                for (section_name, _), section_result in zip(sections, gathered, strict=True):
-                    if isinstance(section_result, BaseException):
-                        # Mirrors the helpers' own embeddable-sub-dict contract
-                        # so an unexpected exception attributes to its section
-                        # instead of bubbling out and dropping the others.
+                # Pre-pass: re-raise anything that must unwind the request
+                # rather than land as an embedded section error. ``gather``
+                # with ``return_exceptions=True`` returns ``CancelledError``
+                # (and any other ``BaseException``) as a result element
+                # instead of propagating, and a ``ToolError`` raised from
+                # inside a helper would otherwise be silently demoted to
+                # ``{"error": "ToolError: …"}`` and break the MCP
+                # ``isError=true`` contract for the whole tool.
+                for section_result in gathered:
+                    if isinstance(section_result, asyncio.CancelledError):
+                        raise section_result
+                    if isinstance(section_result, ToolError):
+                        raise section_result
+                    if isinstance(section_result, BaseException) and not isinstance(
+                        section_result, Exception
+                    ):
+                        # ``KeyboardInterrupt`` / ``SystemExit`` — never demote
+                        # these to a section-level error string.
+                        raise section_result
+                for (section_name, _), section_result in zip(
+                    sections, gathered, strict=True
+                ):
+                    if isinstance(section_result, Exception):
+                        # Last-resort fallback: emit a minimal ``{error: ...}``
+                        # dict so an unexpected exception attributes to its
+                        # section instead of bubbling out and dropping siblings.
+                        # The helpers themselves return richer
+                        # ``{<key>: <baseline>, ..., error: ...}`` shapes on
+                        # their own (caught) failures; this branch is the
+                        # belt-and-suspenders path that fires only on a
+                        # helper-edit regression that lets an exception escape.
                         logger.warning(
                             "Concurrent fetch for section %r raised: %s: %s",
                             section_name,
--- a/src/ha_mcp/tools/tools_system.py
+++ b/src/ha_mcp/tools/tools_system.py
@@ -7,6 +7,7 @@ This module provides tools for Home Assistant system administration including:
 - System health monitoring
 """
 
+import asyncio
 import logging
 from typing import Any
 
@@ -376,22 +377,63 @@ class SystemTools:
             if unknown:
                 result["warning"] = f"Unknown include sections ignored: {', '.join(sorted(unknown))}"
 
-            # Fetch optional sections
-            if "repairs" in includes:
-                result["repairs"] = await self._fetch_repairs(
-                    ws_client,
-                    include_dismissed=include_dismissed_repairs_bool,
-                )
-
+            # Fetch optional sections concurrently. The ws_client serialises
+            # outgoing writes via its internal `_send_lock`, but per-message
+            # futures keyed by message_id let response waits overlap — so this
+            # gives request pipelining instead of head-of-line blocking.
+            #
+            # Each ``_fetch_*`` helper already returns an embeddable sub-dict
+            # with an ``error`` field on backend failure (it never raises);
+            # ``return_exceptions=True`` is belt-and-suspenders against a future
+            # helper edit that lets an exception escape.
             zha_full = "zha_network_full" in includes
             zha_summary = "zha_network" in includes
-            if zha_full or zha_summary:
-                result["zha_network"] = await self._fetch_zha_network(
-                    ws_client, full=zha_full
+            want_repairs = "repairs" in includes
+            want_zha = zha_full or zha_summary
+            want_zwave = "zwave_network" in includes
+
+            sections: list[tuple[str, Any]] = []
+            if want_repairs:
+                sections.append(
+                    (
+                        "repairs",
+                        self._fetch_repairs(
+                            ws_client,
+                            include_dismissed=include_dismissed_repairs_bool,
+                        ),
+                    )
+                )
+            if want_zha:
+                sections.append(
+                    ("zha_network", self._fetch_zha_network(ws_client, full=zha_full))
+                )
+            if want_zwave:
+                sections.append(
+                    ("zwave_network", self._fetch_zwave_network(ws_client))
                 )
 
-            if "zwave_network" in includes:
-                result["zwave_network"] = await self._fetch_zwave_network(ws_client)
+            if sections:
+                gathered = await asyncio.gather(
+                    *(coro for _, coro in sections), return_exceptions=True
+                )
+                for (section_name, _), section_result in zip(sections, gathered, strict=True):
+                    if isinstance(section_result, BaseException):
+                        # Mirrors the helpers' own embeddable-sub-dict contract
+                        # so an unexpected exception attributes to its section
+                        # instead of bubbling out and dropping the others.
+                        logger.warning(
+                            "Concurrent fetch for section %r raised: %s: %s",
+                            section_name,
+                            type(section_result).__name__,
+                            section_result,
+                        )
+                        result[section_name] = {
+                            "error": (
+                                f"{type(section_result).__name__}: {section_result}"
+                            )
+                        }
+                    else:
+                        result[section_name] = section_result
 
             return result
 
--- a/tests/src/unit/test_tools_system.py
+++ b/tests/src/unit/test_tools_system.py
@@ -4,6 +4,7 @@ Regression tests for https://github.com/homeassistant-ai/ha-mcp/issues/612
 ha_restart reports failure when a reverse proxy returns 504 during restart.
 """
 
+import asyncio
 from contextlib import contextmanager
 from unittest.mock import AsyncMock, MagicMock, patch
 
@@ -197,7 +198,7 @@ class TestGetSystemHealthGather:
         mock_zha = AsyncMock(return_value={"devices": [{"name": "A"}]})
         mock_zwave = AsyncMock(return_value={"nodes": []})
 
-        with _patch_health_info_baseline(), patch.object(
+        with _patch_health_info_baseline() as (_, ws_client), patch.object(
             SystemTools, "_fetch_repairs", new=mock_repairs
         ), patch.object(
             SystemTools, "_fetch_zha_network", new=mock_zha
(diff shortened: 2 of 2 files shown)


Patch76:
## Implementation Summary

**Choices Made:**

- **Branched from master, not from #1328's branch** — #1328 (the diagnostics feature that adds a fourth optional section + `diagnostics_fields` / `diagnostics_truncate_at_bytes` params) is still in `CHANGES_REQUESTED` after KP13's review. Stacking this PR on top would have coupled the two merge timelines. Branching from master keeps each iterable independently; once #1328 lands, adding `diagnostics` as a fourth `(name, coro)` tuple in the gather list is mechanical (~10 LOC, documented in PR body as a Future improvement).

- **`asyncio.gather(*coros, return_exceptions=True)` with `isinstance(r, BaseException)` per-result mapping**, not bare `gather`. Sibling pattern: `src/ha_mcp/tools/device_control.py:668` uses the same shape for bulk-device-control aggregation. Each `_fetch_*` helper already returns an embeddable sub-dict with an `error` field on backend failure (verified at `TestFetchRepairs::test_ws_exception_captured_as_error`), so the `return_exceptions=True` is belt-and-suspenders against a future helper edit that lets an exception escape. On hit, the affected section gets `{"error": "<Type>: <message>"}` and a structured `logger.warning(...)` carrying the section name — siblings still populate.

- **WS client concurrency verified live, not assumed.** Read `src/ha_mcp/client/websocket_client.py` to confirm:
  - `_send_lock: asyncio.Lock` serialises outgoing byte writes (no interleaving on the wire).
  - `_pending_requests: dict[message_id, Future]` is keyed by per-call message_id, so response waits can overlap.
  - Per-call `message_id` counter increments under the send_lock.

  Net effect: request pipelining (not raw parallelism) — outbound sends linearise, but the three response waits overlap. Wall-clock collapses from `t_repairs + t_zha + t_zwave` to ≈ `max(t_repairs, t_zha, t_zwave)`, which is the win the issue calls out.

- **`(name, coro)` tuple list + `zip(sections, gathered, strict=True)`** for the dispatch, not a flat positional gather. Three reasons: (a) only-requested-sections are built into the list — unrequested helpers stay un-awaited (verified by `test_unrequested_sections_not_called`); (b) `strict=True` on the zip guards against the section/result counts diverging silently if future-edit splits the list and the gather call; (c) per-section name attribution on raise is trivial.

- **Left `result["warning"]` singular alone** — the singular pattern at L377 is project-wide (~22 sites in 7 modules, tracked under #1332). Migrating it inside this PR would have entangled scope; the migration belongs to #1332's eventual PR.

- **First unit-test class for `ha_get_system_health` on master.** The existing `test_tools_system.py` covered only `_fetch_repairs` directly (`TestFetchRepairs`) and `ha_restart` error handling (`TestHaRestartErrorHandling`) — there was no test for the tool's top-level orchestration. The new `TestGetSystemHealthGather` adds five cases (all-three populated, subset, raising-helper-attribution, `zha_network_full` route, no-include baseline) and a reusable `_patch_health_info_baseline()` contextmanager that other future `ha_get_system_health` tests can build on.

- **`perf:` prefix, no scope.** Matches the conventional-commits matrix (patch bump, user-facing) at AGENTS.md L762. The change is structurally a refactor but the user-visible win is performance — `perf:` is the more accurate signal for changelog readers. No precedent for `perf:` in recent `git log` (last `perf:` is pre-`b7c4dd0`), so this PR sets the precedent shape.

**Problems Encountered:**

- **Issue body assumed #1328 had already landed.** The original wording cited `diagnostics` as one of the four sections to parallelise and `diagnostics_fields` / `diagnostics_truncate_at_bytes` as params to thread through. Re-checking master showed neither exists yet. Resolved by re-scoping to the three sections that are on master (repairs / zha / zwave) and documenting the post-#1328 mechanical re-merge in the PR body.

- **`_fetch_health_info` is private and not easily mockable from outside.** First test draft tried to instantiate `SystemTools` with a fully-faked client and call `ha_get_system_health` end-to-end — failed because `_fetch_health_info` calls `get_connected_ws_client(...)` against a real URL. Resolved by introducing a `_patch_health_info_baseline()` contextmanager that patches the private method directly. Slightly invasive on internals, but the same pattern is already used in the #1328-branch tests, so it's not introducing a new test convention.

- **`zip(sections, gathered, strict=True)` is Python 3.10+.** Project pin is `>=3.13,<3.14` (`pyproject.toml`), so this is safe — but worth a note for future backports if the floor ever drops.

- **After #1328 lands**, extend the `sections` list with a `("diagnostics", fetch_integration_diagnostics(...))` entry threading `diagnostics_fields` + `diagnostics_truncate_at_bytes`. Strictly mechanical — copy the existing tuple shape.

- **Real-world wall-clock measurement.** The win is structurally sound (request pipelining, verified ws_client concurrency primitives) but quantifying it against a live HA install with a real ZHA + Z-Wave + dismissed-repairs payload would close the loop. Left as a manual BAT-style verification — the testcontainers e2e runner doesn't replicate the slow-backend characteristics this targets.

- **Extend the pattern to other tools with multi-fetch orchestration**, if any. `ha_get_overview` is a candidate (multiple optional sections), as is `ha_get_integration` with its options + schema branches. Worth a separate scoping pass before opening — same `(name, coro)` shape would apply.

Closes #1331.

kingpanther13 (changes requested):
Concurrency claim and embeddable-error fallback both check out (verified `_send_lock` + `_pending_requests` futures in `client/websocket_client.py`, and all three `_fetch_*` helpers wrap their WS sends in `try/except Exception` on master). Two correctness issues and one over-claiming comment to address before this lands.

## Required changes

### 1. `isinstance(section_result, BaseException)` swallows `CancelledError`

`asyncio.gather(return_exceptions=True)` returns `CancelledError` as a result element rather than re-raising it. The current `BaseException` branch converts it to `{"error": "CancelledError: "}` and the tool returns `success=True` instead of unwinding. Same hazard for `KeyboardInterrupt` / `SystemExit`.

Narrow to `Exception`, or re-raise the BaseException-but-not-Exception case explicitly:

```python
if isinstance(section_result, asyncio.CancelledError):
    raise section_result
if isinstance(section_result, Exception):
    ...
```

### 2. A captured `ToolError` is silently demoted to a dict

The outer `except ToolError: raise` only fires for `_fetch_health_info` today. Any future `ToolError` raised from inside a helper (or any existing `raise_tool_error` call that ends up reachable from the helpers) is captured by `gather` and stuffed into `result[section_name] = {"error": "ToolError: …"}` — `success=True` returned to the caller, violating the MCP `isError=true` contract.

Add a pre-pass before populating `result`:

```python
for _, sr in zip(sections, gathered, strict=True):
    if isinstance(sr, ToolError):
        raise sr
```

### 3. The "mirrors the helpers' own embeddable-sub-dict contract" comment overpromises

The helpers return populated baselines on failure:

- `_fetch_repairs` → `{"issues": [], "count": 0, "error": "…"}`
- `_fetch_zha_network` → `{"devices": [], "count": 0, "total_count": 0, "error": "…"}`
- `_fetch_zwave_network` → `{"controller": {}, "nodes": [], "count": 0, "total_count": 0, "error": "…"}`

The fallback emits only `{"error": "RuntimeError: …"}` — no baseline scaffold. Downstream consumers expecting `repairs["count"]` or `zha_network["devices"]` would `KeyError` if this path ever fires.

Either soften the comment ("emits a minimal `{error: …}` last-resort dict; helpers themselves return richer baseline+error dicts") or build a per-section skeleton in the fallback so the contract really matches.

## While you're in there

### 4. No test proves the gather actually runs in parallel

`TestGetSystemHealthGather` would pass equally well if the code reverted to serial `await`s. Worth one test that gives two helpers an `asyncio.Event` (helper A waits on the event; helper B sets it) and asserts both can be in-flight simultaneously — otherwise the perf property has zero regression coverage.

### 5. Mixed-failure ordering only covers "first position fails"

`test_one_section_raising_does_not_block_siblings` raises on `repairs` (position 0). Parametrize over which section raises so a future zip/mis-attribution regression gets caught regardless of position.

## Nits (optional)

- `list[tuple[str, Any]]` for `sections` — the second slot is a coroutine; `Coroutine[Any, Any, dict[str, Any]]` is more precise.
- Generator into `gather(*(…))` is correct but a list-comp reads more obviously given that `zip(sections, gathered, strict=True)` is the next line.
- `ws_client.disconnect.assert_awaited_once()` in one existing test would catch a future regression where the gather path skips `finally` cleanup.

Patch76:
All 5 main items + 3 nits addressed in `85441b5`. Per-item below.

## Required changes

**1. `CancelledError` no longer swallowed** — `tools_system.py:415-441`. Added a pre-pass before the section-populating loop that explicitly re-raises any `asyncio.CancelledError`, any `ToolError`, and any `BaseException` that is not an `Exception` (covers `KeyboardInterrupt` / `SystemExit`). The Exception-branch in the populating loop is now narrowed to `isinstance(section_result, Exception)` so anything that needs to unwind the request actually does. New `test_cancelled_error_propagated_not_demoted` pins the behaviour.

**2. Captured `ToolError` re-raised, not demoted** — same pre-pass at `tools_system.py:431-432`. Any future `ToolError` raised from inside a helper (or anything reachable from the helpers) now propagates out of `ha_get_system_health` rather than landing as `result[section_name] = {"error": "ToolError: …"}` with `success=True`. New `test_tool_error_from_helper_re_raised_not_demoted` pins the behaviour by passing in the exact `ToolError` instance and asserting it comes back out of `pytest.raises(...)`.

**3. Last-resort fallback comment softened** — `tools_system.py:443-451`. Replaced the over-claiming "mirrors the helpers' own embeddable-sub-dict contract" framing with: "emits a minimal `{error: …}` last-resort dict; helpers themselves return richer `{<key>: <baseline>, ..., error: ...}` shapes on their own (caught) failures; this branch is the belt-and-suspenders path that fires only on a helper-edit regression that lets an exception escape." Kept the fallback shape minimal rather than building per-section skeletons — the helpers' own try/except paths are the contract-correct emitter and this branch is unreachable on master.

## While you're in there

**4. Parallel execution actually proven** — new `test_helpers_actually_run_concurrently`. Helper A waits on an `asyncio.Event` with a 2s timeout; helper B sets the event. Serial regression would deadlock A (waiting for an event B has no chance to set, since gather would not have started B yet) and the timeout fires the test. Parallel path completes both in microseconds. A revert to serial `await`s would fail this test fast.

**5. Mixed-failure ordering parametrized** — `test_one_section_raising_does_not_block_siblings` is now `@pytest.mark.parametrize("raising_section", ["repairs", "zha_network", "zwave_network"])`. Each variant raises in a different position in the sections list and asserts both (a) the raising section's error attributes to its own name (zip alignment) and (b) the two siblings carry their full success payloads. A future zip/mis-attribution regression now fails at the position where the bug surfaces.

## Nits

**6. `list[tuple[str, Coroutine[Any, Any, dict[str, Any]]]]`** — `tools_system.py:395`. Added `from collections.abc import Coroutine` and tightened the second slot from `Any`. More precise about what's actually being collected.

**7. Generator → list-comp** — `tools_system.py:417` now reads `*[coro for _, coro in sections]` (was `*(coro for _, coro in sections)`). Symmetric with the `zip(sections, gathered, strict=True)` on the next line.

**8. `ws_client.disconnect.assert_awaited_once()`** — added to `test_all_three_sections_populated_when_all_requested` via the `as (_, ws_client)` unpacking on `_patch_health_info_baseline()`. A future regression where the gather path skips the `finally` cleanup (e.g. a rewrite that returns early before the `disconnect()`) now fails this test.

---

Local pytest: 18/18 pass on `tests/src/unit/test_tools_system.py`. Ruff clean.

kingpanther13 (approved):
All 5 main items + 3 nits addressed in 85441b5. Pre-pass cleanly separates re-raise (Cancelled/ToolError/non-Exception BaseException) from demote (Exception); concurrency test uses an asyncio.Event with timeout so a serial regression fails fast; parametrized raising_section catches zip mis-attribution at any position. LGTM.

Issue #1194, opened 2026-05-09
[FEATURE] Client-level lovelace/dashboards/list cache with lovelace_updated invalidation (Option 4 from #1085)


Patch76:
## What would you like to see?

Add a client-level cache for `lovelace/dashboards/list` results, invalidated by HA's `lovelace_updated` event subscription.

Shape sketch:

- The MCP WebSocket client subscribes to `lovelace_updated` once at startup.
- A simple `client.dashboards_list()` accessor returns the cached list (TTL'd or event-invalidated). Cache miss + event-fired both trigger a fresh fetch.
- All current `lovelace/dashboards/list` call sites (and any future ones) route through the accessor instead of issuing the WS call directly.

This was Option 4 in #1085's design discussion — explicitly deferred there because the lifecycle and out-of-scope concerns make it a bigger conversation than the round-trip dedup #1085 closes.

## Why do you need this?

PR #1191 (closes #1085) deduplicated `ha_config_set_dashboard`'s internal-id branch from 2 → 1 round-trip via in-call list-passthrough. That helps one tool, one branch. Four other places still each fetch the list independently:

- `_resolve_dashboard` — `tools_config_dashboards.py:300-315`
- `ha_config_set_dashboard` existence-check fallback — `tools_config_dashboards.py:1162-1180`
- `ha_config_get_dashboard(list_only=True)` — `tools_config_dashboards.py:531-540`
- YAML-mode collision check — `tools_yaml_config.py:48-50`

A consecutive `ha_config_set_dashboard` → `ha_config_set_dashboard` → `ha_config_get_dashboard(list_only=True)` sequence pays three round-trips for what is, almost always, the same registry snapshot. A client-level cache cuts cross-tool redundancy that no in-call passthrough can reach.

## Additional context

Open questions worth resolving before implementation:

- **Cache lifecycle:** TTL only (5–10 s) vs. `lovelace_updated`-event-invalidated vs. both. Event subscription is the cleanest signal but only catches HA-managed changes — out-of-band edits (manual `.storage/lovelace_dashboards` edits, container restart with a different snapshot) won't fire it.
- **Stale-read tolerance:** does `ha_config_set_dashboard`'s existence-check tolerate a slightly stale list (and create a duplicate dashboard), or should writes always force-fetch? Probably the latter; reads can use the cache.
[...]

PR #1344, opened 2026-05-17
refactor(internal): consolidate lovelace/dashboards/list through shared helper

Patch76:
## What does this PR do?

Routes the two remaining raw `lovelace/dashboards/list` call sites through the existing `fetch_dashboards_list` helper:

- `tools_yaml_config.py:49` (`_check_storage_mode_dashboard_collision`)
- `smart_search.py:1778` (`ha_deep_search` dashboard branch)

Renames `_fetch_dashboards_list` → `fetch_dashboards_list` since it's now a cross-module helper (3 internal callers in `tools_config_dashboards.py` at L383/L600/L1216 + 1 docstring in `test_config_get_dashboard_search_error.py:119` updated).

Behaviour note: `smart_search`'s previous `.get("success")` gate silently treated bare-list HA responses as empty; the helper accepts both dict-with-`result`-key and bare-list shapes, so the branch now correctly populates from either.

Refs #1194 (Option B from [comment 4468552918](https://github.com/homeassistant-ai/ha-mcp/issues/1194#issuecomment-4468552918)). The client-level cache layer (the issue's actual subject) stays open in #1194 for separate decision.

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [x] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

No new tests — `test_yaml_config_tool.py` (5 collision-check scenarios) and `test_config_get_dashboard_search_error.py` mock `send_websocket_message` directly, so existing fixtures flow through the helper routing unchanged. Full local unit suite: 2641 passed, 1 skipped, 6 deselected.

## Future improvements

- Client-level cache layer (the actual subject of #1194) — out of scope; 4 open design questions still apply.
- `ha_deep_search` dashboard branch has no direct unit test — focused test exercising the bare-list path would close a real gap.

## Checklist

### DIFF
--- a/src/ha_mcp/tools/smart_search.py
+++ b/src/ha_mcp/tools/smart_search.py
@@ -22,6 +22,7 @@ from ..utils.fuzzy_search import (
     tokenize,
 )
 from .helpers import exception_to_structured_error, safe_info, safe_progress
+from .tools_config_dashboards import fetch_dashboards_list
 
 logger = logging.getLogger(__name__)
 
@@ -1773,14 +1774,9 @@ class SmartSearchTools:
             if "dashboard" in search_types:
                 try:
                     # List all storage-mode dashboards
-                    dash_list_resp = await self.client.send_websocket_message(
-                        {"type": "lovelace/dashboards/list"}
+                    dashboard_entries: list[dict[str, Any]] = (
+                        await fetch_dashboards_list(self.client) or []
                     )
-                    dashboard_entries: list[dict[str, Any]] = []
-                    if isinstance(dash_list_resp, dict) and dash_list_resp.get(
-                        "success"
-                    ):
-                        dashboard_entries = dash_list_resp.get("result", [])
 
                     # Build list of dashboards to search (include default)
                     dashboards_to_search: list[tuple[str, str]] = [
--- a/src/ha_mcp/tools/tools_config_dashboards.py
+++ b/src/ha_mcp/tools/tools_config_dashboards.py
@@ -322,7 +322,7 @@ def _should_lazy_resolve(error_msg: str) -> bool:
     return _LAZY_RESOLVE_TRIGGER in error_msg
 
 
-async def _fetch_dashboards_list(
+async def fetch_dashboards_list(
     client: Any,
 ) -> list[dict[str, Any]] | None:
     """Fetch and normalise the lovelace/dashboards/list WebSocket response.
@@ -380,7 +380,7 @@ async def _resolve_dashboard(
       to the registry id before issuing the delete. Discards
       ``dashboards``.
     """
-    dashboards = await _fetch_dashboards_list(client)
+    dashboards = await fetch_dashboards_list(client)
     if dashboards is None:
         return None, None
 
@@ -597,7 +597,7 @@ def register_config_dashboard_tools(mcp: Any, client: Any, **kwargs: Any) -> Non
         try:
             # List mode
             if list_only:
-                dashboards = await _fetch_dashboards_list(client) or []
+                dashboards = await fetch_dashboards_list(client) or []
                 return {
                     "success": True,
                     "action": "list",
@@ -1213,9 +1213,7 @@ def register_config_dashboard_tools(mcp: Any, client: Any, **kwargs: Any) -> Non
             if pre_fetched_dashboards is not None:
                 existing_dashboards = pre_fetched_dashboards
             else:
-                existing_dashboards = (
-                    await _fetch_dashboards_list(client) or []
-                )
+                existing_dashboards = await fetch_dashboards_list(client) or []
             dashboard_exists = any(
                 d.get("url_path") == url_path for d in existing_dashboards
             )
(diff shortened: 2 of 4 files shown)


Patch76:
## Implementation Summary

**Choices Made:**

- **Option B from #1194 comment [4468552918](https://github.com/homeassistant-ai/ha-mcp/issues/1194#issuecomment-4468552918)** — land the helper-routing consolidation now, keep the issue open for the actual client-level cache decision. The 4 design questions enumerated in #1194's body (cache lifecycle / force-fetch on writes / multi-instance drift / tools↔client coupling) remain unresolved and need their own conversation; the consolidation is unambiguously good and unblocks any future cache decision by giving every consumer a single entry point.

- **Renamed `_fetch_dashboards_list` → `fetch_dashboards_list`** (Boy-Scout). The underscore prefix was correct when the helper was module-internal (introduced by #1207). With cross-module consumers added in this PR, the prefix becomes misleading — Python convention for `_`-prefixed names is "module-internal API, don't import elsewhere". Renaming makes the public/private contract honest. Four conditions for Boy-Scout inclusion all met: (a) same file family — touches the helper's own module plus the 2 new consumer modules, (b) one-line per site, (c) sibling-match — aligns with other public helpers in `tools_config_dashboards.py`, (d) same anti-pattern as live edit — the rename is a direct consequence of the scope expansion this PR introduces.

- **Dropped `success`-key gate in `smart_search.py:1778`.** The old code required `isinstance(dash_list_resp, dict) and dash_list_resp.get("success")` before populating dashboard_entries. This was over-defensive — `send_websocket_message` already filters protocol-level failures, and the gate silently treated bare-list HA responses (a documented valid shape per the helper's docstring) as empty. Routing through the helper restores correct handling of both shapes. Net: a hidden bug fix; not flagged as a fix-prefix commit because it's a side effect of the consolidation, not the primary delta.

- **Kept outer try/except wrapper in `tools_yaml_config.py`** around the helper call. The helper itself does not catch exceptions — by design, callers decide the policy. The yaml-config collision check has a fail-open contract on WS failures (per its docstring: "A WebSocket failure or unexpected response shape warns and skips the check"), so the wrapper stays.

- **No new tests added.** Existing coverage already exercises the refactored paths: `test_yaml_config_tool.py` has 5 scenarios that mock `client.send_websocket_message` directly (storage-collision blocks, no-collision dispatches, non-dashboard path skips, WS failure skips, bare-list response triggers collision). Those mocks intercept calls regardless of whether the inline code or the helper makes them, so they continue to verify behaviour. `test_config_get_dashboard_search_error.py` covers the helper's unexpected-shape warn-and-return-None path. Adding redundant tests for the new routing would be cargo-culting, not real coverage.

**Problems Encountered:**

- **No production blockers.** Ruff check clean on all 4 touched files; mypy clean on the 3 src files; full local unit suite green (2641 passed, 1 expected numpy skip, 6 slow-deselects).

- **Ruff format diff on `tools_config_dashboards.py:L1213-1217` was a direct consequence of the rename.** Shortening `_fetch_dashboards_list` → `fetch_dashboards_list` brought the call site within 88 chars, collapsing what was previously a multi-line wrap into a single line. Applied the format change since it stems from this PR's edit. Pre-existing format debt on untouched lines in `tools_yaml_config.py` and `smart_search.py` left in place — per repo norm (ruff format is not in `pr.yml`'s required checks list; #1318 is the canonical format-sweep tracker).

- **Live-state drift since #1194 comment 2 (2026-05-11).** Comment 2 listed `_fetch_dashboards_list` at L271-281; the helper has since moved to L325 (likely via the #1207 merge that introduced it formally). Comment 3 (2026-05-16) re-anchored to L325. I re-verified the live line numbers before editing — the helper is now at L325 in current master.

**Suggested Improvements (Optional Follow-ups):**

- **Client-level cache layer** — the actual subject of #1194. The 4 design questions in the issue body still apply: cache lifecycle (TTL only / `lovelace_updated` event-invalidated / both), force-fetch on writes vs. stale-tolerant reads, multi-instance drift acceptance, tools↔client coupling contract. Comment 3's mild lean was toward Option B (this PR + issue stays open for that decision); whether the cache layer adds enough on top of the now-completed consolidation to be worth the design work is the next call.

- **Smart_search dashboard-branch direct test coverage** — `ha_deep_search` with `search_types=["dashboard"]` has no focused unit test. Adding one to pin the bare-list-shape handling the helper now correctly supports would close a real gap. Fixture surface is large (`SmartSearchTools` + full HA-client mock + tier-3 dependencies), so a separate PR.

- **Helper relocation question** — `fetch_dashboards_list` lives in `tools_config_dashboards.py` and is now imported by `tools_yaml_config.py` and `smart_search.py`. The cross-module imports work fine (no circular-import risk; both files already do cross-module imports) but architecturally the helper might fit better in a shared `dashboards_helpers.py` or in `util_helpers.py`. Over-engineering for one function — but worth noting if more cross-module dashboards utilities accumulate.

kingpanther13 (approved):

PR #1387, opened 2026-05-20
ci: add ruff format --check on changed Python files

Patch76:
## What does this PR do?

Adds a new step to `pr.yml`'s Ruff Lint job that runs `uv run ruff format --check` only on the `.py` files touched by the PR (diffed against `$GITHUB_BASE_REF`). Skips cleanly when no Python files are changed, and on `workflow_dispatch` when no base ref is set.

**Maintainer-chosen scope** (per [@kingpanther13's 2026-05-20 reply on #1318](https://github.com/homeassistant-ai/ha-mcp/issues/1318#issuecomment-4498146534)): gate on changed files only, no repo-wide sweep. Pre-existing format-debt across the tree stays grandfathered; new edits must pass `ruff format --check`. Closes the loop on the three deferred format-debt comments (#1254, #1258, #1303) without the disruption a sweep would impose on the open PRs currently in flight against master.

This PR itself touches only `pr.yml` (no `.py` files) so the new step exercises its own no-Python-changed exit path during validation. The actual `ruff format --check` codepath is end-to-end-verified on the first subsequent PR that touches a `.py` file.

Closes #1318

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [x] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing
- [ ] I have tested these changes with a LLM agent
- [x] All automated tests pass (`uv run pytest`)
- [x] Code follows style guidelines (`uv run ruff check`)

## Checklist
- [ ] I have updated documentation if needed

PR #1386, opened 2026-05-20
refactor: align dashboards 404 shape with sibling config tools

Patch76:
## What does this PR do?

Replaces the last `create_resource_not_found_error` call site in `tools_config_dashboards.py:1536` (the synthesised 404 path inside `ha_config_delete_dashboard` when `_resolve_dashboard` returns `None`) with a direct `create_error_response(RESOURCE_NOT_FOUND, …, context={"action": "delete", "url_path": url_path})`. Drops the now-unused `create_resource_not_found_error` import.

After #1289 cleaned out the analogous call sites in `tools_config_automations.py:318` and `:1010`, the four config tool families (automations / scripts / scenes / dashboards) were 3-vs-1 on 404-helper choice — three siblings already passed per-tool `context` keys (`identifier` / `script_id` / `scene_id`) through `create_error_response`, while dashboards alone still hard-coded `resource_type` + `identifier` via the per-helper path. This PR converges dashboards to the sibling shape (Option B in the issue analysis).

The new top-level shape (`action` + `url_path`) also matches the idempotent already-deleted success branch on the same function, so the success and failure shapes within `ha_config_delete_dashboard` are now self-consistent.

Adds `TestDeleteDashboardNotFoundShape` (1 test) pinning the new shape: `RESOURCE_NOT_FOUND` code, `action="delete"`, `url_path=<input>`, and explicit absence of the dropped `resource_type` / `identifier` top-level keys.

Closes #1300

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [x] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing
- [ ] I have tested these changes with a LLM agent
- [x] All automated tests pass (`uv run pytest`)
- [x] Code follows style guidelines (`uv run ruff check`)

## Checklist
- [ ] I have updated documentation if needed

Issue #1389, opened 2026-05-20
[FEATURE] Measure HA Core entity-registration p50/p99 to validate _poll_for_automation_entity 0.1s floor


Patch76:
> **Status (2026-05-22)**: ~~Blocked on PR #1384~~ — #1384 merged at `facd7af`; the post-merge `_POLL_CADENCE = (0.1, 1.0, 4.9)` shape now lives in master. PR #1398 carries both halves of the decision rule: the measurement infrastructure AND the `_POLL_CADENCE = (0.025, 1.0, 4.975)` retune the data triggered (same 6.0 s total budget, first-poll pulled from 100 ms to 25 ms). Captured data: p50 ≈ 104 ms across two `E2E Validation` jobs → **`RETUNE NEEDED`** verdict; see [measurement comment](https://github.com/homeassistant-ai/ha-mcp/issues/1389#issuecomment-4509091450). This issue auto-closes when #1398 merges.

### What would you like to see?

A short instrumentation pass that measures HA Core's actual automation-entity-registration latency (p50/p99 across N=10 runs) **with a pre-committed decision rule** for what to do with the data, so the work has a defined endpoint rather than evaporating as an open issue.

**Concrete benefit (the part this issue commits to):**

After collecting the data, this issue resolves down ONE of two terminal paths — no open-ended "we measured it" state:

- **p50 < 100 ms AND p99 < 1.0 s** → cadence is empirically well-tuned. Close this issue with the measurement table as the validation record. No code change.
- **p50 ≥ 100 ms OR p99 ≥ 1.0 s** → open a follow-up PR retuning `_POLL_CADENCE` against the data (e.g. shift the first-poll position, redistribute the second-poll budget). The retune ships with the measurement table as its empirical justification.

Either outcome is concrete and visible.

**Sequencing:**

This is blocked on PR #1384 landing — the issue validates the new `_POLL_CADENCE = (0.1, 1.0, 4.9)` shape, which doesn't exist in master yet (`rest_client.py:929` still uses `1 * (attempt + 1)`).

**Proposed protocol:**

1. Add a `logger.debug("entity-registration-elapsed: %.1fms", elapsed_ms)` line in `_poll_for_automation_entity` measuring time between POST return and the first successful `get_states()` hit on the new entity.
[...]

PR #1397, opened 2026-05-21
fix: error-shape consistency for non-entity not-found (closes #1297)

### LINKED ISSUE #1297: [FEATURE] Error-shape consistency across CRUD tool families (follow-up to #521)

A canonical error-shape contract for CRUD-style tool families (`tools_config_{automations,scripts,scenes,dashboards,helpers}.py`, `tools_areas.py`, `tools_labels.py`, `tools_categories.py`), specifically converging on one shape per error class (404, 400, validation, etc.) so an agent caller doesn't have to branch on `family` to parse the response.

Follow-up to #521 (closed; suggested fix: "Audit remaining tool modules for any other ad-hoc error dict construction and standardize on the `errors.py` utilities."). That audit surfaces four distinct drift dimensions, each verified live against my RBO Home Assistant instance today (pure read-only GETs of nonexistent identifiers, no mutations).

#### Dimension 1 — 404 classification differs by family

Calling each `ha_config_get_<family>` with a nonexistent ID and capturing the structured-error response:

| Family | `error.code` | Suggestions | Top-level extras |
|---|---|---|---|
| `automations` | `RESOURCE_NOT_FOUND` | **0** | `resource_type, identifier, action, reason` |
| `scripts` | `RESOURCE_NOT_FOUND` | 3 | `script_id` |
| `dashboards` | **`SERVICE_CALL_FAILED`** ⚠️ | 3 | `action, url_path` |
| `labels` | `ENTITY_NOT_FOUND` ⚠️ | 1 | `label_id, available_label_ids` |
| `categories` | `ENTITY_NOT_FOUND` ⚠️ | 1 | `category_id, scope, available_category_ids` |
| `scenes` (verified via unit test + mock probe; no read-tool registered in my setup) | `ENTITY_NOT_FOUND` | 3 | `entity_id` |

Two observations stand out:

**(a) `dashboards` 404 is misclassified as `SERVICE_CALL_FAILED`.** HA's WS bridge returns `"Command failed: Unknown config specified: <url_path>"`. The string contains neither `"404"` nor `"not found"`, so the classifier in `src/ha_mcp/tools/helpers.py:L174-208` falls past the resource-not-found branch and lands in the `command failed:` SERVICE_CALL_FAILED fallback. The 404 is structurally identifiable but lost.

**(b) `labels` and `categories` use `ENTITY_NOT_FOUND`** at `src/ha_mcp/tools/tools_labels.py:L102-107` and `src/ha_mcp/tools/tools_categories.py:L121-134`. Labels and categories aren't entities — they're registry metadata — and the auto-injected `create_entity_not_found_error` suggestions ("Use `ha_search_entities()` to find correct entity ID") point agent retry logic at the wrong tool. `RESOURCE_NOT_FOUND` (which exists and would route correctly) is the right code for non-entity registry objects.

#### Dimension 2 — 404 detection mechanism differs

Three mechanisms coexist:

- **Manual string-match + `create_resource_not_found_error`** (`tools_config_automations.py:L307-325`, `:L1009-1029`; `tools_config_dashboards.py:L1479-1488`)
- **`exception_to_structured_error` with `context={"entity_id": ...}`** (`tools_config_scenes.py:L267-283`) — lets the classifier pick `ENTITY_NOT_FOUND` for entity-domain resources
- **`exception_to_structured_error` plain** (`tools_config_scripts.py:L562-580`; `tools_areas.py`; flow-helper path in `tools_config_helpers.py`) — produces `RESOURCE_NOT_FOUND` with no entity-context upgrade

The classifier at `src/ha_mcp/tools/helpers.py:L115-128` (status_code switch) and `:L174-234` (message-pattern fallback, `_classify_by_message`) already does the work; the manual string-match blocks in automations/dashboards duplicate it with subtle differences (extra `resource_type`+`reason` fields, custom message templates).

#### Dimension 3 — Pre-flight identifier validation only in scenes

Live-tested whitespace-only identifier `" "` against `ha_config_get_automation`, `ha_config_get_script`, `ha_config_get_label`, `ha_config_get_category`: all four passed the input through to HA, made an unnecessary round-trip, and returned `"<Family> not found:  "` messages. Only `tools_config_scenes.py:L225, L524, L946` carry the explicit pre-flight guard (added in #1168 R6).

(Already filed as #1294 for the family-wide direction decision — relevant here only as confirmation that the drift is observable across the same surface.)

#### Dimension 4 — `available_*_ids` only on labels/categories

`tools_labels.py:L101` and `tools_categories.py:L120` slice the first 10 valid IDs into the response top-level when the lookup misses. This is the best agent-recovery surface in the family — the caller can self-correct without a second list-call. Other CRUD families (`automations`, `scripts`, `dashboards`) don't expose this even though the data is cheap to surface.

---

### Why do you need this?

Each dimension has a concrete agent-UX cost:

- **D1 dashboards misclassification:** agent retry logic branching on `RESOURCE_NOT_FOUND` doesn't fire for dashboards. They get `SERVICE_CALL_FAILED` and fall into a generic-failure handler instead of the targeted "look up valid IDs and retry" path.
- **D1 labels/categories miscoding:** auto-injected suggestions point at `ha_search_entities`, which doesn't list labels or categories. Agent follows the suggestion → empty result → wrong-tool spiral.
- **D1 automations 404:** zero suggestions in the structured response — the agent has no actionable next-step beyond reading the message. The Pattern-A `create_resource_not_found_error` path on `:L318` doesn't inject suggestions like the `exception_to_structured_error` path does.
- **D2 manual string-match:** four sites in automations + one in dashboards duplicate logic that lives canonically in `helpers.py`. Adding a sixth family or refining the classifier requires patches at all the duplicate sites instead of one.
- **D3 whitespace passthrough:** unnecessary HA round-trip per call, plus user-irritating `" not found: "` messages with the literal whitespace echoed.
- **D4 missing `available_*_ids`:** would be free-to-add on tools that already fetch lists for cache-validation, and gives small models a clean recovery handle.

The styleguide already locks the principle (`.gemini/styleguide.md:L86-99`: "All error handling MUST use structured errors from `errors.py`. Flag HIGH severity if errors use plain exceptions or dict returns") — the drift here is *between* legitimate uses of those utilities, not against them. Per the same styleguide (`L172-179`), response-shape unification is explicitly endorsed as non-breaking and encouraged.

---

### Additional context

**Live-test methodology.** All claims above verified against the current master tree (no commits ahead of `upstream/master` at the time of writing). Two probe types:

1. **Pure error-builder probe** — imported `exception_to_structured_error`, `create_resource_not_found_error`, `create_entity_not_found_error` directly; fed each synthetic `HomeAssistantAPIError` (status 400/401/403/404/500), generic `Exception`, `TimeoutError`, etc. with and without `entity_id` in context; captured the classifier's output. Confirmed: `entity_id` in context is the only differentiating signal at 404; other status codes are context-invariant.

2. **Live RBO HA GET probe** — called each `ha_config_get_<family>` with `"<family>.scope_c_probe_nonexistent_xyz_2026_05_15"` and then with whitespace `" "`. Read-only; no entities created, no backup needed. Results in the tables above.

3. **Mock E2E probe** — instantiated `AutomationConfigTools`/`ConfigScriptTools`/`ConfigSceneTools` with `AsyncMock(side_effect=HomeAssistantAPIError(..., status_code=404))` on the underlying client; verified the public-method output shape matches the live RBO result for the families I could test live, confirming mock fidelity.

**Decision points (one direction per dimension):**

- **D1 dashboards classification:** extend the classifier's not-found branch at `helpers.py:L203` to include `"unknown config specified"`, **or** keep dashboards' current manual string-match block at `tools_config_dashboards.py:L1479-1488` as the canonical fix site. Either works; one location is cleaner.
- **D1 labels/categories code choice:** swap `ENTITY_NOT_FOUND` → `RESOURCE_NOT_FOUND` at `tools_labels.py:L103` and `tools_categories.py:L123`, **or** add label-/category-specific suggestions to override the auto-injected entity ones. First option is one-line per file and semantically clean.
- **D2 mechanism convergence:** drop Pattern-A manual blocks in `automations.py` (4 sites) + `dashboards.py` (1 site); rely on `exception_to_structured_error` + pass `context={"entity_id": ...}` for entity-domain families. Pattern B becomes canonical.
- **D3 pre-flight validation:** already deferred to #1294 — direction TBD there.
- **D4 `available_*_ids` surface:** opt-in expansion; cost is one list-fetch per get-call (already cached for some clients). Not a blocker for D1-D3.

**Implementation effort, assuming Pattern B becomes canonical:**

| Track | Files | LOC | Tests |
|---|---|---|---|
| D1 dashboards fix | `helpers.py` + 1 `tests/src/unit/test_helpers.py` test | ~3 + 15 | regression |
| D1 labels/categories code-swap | `tools_labels.py:L103`, `tools_categories.py:L123` | ~2 | regression × 2 |
| D2 strip Pattern A | `tools_config_automations.py` (2 fns), `tools_config_dashboards.py` (1 fn) | ~30 removed, ~15 added (entity_id context) | new + existing rest_client tests parallel to `TestSceneRestClientErrorMapping` for automations and scripts |
| D4 (optional) | as scoped | — | — |

Total ~50 LOC plus a handful of regression tests, splittable into 2-3 PRs by dimension.

**Out of scope for this issue:** `tools_config_helpers.py` response-shape inconsistency is filed separately as #1293; the `helpers.py` flow-error helper (Pattern D, `_raise_flow_api_error`) is a distinct contract worth keeping; `tools_code.py` manual error dicts are sandbox-runtime contract (deliberately not MCP error shape, documented at `tools_code.py:L632-643`).

Patch76:
## What does this PR do?

Closes #1297.

Reclassifies five non-entity not-found responses from `ENTITY_NOT_FOUND` to `RESOURCE_NOT_FOUND`. After the related #1386 (D2 dashboards/resources alignment) and #1345 (D1 dashboards classifier) landed, this PR finishes the original D1 audit by covering the remaining non-entity sites plus a zones sibling discovered during self-review.

**Why it matters**: an agent branching on `error.code == "ENTITY_NOT_FOUND"` retries via `ha_search_entities()` — which doesn't list labels, categories, devices, or zones (when addressed by registry-internal ID), creating a wrong-tool spiral. The explicit per-site suggestion (e.g. *"Use ha_config_get_label() without label_id..."*) was always correct; only the code was mis-routing structured retry logic.

**Sites changed** (all 1-line `ErrorCode` swaps):

| Tool | File:Line | Why non-entity |
|---|---|---|
| `ha_config_get_label` | `tools_labels.py:123` | Labels live in the label registry, addressed by `label_id` |
| `ha_config_get_category` | `tools_categories.py:140` | Categories live in the category registry, scoped + addressed by `category_id` |
| `ha_get_device` | `tools_registry.py:379` | Devices live in the device registry, addressed by `device_id` UUID (not entity_id) |
| `ha_remove_device` | `tools_registry.py:746` | Same — sibling-bug found during the cross-family audit |
| `ha_get_zone` | `tools_zones.py:100` | Zones are addressed here by registry-internal `zone_id` (not by `zone.<name>` entity_id) — Boy-Scout fold |

**Style commit**: `ruff format` on the files touched by the fix, forced by the new `ruff format --check` gate added in #1387. Cross-link #1318 (per-directory ruff format sweep tracker).

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

5 new regression tests in `tests/src/unit/test_error_code_consistency_1297.py` pin the `RESOURCE_NOT_FOUND` code per site. Three E2E tests pinning the prior `ENTITY_NOT_FOUND` updated to match; zones E2E test docstring + assertion updated alongside the zones fold. Broad-unit-suite clean locally.

## Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -705,7 +705,7 @@ results.append(create_error_response(
 
 Only use `raise_error=False` on `exception_to_structured_error` when you need to mutate the dict before raising. Never add `add_timezone_metadata` to errors.
 
-`exception_to_structured_error` auto-classifies 404s, auth errors, timeouts by exception type. Pass `context={"entity_id": ...}` for automatic `ENTITY_NOT_FOUND` on 404s. Available helpers: `create_entity_not_found_error`, `create_connection_error`, `create_auth_error`, `create_service_error`, `create_validation_error`, `create_config_error`, `create_timeout_error`, `create_resource_not_found_error`, `create_error_response`.
+`exception_to_structured_error` auto-classifies 404s, auth errors, timeouts by exception type. Pass `context={"entity_id": ...}` for automatic `ENTITY_NOT_FOUND` on 404s. Available helpers: `create_entity_not_found_error`, `create_connection_error`, `create_auth_error`, `create_service_error`, `create_validation_error`, `create_config_error`, `create_timeout_error`, `create_error_response`.
 
 ### Return Values
 ```python
--- a/rules/no-return-error-response.yml
+++ b/rules/no-return-error-response.yml
@@ -11,7 +11,6 @@ rule:
     - pattern: return create_validation_error($$$)
     - pattern: return create_config_error($$$)
     - pattern: return create_timeout_error($$$)
-    - pattern: return create_resource_not_found_error($$$)
   not:
     inside:
       kind: function_definition
--- a/src/ha_mcp/tools/tools_categories.py
+++ b/src/ha_mcp/tools/tools_categories.py
@@ -133,7 +133,7 @@ class CategoryTools:
                 available_ids = [cat.get("category_id") for cat in categories[:10]]
                 raise_tool_error(
                     create_error_response(
-                        ErrorCode.ENTITY_NOT_FOUND,
+                        ErrorCode.RESOURCE_NOT_FOUND,
                         f"Category not found: {category_id}",
                         context={
                             "category_id": category_id,
--- a/src/ha_mcp/tools/tools_categories.py
+++ b/src/ha_mcp/tools/tools_categories.py
@@ -36,7 +36,11 @@ class CategoryTools:
     @tool(
         name="ha_config_get_category",
         tags={"Labels & Categories"},
-        annotations={"idempotentHint": True, "readOnlyHint": True, "title": "Get Category"},
+        annotations={
+            "idempotentHint": True,
+            "readOnlyHint": True,
+            "title": "Get Category",
+        },
     )
     @log_tool_usage
     async def ha_config_get_category(
@@ -276,7 +280,11 @@ class CategoryTools:
     @tool(
         name="ha_config_remove_category",
         tags={"Labels & Categories"},
-        annotations={"destructiveHint": True, "idempotentHint": True, "title": "Remove Category"},
+        annotations={
+            "destructiveHint": True,
+            "idempotentHint": True,
+            "title": "Remove Category",
+        },
     )
     @log_tool_usage
     async def ha_config_remove_category(
--- a/src/ha_mcp/tools/tools_categories.py
+++ b/src/ha_mcp/tools/tools_categories.py
@@ -251,6 +251,22 @@ class CategoryTools:
                     "message": f"Successfully {action_past} category: {name}",
                 }
             else:
+                error_str = str(result.get("error", "")).lower()
+                if "not found" in error_str:
+                    raise_tool_error(
+                        create_error_response(
+                            ErrorCode.RESOURCE_NOT_FOUND,
+                            f"Category not found: {category_id}",
+                            context={
+                                "name": name,
+                                "scope": scope,
+                                "category_id": category_id,
+                            },
+                            suggestions=[
+                                f"Use ha_config_get_category('{scope}') without category_id to see all categories",
+                            ],
+                        )
+                    )
                 raise_tool_error(
                     create_error_response(
                         ErrorCode.SERVICE_CALL_FAILED,
@@ -341,6 +357,18 @@ class CategoryTools:
                     "message": f"Successfully deleted category: {category_id}",
                 }
             else:
+                error_str = str(result.get("error", "")).lower()
+                if "not found" in error_str:
+                    raise_tool_error(
+                        create_error_response(
+                            ErrorCode.RESOURCE_NOT_FOUND,
+                            f"Category not found: {category_id}",
+                            context={"category_id": category_id, "scope": scope},
+                            suggestions=[
+                                f"Use ha_config_get_category('{scope}') without category_id to see all categories",
+                            ],
+                        )
+                    )
                 raise_tool_error(
                     create_error_response(
                         ErrorCode.SERVICE_CALL_FAILED,
--- a/src/ha_mcp/tools/tools_categories.py
+++ b/src/ha_mcp/tools/tools_categories.py
@@ -252,7 +252,7 @@ class CategoryTools:
                 }
             else:
                 error_str = str(result.get("error", "")).lower()
-                if "not found" in error_str:
+                if "not found" in error_str or "doesn't exist" in error_str:
                     raise_tool_error(
                         create_error_response(
                             ErrorCode.RESOURCE_NOT_FOUND,
@@ -358,7 +358,7 @@ class CategoryTools:
                 }
             else:
                 error_str = str(result.get("error", "")).lower()
-                if "not found" in error_str:
+                if "not found" in error_str or "doesn't exist" in error_str:
                     raise_tool_error(
                         create_error_response(
                             ErrorCode.RESOURCE_NOT_FOUND,
--- a/src/ha_mcp/tools/tools_config_automations.py
+++ b/src/ha_mcp/tools/tools_config_automations.py
@@ -280,7 +280,9 @@ class AutomationConfigTools:
                 ):
                     return str(state["entity_id"])
         except Exception as e:
-            logger.debug(f"Failed to resolve entity_id for automation {identifier}: {e}")
+            logger.debug(
+                f"Failed to resolve entity_id for automation {identifier}: {e}"
+            )
         return None
 
     @tool(
@@ -335,13 +337,17 @@ class AutomationConfigTools:
                     "Use ha_search_entities(domain_filter='automation') to list automations",
                 ],
             )
-            normalized_config, config_hash = await self._get_automation_config_internal(identifier)
+            normalized_config, config_hash = await self._get_automation_config_internal(
+                identifier
+            )
 
             # Resolve entity_id and fetch category from entity registry
             # (injected after hash so transient registry failures don't affect the hash)
             entity_id = await self._resolve_automation_entity_id(identifier)
             if entity_id:
-                cat_id = await fetch_entity_category(self._client, entity_id, "automation")
+                cat_id = await fetch_entity_category(
+                    self._client, entity_id, "automation"
+                )
                 if cat_id:
                     normalized_config["category"] = cat_id
 
@@ -645,7 +651,10 @@ class AutomationConfigTools:
                                 "Provide the automation entity_id or unique_id",
                                 "Use ha_search_entities(domain_filter='automation') to find automations",
                             ],
-                            context={"action": "python_transform", "identifier": identifier},
+                            context={
+                                "action": "python_transform",
+                                "identifier": identifier,
+                            },
                         )
                     )
                 if config_hash is None:
@@ -657,7 +666,10 @@ class AutomationConfigTools:
                                 "Call ha_config_get_automation() first",
                                 "Use the config_hash from that response",
                             ],
-                            context={"action": "python_transform", "identifier": identifier},
+                            context={
+                                "action": "python_transform",
+                                "identifier": identifier,
+                            },
                         )
                     )
 
@@ -676,7 +688,10 @@ class AutomationConfigTools:
                             ErrorCode.VALIDATION_FAILED,
                             message,
                             suggestions=suggestions,
-                            context={"action": "python_transform", "identifier": identifier},
+                            context={
+                                "action": "python_transform",
+                                "identifier": identifier,
+                            },
                         )
                     )
 
@@ -699,11 +714,20 @@ class AutomationConfigTools:
 
                 # Re-apply category if present
                 entity_id = result.get("entity_id")
-                if not entity_id and identifier and identifier.startswith("automation."):
+                if (
+                    not entity_id
+                    and identifier
+                    and identifier.startswith("automation.")
+                ):
                     entity_id = identifier
                 if transform_category and entity_id:
                     await apply_entity_category(
-                        self._client, entity_id, transform_category, "automation", result, "automation"
+                        self._client,
+                        entity_id,
+                        transform_category,
+                        "automation",
+                        result,
+                        "automation",
                     )
 
                 response: dict[str, Any] = {
@@ -763,7 +787,9 @@ class AutomationConfigTools:
                 self._client, config_dict
             )
 
-            result = await self._client.upsert_automation_config(config_dict, identifier)
+            result = await self._client.upsert_automation_config(
+                config_dict, identifier
+            )
 
             # If the client could not verify the entity was registered, warn but don't hard-fail.
             if result.get("entity_not_verified"):
@@ -785,7 +811,9 @@ class AutomationConfigTools:
             if wait_bool and entity_id:
                 action_word = "created" if identifier is None else "updated"
                 try:
-                    registered = await wait_for_entity_registered(self._client, entity_id)
+                    registered = await wait_for_entity_registered(
+                        self._client, entity_id
+                    )
                     if not registered:
                         result.setdefault("warnings", []).append(
                             f"Automation {action_word} but {entity_id} not yet queryable. "
@@ -799,7 +827,12 @@ class AutomationConfigTools:
             # Apply category to entity registry if provided
             if effective_category and entity_id:
                 await apply_entity_category(
-                    self._client, entity_id, effective_category, "automation", result, "automation"
+                    self._client,
+                    entity_id,
+                    effective_category,
+                    "automation",
+                    result,
+                    "automation",
                 )
 
             if bp_warnings:
@@ -852,9 +885,7 @@ class AutomationConfigTools:
                 {"type": "config/entity_registry/list"}
             )
         except Exception as e:
-            logger.debug(
-                "Failed to list automation entity_ids from registry: %s", e
-            )
+            logger.debug("Failed to list automation entity_ids from registry: %s", e)
             return []
         entries = result.get("result", []) if isinstance(result, dict) else result
         if not isinstance(entries, list):
@@ -913,7 +944,9 @@ class AutomationConfigTools:
         Returns the current normalized config dict.
         Raises ToolError if the hash does not match (conflict).
         """
-        current_config, current_hash = await self._get_automation_config_internal(identifier)
+        current_config, current_hash = await self._get_automation_config_internal(
+            identifier
+        )
         if current_hash != config_hash:
             raise_tool_error(
                 create_error_response(
@@ -934,22 +967,26 @@ class AutomationConfigTools:
         try:
             parsed_config = parse_json_param(config, "config")
         except ValueError as e:
-            raise_tool_error(create_error_response(
-                code=ErrorCode.VALIDATION_INVALID_JSON,
-                message=f"Invalid config parameter: {e}",
-                suggestions=[
-                    "Pass 'config' as a dict, not a JSON string, to avoid escaping issues.",
-                    "Check for JSON syntax errors: unquoted keys, trailing commas, or invalid escape sequences.",
-                ],
-                context={"parameter": "config"},
-            ))
+            raise_tool_error(
+                create_error_response(
+                    code=ErrorCode.VALIDATION_INVALID_JSON,
+                    message=f"Invalid config parameter: {e}",
+                    suggestions=[
+                        "Pass 'config' as a dict, not a JSON string, to avoid escaping issues.",
+                        "Check for JSON syntax errors: unquoted keys, trailing commas, or invalid escape sequences.",
+                    ],
+                    context={"parameter": "config"},
+                )
+            )
 
         if parsed_config is None or not isinstance(parsed_config, dict):
-            raise_tool_error(create_validation_error(
-                "Config parameter must be a JSON object",
-                parameter="config",
-                details=f"Received type: {type(parsed_config).__name__}",
-            ))
+            raise_tool_error(
+                create_validation_error(
+                    "Config parameter must be a JSON object",
+                    parameter="config",
+                    details=f"Received type: {type(parsed_config).__name__}",
+                )
+            )
 
         return cast(dict[str, Any], parsed_config)
 
@@ -978,24 +1015,28 @@ class AutomationConfigTools:
                 context: dict[str, Any] = {"missing_fields": missing_fields}
                 if identifier:
                     context["identifier"] = identifier
-                raise_tool_error(create_error_response(
-                    code=ErrorCode.CONFIG_MISSING_REQUIRED_FIELDS,
-                    message=f"Missing required fields: {', '.join(missing_fields)}",
-                    details=(
-                        "Config contains 'sequence', which belongs to scripts. "
-                        "Automations use 'trigger' and 'action'; scripts use 'sequence'."
-                    ),
-                    suggestions=[
-                        "Did you mean ha_config_set_script? Scripts use 'sequence' directly.",
-                        "For an automation, replace 'sequence' with 'action' and add a 'trigger'.",
-                    ],
-                    context=context,
-                ))
-            raise_tool_error(create_config_error(
-                f"Missing required fields: {', '.join(missing_fields)}",
-                identifier=identifier,
-                missing_fields=missing_fields,
-            ))
+                raise_tool_error(
+                    create_error_response(
+                        code=ErrorCode.CONFIG_MISSING_REQUIRED_FIELDS,
+                        message=f"Missing required fields: {', '.join(missing_fields)}",
+                        details=(
+                            "Config contains 'sequence', which belongs to scripts. "
+                            "Automations use 'trigger' and 'action'; scripts use 'sequence'."
+                        ),
+                        suggestions=[
+                            "Did you mean ha_config_set_script? Scripts use 'sequence' directly.",
+                            "For an automation, replace 'sequence' with 'action' and add a 'trigger'.",
+                        ],
+                        context=context,
+                    )
+                )
+            raise_tool_error(
+                create_config_error(
+                    f"Missing required fields: {', '.join(missing_fields)}",
+                    identifier=identifier,
+                    missing_fields=missing_fields,
+                )
+            )
 
         # Issue #1169: reject configs that wrap ``scene.create`` in an
         # automation with no functional trigger. Models occasionally produce
@@ -1020,31 +1061,33 @@ class AutomationConfigTools:
                 if _action_contains_scene_create(a)
             ]
             if scene_create_indices:
-                raise_tool_error(create_error_response(
-                    code=ErrorCode.VALIDATION_INVALID_PARAMETER,
-                    message=(
-                        "Empty trigger paired with a scene.create action — "
-                        "this automation can never fire. For a state snapshot "
-                        "of one or more entities, use ha_config_set_scene "
-                        "directly instead of wrapping scene.create in an "
-                        "automation."
-                    ),
-                    suggestions=[
-                        "ha_config_set_scene(scene_id='...', config={'name': "
-                        "'...', 'entities': {'<entity_id>': {...}}}) creates "
-                        "a scene without a trigger.",
-                        "If the snapshot really should be the result of an "
-                        "event, add the trigger that should fire it and keep "
-                        "the automation.",
-                        "For a state-derived value that recomputes when its "
-                        "inputs change, use "
-                        "ha_config_set_helper(helper_type='template') instead.",
-                    ],
-                    context={
-                        "scene_create_action_indices": scene_create_indices,
-                        "identifier": identifier,
-                    },
-                ))
+                raise_tool_error(
+                    create_error_response(
+                        code=ErrorCode.VALIDATION_INVALID_PARAMETER,
+                        message=(
+                            "Empty trigger paired with a scene.create action — "
+                            "this automation can never fire. For a state snapshot "
+                            "of one or more entities, use ha_config_set_scene "
+                            "directly instead of wrapping scene.create in an "
+                            "automation."
+                        ),
+                        suggestions=[
+                            "ha_config_set_scene(scene_id='...', config={'name': "
+                            "'...', 'entities': {'<entity_id>': {...}}}) creates "
+                            "a scene without a trigger.",
+                            "If the snapshot really should be the result of an "
+                            "event, add the trigger that should fire it and keep "
+                            "the automation.",
+                            "For a state-derived value that recomputes when its "
+                            "inputs change, use "
+                            "ha_config_set_helper(helper_type='template') instead.",
+                        ],
+                        context={
+                            "scene_create_action_indices": scene_create_indices,
+                            "identifier": identifier,
+                        },
+                    )
+                )
 
         # HA accepts conditions with 'platform' (trigger syntax) but then crashes
         # with an unhelpful 500 rather than a 400 validation error.
@@ -1052,30 +1095,34 @@ class AutomationConfigTools:
             if not isinstance(cond, dict):
                 continue
             if "platform" in cond and "condition" not in cond:
-                raise_tool_error(create_error_response(
-                    code=ErrorCode.VALIDATION_INVALID_PARAMETER,
-                    message=(
-                        f"Condition at index {idx} uses 'platform' (trigger syntax). "
-                        "Conditions use 'condition', not 'platform'."
-                    ),
-                    suggestions=[
-                        f"Replace 'platform' with 'condition': "
-                        f"{{'condition': '{cond['platform']}', ...}}",
-                        "Triggers use 'platform'; conditions use 'condition'.",
-                    ],
-                    context={"condition_index": idx, "found_key": "platform"},
-                ))
+                raise_tool_error(
+                    create_error_response(
+                        code=ErrorCode.VALIDATION_INVALID_PARAMETER,
+                        message=(
+                            f"Condition at index {idx} uses 'platform' (trigger syntax). "
+                            "Conditions use 'condition', not 'platform'."
+                        ),
+                        suggestions=[
+                            f"Replace 'platform' with 'condition': "
+                            f"{{'condition': '{cond['platform']}', ...}}",
+                            "Triggers use 'platform'; conditions use 'condition'.",
+                        ],
+                        context={"condition_index": idx, "found_key": "platform"},
+                    )
+                )
 
         # Prevent duplicate creation when config contains an existing automation id
         if identifier is None and "id" in config_dict:
             existing_id = config_dict["id"]
-            raise_tool_error(create_validation_error(
-                f"Config contains 'id' field ('{existing_id}') but no identifier was provided. "
-                "This would create a duplicate automation instead of updating the existing one.",
-                parameter="identifier",
-                details=f"To update, pass identifier='{existing_id}' (or the automation's entity_id). "
-                "To create a genuinely new automation, remove the 'id' field from the config.",
-            ))
+            raise_tool_error(
+                create_validation_error(
+                    f"Config contains 'id' field ('{existing_id}') but no identifier was provided. "
+                    "This would create a duplicate automation instead of updating the existing one.",
+                    parameter="identifier",
+                    details=f"To update, pass identifier='{existing_id}' (or the automation's entity_id). "
+                    "To create a genuinely new automation, remove the 'id' field from the config.",
+                )
+            )
 
     @tool(
         name="ha_config_remove_automation",
@@ -1140,7 +1187,9 @@ class AutomationConfigTools:
             wait_bool = coerce_bool_param(wait, "wait", default=True)
             if wait_bool and entity_id_for_wait:
                 try:
-                    removed = await wait_for_entity_removed(self._client, entity_id_for_wait)
+                    removed = await wait_for_entity_removed(
+                        self._client, entity_id_for_wait
+                    )
                     if not removed:
                         result.setdefault("warnings", []).append(
                             f"Deletion confirmed by API but {entity_id_for_wait} may still appear briefly."
--- a/src/ha_mcp/tools/tools_config_automations.py
+++ b/src/ha_mcp/tools/tools_config_automations.py
@@ -13,6 +13,7 @@ from fastmcp.tools import tool
 from pydantic import Field
 
 from ..client.rest_client import (
+    HomeAssistantAPIError,
     HomeAssistantAuthError,
     HomeAssistantConnectionError,
 )
@@ -838,6 +839,34 @@ class AutomationConfigTools:
                 suggestions=suggestions,
             )
 
+    async def _list_automation_entity_ids(self) -> list[str]:
+        """Best-effort list of automation entity_ids from the entity registry.
+
+        Used to populate ``available_automation_ids`` in RESOURCE_NOT_FOUND
+        error context. Returns an empty list on any failure — caller treats
+        absence as "no IDs to report" rather than failing the structured
+        error raise.
+        """
+        try:
+            result = await self._client.send_websocket_message(
+                {"type": "config/entity_registry/list"}
+            )
+        except Exception as e:
+            logger.debug(
+                "Failed to list automation entity_ids from registry: %s", e
+            )
+            return []
+        entries = result.get("result", []) if isinstance(result, dict) else result
+        if not isinstance(entries, list):
+            return []
+        return [
+            entry["entity_id"]
+            for entry in entries
+            if isinstance(entry, dict)
+            and isinstance(entry.get("entity_id"), str)
+            and entry["entity_id"].startswith("automation.")
+        ]
+
     async def _get_automation_config_internal(
         self, identifier: str
     ) -> tuple[dict[str, Any], str]:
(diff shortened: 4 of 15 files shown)


Patch76:
## Implementation Summary

Closes #1297. Reclassifies five non-entity not-found responses from `ENTITY_NOT_FOUND` to `RESOURCE_NOT_FOUND` — labels, categories, devices (×2), and zones (Boy-Scout fold during self-review).

### Choices Made

- **Bundled into one PR rather than three** per the maintainer #1297 review ack ("would rather you just bundle them into one PR, will be less noise"). My backlog-pruning comment originally proposed splitting D1 labels/categories, D1 devices, and D2 dashboards/resources — the maintainer explicitly preferred bundling.
- **Scope narrowed mid-PR after #1386 merged**. PR #1386 landed during this PR's prep and absorbed the D2 work (dashboards + resources Pattern-A removal + suggestions); I dropped my parallel D2 commits and rebased. Final scope is D1-only: five `ENTITY_NOT_FOUND → RESOURCE_NOT_FOUND` swaps (labels, categories, devices ×2, zones) where the lookup is by registry-internal ID, not by `entity_id`.
- **Zones folded in during self-review** per Boy-Scout. `tools_zones.py:100` had the exact same anti-pattern as labels/categories (registry-internal `zone_id` lookup, explicit `ha_get_zone()` recovery suggestion, `available_zone_ids` payload) but classified as `ENTITY_NOT_FOUND`. All four Boy-Scout conditions clear (same dir/family, 1-line src + 1-line E2E test, sibling-match, same anti-pattern as the live bug).
- **Helper signature unchanged.** I considered widening `create_resource_not_found_error` with a `suggestions` parameter early on, but #1386 deleted that helper entirely in favor of `create_error_response(...)` direct calls. Rebase dropped the helper-widening hunks.
- **Style commit kept separate** (`50a71491`). `#1387`'s new `ruff format --check` gate fires on file-level, so format-debt on lines I didn't touch (tools_labels.py annotations, test_label_crud.py unrelated breaks, ~150 LOC) had to be cleaned. Split as `style:` for reviewable separation; cross-linked to #1318 (per-dir ruff format sweep tracker).

### Problems Encountered

- **First CI run red on 3 E2E tests pinning the old `ENTITY_NOT_FOUND`.** My pattern-family grep covered `src/` but missed `tests/`. Fixed in commit `2fd1e4ce` (`test(e2e): update assertions to match RESOURCE_NOT_FOUND ...`); the docstrings on those 3 tests now name `RESOURCE_NOT_FOUND` with a one-line `#1297` rationale. Second CI run green.
- **PR body fact-drift, twice.** Initial body claimed "five non-entity not-found responses" against a four-row table; patched to "four" via `gh api -X PATCH` (gh pr edit silent-fail workaround). Then the zones Boy-Scout fold added a fifth swap, so I re-patched body to "five" with the zones row added and line anchors refreshed to their post-format positions.
- **Rebase conflict surface from #1386.** Both PRs touched `errors.py`, `tools_config_dashboards.py`, `tools_resources.py`, `test_errors.py`. Took master side on all four (drop my D2 work, absorbed by #1386). Auto-merged D1 changes in labels/categories/registry cleanly.
- **No Docker locally** ruled out running the e2e suite before push. Relied on the broad unit suite (`pytest src/unit/ -n auto -m "not slow"`, clean) + the dedicated unit regression tests for each changed site to gate the push. CI's e2e run was the first real verification — surfaced the assertion drift above.

kingpanther13 (changes requested):
Six required, one nit. All in the same classifier-shape audit family this PR is doing on the read side.

**1. `src/ha_mcp/tools/tools_zones.py:128`** — the `except Exception` fallback in `ha_get_zone` still carries `"Use ha_search_entities(domain_filter='zone') as alternative"`. Same wrong-tool spiral this PR is closing on the not-found path — zone `entity_id`s returned by entity-search don't feed into `ha_get_zone(zone_id=)`. Drop the third suggestion or replace with `"Use ha_get_zone() without zone_id to see all available zones"`.

**2. `src/ha_mcp/tools/tools_registry.py:379` and `:746`** — the new `RESOURCE_NOT_FOUND` paths only emit `device_id` in context. The four sibling sites in this PR emit `available_label_ids` / `available_category_ids` / `available_zone_ids`. Add `available_device_ids` (first-10 truncation like the others) for parity. Without it the device unit tests can only pin `error[\"code\"]` while sibling tests pin code + suggestion + payload.

**3. `tests/src/unit/test_error_code_consistency_1297.py:163-201`** — after #2 lands, bring the device tests up to sibling depth: assert `ha_get_device()` in suggestion text and `available_device_ids` in payload.

**4. Sibling mutation paths in the same registry families** — `tools_zones.py:318` (`ha_set_zone` update branch), `tools_zones.py:393` (`ha_remove_zone`), plus the equivalent set/remove branches in `tools_labels.py` and `tools_categories.py`, still raise `SERVICE_CALL_FAILED` when the registry-internal id doesn't exist. Same not-found classifier audit family — fold in. Needs per-site WS \"not found\" string match rather than a 1-line `ErrorCode` swap, but it's the same audit.

**5. `AGENTS.md:708`** — Error Handling section still lists `create_resource_not_found_error` in \"Available helpers\". Function was removed by #1386. Drop from the list.

**6. `rules/no-return-error-response.yml:14`** — same removal: drop the `return create_resource_not_found_error(\$\$\$)` pattern.

**While you're in there**

**7. `tests/src/unit/test_error_code_consistency_1297.py:6-9`** — module docstring's \"auto-routed agent retry to \`ha_search_entities()\`\" overstates the mechanism. PR description's own phrasing (\"an agent branching on \`error.code == ENTITY_NOT_FOUND\` retries via \`ha_search_entities()\`\") is sharper. Either match that, or expand to point at `errors.py:129-133` where the default-suggestions table attaches the search hint when callers don't supply explicit suggestions.

Patch76:
Addressed in 70b9edf. All six required + the nit.

## 1. `tools_zones.py:128` — ha_search_entities suggestion replaced

Third suggestion in `ha_get_zone`'s `except Exception` fallback now reads `"Use ha_get_zone() without zone_id to see all available zones"` instead of the entity-search detour. Matches the inner-not-found branch's own recovery hint.

## 2. `tools_registry.py:379` / `:746` — available_device_ids parity

`ha_get_device` single-lookup mode (`tools_registry.py:377-390`) and `ha_remove_device` (`:750-763`) now compute `available_device_ids = [d.get("id") for d in <list>[:10] if d.get("id")]` and emit it in the `RESOURCE_NOT_FOUND` context. Both sites already had the device list in scope, so this is zero-cost — no extra WS round-trip.

## 3. Device tests at sibling depth

`TestDeviceLookupMissingReturnsResourceNotFound` (`tests/src/unit/test_error_code_consistency_1297.py:181-255`) now pins `ha_get_device` in the suggestion text and `available_device_ids` in the payload. Mocks now seed two device entries so the truncated list has shape worth asserting against (`["dev-existing-1", "dev-existing-2"]`).

## 4. Mutation sites: zones, labels, categories set/remove

Six new branches added per the same `tools_entities.py:1377` shape — per-site WS `"not found"` substring match before the catch-all `SERVICE_CALL_FAILED`:

- `tools_zones.py:317` (`ha_set_zone` update branch), `:404` (`ha_remove_zone`)
- `tools_labels.py:249` (`ha_config_set_label` update branch), `:338` (`ha_config_remove_label`)
- `tools_categories.py:255` (`ha_config_set_category` update branch), `:361` (`ha_config_remove_category`)

Each routes to `RESOURCE_NOT_FOUND` with the same list-tool suggestion the GET-path siblings use. Six new tests in `TestLabel/Category/ZoneMutationRoutesNotFoundToResourceNotFound` (`test_error_code_consistency_1297.py:256-398`) pin the new branch shape.

E2E follow-on: `tests/src/e2e/workflows/categories/test_category_crud.py::test_delete_nonexistent_category` previously pinned `SERVICE_CALL_FAILED` for the now-`RESOURCE_NOT_FOUND` branch; updated to match.

## 5. `AGENTS.md:708` — helpers list

`create_resource_not_found_error` dropped from the "Available helpers" list. Function was removed by #1386.

## 6. `rules/no-return-error-response.yml`

Same removal: dropped the `return create_resource_not_found_error($$$)` pattern. File is now 28 lines (was 29).

## 7. Test module docstring — nit

`test_error_code_consistency_1297.py:1-16` docstring rewritten. The original "auto-routed agent retry to `ha_search_entities()`" overstated the mechanism — explicit `suggestions` at every changed site means the `ENTITY_NOT_FOUND` default-suggestion table in `errors.py:129-133` never surfaces. New phrasing names the actual leak path (agent code branching on `error.code == "ENTITY_NOT_FOUND"`) and points at the default-suggestions table as the mechanism reference. Zones added to the family list since item #4 brought zones into the audit on the mutation side.

Branch HEAD is `466ce2c` — `70b9edf` covered the six review items + nit; `466ce2c` widens the item-4 substring check after CI surfaced that HA Core's category-registry uses the `Category ID doesn't exist` phrasing rather than `not found`. New unit test `test_remove_with_doesnt_exist_phrasing` pins the second phrasing.

kingpanther13 (changes requested):
Prior 7 all addressed cleanly — CI green. One more to fold in.

**D4 — `available_*_ids` parity for automations / scripts / dashboards.**

You marked this deferred in the implementation summary, but the Anti-Noise Gate says don't file/track follow-ups for handful-of-lines work in the same audit family. D4 is the same classifier-shape audit this PR just closed for the four sibling registries. Verified not already opened anywhere — no open PR, no open issue tracks the sites — and `Closes #1297` on this PR leaves the gap unclosed otherwise.

**Three sites:**

1. `src/ha_mcp/tools/tools_config_dashboards.py:1537` (`ha_config_delete_dashboard`) — `_resolve_dashboard` already fetches the dashboard list; splice first-10 IDs into context as `available_dashboard_ids`. ~5-10 LOC.

2. `src/ha_mcp/tools/tools_config_automations.py:841` (`_get_automation_config_internal` central not-found raise) — match the labels/categories shape: direct `raise_tool_error(create_error_response(ErrorCode.RESOURCE_NOT_FOUND, ..., context={\"automation_id\": ..., \"available_automation_ids\": [...]}, suggestions=[...]))` rather than the current `exception_to_structured_error` route. ~30-50 LOC + tests.

3. `src/ha_mcp/tools/tools_config_scripts.py:173` (`_get_script_config_internal`) — mirror the automations shape. ~30-50 LOC + tests.

Tests should hit the same assertion depth as `TestLabelGetMissingReturnsResourceNotFound` etc.: `error[\"code\"]` + suggestion text mentioning the list-tool + `available_*_ids` payload.

---

**On the implementation-summary format**: AGENTS.md § PR Workflow → Final reporting prescribes two sections only — *Choices Made* and *Problems Encountered*. There's no *Suggested Improvements* section in the template, and the Anti-Noise Gate covers why: anything inline-foldable belongs in the PR, anything truly large needs explicit user sign-off via the verification template — not a unilateral *follow-ups, not blockers* bucket in the summary. Drop the section in future summaries.

Patch76:
Addressed in 8e96914 (logic + tests) and 993c332 (ruff format sweep on the touched files, mirroring 50a7149's pattern for the in-scope format gate).

## D4 — `available_*_ids` parity (3 sites)

### Site 1 — `tools_config_dashboards.py:1533-1555` (`ha_config_delete_dashboard`)

`_resolve_dashboard` already returns the dashboards list alongside the match — previously discarded via `resolved, _ = …`. Now captured as `resolved, dashboards = …` (`:1533`) and spliced via `[d.get("url_path") for d in (dashboards or [])[:10] if d.get("url_path")]` (`:1535-1539`) into the `RESOURCE_NOT_FOUND` context at `:1552`. Zero extra round-trip.

### Site 2 — `tools_config_automations.py:875-937` (`_get_automation_config_internal`)

Central not-found raise moved into the internal helper per your anchor. New `_list_automation_entity_ids` helper (`:875-898`, best-effort `config/entity_registry/list` filtered by `automation.` prefix) feeds the `available_automation_ids` payload. The fetch helper at `:901-937` now wraps the REST call in `try/except HomeAssistantAPIError`: `e.status_code == 404` triggers a structured `RESOURCE_NOT_FOUND` raise (`:917-933`); other `HomeAssistantAPIError` instances re-raise unchanged so the caller's `exception_to_structured_error` route still handles them.

`_fetch_and_verify_hash` (`:947`) and the python_transform refetch (`:712`) inherit the new behavior since both flow through the same helper — set/remove paths now also surface `RESOURCE_NOT_FOUND` with available IDs when the hash-verify hits a 404, matching the audit family's intent.

### Site 3 — `tools_config_scripts.py:174-246`

Scripts had a different architecture from automations (the get tool called REST directly at `:126`; the internal helper only fed `_fetch_and_verify_hash`). Centralized 404 handling in a new private helper `_fetch_script_config_envelope` (`:200-229`) that both `ha_config_get_script` (now `:126`) and `_get_script_config_internal` (`:231-246`) route through. Same shape as automations: 404 → `RESOURCE_NOT_FOUND` with `available_script_ids` from `_list_script_entity_ids` (`:174-198`).

### Regression tests (`tests/src/unit/test_error_code_consistency_1297.py:411-624`)

Five new tests across three sibling classes:

- `TestDeleteDashboardNotFoundSurfacesAvailableIds::test_missing_dashboard_surfaces_available_dashboard_ids` — pins code + `action`/`url_path` + `available_dashboard_ids` (test seeds 3 dashboards; the `[:10]` cap is exercised in production)
- `TestGetAutomationMissingSurfacesAvailableIds::test_missing_automation_surfaces_available_automation_ids` — pins code + `automation_id` + `available_automation_ids` filtered to `automation.` prefix (script/light entries in the mock registry are skipped) + `ha_search_entities(domain_filter='automation')` suggestion
- `TestGetAutomationMissingSurfacesAvailableIds::test_registry_list_failure_yields_empty_available_ids` — best-effort fallback: a registry-list WS failure yields `available_automation_ids: []` rather than masking the 404
- `TestGetScriptMissingSurfacesAvailableIds::test_missing_script_surfaces_available_script_ids` — mirror of automations test
- `TestGetScriptMissingSurfacesAvailableIds::test_registry_list_failure_yields_empty_available_ids` — mirror

All 17 tests in `test_error_code_consistency_1297.py` green locally; sibling unit suites (`test_tools_config_{automations,scripts,dashboards}.py`, 48 tests) also green — no regression.

## Implementation Summary format

Comment 4507758082 patched to drop the `Suggested Improvements` section — AGENTS.md L305-312 prescribes two sections only (Choices Made + Problems Encountered). The deferred-D4 line in the dropped section was superseded by 8e96914 anyway; the two cross-link mentions (#1399 classifier-bypass sites, PR #1398 / #1389 poll-cadence) live in their own threads. Future IS strictly two-section.

kingpanther13 (changes requested):
D4 fold-in landed at the three sites cleanly — sibling-site shape, list-helper best-effort contract, and `ToolError` guards all verified. CI lanes green (HAOS still pending). Three findings plus a few minor folds.

**1. Mutation-path parity gap for automations/scripts.** Same audit family as `70b9edf4`.

The new `_list_automation_entity_ids` / `_list_script_entity_ids` helpers are wired into the GET paths only. Mutation paths still route 404s through `exception_to_structured_error` → `_classify_api_status`, which surfaces `RESOURCE_NOT_FOUND` *without* `available_*_ids`:

- `src/ha_mcp/tools/tools_config_automations.py` — `ha_config_remove_automation`, `ha_config_set_automation` (update branch)
- `src/ha_mcp/tools/tools_config_scripts.py` — `ha_config_remove_script`, `ha_config_set_script` (update branch)

Exact gap `70b9edf4` closed for labels/categories/zones siblings — helpers are right there, wire them in.

**2. Test docstring factual error — `tests/src/unit/test_error_code_consistency_1297.py:480-484`**

Says: *\"Pre-#1397-D4 the 404 fell through the generic \`exception_to_structured_error\` route, surfacing as \`INTERNAL_ERROR\` without an actionable payload.\"*

This is wrong. \`helpers.py::_classify_api_status\` (lines 199-205) was already mapping \`HomeAssistantAPIError(status_code=404)\` → \`RESOURCE_NOT_FOUND\`. The pre-fix gap was *missing payload + missing list-tool suggestion*, not wrong code. Rewrite:

> \"Pre-#1397-D4 the 404 surfaced as \`RESOURCE_NOT_FOUND\` via the generic \`_classify_api_status\` route, but without \`available_automation_ids\` or the list-tool suggestion that sibling sites populate.\"

Future maintainer reading the current text would chase a non-existent classification regression.

**3. \`available_script_ids\` payload shape mismatch — \`src/ha_mcp/tools/tools_config_scripts.py:221\`**

Helper populates the payload with full \`entity_id\` form (\`script.morning_routine\`), but \`ha_config_get_script(script_id=...)\` takes the bare \`script_id\` (\`morning_routine\`). Agent reading \`available_script_ids\` would have to strip the \`script.\` prefix before retrying. Either strip in \`_list_script_entity_ids\` (preferred — mirrors what the labels/categories sites do with registry-internal IDs) or note the prefix in the suggestion text. Automation sibling is fine because \`ha_config_get_automation\` accepts both forms.

**While you're in there**

**4.** \`_resolve_dashboard\` failure path untested. Current dashboard test only covers the happy \`_resolve_dashboard returns N dashboards\` case. Add a counterpart asserting \`available_dashboard_ids == []\` when \`_resolve_dashboard\` returns \`(None, None)\` — the \`(dashboards or [])[:10]\` guard at \`tools_config_dashboards.py:1535\` already supports it.

**5.** \`[:10]\` truncation lives at every caller, not in the helpers. If a new caller forgets, response balloons. Move the slice into \`_list_automation_entity_ids\` / \`_list_script_entity_ids\` (return \"up to 10 IDs\"), or document the caller-side contract on the helpers.

**6.** Suggestion-string verb inconsistency across the three new sites. Dashboard: \`\"Use ha_config_get_dashboard(list_only=True)\"\`. Automations: \`\"Verify the entity_id or unique_id is correct\"\`. Scripts: \`\"Verify the script identifier is correct\"\`. Sibling registries (labels/categories/zones/devices) consistently use the list-tool form. Align scripts/automations to \`\"Use ha_config_get_<thing>() to find valid <noun> IDs\"\`.

Patch76:
Addressed in f679dfd. All 3 required + 3 folds.

## 1. Mutation-path parity (automations + scripts)

Extracted the `RESOURCE_NOT_FOUND` raise into single-source helpers `_raise_automation_not_found` (`tools_config_automations.py:909-932`) and `_raise_script_not_found` (`tools_config_scripts.py:204-226`), wired into the four mutation sites alongside the two GET-path sites:

| Site | Anchor | Catch |
|---|---|---|
| `_get_automation_config_internal` | `tools_config_automations.py:947-952` | `except HomeAssistantAPIError` (404) |
| `ha_config_set_automation` (update branch) | `tools_config_automations.py:856-863` | guarded by `identifier and …` so create stays untouched |
| `ha_config_remove_automation` | `tools_config_automations.py:1229-1231` | `isinstance(e, HomeAssistantAPIError) and e.status_code == 404` |
| `_fetch_script_config_envelope` | `tools_config_scripts.py:236-243` | `except HomeAssistantAPIError` (404) |
| `ha_config_set_script` (update branch) | `tools_config_scripts.py:745-746` | `isinstance(e, HomeAssistantAPIError) and e.status_code == 404` |
| `ha_config_remove_script` | `tools_config_scripts.py:823-824` | `isinstance(e, HomeAssistantAPIError) and e.status_code == 404` |

The create branch on `ha_config_set_automation` is guarded with `identifier and …` so a 404 with `identifier=None` (create path) still falls through to the original `exception_to_structured_error` route — only updates against missing IDs surface as `RESOURCE_NOT_FOUND`.

## 2. Test docstring rewrite (`tests/src/unit/test_error_code_consistency_1297.py:499-508`)

`TestGetAutomationMissingSurfacesAvailableIds` docstring rewritten per your phrasing. Verified `helpers.py::_classify_api_status` at L199-205 — the 404→`RESOURCE_NOT_FOUND` mapping was already there; the pre-#1397-D4 gap was the missing `available_*_ids` payload and the missing list-tool suggestion, not the classification. New docstring names the actual gap and points at `helpers.py:199-205` as the reference.

## 3. `available_script_ids` shape fix (`tools_config_scripts.py:196-202`)

`_list_script_entity_ids` now returns bare storage keys (e.g. `morning_routine`), stripping the `script.` entity_id prefix via `entry["entity_id"][len("script.") :]` (`:197`). Matches what `ha_config_get_script(script_id=...)` / `ha_config_set_script` / `ha_config_remove_script` take, so an agent reading `available_script_ids` retries directly without a prefix-strip step. Automations are untouched — `ha_config_get_automation` accepts both forms per its docstring + the `_resolve_automation_entity_id` fallback. Test expectations updated to `["morning_routine", "evening_routine"]`.

## 4. Dashboard resolver-failure test (`tests/src/unit/test_error_code_consistency_1297.py:477-496`)

`test_resolver_unexpected_shape_yields_empty_available_ids` added. WS-response returns the literal string `"garbage"`, triggering the `lovelace/dashboards/list returned an unexpected shape` warning path in `_fetch_dashboards_list`, which propagates as `_resolve_dashboard` returning `(None, None)`. Asserts the raise still fires with `available_dashboard_ids == []` — the `(dashboards or [])[:10]` guard at the raise site already handled it; this just pins the contract.

## 5. `[:10]` truncation moved into helpers

`_list_automation_entity_ids` (`tools_config_automations.py:882-907`, cap at `:907`) and `_list_script_entity_ids` (`tools_config_scripts.py:174-202`, cap at `:202`) both now append `[:10]` to the comprehension and document the 10-entry cap in their docstrings. Callers (`_raise_*_not_found` helpers) just pass `available_ids` through to context. New call sites can't bloat the payload without explicitly slicing wider.

## 6. Suggestion-string verb alignment

Dropped the redundant `"Verify the entity_id or unique_id is correct"` (automations, `_raise_automation_not_found` at `tools_config_automations.py:927-929`) / `"Verify the script identifier is correct"` (scripts, `_raise_script_not_found` at `tools_config_scripts.py:220-222`) lines, leaving the single list-tool suggestion (`Use ha_search_entities(domain_filter='…') to find existing …`). Matches the sibling labels / categories / zones / devices single-suggestion shape.

## Test coverage

5 new tests in `test_error_code_consistency_1297.py`:

- `TestDeleteDashboardNotFoundSurfacesAvailableIds::test_resolver_unexpected_shape_yields_empty_available_ids` (`:477-496`) — Item 4
- `TestSetAutomationMissingSurfacesAvailableIds::test_update_with_missing_identifier_surfaces_available_ids` (`:673-736`) — Item 1
- `TestRemoveAutomationMissingSurfacesAvailableIds::test_remove_with_missing_identifier_surfaces_available_ids` (`:738-781`) — Item 1
- `TestSetScriptMissingSurfacesAvailableIds::test_update_with_missing_script_id_surfaces_available_ids` (`:783-837`) — Item 1
- `TestRemoveScriptMissingSurfacesAvailableIds::test_remove_with_missing_script_id_surfaces_available_ids` (`:839-880`) — Item 1

All 22 tests in `test_error_code_consistency_1297.py` green locally; sibling unit suites (`test_tools_config_{automations,scripts,dashboards}.py`, 48 tests) also green — no regression.

kingpanther13 (changes requested):
Five of six items from the third-pass landed cleanly. One residual at the script GET tool — fix here, then this is good to merge.

**\`src/ha_mcp/tools/tools_config_scripts.py:91\`** — make \`ha_config_get_script(script_id: str)\` accept both \`script.morning_routine\` (entity_id form) and \`morning_routine\` (bare form), mirroring \`ha_config_get_automation\` which already accepts both forms. Strip a leading \`script.\` prefix at the top of the function; the rest of the body runs unchanged.

The suggestion in \`_raise_script_not_found\` routes to \`ha_search_entities(domain_filter='script')\`, which returns entity_id form (\`script.morning_routine\`). The payload (\`available_script_ids\`) is bare form after the third-pass item-3 fix. An agent that follows the suggestion text literally — the class of agent this PR is built for (smaller / suggestion-literal models) — gets entity_ids back, then fails against the bare-only \`ha_config_get_script\`. That reseeds the wrong-tool spiral at one site.

Closing it via the suggestion text would require either spelling out the prefix-strip inline (\"call \`ha_config_get_script\` with the bare ID, e.g. 'morning_routine' for 'script.morning_routine'\") or routing through a tool that doesn't exist. Making the GET tool accept both forms (symmetry with automations) closes it at the source.

Once this lands and CI is green, PR is good to merge.

Patch76:
Fourth-pass fix on 4740252.

`ha_config_get_script` now strips a leading `script.` prefix at the top of the function (after `validate_identifier_not_empty`), so the entity_id form (`script.morning_routine`) and the bare storage key (`morning_routine`) both route to the same `get_script_config` call. Mirrors `ha_config_get_automation`. The function body is otherwise unchanged.

Closed the wrong-tool spiral you described: `_raise_script_not_found` suggests `ha_search_entities(domain_filter='script')` which returns entity_ids — those now feed back into the GET tool successfully and return the canonical bare `script_id` for subsequent `ha_config_set_script` / `ha_config_remove_script` calls.

Docstring + Field description updated to spell out both forms. The existing `TestGetScriptMissingSurfacesAvailableIds` class docstring also dropped its now-outdated "scripts only accept the bare storage key" line.

3 new unit tests in `TestGetScriptAcceptsEntityIdForm` (`tests/src/unit/test_error_code_consistency_1297.py`):

- `test_entity_id_form_routes_to_bare_storage_key` — `script.foo` arrives at `get_script_config` as `foo`
- `test_bare_form_unchanged` — `foo` arrives unchanged
- `test_only_leading_prefix_is_stripped` — a bare key with `script.` mid-string (`my_script.backup`) passes through untouched

kingpanther13 (changes requested):
Apologies before the asks: #1 (set/remove parity) is pre-existing, not caused by this commit. It was true at the 3rd-pass when I noted that `available_script_ids` should be bare to match `ha_config_get_script(script_id=)`, and I contrasted with automations accepting both forms — that was the moment to flag set/remove. Missed it then, missed it again at 4th-pass when asking only for the GET strip. Catching it now on the 5th, with apologies for the drip-feed. All three fold in.

**1. Parity gap on the same family — `ha_config_set_script` and `ha_config_remove_script` don't strip the prefix (pre-existing).**

The wrong-tool spiral the 4th-pass closed on GET is still open at the sibling sites:

- `src/ha_mcp/tools/tools_config_scripts.py:362` (`ha_config_set_script`) — `script_id` passes through unchanged. Agent that gets `script.foo` from `ha_search_entities` and calls `ha_config_set_script(\"script.foo\", ...)` hits `upsert_script_config(\"script.foo\", ...)`. HA's script collection keys writes by the literal `script_id`, creating a phantom `script.foo` key that won't match a subsequent `ha_config_get_script(\"foo\")`.
- `src/ha_mcp/tools/tools_config_scripts.py:765` (`ha_config_remove_script`) — same passthrough; also constructs `entity_id = f\"script.{script_id}\"` at the wait step (line 819) → becomes `script.script.foo`, breaking the `wait_for_entity_removed` watcher contract.

Apply the same `script_id = script_id.removeprefix(\"script.\")` strip at the top of both bodies. Place it BEFORE `validate_identifier_not_empty` — see #3.

**2. \"Mirrors `ha_config_get_automation`\" claim is factually wrong (introduced by 4th-pass).**

`ha_config_get_automation` doesn't strip a prefix — it routes through `_resolve_automation_entity_id` (`tools_config_automations.py:268-286`) which passes `automation.*` as-is and resolves bare identifiers via state-search unique_id lookup. Behavioral parity, different mechanism. Reword in:

- `tools_config_scripts.py:107` (\"symmetry with `ha_config_get_automation`\") → \"behavioral parity with `ha_config_get_automation` (mechanism differs — automations resolve via state lookup; scripts strip the prefix)\"
- `tools_config_scripts.py:131-136` inline comment (\"mirrors `ha_config_get_automation`\") — same correction
- `tests/src/unit/test_error_code_consistency_1297.py:587-589` class docstring — drop the \"leading-prefix strip at the function head\" claim for automations

**3. Validator bypass via strip — `\"script.\"` slips past `validate_identifier_not_empty` (introduced by 4th-pass).**

`\"script.\"` passes `validate_identifier_not_empty` (non-empty pre-strip) → `removeprefix` makes it `\"\"` → `get_script_config(\"\")` is called. Pick a contract: re-validate after strip, OR (cleanest) place the strip BEFORE `validate_identifier_not_empty` so empty post-strip surfaces as `VALIDATION_INVALID_PARAMETER` not a downstream 404. Apply across all 3 strip sites (GET, set after #1, remove after #1). Pin with a regression test.

**While you're in there**

- Pin the nested-prefix one-layer-strip behavior in `TestGetScriptAcceptsEntityIdForm`: `\"script.script.foo\"` → `\"script.foo\"`.
- Field descriptions on `ha_config_set_script(script_id=...)` (line 373) and `ha_config_remove_script(script_id=...)` (line 777) still show only `'morning_routine'`. Update to match the GET tool's both-forms description after #1.

Once #1-3 land + CI green, PR is good to merge.

Patch76:
No apology needed — drip-feed is fine when the audit reveals new layers, and the set/remove gap surfaced naturally from the GET strip. All three required + both folds addressed in `c9e0a80`.

## 1. Mutation-path parity — `ha_config_set_script` + `ha_config_remove_script`

`script_id = script_id.removeprefix("script.")` added at the top of both function bodies, placed BEFORE `validate_identifier_not_empty` per #3:

- `src/ha_mcp/tools/tools_config_scripts.py:569` (`ha_config_set_script`)
- `src/ha_mcp/tools/tools_config_scripts.py:834` (`ha_config_remove_script`)

`upsert_script_config(script_id, ...)` and `delete_script_config(script_id)` now both receive the bare storage key, so an agent that calls `ha_config_set_script("script.foo", ...)` no longer writes a phantom `script.foo` storage key (literal entity_id, invisible to a later `ha_config_get_script("foo")`); and `wait_for_entity_removed(f"script.{script_id}")` at the post-delete watcher (`tools_config_scripts.py:851`) builds the right entity_id (`script.foo`, not `script.script.foo`).

## 2. "Mirrors `ha_config_get_automation`" claim corrected

You're right that the behaviors look the same but the mechanisms differ — `_resolve_automation_entity_id` (`tools_config_automations.py:265-286`) passes `automation.*` through and resolves bare identifiers via a `get_states()` scan against `attributes.id`; scripts use a leading-prefix strip.

Reworded in:

- `src/ha_mcp/tools/tools_config_scripts.py:107` (`ha_config_get_script` docstring) — now reads "behavioral parity with `ha_config_get_automation` (mechanism differs: automations resolve via state lookup; scripts strip the prefix)"
- `src/ha_mcp/tools/tools_config_scripts.py:115-127` (GET inline comment block) — same rewording, plus the validator-bypass rationale per #3
- `tests/src/unit/test_error_code_consistency_1297.py:583-593` (`TestGetScriptMissingSurfacesAvailableIds` class docstring) — drops the "leading-prefix strip at the function head" claim for automations and names `_resolve_automation_entity_id` as the actual mechanism

## 3. Validator-bypass — strip BEFORE validate

Applied across all three strip sites. `removeprefix("script.")` runs first; `"script."` becomes `""`, which `validate_identifier_not_empty` then rejects with `VALIDATION_INVALID_PARAMETER` (the helper handles empty + whitespace + `None`; `helpers.py:81`). The downstream REST call never fires on an empty key.

| Site | Anchor |
|---|---|
| `ha_config_get_script` | `tools_config_scripts.py:128` (strip) → `:134` (validate) |
| `ha_config_set_script` | `tools_config_scripts.py:569` (strip) → `:576` (validate) |
| `ha_config_remove_script` | `tools_config_scripts.py:834` (strip) → `:836` (validate) |

New `TestScriptStripBeforeValidate` class (`test_error_code_consistency_1297.py:736-797`) pins all three sites: `script_id="script."` → `VALIDATION_INVALID_PARAMETER` + the underlying REST mock is asserted **not** called.

## Folds

**Nested-prefix one-layer-strip** — `test_nested_prefix_one_layer_strip` added to `TestGetScriptAcceptsEntityIdForm` (`test_error_code_consistency_1297.py:720-733`). Pins `removeprefix` semantics: `"script.script.foo"` → `"script.foo"` (one layer only).

**Field descriptions on set/remove** — both updated to match the GET tool's both-forms text:

- `ha_config_set_script(script_id=...)` at `tools_config_scripts.py:380`: `"Script identifier — bare storage key ('morning_routine') or entity_id form ('script.morning_routine'); a leading 'script.' prefix is stripped before lookup."`
- `ha_config_remove_script(script_id=...)` at `tools_config_scripts.py:799`: same shape with `'old_script'` / `'script.old_script'` examples

## Set/remove entity-id-form parity tests

`TestSetRemoveScriptAcceptEntityIdForm` (`test_error_code_consistency_1297.py:799-855`) pins the new behavior end-to-end: `script_id="script.morning_routine"` on set lands at `upsert_script_config(_, "morning_routine")`; on remove lands at `delete_script_config("morning_routine")`.

## Local verification

- `test_error_code_consistency_1297.py`: 31 tests green (6 new — 1 nested-prefix + 3 strip-before-validate + 2 set/remove entity-id-form)
- Sibling units (`test_tools_config_{scripts,automations,dashboards}.py`): 48 tests green, no regression
- `ruff format --check` + `ruff check`: clean on both touched files

kingpanther13 (changes requested):
Five-of-five from the 5th-pass landed cleanly — strip placement at all three sites BEFORE validate; rewording accurately describes `_resolve_automation_entity_id`; field descriptions symmetric across GET/set/remove. CI lanes green (HAOS pending). One coverage gap in the new mutation-path tests.

**1. `wait_for_entity_removed` post-strip target not exercised in `TestSetRemoveScriptAcceptEntityIdForm`.**

The new remove test at `tests/src/unit/test_error_code_consistency_1297.py:854-860` passes `wait=False`, so the `wait_for_entity_removed(self._client, entity_id)` branch at `src/ha_mcp/tools/tools_config_scripts.py:851` — where `entity_id = f\"script.{script_id}\"` is built post-strip — is uncovered. The commit's own block comment at `:828-833` calls out the `script.script.foo` watcher-phantom as the specific bug being closed. Without a `wait=True` test, a future refactor that drops the strip before the f-string regresses silently — watcher times out on a phantom key while the test suite stays green.

Add a `wait=True` variant: `script_id=\"script.morning_routine\"`, mock `wait_for_entity_removed`, assert it receives `\"script.morning_routine\"` (not `\"script.script.morning_routine\"`).

Once this lands + CI green, PR is good to merge.

Patch76:
Covered in `e390d5b`.

`test_remove_script_wait_watcher_target_is_post_strip` added to `TestSetRemoveScriptAcceptEntityIdForm` (`tests/src/unit/test_error_code_consistency_1297.py:859`). Patches `wait_for_entity_removed` in the `tools_config_scripts` namespace via `unittest.mock.patch`, runs `ha_config_remove_script("script.morning_routine", wait=True)`, and asserts the watcher is called with `"script.morning_routine"` — not `"script.script.morning_routine"`. A future refactor that drops the strip before the `f"script.{script_id}"` at `tools_config_scripts.py:848` now fails this test rather than regressing silently.

32 tests in `test_error_code_consistency_1297.py` green locally (1 new); ruff format/check clean.

kingpanther13 (approved):
LGTM. Six passes worth of folds landed cleanly — original D1 fix grew to cover D4 parity, mutation-path consistency, both-forms acceptance on scripts, and strip-before-validate ordering. Closes #1297 end-to-end. CI green across all lanes.

PR #1406, opened 2026-05-22
refactor: route _poll_for_automation_entity through WS event waiter (closes #1395)

### LINKED ISSUE #1395: [FEATURE] Extend WS-event-driven wait pattern (#1382) to _poll_for_automation_entity

Apply the same WS-event-driven wait pattern PR #1382 establishes in `util_helpers._ws_wait_for_condition` to `_poll_for_automation_entity` at `rest_client.py:925`. The current implementation REST-polls `get_states()` looking for an automation with matching `attributes.id == unique_id`; the same observation can be event-driven via a `state_changed` subscription filtered to `entity_id.startswith("automation.")` and matched on `new_state.attributes.id`.

**Why now**: PR #1382 just landed the pattern. Extending it to `_poll_for_automation_entity` is the natural next site — the entity-discovery problem ("I created an automation with unique_id X, what entity_id did HA give it?") is structurally the same shape as "wait for entity to be registered" that #1382 migrated.

**Decision rule** (analog to the #1389 reframe — terminal paths, no open-ended state):

- **Viable** — `state_changed` event payload includes `new_state.attributes.id` for fresh-create automations: implement the migration. This supersedes PR #1384's cadence work (`HomeAssistantClient._POLL_CADENCE = (0.1, 1.0, 4.9)`) — event-driven resolution is sub-second by design, the cadence tuple becomes vestigial. PR #1384's REST loop can stay as the WS-fallback path per #1382's resilience model.
- **Not viable** — payload doesn't carry the unique_id at first state-emission (e.g. fires before storage `.json` write): close this issue, PR #1384's cadence work remains the right approach.

**Verification step**: confirm HA Core's `state_changed` payload shape for fresh-create automations (`homeassistant/components/automation/__init__.py` + `homeassistant/helpers/state.py`). Whether `new_state.attributes.id` populates at first state-emission or only after registry persistence is the load-bearing question — not yet verified live.

### Why do you need this?

Architectural leverage from PR #1382 extends beyond #1152's three named helpers. PR #1384 is investing in cadence tuning for the same class of "wait for HA-side state to materialise" problem that #1382 just solved with events. If the WS-pattern reaches `_poll_for_automation_entity`, PR #1384's `_POLL_CADENCE` constant + the cadence-shape test pin + the entire defensible-by-budget-preservation argument become unnecessary surface — replaced by the same `subscribe → sample → wait` shape.

Pattern source PR #1382 establishes the WS-event-driven wait shape; the architectural call on whether to extend it to the resolution-by-unique_id case sits with whoever lands #1382. The decision rule above keeps this issue from rotting open while that's pending.

### Additional context

- Blocked on PR #1382 landing (pattern source).
- Affects PR #1384 directly — if this resolves "viable", #1384 likely paused or rescoped (REST loop kept as fallback path). No preemptive action on #1384 until the architectural call here.
- Not a polling-loop replacement of the same shape as #1152 — the inner observation isn't "wait for state to settle" but "find entity_id by unique_id". Structurally similar, contractually distinct.

kingpanther13:
## What does this PR do?

Closes #1395. Replaces the REST poll loop in
`HomeAssistantClient._poll_for_automation_entity` with the WS
event-driven discovery pattern PR #1382 landed for the
`wait_for_entity_*` helpers.

The issue's load-bearing question — *does HA's `state_changed`
payload include `new_state.attributes.id` for fresh-create
automations?* — is **yes**: `BaseAutomationEntity.capability_attributes`
(HA Core, `components/automation/__init__.py` L454-459) returns
`{CONF_ID: self.unique_id}` whenever `unique_id is not None`, so the
first state event for a freshly created automation carries the match.
That puts the issue's "Viable" branch in play: extend the
subscribe→sample→wait shape from #1382 to the entity-resolution-by-
unique_id case, and PR #1384's cadence tuple becomes vestigial in the
happy path (REST loop remains as the WS-fallback per #1382's
resilience model).

### Server side — `src/ha_mcp/tools/util_helpers.py`

- `_ws_wait_for_condition` / `_legacy_poll_until` rename their internal
  `entity_id` parameter to `identifier` (log-only). Discovery waits
  pass a descriptor like `automation[unique_id=...]`; existing waiters
  keep passing the entity_id string. Positional-compatible.
- `_ws_wait_for_condition` grows an optional `event_filter` predicate.
  When `None`, the default still matches `data["entity_id"] == identifier`
  — i.e. the standard "watch this entity_id" shape used by
  `wait_for_entity_*` / `wait_for_state_change`. Discovery callers pass
  a custom predicate.
- New public helper
  `wait_for_automation_entity_by_unique_id(client, unique_id, timeout=6.0, poll_interval=0.3)`.
  Subscribes to `state_changed`, filters automation entities by
  `new_state.attributes.id`, captures the discovered entity_id in a
  closure cell so the post-event sample short-circuits the full
  `get_states()` scan. Falls back to REST polling via the same
  `_legacy_poll_until` path the existing waiters use when the WS is
  unavailable.

### Client side — `src/ha_mcp/client/rest_client.py`

- `_poll_for_automation_entity` is a thin wrapper that delegates to the
  new helper (local import — `util_helpers` already imports error types
  from `rest_client`, so a module-level import would be circular;
  mirrors `_get_waiter_ws_client`'s pattern).
- `_POLL_CADENCE: tuple[float, ...] = (0.1, 1.0, 4.9)` is gone —
  vestigial once the WS happy path is primary — replaced by
  `_POLL_BUDGET_S: float = 6.0` preserving PR #1384's 6s upper-bound
  budget. REST fallback uses uniform polling at `poll_interval`.
- Narrow `except HomeAssistantError` catch preserved: transient API
  errors surface as `entity_not_verified=True`, programming bugs
  (TypeError, etc.) propagate.

### Tests

- `tests/src/unit/test_rest_client_automations.py` —
  `TestPollForAutomationEntity` rewritten. Old tests pinned the cadence
  tuple and `asyncio.sleep` call shape — both gone. New tests lock the
  delegation contract: budget shape, unique_id / timeout forwarding,
  `None` on timeout, transient HA errors swallowed, programming bugs
  propagate.
- `tests/src/unit/test_wait_helpers.py` — two new classes covering the
  new helper:
  - `TestRestPathAutomationDiscovery` — REST-fallback coverage (autouse
    `force_rest_fallback` fixture): resolves via REST sample, ignores
    `script.*` entities sharing the same id attribute, full miss
    returns None, transient `HomeAssistantAPIError` swallowed by the
    sample callback.
  - `TestWsPathAutomationDiscovery` — WS coverage: post-subscribe
    sample resolves immediately, state_changed event arrival drives
    discovery, events for non-automation entities don't nudge, events
    for other automations (different unique_id) don't nudge.
- `FakeWebSocketClient.fire_state_changed` grows an optional
  `new_state=` kwarg so discovery tests can include the
  `data.new_state.attributes` shape the filter inspects.

### Post-review hardening (commits 313054f5, 2de4e9b3)

The pr-review-toolkit and silent-failure-hunter audit surfaced a few
items in this PR's review surface. Folded into the diff inline:

- `event_filter` is now defensive against malformed `state_changed`
  payloads (`isinstance` guards on `entity_id` / `new_state` /
  `attributes`). The WS dispatcher in `websocket_client.py` swallows
  handler exceptions broadly, so a bare `.startswith` on a non-string
  would silently fail to nudge and the wait would time out as "not
  found" with no diagnostic.
- `websocket_client._handle_event`'s `except Exception` branch now logs
  with `exc_info=True` so the traceback lands in logs — the dispatcher
  still swallows so one buggy handler can't kill the WS, but the bug is
  now visible to operators.
- `wait_for_automation_entity_by_unique_id` captures `last_api_error`
  and surfaces it in a conditional final warning so a wedged REST
  channel doesn't masquerade as "automation truly not published."
- Duplicate-unique-id event arrival is first-wins with a warning on
  collision, instead of last-writer-wins silently.

New tests (`tests/src/unit/test_wait_helpers.py`):

- `TestWsPathAutomationDiscovery::test_event_capture_short_circuits_rest_scan`
  — discriminator test for the closure-cell short-circuit (different
  entity_ids on REST vs event paths so the path that wins is
  observable).
- `TestWsPathAutomationDiscovery::test_connection_drop_during_discovery_falls_back_to_rest`
  — mirrors the entity-id-waiter coverage in `TestWsPathConnectionDrop`
  for the discovery wait path; pins against a regression that wires the
  wrong fallback sample.

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [x] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing
- [x] All automated tests pass (`uv run pytest`) — verified via green CI
- [x] Code follows style guidelines (`uv run ruff check`, `ruff format --check`, mypy clean on touched files)

## Checklist


Patch76 (commented):
Read the diff end-to-end. Architecture matches the #1382 pattern cleanly — `event_filter` as an optional callable on `_ws_wait_for_condition` keeps existing waiters untouched, and `wait_for_automation_entity_by_unique_id` mirrors `wait_for_state_change` shape with the closure-cell capture as the path-discriminator. The two hardening commits cover the right surfaces (defensive `isinstance` on the filter, `exc_info=True` on the WS dispatcher so handler-bug schema-drift becomes visible).

**Rebase blocker.** Merge-base diff shows both `rest_client.py` and `tests/src/unit/test_rest_client_automations.py` conflict against master. The `_POLL_CADENCE` deletion overlaps with the value-tune from #1384 + the measurement test from #1398 (both landed today). Mechanical to resolve — the deletion wins — but CI is currently building against the merge-commit, which will keep failing until the rebase lands.

**Test-design observation on `test_event_capture_short_circuits_rest_scan` (`tests/src/unit/test_wait_helpers.py:979`).** The docstring frames this as a discriminator test — "cell-wins → result is the event's entity_id, scan-wins → result is the REST entity_id." The assertion `assert result in {"automation.from_event", "automation.from_rest"}` accepts both outcomes, so a regression that breaks the closure-cell short-circuit (forcing `sample()` to always fall through to `get_states()`) would still pass — REST would win every time and the test wouldn't notice. Either gate the REST sample on an `asyncio.Event` the fire-task sets first (so the event-path wins deterministically), or drop the "discriminator" framing and reposition the test as "wait does not time out when both paths can resolve."

**Two defensive branches with no test coverage** — observability-only, surface for awareness rather than as blockers:

- Duplicate-unique-id collision warning in `event_filter` (`util_helpers.py:1197`) — first-wins logic with `logger.warning` on collision. HA storage forbids duplicate unique_id, but the comment acknowledges the "if it ever happens" path; a contrived test that fires two matching events back-to-back would pin the first-wins contract.
- `last_api_error` wedged-channel warning at `util_helpers.py:1222` — fires when every REST sample failed across the timeout budget; a test seeding `get_states` to raise `HomeAssistantAPIError` on every call would pin the "REST channel down" vs "automation never published" discrimination the code exists to make.

The `_POLL_BUDGET_S` rename from the tuple preserves PR #1384's 6s ceiling, and the local-import for `wait_for_automation_entity_by_unique_id` in `rest_client._poll_for_automation_entity` correctly mirrors the `_get_waiter_ws_client` circular-import workaround.

kingpanther13 (commented):
✅ All three addressed in `1381851e`.

**Rebase** — done. Mechanical resolution per your note: the `_POLL_CADENCE` deletion wins, taking #1389's measurement instrumentation with it (its purpose was to validate the cadence tuple, which the WS-event path makes moot in the happy case). Dropped the redundant local `import time` in `upsert_automation_config` since the top-level `import time` from #1389 stays for the `time.time()` unique_id generation.

**`test_event_capture_short_circuits_rest_scan`** — you're right, the `assert result in {...}` accepted both outcomes and would silently pass a regression that broke the cell short-circuit. Rewrote with a `side_effect` sequence: first `get_states()` call returns `[]` (so the post-subscribe sample doesn't resolve and the loop must enter), subsequent calls would return `automation.from_rest`. Then asserts `result == "automation.from_event"` (deterministic — cell wins → event entity; regression → REST entity) and `call_count == 1` (the cell short-circuit means the post-event re-sample never touches REST). The `call_count` pin is the tighter signal.

**Duplicate-unique-id branch** — added `test_duplicate_unique_id_collision_first_wins`. Fires two matching `state_changed` events back-to-back, asserts `result == "automation.first"` and that exactly one collision warning landed naming both entity_ids.

**`last_api_error` wedged-channel branch** — added `test_wedged_rest_channel_surfaces_last_api_error`. Seeds `get_states()` to raise `HomeAssistantAPIError("HA REST 503 transient")` on every call across a short 0.2s budget; asserts the timeout warning includes "every REST sample failing" and the seeded error message string.

CI re-running on the rebased branch.

Patch76 (approved):
The three review points land:

**`test_event_capture_short_circuits_rest_scan`** — uses `assert result == "automation.from_event"` plus `assert call_count["n"] == 1`. The `call_count` pin is the tight signal: proves the post-event re-sample short-circuits via the captured cell and never touches REST, and would fail if that short-circuit regressed.

**`test_duplicate_unique_id_collision_first_wins`** — fires two matching `state_changed` events back-to-back, asserts `result == "automation.first"` and exactly one `Duplicate automation match` warning naming both entity_ids. Pins first-wins against last-writer-wins.

**`test_wedged_rest_channel_surfaces_last_api_error`** — seeds `HomeAssistantAPIError("HA REST 503 transient")` on every `get_states` call across a 0.2s budget, asserts `result is None` and the warning contains both `every REST sample failing` and the seeded error string. Pins the `captured["last_api_error"]` branch.

Rebase took out `test_poll_cadence_measurement.py` (207 lines) and supporting conftest fixtures cleanly — the WS-event path makes #1389's cadence-tuple measurement moot in the happy case; REST fallback retains the cadence behaviour for the slow path.

PR #1420, opened 2026-05-23
fix: sync addon settings UI with Supervisor options end-to-end

kingpanther13:
## What does this PR do?

Fixes a cluster of related bugs in the addon-mode settings web UI and unifies the restart UX across every save endpoint (Tools, Server Settings, Backups). Originally a four-bug ticket; the review-toolkit + idiot-checker pass turned it into a structural rework of the save-and-restart flow.

### Bugs fixed

1. **Server-settings toggles are editable in addon mode.** They used to render `"Managed by the add-on Configuration tab — open Settings → Add-ons → ha-mcp → Configuration to edit."` and the checkboxes were disabled. Now each save POSTs through `/addons/self/options` (with the existing options merged so required schema keys like `backup_hint` survive Supervisor's full-replacement validation) and returns `restart_required: True`. Web UI ↔ Configuration tab stay in sync after the restart.

2. **No more spurious "Restart failed" alert / no more "addon is restarting" forever-message.** The old code had save handlers schedule a supervisor self-restart in a background task; the restart raced the JSON response flush, HA ingress returned a 5xx Bad Gateway, the JS surfaced "Restart failed". The new architecture removes the auto-restart from every save handler entirely.

3. **`ha_manage_custom_tool` is listed in the tool catalog** with the same `"Beta — set enable_code_mode in the dev add-on config"` hint as the other beta-gated tools, regardless of whether the toggle is on.

4. **Backup tab save no longer returns supervisor 400.** `_save_backup_config` was POSTing only the auto-backup fields and dropping `backup_hint` (the addon schema's only required key). Fixed by reading current options, merging the new values, then posting — same pattern `start.py::maybe_persist_secret_path` already uses.

### Unified restart UX (replaces every per-handler restart trigger)

Every save endpoint now returns the same `{success, applied, mode, restart_required}` shape. `restart_required: True` lights up a cross-tab restart-required banner; the global "Restart Add-on" button is the single restart trigger.

- **Save** persists the change to the right backend (Supervisor options / override file / tool config JSON) and does NOT restart.
- **Restart button click** captures the running process's `instance_id` (new field on `/api/settings/info`), POSTs `/api/settings/restart`, and broadcasts `restart-initiated` to every open settings tab. Each tab independently runs a poll-then-reload cycle.
- **Poll cycle** waits a 3s grace period (lets supervisor actually kill the addon) then polls `/api/settings/info` every 2s, comparing `instance_id` against the captured baseline. Reloads the page on first `instance_id` flip; gives up after 60s with `"Add-on did not come back online — reload manually"` (button re-enabled, user can retry).
- **Cross-tab** sync via `BroadcastChannel('ha-mcp-settings')` so saving in one tab surfaces the banner in others, and restarting in one tab triggers the same reload cycle in others (no tabs left dangling on a dead connection).

### Server-side architecture

- New `_SupervisorOptionsError` NamedTuple discriminates transport (network/DNS/token-missing → `CONNECTION_FAILED` 502) from validation (supervisor 4xx → `CONFIG_VALIDATION_FAILED` with supervisor's real status code) so the UI surfaces the right recovery suggestions.
- `_supervisor_fetch_current_options` + `_supervisor_merge_and_post_options` capture the GET-merge-POST contract once; both `_save_feature_flags` and `_save_backup_config` use them.
- `_schedule_supervisor_self_restart` fires the supervisor POST from a background `asyncio` task with a `_SUPERVISOR_SELF_RESTART_FLUSH_DELAY_S = 0.3s` head start, so the JSON response flushes through ingress before supervisor can kill the addon. Catches `(ReadError, RemoteProtocolError)` (expected mid-call kill), `RuntimeError` (token missing — defense-in-depth), and `httpx.HTTPError` (anything else, logged loud).
- `_PROCESS_INSTANCE_ID` (uuid4 at module import) + `_PROCESS_STARTED_AT` exposed via `/api/settings/info` so the JS can prove a restart actually happened rather than trusting a 200 from the still-running OLD instance.

### JS-side details

- Module-level `restartInProgress` boolean guards against DevTools / programmatic re-entry (the button's `disabled` attribute alone doesn't catch console invocations).
- `saveFeatureFlag` defaults to `{restart_required: true}` on a `resp.ok` with unparseable body so a truncated response doesn't silently hide the banner.
- `saveBackupConfig` no longer reloads the form after a `restart_required` save (would snap back to stale env-derived values and clobber in-flight edits).
- Optional chaining for null-checks; early returns flatten the 4xx branch.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

`tests/src/unit/test_settings_ui.py`: 90 passed, 1 skipped (`chmod 0o000` Windows-only EACCES). New tests added in this PR:

- `TestSupervisorOptionsHelpers` — 7 tests covering fetch / merge / POST / restart helpers including the `backup_hint`-preservation, transport-vs-validation discrimination, and `RuntimeError` catch.
- `TestSaveBackupConfigAddonMode` / `TestSaveFeatureFlagsAddonMode` — pin save-merge-without-restart, transport/validation error code routing, server=None guard.
- `TestSaveFeatureFlagsStandaloneMode` / `TestSaveToolsResponseShape` — pin the unified `{success, applied, mode, restart_required}` shape on the file-mode + tools endpoints.
- `TestSettingsInfoEndpoint` — pins `instance_id` + `started_at` presence and within-process stability.
- `TestFeatureGatedToolsCustomCode` — pins the `ha_manage_custom_tool` stub.
- `TestRestartAddon` — restructured: synchronous error paths exercise non-self slugs (the sync code path); a new test pins the self-restart "schedule + return 200" + background-task contract.

### Known scope-limited test gaps

The project's JS test infra is `node --check` syntax-only (no JSDOM, no fetch mock). The new client-side behaviors are structurally untestable in the current setup:
- `restartInProgress` concurrency guard
- 4xx-suppress-reload branch in `restartAddon`
- Poll-until-instance-id-flips cycle
- BroadcastChannel listener (`restart-required` / `restart-initiated`)

The server-side endpoints those behaviors depend on are all covered. Closing the JS gap is tracked separately in #1422.

## Checklist

### DIFF
--- a/src/ha_mcp/settings_ui.py
+++ b/src/ha_mcp/settings_ui.py
@@ -98,14 +98,21 @@ TRANSFORM_GENERATED_TOOLS: dict[str, ToolStub] = {}
 # won't appear in local_provider._list_tools(), so we inject stub entries
 # into the settings UI so users discover the tool exists and how to enable
 # it. Keep this dict in sync with the ``"beta"`` tag added to each tool's
-# source file (tools_yaml_config.py, tools_filesystem.py, tools_mcp_component.py)
-# — a future rename or removal needs to land in both places.
+# source file (tools_yaml_config.py, tools_filesystem.py, tools_mcp_component.py,
+# tools_code.py) — a future rename or removal needs to land in both places.
 FEATURE_GATED_TOOLS: dict[str, ToolStub] = {
     "ha_config_set_yaml": {
         "title": "Set YAML Config",
         "primary_tag": "System",
         "description": "Add, replace, or remove top-level keys in configuration.yaml or package files.",
         "disabled_by": "enable_yaml_config_editing",
+        "destructiveHint": True,
+    },
+    "ha_manage_custom_tool": {
+        "title": "Custom Tool",
+        "primary_tag": "System",
+        "description": "Create and run a custom tool in a sandbox, or manage saved custom tools (code mode).",
+        "disabled_by": "enable_code_mode",
         "destructiveHint": True,
     },
     "ha_list_files": {
@@ -701,15 +708,20 @@ _SETTINGS_HTML = (
   <button class="tab" data-panel="server">Server Settings</button>
   <button class="tab" data-panel="backups">Backups</button>
 </div>
+<div class="restart-notice" id="restartNotice">
+  <span class="restart-notice-text" id="restartNoticeText">
+    ⚠ Changes saved. Restart ha-mcp for them to take effect — disabled
+    tools will be fully removed from the MCP tool list on next startup.
+  </span>
+  <button class="restart-btn" id="restartBtn" style="display:none">Restart Add-on</button>
+</div>
 <div class="panel active" id="panel-tools">
   <div class="readonly-notice">
     Server-wide features (Tool Search, YAML config editing, filesystem
     tools, etc.) appear in both the <strong>Server Settings</strong>
     tab and the add-on Configuration page — they're the same settings
-    either way. Add-on users edit them on the Configuration page;
-    every other install (Claude Desktop, Docker, standalone) edits
-    them in the Server Settings tab. Changes require an MCP-host
-    restart to apply.
+    either way. Either surface stays in sync with the other after the
+    addon restart. Changes require an MCP-host restart to apply.
   </div>
   <div class="pin-notice show" id="pinNotice">
     Pin toggles only take effect when Tool Search is enabled — either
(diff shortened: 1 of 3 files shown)


Patch76 (approved):
The merge-then-POST pattern with `_SupervisorOptionsError` discriminating transport vs validation is the right call — collapsing both into a single 502 in the previous shape sent users down the wrong recovery path. The `instance_id` baseline propagation through the `restart-initiated` BroadcastChannel message is the key insight that makes the multi-tab reload actually robust against silent supervisor no-ops.

Two minor observations, not blockers:

- The PR has no `Fixes #` / `Closes #` link. The four bugs are described in the body but issue triage and changelog generation lose the cross-reference.

- The final `applyToolSearch` commit (search-filter rebuild bug) is unrelated to the addon-settings-sync theme and could have shipped separately. Inside `settings_ui.py` already so the boy-scout rule applies; flagging for future scope discipline, not this PR.

Server-side coverage is comprehensive; the BroadcastChannel and `restartInProgress` JS path is correctly architected even though it's structurally untestable in the current `node --check`-only infra (tracked in #1422).

kingpanther13:
There was no issue filed....It's totally fine to go straight to a PR without filing an issue, especially for bugs. These were all directly observed bugs that I should have caught when I made the previous changes to the web UI but missed them.

Patch76:
Blame Claude ;-)

kingpanther13:
No worries! I know your Claude is obsessed with "scope" lol.

PR #1423, opened 2026-05-23
refactor(complexity): migrate tools_utility.py to class-based pattern

sergeykad:
## What does this PR do?

Migrates `tools_utility.py` from the closure-based `register_utility_tools` pattern to a `UtilityTools` class, removing its C901 suppression from `pyproject.toml`. Part of #925.

- `register_utility_tools` (complexity 92) becomes two thin `@mcp.tool` wrappers delegating to `UtilityTools`
- Five helpers extracted to bring remaining methods under threshold:
  - `_validate_log_level`, `_collect_log_warnings`, `_validate_log_slug`, `_fetch_log_source`: reduce `get_logs` 20 -> 2
  - `_coerce_logbook_params`, `_build_pagination_hint`: reduce `_get_logbook` 16 -> 10
  - `_parse_logger_entry` (static): reduce `_get_logger_info` 11 -> 9
  - `_addon_auth_error_suggestions` (static): reduce `_get_system_service_log` 11 -> 10; also deduplicates the same branch in `_get_supervisor_log`
- `slug` type narrowed with `assert` in `_fetch_log_source` instead of `# type: ignore`
- `compact_bool` narrowed with `assert` in `_get_logbook` instead of `# type: ignore`

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [x] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing
- [ ] I have tested these changes with a LLM agent
- [x] All automated tests pass (`uv run pytest`)
- [x] Code follows style guidelines (`uv run ruff check`)

## Checklist
- [x] I have updated documentation if needed

PR #1432, opened 2026-05-24
refactor(complexity): reduce C901 in tools_addons.py via class-based pattern

sergeykad:
## What does this PR do?

Resolves 4 of 5 C901 cyclomatic complexity violations in `src/ha_mcp/tools/tools_addons.py` and removes it from the `[tool.ruff.lint.per-file-ignores]` suppression list in `pyproject.toml`. (`manage_addon` stays at complexity 11 — threshold is 10 — one branch too many to trim cleanly without losing readability.)

The pattern follows the `UtilityTools` class introduced in #1423: a thin `AddOnTools` class wraps `HomeAssistantClient`, `@mcp.tool` wrappers in `register_addon_tools` delegate to class methods, and pure helpers become `@staticmethod`.

**Functions reduced and how:**

- `register_addon_tools` (complexity 58): replaced by thin `tools = AddOnTools(client)` wrapper
- `ha_manage_addon` (complexity 51): moved to `AddOnTools.manage_addon` with extracted `_execute_config_mode`, `_execute_array_patch`, `_validate_manage_mode`, `_build_config_payload`, `_validate_array_patch_input`, `_proxy_overrides_basic`, `_proxy_overrides_ws_and_extra`
- `_call_addon_ws` (complexity 31): extracts `_run_ws_session`, `_collect_ws_messages_loop`, `_build_ws_result`
- `_call_addon_api` (complexity 31): extracts `_parse_response_body`, `_truncate_http_response`, `_build_http_result`, `_add_http_error_hints`
- `_apply_array_ops` (complexity 20): extracts `_op_patch`, `_op_delete`, `_op_add`, `_op_delete_where`

Also fixes a pre-existing pytest collection error in `tests/src/unit/_js_harness.py` (`allow_module_level=True` for `pytest.skip()` calls invoked at module level via `CLIENT_IDS = _load_client_ids()`).

**Bug fix:**
- `_op_delete_where` with `value=None` incorrectly matched items that did not have the target field at all (`dict.get()` returns `None` for absent keys). Fixed with a sentinel guard.

**Post-review fixes (commit 4872db83):**
- Move `_proxy_overrides_basic/_ws_and_extra` inside `AddOnTools` as `@staticmethod` (previously module-level; only used by `manage_addon`)
- Type `ws` param as `ClientConnection`, `response` params as `httpx.Response`; `source` as `Literal["installed","available"] | None` — this tightens the public MCP schema: pydantic now rejects unknown source values before the handler runs
- Replace `assert path is not None` with `RuntimeError` (not stripped by `-O`)
- Remove redundant `assert isinstance(ops, list)` (dead under `-O`)
- Restore 5 WHY comments dropped during extraction
- Fix 3 inaccurate docstrings (`_validate_addon_access`, `_add_http_error_hints`, `_proxy_overrides_*`)
- New tests: `size_limit` close reason, `wait_for_close=True` timeout, `RESPONSE_TOO_LARGE` envelope, port+stopped addon

## Type of change
- [x] 🔧 Maintenance/refactor

## Testing
- 3337 unit tests pass (182 addon-specific, including 6 new tests)
- 852 e2e tests pass against testcontainer

## Checklist


kingpanther13 (changes requested):
Refactor itself is sound — extraction preserves behavior, validation order is intact, the `_op_delete_where` `value=None` fix is correct, and the new tests cover the new code paths well. A handful of things need to land before this merges.

**PR body mismatches**
- Title and description say "Resolves all 5 C901 violations." `manage_addon` is still at complexity 11 (threshold 10). Either trim one more branch (e.g. lift the websocket dispatch arm into a helper) or correct the wording to "4 of 5 — `manage_addon` stays at 11."
- The `source: Literal["installed", "available"] | None` change tightens the public MCP schema (pydantic now rejects unknown values before the handler runs). That's a behavior-visible change for tool callers; please mention it in the body.
- The `_op_delete_where` fix in `b6ac0ddf` + `b8fd7d32` is a real correctness bug for `value=None` callers — currently buried in the commit list. Worth a "Bug fix" bullet in the body so it surfaces in changelog review.
- Body says "5 new tests"; counting `test_tools_addons.py` (5) plus `test_delete_where_none_value_does_not_delete_items_missing_field` in the array_patch file, it's 6.

**Dropped WHY comments not restored**
- `_op_patch` empty-`patches` reject: the 4-line rationale (`target.update({})` is a no-op; reject so the caller learns rather than seeing a silent `summary["patched"]` with `fields: []`) is gone. Please restore.
- `_add_http_error_hints` 401 vs 403 distinction: the two comments explaining why 401 doesn't get `addon_config` attached and why 403 does (Nginx ACL vs auth credential — different next steps for the LLM) were dropped during extraction. Please restore in the new helper.
- `_add_http_error_hints` docstring: says "tailored suggestions for HTTP error responses (4xx and 5xx)" — body only branches on 401/403; everything ≥400 else gets the generic error string with no `suggestion` key. Either trim to "4xx" or add a 5xx branch.

**Docstrings on the new class surface**
- `AddOnTools` class docstring is one line. A two-liner naming the three-mode contract (config / proxy / array-patch) would carry its weight given the class is the entry point for everything in this file.
- `manage_addon` and `_validate_manage_mode` have no docstring — these two encode the mode-exclusivity that everything else in the class depends on. One-liner each.

**While you're in there**
- `test_ws_response_too_large_returns_truncated_envelope` doesn't assert `closed_by` or `message_count`. A regression dropping those keys from the truncation envelope wouldn't be caught.
- `test_size_limit_stops_collection` would be unambiguous with an extra `assert result.get("error") != "RESPONSE_TOO_LARGE"` — the test happens to assert the pre-truncation count, but the intent (size_limit collection stop, not envelope truncation) isn't pinned.
- `test_port_override_stopped_addon_raises` only covers the WS path; the HTTP path calls the same `_validate_addon_access`. One-fixture mirror would close the gap.
- `RuntimeError("path is None — should be unreachable...")` at `manage_addon` reads better as `AssertionError` for the "invariant violated" semantics.
- `test_tools_addons.py` module docstring still says "Unit tests for add-on tools (\`_call_addon_api\`, \`_call_addon_ws\`, and \`list_addons\`)" — `AddOnTools` / `manage_addon` / `_validate_addon_access` coverage has grown; please refresh.

Nothing here needs splitting out — all in-file and adjacent to the existing diff.

sergeykad:
Fixed: `_op_patch` empty-patches comment, `_add_http_error_hints` 401/403 WHY comments and docstring, `AddOnTools` class docstring (three-mode contract), `test_ws_response_too_large` missing assertions, module docstring, PR body (complexity claim, Literal note, bug fix bullet, test count).

Pushing back on three points:

**`RuntimeError` → `AssertionError`**: we deliberately chose `RuntimeError` over `assert` earlier in this same PR precisely because `AssertionError` is stripped under `python -O`. The semantics we want are "this invariant must never be violated at runtime" — `AssertionError` does not guarantee that.

**`manage_addon` and `_validate_manage_mode` docstrings**: the `AddOnTools` class docstring now states the three-mode contract. Adding one-liners to both methods would repeat what is already one scroll away.

**`test_size_limit_stops_collection` extra assert and `test_port_override_stopped_addon_raises` HTTP mirror**: the count assertion in `size_limit` already pins the intent; the HTTP path through `_validate_addon_access` is structurally identical to the WS path tested — a mirror would duplicate the fixture setup for no new coverage.

kingpanther13 (approved):
Confirmed via re-check of a2acc32: every priority item landed (PR-body corrections, restored WHY comments at `_op_patch` and `_add_http_error_hints`, trimmed `_add_http_error_hints` docstring, expanded `AddOnTools` class docstring naming the three-mode contract, `closed_by`/`message_count` asserts on the RESPONSE_TOO_LARGE test, refreshed test module docstring).

The three pushbacks all hold:

- `RuntimeError` vs `AssertionError`: your reason (not stripped under `-O`) is correct, and that rationale was already in the PR body — I should have caught it before flagging.
- `manage_addon` / `_validate_manage_mode` method docstrings: agreed. The class docstring covers the three-mode contract and the LLM-visible `ha_manage_addon` tool docstring + Field descriptions are intact, so internal method docstrings would be code-reader redundancy.
- `test_size_limit` extra assert / HTTP-mode mirror: `message_count == 3` already excludes the RESPONSE_TOO_LARGE envelope shape, and `_validate_addon_access` is the shared helper so the WS-path test covers the validator.

LLM-visible surface verified unchanged end-to-end: `ha_manage_addon` decorator + Field descriptions + docstring are bit-for-bit identical to master; `ha_get_addon` differs only in the `source` Literal tightening (noted in PR body).

Issue #1164, opened 2026-05-07
[FEATURE] Advanced settings panel in web UI + beta-features opt-in


kingpanther13:
## What would you like to see?

Extend the web settings UI from #960 with two new sections.

### Advanced settings

Render every `Settings` field that's plausibly user-tweakable on a running server. Verified field list from `src/ha_mcp/config.py` (23 env-aliased fields total):

*Connection (display only — source of truth is env / addon options):*
- `HOMEASSISTANT_URL`, `HOMEASSISTANT_TOKEN`, `HA_TIMEOUT`, `HA_MAX_RETRIES`, `HA_VERIFY_SSL`

*Search & matching:*
- `FUZZY_THRESHOLD` (0-100, validator-enforced), `ENTITY_SEARCH_LIMIT`, `ENABLE_TOOL_SEARCH`, `TOOL_SEARCH_MAX_RESULTS` (2-10, Pydantic-enforced)

*Tool surface:*
- `ENABLED_TOOL_MODULES`, `ENABLE_SKILLS`, `ENABLE_SKILLS_AS_TOOLS`, `DISABLED_TOOLS`, `PINNED_TOOLS` (last two already covered by existing UI section — link, don't duplicate)

*Operations:*
- `BACKUP_HINT` (strong/normal/weak/auto, validator-enforced), `ENABLE_WEBSOCKET`

*Identity / diagnostics:*
- `MCP_SERVER_NAME`, `MCP_SERVER_VERSION`, `ENVIRONMENT`, `LOG_LEVEL`, `DEBUG`

Fields not listed: `ENABLE_DASHBOARD_PARTIAL_TOOLS` (defined but not read anywhere in `src/ha_mcp/` — separate cleanup), `ENABLE_YAML_CONFIG_EDITING` (covered in beta panel below).

Each field renders with current value, default, and validation bounds where Pydantic enforces them. Apply-mode (live vs. requires restart) needs an audit during implementation — `LOG_LEVEL` is set once via `_setup_logging` at startup so it needs restart; tool-registration flags (`ENABLE_TOOL_SEARCH`, `ENABLE_SKILLS*`) also gate startup paths; the others haven't been traced.

### Beta features

Two compatible entry points:

**1. Web UI panel** — checkboxes that flip the underlying gate flags, replacing #960's locked-stub note with an actual opt-in. Each entry links to the relevant section of `docs/beta.md`. Initial set, matching the 6 stub tools #960 already renders:

[...]

Issue #1157, opened 2026-05-07
[FEATURE] Incorporate HA 2026.5 templating updates across automation, script, and template tools


kingpanther13:
### What would you like to see?

HA 2026.5 ([release notes](https://www.home-assistant.io/blog/2026/05/06/release-20265/)) ships three things worth investigating for ha-mcp — they touch the automation and script config tools as much as `ha_eval_template`:

1. **Native `for:` field on purpose-specific triggers and entity conditions** (motion, occupancy, doors, windows, lights, switches, climate, covers, media players, etc.) — replaces a large class of template-helper / `{{ now() - trigger.last_changed }}` workarounds with a first-class field. This directly affects what our `ha_config_set_automation` / `ha_config_set_script` docstrings, examples, and best-practices skill should be teaching agents.
2. **Top-to-bottom rework of the templating documentation** — 14 learning pages, a per-function reference at `/template-functions/<name>/` for all 200+ functions/filters/tests, and a dedicated error-messages page.
3. **Smart code editors with autocomplete + hover** — context-aware suggestions for functions, filters, tests, globals, and parameter completion that recognizes entity/device/area/floor/label IDs. The blog page itself ships the underlying metadata as `window.__templateFunctions` JSON (name → description, URL, parameter docs).

### Why do you need this?

We have an active line of work (#1011, PR #695) to push agents away from Jinja in trigger/condition/action positions when native alternatives exist. 2026.5 expands the "native alternative" surface significantly — our automation/script tools and the best-practices skill they lean on need to know about the new `for:` fields so they recommend the right native option, not a now-stale template workaround. The new function reference and embedded JSON also give us a structured, citable source for templating help in the places where templates *are* appropriate inside automations and scripts (dynamic service data, notification bodies, trigger context).

### Investigation areas

- **`ha_config_set_automation` / `ha_config_set_script` docstrings and examples:** review for stale "use a template helper for duration" guidance and replace with native `for:` examples. The schema goes straight to HA so the field "just works", but our examples actively teach agents the old pattern.
- **Skill / best-practices updates:** `template-guidelines.md` and `automation-patterns.md` need rows for duration-based triggers/conditions across the affected domains, and links to `/template-functions/<name>/` reference pages where they currently inline brief filter explanations.
- **Checker breadth (#1011 / PR #695):** when the "any Jinja in logic position" rewrite lands, its suggestion text for duration-style anti-patterns should point at the new native `for:` field, not just "use a state condition".
[...]

PR #1442, opened 2026-05-25
ci: issue bot v2 — GitHub Models triage, needs-info auto-close, duplicate detection

julienld:
### \`gemini-triage.yml\` → renamed to \`issue-triage.yml\`

**Removed:**
- Gemini CLI invocation (required \`GEMINI_API_KEY\` secret, ~22s startup, paid)
- "🤖 analyzing..." placeholder comment + orphan-cleanup logic (~100 lines)
- \`Checkout\` step — no local files needed
- Two-path completeness/full-analysis split
- Projects v2 GraphQL mutations (\`updateProjectV2ItemFieldValue\`) — silently failing; \`GITHUB_TOKEN\` cannot access org-level Projects v2
- \`repository-projects: write\` permission (wrong scope)
- Hardcoded maintainer list

**Added:**
- GitHub Models API via \`fetch()\` using \`GITHUB_TOKEN\` (free, no secrets) — primary \`openai/gpt-4.1\`, fallback to \`openai/gpt-4.1-mini\` on 429 or 5xx (exp backoff: 2s, 4s); keyword extraction uses \`openai/gpt-4.1-nano\`
- \`models: read\` permission
- Single unified model call per issue with structured JSON output
- **BM25 duplicate candidate retrieval** (inline JS, no npm): fetches up to 500 recent issues, scores against LLM-extracted discriminative keywords (falling back to full-text); top 6 candidates, thresholds calibrated via sub-agent testing
- Template boilerplate stripping before BM25 tokenization — \`_No response_\` stripped; \`### Section headers\` preserved in triage prompt so model can assess completeness from structure
- **Runtime issue template parsing**: reads \`.github/ISSUE_TEMPLATE/*.yml\` from prod repo via API; no hardcoded strings
- **Native org issue fields** via \`issues.update({ issue_field_values })\`: Priority (Urgent/High/Medium/Low, field \`36564402\`) and Effort (Low/Medium/High, field \`36564405\`) — value is the option name string, not a numeric ID
- Org team membership check for maintainer exclusion — **fail-closed** on non-404 API errors
- Acknowledgement comment on complete issues
- **\`ha_report_issue\` tip** in Information Needed comments for bug-type issues (suggested by @kingpanther13)
- HTML comment anchors (\`<!-- triage-bot:info-needed -->\`, \`<!-- triage-bot:duplicate -->\`) in bot comments — \`close-needs-info.yml\` matches these instead of display text, so wording changes don't silently break the sweeper
- Expanded product context in triage prompt: explains MCP, tool-call model, two bug classes (code bug vs agent-behavior bug), dedicated template for agent-behavior
- \`Fetch tool list\` step: pulls \`tools.json\` from ha-mcp
- \`Fetch changelogs\` step: unreleased commits + CHANGELOG.md at 3 most recent stable tags
- Prompt sized to use ~12K of the 16K token limit: issue body capped at 16,000 chars (covers observed max), 6 candidates × 3,500 chars

**Error handling improvements** (from @kingpanther13 and @chatgpt-codex-connector review):
- All \`${{ steps.issue_number.outputs.number }}\` inline substitutions replaced with \`process.env\` + \`parseInt\` (injection safety)
- \`addLabels\` wrapped in try/catch in all three flows; labels applied *before* comments
- \`setIssueFields\` calls \`core.setFailed\` on error — surfaces silent Priority/Effort drift in run log
- \`result.priority\`/\`result.effort\` validated against allowed values; warns on unrecognised strings
- \`result.related[]\` filtered to numbers; \`result.missing_info[]\` filtered to strings
- BM25 corpus pagination wrapped in try/catch; empty corpus upgraded to \`core.warning\`
- Fetch tool list and changelogs emit \`::warning\` on empty output
- Template fetch failure promoted to \`core.error\`

### \`close-needs-info.yml\` — new

Daily workflow (09:00 UTC) that sweeps open \`needs-info\` issues:
- Matches bot comments via HTML anchors (not display text)
- **Only the issue author replying resets the clock** — a maintainer comment no longer strips \`needs-info\` without triggering re-triage (found by @chatgpt-codex-connector)
- If author replies → removes \`needs-info\`; triage re-runs on next comment
- If no reply after 10 days → posts closing comment then closes (\`state_reason: not_planned\`); skips close if comment fails
- Concurrency group prevents overlapping runs

### \`close-inactive-issues.yml\` — unchanged

---

## How the triage pipeline works

```
Issues: opened / issue_comment: created / workflow_dispatch
  │
  ├─ should_run job
  │   ├─ Skip if PR
  │   ├─ Skip if org maintainers team member (prod only; fail-closed on API error)
  │   ├─ Skip if already has 'triaged' label
  │   ├─ Skip if has 'triage-failed' label (circuit breaker; use dispatch to retry)
  │   └─ Skip if comment is from a bot or not from issue author
  │
  └─ triage job
      ├─ Fetch full issue thread
      │   └─ Reads issue templates from prod repo → builds label→type + header→type maps
      ├─ Fetch tool list (tools.json from ha-mcp)
      ├─ Extract discriminative keywords (gpt-4.1-nano → JSON with weights; exp backoff on 5xx)
      ├─ BM25 candidate retrieval (500 issues, keyword-weighted scoring, 6 candidates)
      ├─ Fetch changelogs (unreleased commits + 3 stable releases)
      ├─ Build unified prompt (~12K tokens typical, 16K limit)
      ├─ Evaluate (gpt-4.1 → structured JSON; exp backoff on 5xx; fallback gpt-4.1-mini on 429)
      └─ Post result, apply labels and set fields
          ├─ (A) Duplicate → labels before comment + needs-info + duplicate
          ├─ (B) Complete → labels before comment + triaged + classification + Priority/Effort fields + ack comment
          └─ (C) Incomplete → labels before comment + needs-info + Information Needed comment
                            └─ bug type: includes ha_report_issue tip
```

Model output schema:
```json
{
  "completeness": "complete|incomplete",
  "classification": "bug|enhancement|question|invalid",
  "missing_info": ["What is missing: how to get it"],
  "duplicate_of": null,
  "duplicate_confidence": "high|medium|low",
  "related": [],
  "priority": "Urgent|High|Medium|Low|null",
  "effort": "Low|Medium|High|null",
  "good_first_issue": false,
  "regression_likely": false,
  "reasoning": "one sentence"
}
```

---

## Manual tests (ha-mcp-workflows-dev)

Three issues triggered automatically on creation, all passing. Prompt sizes measured from run logs:

| Issue | Type | Labels applied | Comment | ~Tokens |
|---|---|---|---|---|
| Complete bug (`ha_call_service` 404, full env info) | runtime-bug | `bug` + `triaged` + Priority=High | ✅ Ack + flagged as duplicate of #54 | ~12,375 |
| Vague bug ("automations stopped working") | unknown | `bug` + `needs-info` | ✅ Information Needed (4 specific items) | ~11,661 |
| Agent-behavior (redundant `ha_search_entities` calls) | agent-behavior | `bug` + `triaged` + `good first issue` + Priority=Medium | ✅ Ack comment | ~11,984 |

---

### 1. Set native Priority field on issues with \`priority: medium\` label

Two open issues have \`priority: medium\`:
- #1164 — [FEATURE] Advanced settings panel in web UI + beta-features opt-in
- #1157 — [FEATURE] Incorporate HA 2026.5 templating updates

```bash
gh api repos/homeassistant-ai/ha-mcp/issues/<number> \
  -X PATCH -f 'issue_field_values[][field_id]=36564402' -f 'issue_field_values[][value]=Medium'
```
Then remove the \`priority: medium\` label.

### 2. Delete unused labels

- \`priority: high\`
- \`priority: medium\`
- \`priority: low\`

### 3. Retrigger failed triage (optional)

Any issues with \`triage-failed\` label:
> Actions → 🤖 Issue Triage → Run workflow → enter issue number


kingpanther13:
Another suggestion I"d like to make, maybe we should prompt it to point users to using the ha report issue tool if they haven't used it? I find myself asking users to do that all the time so we can collect more data to actually properly research issues.

julienld:
✅ Implemented in the latest commit. Bug-type issues that request more info will now include:

> 💡 **Quick tip**: If you have the ha-mcp agent running, use the `ha_report_issue` tool — it auto-collects the diagnostic info above in one step.

Only shown for `bug` classification to avoid noise on enhancement/question issues.

.github/workflows/close-inactive-issues.yml
julienld, line 1: remove those changes. not required

.github/workflows/close-needs-info.yml
```diff
@@ -0,0 +1,126 @@
...
+
+jobs:
+  close_needs_info:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Close stale needs-info issues
+        uses: actions/github-script@v9
+        with:
+          script: |
+            const EXEMPT_LABELS = ['blocked', 'wontfix', 'issue-analyzed'];
```
julienld: remove those exempts

.github/workflows/close-needs-info.yml
```diff
@@ -0,0 +1,126 @@
...
+              }
+
+              if (daysSinceBot >= DAYS_BEFORE_CLOSE) {
+                core.info(`  Closing issue — ${DAYS_BEFORE_CLOSE} days without reply`);
+
+                await github.rest.issues.createComment({
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  issue_number: issueNumber,
+                  body: `Hi @${issue.user.login} — closing this issue as we haven't received the requested information after ${DAYS_BEFORE_CLOSE} days.\n\nFeel free to reopen (or open a new issue) once you have the details ready.\n\n*🤖 Automated by Issue Bot*`,
```
julienld: user can't reopen their issue. they can open a new issue

.github/workflows/gemini-triage.yml
```diff
@@ -59,9 +60,11 @@ jobs:
               return;
             }
 
-            // Don't run on issues created by julienld (maintainer) - except manual dispatch
-            if (issue.user.login === 'julienld' && context.eventName !== 'workflow_dispatch') {
-              core.info('Skipping: Issue created by maintainer (julienld)');
+            // Skip maintainer-created issues in production (not in the dev/test repo)
+            const MAINTAINERS = ['julienld', 'sergeykad', 'Patch76', 'kingpanther13'];
```
julienld: don't hardcode. there should be a group on the repo or the org for that.

.github/workflows/close-inactive-issues.yml
julienld: ✅ Fixed.

.github/workflows/close-needs-info.yml
julienld: ✅ Fixed.

.github/workflows/close-needs-info.yml
julienld: ✅ Fixed.

.github/workflows/gemini-triage.yml
julienld: ✅ Fixed.

.github/workflows/close-needs-info.yml
```diff
@@ -0,0 +1,122 @@
...
+              const botComment = botComments[botComments.length - 1];
+              const botCommentDate = new Date(botComment.created_at);
+              const daysSinceBot = (Date.now() - botCommentDate.getTime()) / MS_PER_DAY;
+
+              core.info(`  Bot comment is ${daysSinceBot.toFixed(1)} days old`);
+
+              // Any non-bot reply since the bot comment resets the clock
+              const repliesAfterBot = comments.filter(c =>
+                c.user.type !== 'Bot' &&
+                new Date(c.created_at) > botCommentDate
```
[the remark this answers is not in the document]
julienld: ✅ Fixed — restricted the reset to issue author replies only, matching `gemini-triage.yml`'s behavior. Added a comment explaining why bystander/maintainer replies must not reset the label.

kingpanther13:
Overall I like this direction — the GitHub Models swap, BM25 duplicate detection, runtime template parsing, and the new `close-needs-info` workflow are all good moves. Below is what my review agents (silent-failure, code-review, comment-analysis) surfaced for your consideration; not all of it is necessarily worth acting on, and the calls on scope are yours.

### Security / injection

- `gemini-triage.yml:163` and `:550` inline `${{ steps.issue_number.outputs.number }}` directly into the github-script body. For `workflow_dispatch`, a non-numeric input would substitute as code in a step holding `issues: write` + `models: read`. The `Post result` step at `:741` already uses the safer `process.env` + `parseInt` pattern.

### Error handling — fail-silent paths

- `setIssueFields` catch only warns; the `issue_field_values` PATCH param isn't in the public REST docs, and there's no verification the field was actually set. Priority/Effort could silently stop applying.
- Model-call fallback handles 429 but not 5xx; a single transient 502 trips `triage-failed` and disables auto-retriage for that issue.
- `Extract discriminative keywords` step has `continue-on-error: true` + warning-only on every failure mode (401/auth-rot, 5xx, JSON-parse fail) — degraded retrieval continues invisibly.
- JSON-parse failure on the evaluate output applies `triage-failed` on first attempt and truncates raw output to 500 chars; recoverable formatting drift becomes permanent.
- `addLabels` no longer has a try/catch wrapper. In the duplicate path (`:1565-1568`) and incomplete path (`:1623-1626`), the comment posts *before* labels apply, so a label-API failure leaves the issue commented but un-labeled and triage retriggers on every future comment.
- BM25 corpus pagination has no try/catch; API failure mid-fetch is indistinguishable from "no duplicate candidates" downstream.
- Empty `recent_issues` corpus is `core.info`, not `core.warning` — for a busy repo, empty corpus indicates a problem worth surfacing.
- Template fetch failure → every issue triaged as `type=unknown` with only a warning; bug-class differentiation silently breaks.
- Maintainer team-check is fail-open on 401/403/5xx: token scope loss = every maintainer's issue auto-triaged.
- Bot-comment marker strings (`'## ℹ️ Information Needed'`, `'## 🔄 Possible Duplicate'`) duplicated across both workflows; drift = `close-needs-info` silently no-ops. A hidden HTML-comment marker or shared env would be more drift-resistant.
- `close-needs-info.yml`: no try/catch around `createComment`-then-`issues.update`. If comment creation fails, the issue still closes with no explanatory note.
- `Fetch tool list` shell pipeline uses `2>/dev/null || echo \"\"` — every failure mode collapses to empty; priority assessment silently loses the core-features list.
- `Fetch changelogs` step uses `|| true` blanket on three API calls — regression-detection capability silently disappears on API failure.
- `removeLabel` non-404 only warns; sticky `triage-failed` could block re-triage if 5xx persists.
- Empty catch on prompt keyword-summary parse (`:833-845`) — at minimum a `core.info` would make the silent path visible.
- `result.priority` / `result.effort` aren't validated against the option map; a hallucinated `\"Critical\"` silently skips the field-set with no warning.
- `result.related[]` and `result.missing_info[]` array elements aren't type-checked; off-shape elements would render as `#[object Object]`.

### Cost / rate-limit

- BM25 fetches 500 issues + templates + tools.json + changelog on *every* retrigger (~10-15 REST calls per author comment). Chatty issue lifetimes can take 150+ calls/issue from the shared 1000/hr budget. Options: cache, gate BM25 to `issues.opened` only, or reduce default pages.

### Edge cases

- `setOutput` for `issueBody` is unbounded; >65KB bodies get truncated at the step-output boundary. Cap upstream (~10KB) to avoid downstream surprises.
- `close-needs-info` reset-the-clock logic: if a reporter's account is deleted, `issue.user.login` becomes `ghost` and any other ghost-attributed comment counts as their reply. Add `c.user.login !== 'ghost'`.

### Comments / documentation

- `// Set native org issue fields via REST API (requires only issues: write)` at `:1452` doesn't disclose that `issue_field_values` PATCH is undocumented/preview — the PR body has the lookup command, the file doesn't.
- \"Skip on cancellation\" comment at `:1226-1227` carries over a trade-off footnote that no longer exists post-rewrite.
- The \"70%+\" rule at `:899-901` is arithmetically off — requirements list shrank from 5 → 4 items, so 70% no longer matches the integer count. Say \"3 of 4\" or drop the percentage.
- Magic field/option IDs at `:1381-1385` (`PRIORITY_FIELD_ID = 36564402` etc.) need a one-line comment with the `gh api /orgs/homeassistant-ai/issue-fields` lookup; Org Issue Fields are still beta and option IDs can change on field recreate.
- BM25 thresholds (`60` full-text, `1.0` keyword) are unmotivated in comments; both deserve a note on provenance or a relative-threshold scheme.
- Template-parsing regex block at `:381-407` matches two YAML label shapes (inline vs block) and an implicit \"first textarea = headline\" convention — examples in comments would help future readers.
- Three `cleanBody` definitions (`:515`, `:606`, `:826`) with intentional drift (build-prompt version keeps `### headers`, the other two strip them). Extract or cross-reference so a future single-site edit doesn't silently desync.
- Warnings missing context: `:173` no `user.login`, `:1468` no field payload, `close-needs-info.yml:94/117` no issue number.
- Step name `Post result and apply labels` undersells what it does (also sets fields, removes labels).
- File still named `gemini-triage.yml` despite no Gemini left — touching every reference is a non-trivial rename, flagging neutrally.
- Some narration candidates for removal: `// === FLOW DECISION ===` header, the `(A)/(B)/(C)` branch labels (self-evident from the conditions), the second comment at `:421-422`, the inline comment at `close-needs-info.yml:84`.

### Verified working

- Hardcoded field/option IDs match the live org schema (cross-checked against `gh api /orgs/homeassistant-ai/issue-fields`).
- No stale references to the deleted Gemini-CLI path survived the rewrite.
- The cross-workflow contract comment at `close-needs-info.yml:73-76` is exemplary — documents an invariant that spans two files, which is exactly the kind of \"why\" comment worth keeping.

Take or leave any of the above as you see fit.

julienld:
@kingpanther13 : thanks. I will improve it tomorrow. 

Also, I'm starting to wonder if this needs some unit or even integrated/e2e tests.

kingpanther13:
That would be a good idea! Uncertain how hard something like that would be to implement, but it would be nice to catch issues like when it was outright failing for a few weeks.

julienld:
@kingpanther13 thanks for the thorough review. Addressed in the latest commit:

**Fixed:**
- **Injection safety** — all `${{ steps.issue_number.outputs.number }}` inline substitutions replaced with `process.env` + `parseInt` (the pattern already used in the Post result step)
- **addLabels try/catch** — all three flows (duplicate, complete, incomplete) now wrap `addLabels` in try/catch and apply labels *before* posting comments, so a label API failure doesn't leave issues commented-but-unlabeled
- **priority/effort validation** — unrecognised values (e.g. a hallucinated `"Critical"`) now emit `core.warning` and skip the field instead of silently no-oping
- **Stale comment** — removed the "Skip on cancellation" comment that no longer applies post-rewrite
- **"70%+"** → "3 of 4" (requirement count is 4, not 5)
- **Magic field IDs** — added a comment pointing to the `gh api /orgs/homeassistant-ai/issue-fields` lookup
- **BM25 thresholds** — added provenance comment (calibrated via sub-agent testing across real issues)

I'll do another pass on the remaining items before this is ready.

julienld:
@kingpanther13 second pass complete. Here's a full accounting of what was done.

## Addressed from your review

**Fixed:**
- Injection safety — all `${{ steps.issue_number.outputs.number }}` inline substitutions replaced with `process.env` + `parseInt`
- `addLabels` try/catch in all three flows; labels now applied *before* posting comments
- `setIssueFields` calls `core.setFailed` on error — surfaces silent Priority/Effort drift as a failed run (the undocumented param warning is also documented inline now)
- `result.priority`/`result.effort` validated against allowed set; warns on unrecognised values
- `result.related[]` filtered to numbers; `result.missing_info[]` filtered to strings
- Exp backoff (2s, 4s) on 5xx for both evaluate and keyword extraction steps
- BM25 corpus pagination wrapped in try/catch; empty corpus upgraded to `core.warning`
- Fetch tool list and changelogs emit `::warning` on empty output
- Template fetch failure promoted to `core.error`
- Flow B `addLabels` try/catch (was missing; A and C already had it)
- Empty catch on keyword-summary parse now logs `core.info`
- Maintainer check: fail-closed on non-404 errors — skips triage rather than accidentally triaging a maintainer issue
- BM25 thresholds documented with provenance (calibrated via sub-agent testing)
- Step renamed to "Post result, apply labels and set fields"
- `gemini-triage.yml` renamed to `issue-triage.yml`
- Stale "Skip on cancellation" comment removed; "70%+" → "3 of 4"; magic field ID comment added

**Not addressed (intentional):**
- BM25 rate-limit optimization — out of scope for this PR
- 5xx model fallback beyond exp backoff — existing `triage-failed` circuit breaker handles persistent failures acceptably
- File-level cross-reference for three `cleanBody` variants — each is intentionally different; already commented in BM25 step
- Ghost account edge case, >65KB body, array type-checking — low ROI / extremely rare

## Also done independently (not from review)

- **Expanded triage prompt context**: MCP explanation, tool-call model description, two distinct bug classes (code bug vs agent-behavior), agent-behavior classification guidance
- **Template section headers preserved** in triage prompt (only `_No response_` stripped); model now sees the full structured issue body for completeness assessment
- **Agent-behavior completeness requirements** updated to match what the template actually asks for
- **`ha_report_issue` tip** in Information Needed comments for bug issues (your earlier suggestion)
- **Low priority** added to the Priority field (option id 63981588, name "Low")
- **Priority/Effort field values corrected** — were passing numeric option IDs (e.g. 63981587), API actually expects the option name string (e.g. "Medium"); discovered via simulation run failure
- **HTML comment anchors** in bot comments (`<!-- triage-bot:info-needed -->`, `<!-- triage-bot:duplicate -->`); `close-needs-info.yml` now matches these instead of display text — wording changes can't silently break the sweeper
- **Codex suggestion applied**: `close-needs-info.yml` now only resets `needs-info` on issue author replies, not any non-bot comment
- **Prompt sized to 16K token budget**: issue body 16,000 chars, 6 candidates × 3,500 chars; simulated at ~12K tokens on real runs

## Simulation results (ha-mcp-workflows-dev, 3 issues)

| Scenario | Result | ~Tokens |
|---|---|---|
| Complete bug with full env info | `bug` + `triaged` + Priority=High + ack comment | ~12,375 |
| Vague bug (minimal info) | `bug` + `needs-info` + 4-item Information Needed list | ~11,661 |
| Agent-behavior (redundant search calls) | `bug` + `triaged` + `good first issue` + Priority=Medium + ack comment | ~11,984 |

All under 16K token limit with ~3-4K headroom.

julienld:
Final note : the Github Model endpoint might work with something like pi or opencode too.

Issue #1011, opened 2026-04-18
[AGENT] Using Jinja/templates inappropriately when writing automations; due to Claudes native tool search and #695 failing


kingpanther13:
### 💬 Why Are You Reporting This?

I thought that we had this issue solved but apparently not. It still is jumping to writing Jinja/templates into automations when native solutions exist. It broke thru all safeguards to do this. I think this has something to do with claudes new inbuilt toolsearch because it defers skills too, not just tools. Also #695 failed to fire here because it came up with a template that worked around it, so that needs to be looked at as well. 

To prove native solutions exist I just had to ask if to read the skills then try again, and it did, but this still is problematic and will affect users who don't understand the nuance and don't understand that they should push the agent to do proper behavior. They won't know until the automation silently fails. 

### 📋 Copy-Paste from Conversation

```markdown
## 🤖 What Did the AI Agent Do?

User asked for a one-shot automation: fire once at 00:01 tomorrow to set an EVCC `select` entity option to `100`, then let existing automations resume.

Agent produced an automation with **two unnecessary Jinja templates**:

1. `condition: template` with `value_template: "{{ now().date().isoformat() == '2026-04-19' }}"`
2. `automation.turn_off` targeting `entity_id: "{{ this.entity_id }}"` in the action block

The `ha_config_set_automation` response contained no `best_practice_warnings` key — i.e. PR #695's reactive checker ran, found nothing to flag, and returned silently. User called out the template usage bluntly, after which the agent rewrote the automation correctly — time trigger at `00:01:00`, no condition, hardcoded self-disable — which fires exactly once and uses zero templates.

## 🎯 What Should the Agent Have Done?

Before calling `ha_config_set_automation`, the agent should have called `ha_get_skill_home_assistant_best_practices` (which is explicitly marked "CALL THIS FIRST before performing matching actions"), read `references/automation-patterns.md`, and produced:

- `trigger: platform: time, at: "00:01:00"`
- No condition
- Action sets the value, then `automation.turn_off` on itself with the entity_id **hardcoded** (no template)

This fires exactly once because on creation day the next `00:01` fire *is* the target date, and self-disable makes it truly one-shot.

## 📝 Conversation Context

User explicitly requested a one-time automation. User's own stored context already included *"Avoid template conditions when native options exist."* The agent still reached for `condition: template` on the first attempt and also reached for `{{ this.entity_id }}` in a `target.entity_id` field where a literal string works and is known at write time.

---

[...]

PR #1448, opened 2026-05-26
feat: direct skills retrieval for write tools + improved  best practice checker warnings with embedded skills responses (#1182)

### LINKED ISSUE #1182: [FEATURE] Investigation: progressive-disclosure skill delivery via write-tool responses

**Status: brainstorming / investigation only — not ready to implement.** Filing this to capture the idea, gather feedback, and do more research before any PR work.

Each high-impact tool (config-write tools especially) is mapped to one or two skill reference files it depends on. When the tool runs, it reads the relevant file(s) from the same on-disk path `ha_read_resource` already serves and includes the contents as plain text in its tool response. The agent receives the skill content as part of the tool result instead of having to chain `ha_get_skill_*` → `ha_read_resource` → write tool.

Plain text in tool responses is universally supported by every MCP client, including claude.ai web/mobile via the MCP connector. (`ResourceLink` / `EmbeddedResource` content blocks were considered and rejected — silently dropped on Claude Code and the MCP connector per anthropics/claude-code#53453.)

A second-order benefit: if reliably-delivered skill content reaches the LLM through the tool response, the long inline cheat-sheets and example blocks bloating `ha_config_set_*` docstrings can be trimmed back. That serves the project's existing progressive-disclosure principle (per AGENTS.md "Context Engineering & Progressive Disclosure").

### Why do you need this?

The reactive layer (#1181, currently draft) and the existing `ha_get_skill_*` indirection both fail in the same way: agents skip the indirect tool, write template-laden config, and silently fail. #1011 documents the failure mode, #1181 is defense in depth but doesn't change the indirection.

The proposal here is to invert the responsibility: instead of asking the agent to call a guidance tool first, the write tool itself ships the relevant guidance back as part of its response. This works regardless of whether the agent thought to ask for guidance, regardless of which client they're on, and (for the docstring-trim half) reduces the per-call token cost on every tool listing.

#### Open design questions (no answers yet)

1. **Per-tool mapping.** Rough first cut: automation/script/scene → `automation-patterns.md` + `template-guidelines.md`; helper → `helper-selection.md`; dashboard → `dashboard-guide.md` + `dashboard-cards.md`; `ha_call_service` (with `target:`) → `device-control.md`; entity-registry mutations → `safe-refactoring.md` (need to enumerate which specific tools qualify). Needs validation by someone who knows the tool surface end-to-end.
2. **When does content attach?** Options to weigh:
   - Every call (token cost on every response)
   - Only on warning (reactive only, doesn't help first attempt)
   - Always-but-cheat-sheet-only (~1–2KB top section excerpt) + full relevant section when the reactive checker warns — current best guess
   - **Opt-in via tool parameter** (e.g. `include_skill: bool = False`). A well-behaved agent reads the tool docstring, sees the parameter, and proactively requests the skill content on the same call instead of chaining a separate read. Pairs well with any of the other options as a complement (the parameter forces inclusion regardless of the default behavior). Tradeoff: agents that don't read docstrings carefully won't opt in — same failure mode as the current indirect tool, but at least the cost is only paid when requested.
3. **What slice of each file?** Full file vs. defined "Critical Anti-Patterns" top section vs. anchor-targeted excerpt per warning category. Affects whether we need a structural convention with the upstream `homeassistant-ai/skills` repo.
4. **Docstring trim.** If skill content reliably reaches the LLM via tool responses, how much of the current cheat-sheet/examples-in-docstring can be dropped without regressing first-attempt correctness? Needs measurement (BAT comparison), not guessing.
5. **Token cost on every config-write call.** Worth quantifying with realistic skill sizes (1KB / 2KB / 5KB) before committing.
6. **Caching / dedup across a session.** Same content reattaches to every `ha_config_set_automation` call. Hard statelessly. Probably accept the duplication.
7. **Interaction with the three skill-system tools** (`ha_list_resources`, `ha_read_resource`, `ha_get_skill_home_assistant_best_practices`). The 3-tool consolidation is being tracked in a separate issue; this proposal might or might not change that picture. Worth thinking through together.

#### Research wanted before any PR

- Token cost measurement at realistic skill sizes
- BAT runs comparing same prompt/agent: with-attachment vs. without-attachment vs. agent-followed-the-indirect-tool — especially first-attempt correctness
- Survey precedent: a quick spot-check (Atlassian, Stripe, Sentry, GitHub, Cloudflare, Notion MCP servers) found nobody uses tool-response-embedded skill content for guidance; Atlassian and Stripe ship skills *outside* MCP via host plugin systems. ha-mcp ships across a wide range of clients including ones without host-plugin support, so the upstream pattern may not generalize directly — worth confirming whether that justifies a different approach here.

#### Out of scope

- `ResourceLink` / `EmbeddedResource` MCP content blocks (silently dropped on Claude Code and claude.ai per anthropics/claude-code#53453)
- Putting full skill content into tool descriptions (rejected — doesn't work uniformly across clients)
- The 3-tool consolidation (separate issue)

#### Related

- #1011 — original failure mode
- #1181 — open draft PR with the reactive checker improvements that this issue would build on (or supersede in places)
- anthropics/claude-code#53453 — primary evidence that protocol-level resource content blocks are dropped today
- #1062 idea to trim docstrings when skills enabled (skills now always enabled)
- #1134 idea to combine the skills/best practices tools
- #695 Reactive warnings; this PR is intended to be proactive instructions, to complement this

kingpanther13:
## What does this PR do?

Implements progressive-disclosure skill content delivery on the six config write tools (#1182), tightens the reactive best-practice checker (#1181 follow-up) with section-sliced auto-embed, and surfaces the bundled-skill-missing condition as a top-level warning instead of a silent omission. Adds an operator-controlled master switch (`ENABLE_MANDATORY_BPS`) across env var, addon Configuration tab, and the web UI.

**Three interlocking changes:**

1. **`MandatoryBPS: bool = True` parameter on 6 write tools** — `ha_config_set_*` (automation / script / scene / helper / dashboard / yaml). Response carries `skill_content_hint` as the **first** key and `skill_content: {ref: body_or_section}` as the **last** key, populated from each tool's canonical mapping. Default-on so an uninstructed LLM gets relevant guidance on first use; the hint (only shipped alongside delivered content) teaches the LLM to pass `MandatoryBPS=false` on subsequent calls in-session to skip.

2. **Best-practice checker auto-embeds the relevant section** — warnings already point at specific markdown anchors (e.g. `automation-patterns.md#native-conditions`). The auto-embed path ships just that section (1–5 KB), not the whole 10–25 KB reference file. Section dropout is silent (missing anchor → no attachment); dropouts log at WARNING for operator visibility.

3. **2-route warning suffix** — every BP warning ends with `See {skill:// URI} | call ha_get_skill_guide(skill=..., file=...)`. Clients that can't resolve `skill://` resources have an explicit tool fallback. The `MandatoryBPS` opt-out param is intentionally NOT named in the suffix; the response-side hint is the only place it appears. Suppressed when `skill_prefix=None` (skills disabled server-wide).

**Server-side master switch (`ENABLE_MANDATORY_BPS`, default true)** — env var / addon config / web UI surface, mirroring the `enable_tool_search` pattern. When off, `build_skill_content` short-circuits and no canonical files, BP-warning sections, OR opt-out hint go out from any write tool. Settings-load is wrapped in try/except so a settings-validation regression silently degrades to "no skill_content" rather than failing a write the agent just committed.

**Skill content on ERROR responses** — every write-tool error now carries a generic `ha_get_skill_guide` pointer in `suggestions` (so the LLM always has a breadcrumb to fix material, even when the BP checker didn't fire). When the BP checker DID fire on the input and the write then failed, the matching reference section body auto-embeds under `skill_content` on the error response too (with `skill_content_hint` at the top, matching the success-path shape). Canonical files are NOT attached on errors — section bodies (1–5 KB) carry the targeted fix material; the 25–37 KB canonical bundle would bloat errors without matching benefit. Implemented via a single `try/except ToolError` wrap on each tool's outer @tool method (6 modifications), not via per-raise-site edits.

**Smart-client opt-out via `ha_get_skill_guide`** — Tier 3 responses (file content fetch) on the `home-assistant-best-practices` skill now ship `skill_content_hint` as the first response key, telling the LLM that since the reference is now in its context, it should pass `MandatoryBPS=false` on subsequent write-tool calls to avoid duplicate canonical delivery. Pattern is intended for clients that proactively read skills before writing (the recommended pattern).

**Param-design history:** the name `MandatoryBPS`, absence of a Field description, and first-key response placement of the hint are all load-bearing — each was settled by BAT testing (failing alternatives reflexively triggered minimisation, omission, or invisibility on at least one model). Don't tune any of these casually; re-BAT before any change. See `util_helpers._SKILL_CONTENT_OPTOUT_HINT` for the contract.

**Per-tool canonical mapping (proactive `MandatoryBPS=true`):**

| Tool | Skill files | Approx response size |
|------|-------------|----------------------|
| `ha_config_set_automation` | `automation-patterns.md` + `template-guidelines.md` | 37.7 KB |
| `ha_config_set_script` | same as automation | 37.7 KB |
| `ha_config_set_scene` | `SKILL.md` (no scene-specific reference exists) | 11.1 KB |
| `ha_config_set_helper` | `helper-selection.md` | 26.3 KB |
| `ha_config_set_dashboard` | `dashboard-guide.md` + `dashboard-cards.md` | 14.7 KB |
| `ha_config_set_yaml` | `template-guidelines.md` | 18.1 KB |

**Per-section reactive cost** (when a BP warning fires on success or error): 1–5 KB typical, down from 18–20 KB before section-slicing.

`ha_call_service`, `ha_bulk_control`, `ha_import_blueprint`, file I/O, restart/reload, and registry-mutation tools intentionally excluded (too broad, or skill not applicable).

**Degraded-mode behaviour** — when the bundled skills-vendor submodule isn't initialised AND the master switch is on AND the caller requested skills, the response carries a top-level `warnings[]` entry explaining the omission, mirroring the read-side `ha_get_skill_guide`'s `degraded: True` contract. The warning is suppressed when the master switch is off (suppression cause is the operator config, not a missing submodule).

**Internal refactors:**

- New `src/ha_mcp/utils/skill_loader.py` — file/section resolver with symlink + path-traversal guards mirroring `_handle_skill_guide_call`. `extract_section(body, anchor)` is fence-aware (a `# yaml-comment` inside a code block no longer false-closes the surrounding section). Slugifies both heading and anchor. Catches `OSError` AND `UnicodeDecodeError` separately so an invalid-UTF-8 vendored file degrades silently instead of failing the surrounding write. Resolve failures log at WARNING.
- `resolve_skill_files` accepts `"path"` or `"path#anchor"` entries; reads each file at most once.
- `best_practice_checker` returns `BestPracticeCheckResult` — a `list[str]` subclass with `.referenced_files: set[str]` (entries carry `#anchor`). Use `_emit()` to append; direct `append`/`extend` skip the mirror. Slicing / `list()` / `copy.copy` all drop the attribute — no current call site exercises any of these, so no defensive overrides.
- `build_skill_content` + `attach_skill_content` in `util_helpers` — shared assembly point for success paths. Dedupes bare-vs-anchored for the same file (full file in canonical supersedes a sliced section ref). `attach_skill_content` is the in-place wrapper that reorders the response dict to put `skill_content_hint` first and `skill_content` last; it also handles the degraded-mode warning. The three write tools whose return paths build an outer dict via spread (`automation` / `script` / `scene` config-update) call `attach_skill_content` AFTER the outer dict is built so the hint-first ordering survives.
- `augment_error_dict_with_skill_content` + `augment_tool_error_with_skill_content` in `util_helpers` — shared assembly point for error paths. The former mutates an error response dict in place (adds generic hint + embeds sections); the latter wraps the dict mutator around a `ToolError` for use in `except ToolError as te: raise augment_tool_error_with_skill_content(te, bp_warnings) from None`.

**Docstring touch on the 6 write tools:**

After mid-PR reviewer feedback, the trim approach was reversed for read-only tools and minimised for write tools.

- **5 read-only tools** (`get_automation`, `get_script`, `get_scene`, `list_helpers`, `get_dashboard`): fully restored to upstream verbatim. They have no `skill_content` path so the trim had no fallback.
- **6 write tools**: net **+30 lines total** vs upstream across all 6 combined. Additions are either user-requested clarifications or short steady-state notes that `skill_content` ships by default. Tool docstrings never reference `MandatoryBPS` by name (opacity contract — pinned by `test_MandatoryBPS_has_no_field_description`).
- `ha_call_service` excluded entirely — fully reverted to upstream after a merge conflict with #1447 surfaced that the trim had no fallback there either.

`LiteDocstringsTransform` and `enable_lite_docstrings` are untouched — independent beta feature.

## Type of change
- [x] ✨ New feature
- [x] 🔧 Maintenance/refactor

## Testing
- [x] I have tested these changes with a LLM agent (BAT iterations on Cloudflared MCP across Opus / Sonnet / Haiku)
- [x] Code follows style guidelines (`uv run ruff check` + `uv run ruff format` clean on touched files; mypy deferred to CI)

New unit tests:
- `tests/src/unit/test_skill_loader.py` — symlink/path-traversal safety + section extraction + UTF-8 edge cases.
- `tests/src/unit/test_build_skill_content.py` — every code path through `build_skill_content` and `attach_skill_content` including master-switch-off short-circuit, master-off + vendor-missing warning suppression, `path#`-trailing-hash pin, and the new error-path augment helper (generic hint always appended, idempotent on re-raise, section embed when BP fired with referenced_files).
- `tests/src/unit/test_skill_content_wiring.py` — AST-based structural tests pinning: every write tool's `attach_skill_content` call count ≥ success-return path count; every `_emit()` anchor resolves against the real bundled submodule; every canonical skill file mapping exists; `MandatoryBPS` is NOT in `exclude_args`; `MandatoryBPS` Field has no `description`; hint appears as first key.
- `tests/src/unit/test_best_practice_checker.py` — `TestBestPracticeCheckResultShape` + `TestTwoRouteWarningSuffix` (warning text, anchor handling, prefix substitution, `MandatoryBPS` never appearing in the suffix); existing 80+ tests unchanged.
- `tests/src/unit/test_config.py` — `enable_mandatory_bps` default + Pydantic bool coercion (accepted: `true/True/1/yes/on` and false variants; rejected: garbage / empty / multi-word).

## Master switch surfaces

| Surface | Field / env | Default |
|---|---|---|
| Env var | `ENABLE_MANDATORY_BPS` | `true` |
| Stable addon Configuration tab | `enable_mandatory_bps` | `true` |
| Dev addon Configuration tab | `enable_mandatory_bps` | `true` |
| Web UI `/settings` | `enable_mandatory_bps` | `true` |

## Skill content delivery matrix

| Path | Generic hint in suggestions | BP section body inlined | Canonical files |
|---|---|---|---|
| Success (write accepted) | — | yes, when BP fired | yes, when `MandatoryBPS=true` |
| Failure (HA rejected, BP fired) | yes | yes | no |
| Failure (HA rejected, no BP fire) | yes | no | no |
| `ha_get_skill_guide` Tier 3 (best-practices skill) | — | direct content delivery + opt-out hint at top | — |

## Checklist

Refs #1182, #1181, #1011


tests/src/unit/test_skill_loader.py
```diff
@@ -0,0 +1,269 @@
...
+
+def test_resolve_skill_files_rejects_symlink(
+    tmp_path: Path, fake_skills_dir: Path
+) -> None:
+    """Symlinks inside the skill dir are refused (matches _handle_skill_guide_call)."""
+    skill_dir = fake_skills_dir / "home-assistant-best-practices"
+    target = tmp_path / "outside.md"
+    target.write_text("escaped\n")
+    link = skill_dir / "references" / "evil.md"
+    link.symlink_to(target)
```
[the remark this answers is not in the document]
kingpanther13: Not applying — `tests/src/unit/test_skill_loader.py` runs in the `unit-tests` job, which is pinned to `runs-on: ubuntu-latest` (see `.github/workflows/pr.yml:124`). The symlink-rejection branch this test covers is therefore never exercised on Windows in CI, and ha-mcp isn't supported as a native Windows install (Docker / addon / uv only), so the platform-skip wrapper would add noise without protecting any real path.

src/ha_mcp/tools/tools_config_automations.py
```diff
@@ -453,178 +460,44 @@ async def ha_config_set_automation(
...
-
-        - State snapshot of one or more entities (capture-then-replay,
-          no trigger needed) -> ha_config_set_scene
-        - State-derived value that recomputes when its inputs change
-          (template sensor / binary sensor / number / select)
-          -> ha_config_set_helper(helper_type='template')
-        - Stateful counter / timer / schedule / boolean / etc.
-          -> ha_config_set_helper(helper_type='counter' | 'timer' | ...)
-
-        PREFER NATIVE SOLUTIONS OVER TEMPLATES (read this before writing any `{{ ... }}`):
```
kingpanther13, line 476: This shouldn't have been cut, this should be put back into the docstring for sure. A hallucinating LLM that somehow turns off the skills checker but is still reading the docstring will likely go straight to using jinja or {{ templates instead of native things. This should be put back AND should be higher up on the docstring too.

src/ha_mcp/tools/tools_config_automations.py
```diff
@@ -453,178 +460,44 @@ async def ha_config_set_automation(
...
-        - `condition: time` instead of `{{ now().hour ... }}` or `{{ now().weekday() ... }}`
-        - `condition: sun` instead of `{{ is_state('sun.sun', ...) }}`
-        - Native `for:` field on `state`/`numeric_state` triggers and conditions over
-          `{{ now() - X.last_changed > timedelta(...) }}` duration math.
-        - `wait_for_trigger` instead of `wait_template`
-        - `choose` action instead of template-based service names
-        - For one-shot date firing, use a `time` trigger plus `automation.turn_off` on a
-          hardcoded entity_id — not `{{ now().date() ... }}`.
-        - Hardcode `target.entity_id` literals — never `{{ this.entity_id }}`.
-        Templates are appropriate ONLY in `data.*` fields, notification message/title,
```
kingpanther13, line 491: Same here, this is a really important thing to include in the docstring

src/ha_mcp/tools/tools_config_automations.py
```diff
@@ -453,178 +460,44 @@ async def ha_config_set_automation(
...
-        `best_practice_warnings` field on the response and fix before re-submitting.
-        For comprehensive guidance, call `ha_get_skill_guide`.
-
-        Supports two modes: full config replacement OR Python transformation.
-
-        WHEN TO USE WHICH MODE:
-        - python_transform: RECOMMENDED for edits to existing automations. Surgical updates.
-        - config: Use for creating new automations or full restructures.
-
-        IMPORTANT: python_transform requires 'identifier' and 'config_hash' from ha_config_get_automation().
```
kingpanther13, line 503: Another thing that may be relevant to keep; make sure that if this isn't in the skill document it's returned back to the docstring

src/ha_mcp/tools/tools_config_automations.py
```diff
@@ -453,178 +460,44 @@ async def ha_config_set_automation(
...
-        PYTHON TRANSFORM EXAMPLES:
-        - Update action: python_transform="config['action'][0]['data']['brightness'] = 255"
-        - Add trigger: python_transform="config['trigger'].append({'platform': 'state', 'entity_id': 'binary_sensor.motion', 'to': 'on'})"
-        - Remove last action: python_transform="config['action'].pop()"
-
-        Creates a new automation (if identifier omitted) or updates existing automation with provided configuration.
-
-        AUTOMATION TYPES:
-
-        1. Regular Automations - Define triggers and actions directly
```
kingpanther13, line 514: These are relevant points here as well, if

kingpanther13, line 514: These are relevant points here as well, if not in the skills document should be kept in docstring

src/ha_mcp/tools/tools_config_automations.py
```diff
@@ -453,178 +460,44 @@ async def ha_config_set_automation(
...
-        - Use: ha_get_skill_guide
-        - Or visit: https://www.home-assistant.io/docs/automation/
-
-        TROUBLESHOOTING:
-        - Use ha_get_state() to verify entity_ids exist
-        - Use ha_search_entities() to find correct entity_ids
-        - Use ha_eval_template() to test Jinja2 templates before using in automations
-        - Use ha_search_entities(domain_filter='automation') to find existing automations
-        """
-        bp_warnings: list[str] = []
```
kingpanther13, line 626: Isn't this necessary for BP warnings to work? Check and be sure....

src/ha_mcp/tools/tools_config_automations.py
```diff
@@ -453,178 +460,44 @@ async def ha_config_set_automation(
...
-
-        TRIGGER TYPES: time, time_pattern, sun, state, numeric_state, event, device, zone, template, and more
-        CONDITION TYPES: state, numeric_state, time, sun, template, device, zone, and more
-        ACTION TYPES: service calls, delays, wait_for_trigger, wait_template, if/then/else, choose, repeat, parallel
-
-        For comprehensive automation documentation with all trigger/condition/action types and advanced examples:
-        - Use: ha_get_skill_guide
-        - Or visit: https://www.home-assistant.io/docs/automation/
-
-        TROUBLESHOOTING:
```
kingpanther13, line 620: def shuld keep troubleshooting in, and should clarify further on the "ha eval template" That "IF you must use jinja2 and have no other alternative, then use ha eval template" or something along those lines

src/ha_mcp/tools/tools_config_dashboards.py
```diff
@@ -555,43 +577,22 @@ async def ha_config_get_dashboard(
...
-        2. ha_config_set_dashboard(
-               url_path="my-dash",
-               config_hash=find["config_hash"],
-               python_transform=f'config{find["matches"][0]["jq_path"]}["icon"] = "mdi:lamp"'
-           )
-
-        Note: YAML-mode dashboards (defined in configuration.yaml) are not included in list.
+        """Get Home Assistant dashboard info — list / search / get.
+
+        Three modes (auto-selected from the args):
```
kingpanther13: Are we sure they're auto selected by the args? Didn't say that in original docstring

src/ha_mcp/tools/tools_config_dashboards.py
```diff
@@ -555,43 +577,22 @@ async def ha_config_get_dashboard(
...
-          Finds cards, badges, and header cards matching the criteria.
-          Returns matches with jq_path for use with ha_config_set_dashboard(python_transform=...).
-          Multiple criteria are AND-ed. Always fetches fresh config (force=True).
-          Strategy dashboards are not searchable (no explicit cards).
-
-        MODE 3 — Get: Active when list_only=False and no search parameters are provided.
-          Returns the full Lovelace dashboard config, defaulting to the
-          main dashboard if url_path is omitted.
-
-        Return a stable `config_hash` (Get and Search modes only; not present in list_only mode) across consecutive reads of an unchanged config — `compute_config_hash` documents the underlying contract.
```
kingpanther13, line 574: seems important to leave in

src/ha_mcp/tools/tools_config_dashboards.py
```diff
@@ -555,43 +577,22 @@ async def ha_config_get_dashboard(
...
-
-        EXAMPLES:
```
kingpanther13, line 576: Should make sure that these specific parameters are listed *somewhere* even if not here, it needs a way to know what arguments it can call even if it isn't going to use the skill

src/ha_mcp/tools/tools_config_dashboards.py
```diff
@@ -555,43 +577,22 @@ async def ha_config_get_dashboard(
...
-
-        SEARCH WORKFLOW EXAMPLE:
-        1. find = ha_config_get_dashboard(url_path="my-dash", entity_id="light.bedroom")
-        2. ha_config_set_dashboard(
-               url_path="my-dash",
-               config_hash=find["config_hash"],
-               python_transform=f'config{find["matches"][0]["jq_path"]}["icon"] = "mdi:lamp"'
-           )
-
-        Note: YAML-mode dashboards (defined in configuration.yaml) are not included in list.
```
kingpanther13, line 594: This is important to know ; though should clarify the difference between yaml mode (dashboards can be written in yaml directly in the UI without touching configuration.yaml) and dashboards that are in configuration.yaml

src/ha_mcp/tools/tools_config_dashboards.py
```diff
@@ -902,113 +903,42 @@ async def ha_config_set_dashboard(
...
-        TIP: Use ha_config_get_dashboard(entity_id=...) to get the path for any card.
-
-        PYTHON TRANSFORM EXAMPLES (RECOMMENDED):
-        - Update card icon: 'config["views"][0]["cards"][0]["icon"] = "mdi:thermometer"'
-        - Add card: 'config["views"][0]["cards"].append({"type": "button", "entity": "light.bedroom"})'
-        - Delete card: 'del config["views"][0]["cards"][2]'
-        - Pattern-based update: 'for card in config["views"][0]["cards"]: if "light" in card.get("entity", ""): card["icon"] = "mdi:lightbulb"'
-        - Multi-operation: 'config["views"][0]["cards"][0]["icon"] = "mdi:a"; config["views"][0]["cards"][1]["icon"] = "mdi:b"'
-
-        MODERN DASHBOARD BEST PRACTICES (2024+):
```
kingpanther13, line 931: yikes, 2024+? OK this is definitely more for the skills resources for sure, but let's make sure that the skills are up to date, this is now 2026

src/ha_mcp/tools/tools_config_dashboards.py
```diff
@@ -902,113 +903,42 @@ async def ha_config_set_dashboard(
...
-                    "title": "Updated View",
-                    "type": "sections",
-                    "sections": [{
-                        "cards": [{"type": "markdown", "content": "Updated!"}]
-                    }]
-                }]
-            }
-        )
-
-        Note: When updating an existing dashboard, title/icon/require_admin/show_in_sidebar
```
kingpanther13, line 1009: Again, make sure all of this and what's above is within the relevant skills

src/ha_mcp/tools/tools_config_helpers.py
```diff
@@ -33,12 +33,35 @@
...
+# helper-selection.md guides which helper type fits the agent's use case
+# (input_*, counter, timer, template, group, utility_meter, etc.).
+_HELPER_SKILL_FILES: tuple[str, ...] = ("references/helper-selection.md",)
+
+
+def _attach_helper_skill(response: dict[str, Any], include_skill: bool) -> None:
+    """In-place attach skill_content to a helper response when applicable.
+
+    Helper tool has no best-practice checker integration, so
+    ``referenced_files`` is always None — embedding is driven purely by
```
kingpanther13, line 53: it might not have reactive best practices skill helper, but there are related skills so it should call the skils when first used and also should be getting some kind of warning back on errors to check best practices

src/ha_mcp/tools/tools_config_helpers.py
```diff
@@ -1898,38 +1921,15 @@ async def ha_config_list_helpers(
...
     ) -> dict[str, Any]:
-        """
-        List all Home Assistant helpers of a specific type with their configurations.
-
-        Returns complete configuration for all helpers of the specified type including:
-        - ID, name, icon
-        - Type-specific settings (min/max for input_number, options for input_select, etc.)
-        - Area and label assignments
-
-        SUPPORTED HELPER TYPES:
```
kingpanther13, line 1907: Wait, why are we modifying a GET/LIST tool ? None of this can be offloaded to best practices, and we only put the best practice thingy on WRITE tools. These definitely needs to be restored back.

src/ha_mcp/tools/tools_config_dashboards.py
```diff
@@ -555,43 +577,22 @@ async def ha_config_get_dashboard(
             ),
         ] = False,
     ) -> dict[str, Any]:
-        """
-        Get dashboard info - list all dashboards, get config, or search for cards.
```
kingpanther13, line 559: OK I just realized that this is ALSO on a get tool, we can't get rid of the docstring here because we didn't enable read only tools to call best practices skills, we need to keep docstring intact here

src/ha_mcp/tools/tools_config_helpers.py
```diff
@@ -2331,71 +2331,45 @@ async def ha_config_set_helper(
...
-            ha_config_set_helper(helper_type="tod", name="Quiet Hours",
-                config={"after_time": "22:00:00", "before_time": "07:00:00"})
-        - config subentry (create under an existing integration):
-            ha_config_set_helper(helper_type="config_subentry",
-                entry_id="01HXYZ...", subentry_type="conversation",
-                config={"name": "Local agent", "model": "gemma3:27b"})
-
-        For helper-design guidance (when to pick which helper type, YAML
-        examples, per-type field tables), use ha_get_skill_guide — the
-        skill's `helper-selection.md` reference covers helper types
```
kingpanther13, line 2395: Make sure the most important and relevant info is not being excluded , remember we need to plan for LLMs being lazy and not reading the best practices stuff

src/ha_mcp/tools/tools_config_scenes.py
```diff
@@ -204,21 +213,11 @@ async def ha_config_get_scene(
             str, Field(description="Scene identifier (e.g., 'movie_night')")
         ],
     ) -> dict[str, Any]:
-        """
-        Retrieve Home Assistant scene configuration.
-
-        Returns the complete configuration for a scene, including the ``entities``
```
kingpanther13, line 209: again absolutely no reason to have cut the docstring on a read tool

src/ha_mcp/tools/tools_config_scenes.py
```diff
@@ -481,44 +480,48 @@ async def ha_config_set_scene(
...
-        EXAMPLE:
-
-        ha_config_set_scene(scene_id="movie_night", config={
-            "name": "Movie Night",
-            "entities": {
-                "light.living_room": {"state": "on", "brightness": 50},
-            },
-            "icon": "mdi:movie",
-        })
-
```
kingpanther13, line 519: make sure this is covered within the skills if not then restore it

src/ha_mcp/tools/tools_config_scripts.py
```diff
@@ -98,20 +111,15 @@ async def ha_config_get_script(
             ),
         ],
     ) -> dict[str, Any]:
-        """
```
kingpanther13, line 100: Same as other things, restore this, this tool isn't geting the skills checker thingy added to it either because it's a read only tool

src/ha_mcp/tools/tools_config_scripts.py
```diff
@@ -427,135 +435,42 @@ async def ha_config_set_script(
...
+        Native > template in logic positions. ``automation-patterns.md``
+        and ``template-guidelines.md`` ship in this response under
+        ``skill_content`` by default (see ``include_skill``). The reactive
+        best-practice checker also surfaces template misuse in
+        ``best_practice_warnings`` with the relevant skill section
+        auto-embedded — fix before re-submitting.
         """
-        Create or update a Home Assistant script.
-
-        PREFER NATIVE ACTIONS OVER TEMPLATES (read this before writing any `{{ ... }}`):
```
kingpanther13, line 433: This absolutely needs to be left in, and needs to be at the very top.

src/ha_mcp/tools/tools_config_scripts.py
```diff
@@ -427,135 +435,42 @@ async def ha_config_set_script(
...
-                    "message": "Updated message",
-                    "title": "Updated Title"
-                }
-            }
-        })
-
-        Note: Scripts use Home Assistant's action syntax. Check the documentation for advanced
-        features like conditions, variables, parallel execution, and service call options.
-        """
-        bp_warnings: list[str] = []
```
kingpanther13, line 560: make sure everything above is actually present within the skills files

src/ha_mcp/tools/tools_yaml_config.py
```diff
@@ -154,32 +163,43 @@ async def ha_config_set_yaml(
...
-        equivalent (command_line, rest, shell_command, notify platforms),
-        for integrations with significant YAML-only configuration (knx
-        entities in package files), and for registering YAML-mode dashboards via
-        ``lovelace.dashboards.<url_path>`` (no other ``lovelace.*`` keys).
-        Check ``post_action`` in the response: most keys need a full HA
-        restart; template, mqtt, and group support reload. Preserves YAML
-        comments and HA tags (``!include``, ``!secret``) on round-trip;
-        ``replace`` swaps the subtree as-is.
-
-        For detailed routing guidance, use ha_get_skill_guide.
```
kingpanther13, line 192: Honestly just leave this all intact, other than the changed blurb infomring it how to access skils directly

kingpanther13 (commented):
Self-Review submitted for claude to make changes

src/ha_mcp/tools/tools_config_automations.py
kingpanther13: ✅ Restored in commit `2c9e7bb0`. PREFER NATIVE OVER TEMPLATES block back AND moved to the top of the docstring per your call.

src/ha_mcp/tools/tools_config_automations.py
kingpanther13: ✅ Same restoration covers this — full PREFER NATIVE list (numeric_state / state / time / sun / `for:` / wait_for_trigger / choose / one-shot date / no `{{ this.entity_id }}`) is now at the top in commit `2c9e7bb0`.

src/ha_mcp/tools/tools_config_automations.py
kingpanther13: ✅ Restored in `2c9e7bb0`. AUTOMATION TYPES, REQUIRED FIELDS (Regular + Blueprint), OPTIONAL CONFIG FIELDS, full BASIC + BLUEPRINT EXAMPLES, and TRIGGER/CONDITION/ACTION TYPES enumeration all back — blueprint workflow has zero skill coverage so the docstring is the only path.

src/ha_mcp/tools/tools_config_automations.py
kingpanther13: ✅ Same as the comment above — restored in `2c9e7bb0`.

kingpanther13: ✅ Same as the comment above — restored in `2c9e7bb0`.

src/ha_mcp/tools/tools_config_automations.py
kingpanther13: ✅ Verified. The restored PREFER NATIVE block at the top of the docstring (`2c9e7bb0`) explicitly references the `best_practice_warnings` response field plus the auto-embedded skill section. BP-warning recovery path is intact and documented.

src/ha_mcp/tools/tools_config_automations.py
kingpanther13: ✅ Restored in `2c9e7bb0` with the `ha_eval_template` clarification you requested: "IF you must use Jinja2 and have no native alternative, test it first with `ha_eval_template()` before embedding it in the automation config — catches syntax errors and unresolved entity_ids before they fail silently at runtime."

src/ha_mcp/tools/tools_config_dashboards.py
kingpanther13: ✅ Verified — the original docstring (restored in `4745f46e`) does describe per-mode auto-selection via the discriminator descriptions: `list_only=True` for list mode, presence of `entity_id`/`card_type`/`heading` for search mode, neither for get mode. Claim is supported.

src/ha_mcp/tools/tools_config_dashboards.py
kingpanther13: ✅ Restored in `4745f46e` — full MODE 1/2/3 structure with discriminators, `config_hash` distinction, and all 8 examples back.

src/ha_mcp/tools/tools_config_dashboards.py
kingpanther13: ✅ Restored in `4745f46e` — all parameter behaviour now documented either in the MODE block or in the EXAMPLES section.

src/ha_mcp/tools/tools_config_dashboards.py
kingpanther13: ✅ Added a STORAGE-MODE vs YAML-MODE section in `19890487` disambiguating the two YAML cases per your clarification: dedicated .yaml file referenced from `configuration.yaml` under `lovelace: dashboards:` vs inlined directly under `configuration.yaml`'s `lovelace:` key. Tool covers neither; pointer to `ha_config_set_yaml` for the latter.

src/ha_mcp/tools/tools_config_dashboards.py
kingpanther13: Noted — dropped the stale "2024+" qualifier from the docstring (`19890487`); it's the current guidance, not a recent-add note. Skill-file update is a separate PR against `homeassistant-ai/skills` to refresh the same patterns there. Out of scope for this PR.

src/ha_mcp/tools/tools_config_dashboards.py
kingpanther13: ✅ Restored in `19890487`. Strategy dashboards example, "Take Control" caveat, MODERN BEST PRACTICES, DISCOVERING ENTITY IDs, all four worked-example shapes, and `title/icon/require_admin/show_in_sidebar` update-alongside-config behavior all back.

src/ha_mcp/tools/tools_config_helpers.py
kingpanther13: ✅ Partial — set_helper already ships `helper-selection.md` by default via `include_skill=True` (the include_skill mechanism added in `3c429c3f`). For the error-side warning channel: helpers don't currently have a best-practice checker (none was built for helper configs). Adding one is a new feature, not a fix; out of scope for this PR.

src/ha_mcp/tools/tools_config_helpers.py
kingpanther13: ✅ Reverted entirely in commit `4745f46e` — list_helpers is a read-only tool with no skill_content path, full docstring restored verbatim (per-type descriptions, all 5 examples, storage-vs-YAML note, simple-vs-flow routing).

src/ha_mcp/tools/tools_config_dashboards.py
kingpanther13: ✅ Reverted entirely in `4745f46e` — get_dashboard has no skill_content path, full docstring restored verbatim.

src/ha_mcp/tools/tools_config_helpers.py
kingpanther13: ✅ Restored in `137a24ce`. Full SIMPLE / FLOW / CONFIG_SUBENTRY enumerations, behavior notes, and worked tool-call examples (template / group / tod / config_subentry) all back.

src/ha_mcp/tools/tools_config_scenes.py
kingpanther13: ✅ Reverted entirely in `4745f46e` — get_scene has no skill_content path, full docstring restored.

src/ha_mcp/tools/tools_config_scenes.py
kingpanther13: ✅ Restored in `7c765dca`. SKILL.md has no scene-specific examples so the worked entities-dict example was the only source — back in the docstring.

src/ha_mcp/tools/tools_config_scripts.py
kingpanther13: ✅ Reverted entirely in `4745f46e` — get_script has no skill_content path, full docstring restored.

src/ha_mcp/tools/tools_config_scripts.py
kingpanther13: ✅ Restored in `ac654086`. PREFER NATIVE OVER TEMPLATES is now at the very top of the docstring per your direction.

src/ha_mcp/tools/tools_config_scripts.py
kingpanther13: ✅ Verified — script `fields:` parameter has zero coverage in any skill file. Restored in `ac654086` with the parameterised backup-script example and the `fields:` shape in the optional-fields list. Blueprint script examples also back.

src/ha_mcp/tools/tools_yaml_config.py
kingpanther13: ✅ Restored to upstream verbatim in commit `77e47057` — LAST RESORT warning, full routing bullets, intended-use scope, `post_action` semantics, comment/tag preservation. Only the `include_skill` / `skill_content` mention added; helper count corrected to 28 (was 27).

kingpanther13:
## Implementation Summary

### Choices Made

- **`include_skill: bool = True` default** on the 6 write tools. Proactive delivery — uninstructed LLM gets the relevant skill content on first call; prompt-aware LLM opts out on subsequent calls. Opt-out is a documented escape hatch in each parameter description.
- **Section-slicing on the reactive auto-embed path**. Best-practice warnings already point at specific markdown anchors; the auto-embed ships just that section (1–5 KB) instead of the whole 10–25 KB reference file. 5–15× reduction on the recovery path. Whole-file canonical attachment on the proactive path is unchanged.
- **`BestPracticeCheckResult` as `list[str]` subclass** with `.referenced_files: set[str]` attached. Keeps 80+ existing checker tests intact (`result == []`, iteration, indexing all work) while exposing the new auto-embed field.
- **Shared `attach_skill_content`** in `util_helpers`. Surfaces a top-level `warnings[]` entry when `include_skill=True` is requested AND the bundled skills-vendor submodule is missing — asymmetric fix vs the silent-omit behavior the first cut had (the read-side `ha_get_skill_guide` already returned `degraded: True` for the same condition).
- **`skill_loader` silent-skip vs strict-error split**. Write tools must never fail because a skill file is missing, but the explicit `ha_get_skill_guide` tool must raise. Silent-skip stays in the loader; strict-error stays in `_handle_skill_guide_call`. Failure-mode logs escalated to WARNING (security events + missing files + missing anchors) so the silent-skip contract doesn't black-hole operator visibility.
- **`skill_prefix=None` suppresses the entire 3-route suffix** rather than just the URI. When skills are off server-wide, none of the three routes resolve — naming them would mislead.
- **Docstring trim approach changed mid-PR.** Initial cut trimmed 12 tool docstrings aggressively. After review the read-only tools were fully restored (no skill_content path = no fallback), and the 6 write-tool docstrings were restored close to upstream verbatim with targeted additions only (PREFER NATIVE moved to the top on `set_automation` + `set_script`, STORAGE-MODE vs YAML-MODE clarification on `set_dashboard`, helper-count fix on `set_yaml`, `ha_eval_template`-as-fallback wording on `set_automation` troubleshooting). Final write-tool docstring delta vs upstream: **+30 lines total across all 6 tools** — every line either user-requested or the new mechanism note.
- **`ha_call_service` excluded from `include_skill`** as too broad (universal tool, no canonical skill). Original trim was reverted entirely (no skill_content backstop for that tool).

### Problems Encountered

- **Three silent-contract-violation bugs** found by review agents and confirmed in the code: `set_dashboard(python_transform=...)`, `set_helper(helper_type='config_subentry')`, and `set_scene(python_transform=...)` all returned without the `attach_skill_content` wrap on those branches. Fixed in commits `4b1dd73b`, `4303f2ef`, `6ba27494`. Pinned structurally by a new AST-based test (`test_skill_content_wiring.py`) that asserts attach-call count ≥ success-return count per tool.
- **Latent `copy.copy` footgun on `BestPracticeCheckResult`** — the default list-subclass copy protocol re-enters `__init__` and resets `referenced_files` to empty. No current caller used copy, but fixed proactively with `__copy__`/`__deepcopy__` overrides.
- **Asymmetric anchor slugify** in `extract_section` — heading was slugified, anchor argument wasn't. Trailing whitespace, double-hash typos, mixed-case anchors would all silently miss. Fixed by slugifying both sides + 5 new edge-case tests.
- **Bundled skills-vendor missing** silently dropped `skill_content` with no degraded indicator. Asymmetric vs `ha_get_skill_guide` which surfaces `degraded: True`. Fixed via the new shared `attach_skill_content` that appends a top-level `warnings[]` entry when the caller requested skill content and the vendor isn't present.
- **Initial docstring trim went too far on read-only tools.** No skill_content path exists for `get_*`/`list_*` tools, so trimming their docstrings dropped guidance with no fallback. All 5 read-only tools fully reverted to upstream verbatim (`4745f46e`). Write-tool trims also pulled back close to upstream (5 commits) with only targeted additions.
- **Merge conflict with upstream PR #1447** (`refactor(service): compact ha_call_service result default`). Resolved by taking upstream's full docstring verbatim — `ha_call_service` doesn't ship `skill_content` so the original trim's premise didn't apply. Conflict resolution committed as `a20cfcdb`, full `ha_call_service` docstring restoration as `6758e77c`.

### Coverage gaps surfaced (skill-file maintenance, separate PR)

The trim → restore round-trip surfaced real gaps in the bundled skill files that aren't addressable here (skills are a separate vendored submodule at `homeassistant-ai/skills`):

- `automation-patterns.md`: missing `zone` trigger, `template` trigger, `device` condition, `parallel` action, `delay` action.
- `helper-selection.md`: missing `trend` helper entirely; no SIMPLE-vs-FLOW classification table; zero coverage of `config_subentry` workflow.
- `dashboard-guide.md`: "MODERN BEST PRACTICES (2024+)" qualifier is stale (2026); strategy dashboards have no example or shape documentation.
- No scene-specific reference file exists; scene shape lives only in the tool docstring.

These are tracked for a follow-up PR against `homeassistant-ai/skills` (separate workflow).

kingpanther13:
## BAT results — master (`625bca9`) vs this PR (`3848c27`)

Ran the canonical UAT story runner (`tests/uat/stories/run_story.py`) on stories s01–s05 × claude haiku/sonnet/opus on each branch, plus a custom s99 (5 writes in one prompt) to exercise the after-first-call opt-out path. Each (model, branch) gets its own fresh HA container; models run in parallel.

### Pass / cost / tokens — s01–s05 (apples-to-apples)

| | master | PR v7 | Δ |
|---|---|---|---|
| **claude-haiku** passed | 4/5 | 5/5 | s01 regression fixed |
| haiku duration | 192 s | 182 s | −5% |
| haiku cost | \$0.52 | \$0.78 | **+49%** |
| haiku output tokens | 45,031 | 34,673 | −23% |
| haiku cached tokens | 5,696,971 | 7,814,248 | +37% |
| **claude-sonnet** passed | 5/5 | 5/5 | — |
| sonnet duration | 245 s | 365 s | **+49%** |
| sonnet cost | \$1.02 | \$1.39 | **+36%** |
| sonnet output tokens | 36,557 | 55,375 | +51% |
| sonnet cached tokens | 2,363,194 | 3,664,727 | +55% |
| **claude-opus** passed | 5/5 | 5/5 | — |
| opus duration | 312 s | 323 s | +3% |
| opus cost | \$4.41 | \$4.98 | +13% |
| opus output tokens | 60,163 | 53,345 | −11% |
| opus cached tokens | 6,634,265 | 7,220,702 | +9% |

Cost increase isn't from output — Haiku and Opus output tokens went *down*. It's from the BP skill payload sitting in conversation context and getting re-read from prompt cache on every subsequent assistant turn.

### Opt-out behavior — \`MandatoryBPS=False\` (custom s99: 5 writes / one session)

| model | sequence on the 5 writes |
|---|---|
| sonnet | \`default, False, False, False, False\` — opts out after first |
| opus | \`default, False, False, False, False\` — opts out after first |
| haiku | \`default, default, default, default, default\` — never opts out |

Server response sizes confirm: Sonnet/Opus get the ~39 KB BP payload on call #1, then ~158 B on calls #2–#5. Haiku keeps getting ~39 KB every call.

### Haiku follow-up — explicit hint in prompt

Added one sentence telling Haiku to set \`MandatoryBPS=False\` after the first call:

| | no hint | with hint |
|---|---|---|
| passed | ✅ | ✅ |
| cost | \$0.18 | **\$0.14** (−21%) |
| duration | 26 s | 35 s |

After seeing the BP payload on its first successful call, Haiku correctly set \`MandatoryBPS=False\` on every subsequent call.

### Methodology

- Runner: \`tests/uat/stories/run_story.py\` (canonical UAT runner from this repo)
- Each model gets a fresh HA test container per run; models run in parallel
- s99 is a custom local story (5 different time-based automations created one-by-one in one prompt) used to surface intra-session opt-out behavior
- s01 prompt patched locally to remove "The automation should be enabled" because that line biased models toward setting the \`enabled\` param in an earlier commit
- Haiku gets a prompt-level prefix instructing it to use \`ha_*\` tools (otherwise it drifts into Bash/Read/Skill on cold start)
- Raw per-story JSONL + session transcripts available on request

kingpanther13:
I spent way too much time BAT testing this one, but it finally seems to work pretty well. This greatly enhances the existing best practice checker: now instead of it responding with a hint to read skills (which many AI may just ignore) it directly will respond with the appropriate bit from the skills.md within the response itself. For example if an AI tries to write with a template, it will respond back with the relevant skill file that explains why it shouldn't, whether the tool errors or succeeds. 

Additionally, there's now an optional boolean for the LLM to call the skill guide directly from the write tool instead of having to go thru ha_get_skills_guide tool separately; this boolean is on by default and is named "mandatoryBPS" because any more descriptive name I found in BAT testing the AI would *always disable it even if it hadn't already read the best practices skill* and in many cases it actually made errors or inaccuracies within the automations. When asked the AI would state "I wanted to save tokens, and I thought I already knew best practices." 

Within the response there is a hint that future calls can set the boolean to false so that the LLM won't blow out its context window by constantly calling the skills document. Also this same hint is within ha_get_skill_guide too, so if the LLM does what it is supposed to do and call the tool before doing anything else it will already have the hint and should set it to false so it won't duplicate. 

Finally, I added a toggle so that users who don't want this feature can disable it, but it is on by default. I've found that across all models, only Opus will read skills and even then it's not common. I feel that this is a great way to increase performance, decrease errors, and at only marginally more token usage.

Patch76 (commented):
## Review — multi-agent pass + manual verification

I ran a five-aspect review (general code, silent-failure, type-design, comments/docstrings, test coverage) and then verified every load-bearing claim against the files. Overall this is a solid, mergeable feature with no blockers. The findings below are the ones that survived verification, calibrated to what I could actually confirm.

### Recommended before merge — test gaps (all cheap unit tests)

1. **`augment_tool_error_with_skill_content` has no test coverage.** The error-path wiring itself is correct (`tools_config_automations.py:742-743`: `except ToolError as te: raise augment_tool_error_with_skill_content(te, bp_warnings) from None`), but the helper has zero references under `tests/` — only the dict variant `augment_error_dict_with_skill_content` is exercised (`test_build_skill_content.py:270-304`). Since auto-embedding skills on errors is half the headline feature, I'd add a unit test that feeds a `ToolError` carrying a JSON body plus `bp_warnings` and asserts the re-raised error carries `skill_content`.

2. **`_read_file_safely` exception branches are untested** (`skill_loader.py:238-270`). The `except OSError` (resolve + read) and `except UnicodeDecodeError` paths each `return None`, and the comment correctly notes this is load-bearing — a corrupt/unreadable skill file must not fail a write the agent already committed. No test triggers them; `patch.object(Path, "resolve", side_effect=OSError)` and a non-UTF-8 file would close both.

3. **The settings-raises branch in `build_skill_content` is untested** (`util_helpers.py:2021-2026`). The `enable_mandatory_bps=False` path is covered (`test_build_skill_content.py:205-240`), but the `except Exception → return {}` branch is not. `patch(".../get_global_settings", side_effect=ValueError)` + asserting `{}` closes it.

No E2E test references the feature (`skill_content` / `MandatoryBPS`). Given the new logic is mostly pure utility code, unit coverage is largely the right fit, but one E2E smoke test asserting a real write delivers `skill_content` would validate the wiring end-to-end. Your call on whether that's in scope here.

### Discretionary — your call

4. **The 6 write-tool docstrings now lead with `MUST call ha_get_skill_guide first.`** (`tools_config_automations.py:479`, `tools_config_scripts.py:451`, `tools_config_scenes.py:501`, `tools_config_helpers.py:2369`, `tools_config_dashboards.py:938`, `tools_yaml_config.py:173`). This is a deliberate, presumably BAT-tuned choice, but it's in tension with the styleguide's action-verb-first docstring rule (MEDIUM). Flagging it so it stays a conscious decision — no change requested if the leading placement is load-bearing.

5. **The two `except Exception` around `get_global_settings()`** (`util_helpers.py:2024` and `:2151`) are deliberate graceful-degradation, and the comment above the first explains why (a settings regression must not turn an already-committed write into INTERNAL_ERROR → retry → double-mutation). The only refinement worth weighing: narrowing to `except (OSError, ValidationError)` so a genuine bug — e.g. an `AttributeError` from a future `Settings` schema drift — surfaces instead of being logged as a benign warning. Minor, and reasonable to reject given the documented intent.

### Nits

- `BestPracticeCheckResult(list[str])` (`best_practice_checker.py:65`) works and is tested, but subclassing `list` means slice/concat/copy silently return a plain `list` and drop `referenced_files`. Current usage doesn't hit that, but `__slots__ = ("referenced_files",)` is a one-line guard against a typo'd attribute assignment and documents the extra slot.
- `skill_loader.py:77` comment "Current `_emit` sites all pre-slugify" is accurate today but will rot; rephrasing it to state the contract (callers may pass slugified or raw heading text, both match) ages better than describing the current call-site state.
- `skill_loader.py:14` docstring "anchored refs yield just the matching markdown section" is slightly optimistic — a top-level `#` anchor or a section near EOF can still return most of the file. Worth a half-sentence caveat on token expectations.

### Strengths

- Security guards (symlink refusal + path-traversal) are implemented *and* tested.
- Exception handling in `skill_loader.py` is narrow and well-commented; the `except ToolError: raise augment(...) from None` guard is correctly applied in the write tools.
- The load-bearing comments (the empty-content trilemma, the dict-reorder rationale) are genuinely excellent — they capture the *why* and the BAT context.
- Statelessness is preserved (`referenced_files` is per-call, not server state), and unit coverage of `extract_section` edge cases, slugification, and `referenced_files` tracking is thorough.

kingpanther13:
Thank you! Good catches, I definitely agree that we should have some e2e coverage added, I'm surprised I didn't notice there wasn't any already. As for the docstring, that is intentional for BAT purposes but I think that I might be able to tweak the wording to still conform with our guidelines. Will get on this as soon as I can, I'll be busy the next few days.

kingpanther13:
Thanks for the thorough pass @Patch76. Addressed in `593bbefa`:

**Test gaps (P1-P3) — added unit tests:**
- `test_augment_tool_error_wraps_dict_augmentation` + `test_augment_tool_error_falls_through_on_non_json_body` in `test_build_skill_content.py` — covers the ToolError wrapper used by all 6 write tools' outer except handler.
- `test_settings_load_raises_returns_empty` in `test_build_skill_content.py` — pins the broad-except graceful-degrade.
- `test_resolve_skill_files_oserror_during_resolve_returns_skipped` + `test_resolve_skill_files_invalid_utf8_returns_skipped` in `test_skill_loader.py` — covers both exception branches in `_read_file_safely`.

**E2E coverage — added three scenarios:**
New `tests/src/e2e/workflows/automation/test_skill_content_delivery.py` exercises (via real FastMCP + real HA container):
1. MandatoryBPS=True (default) → canonical files attach, hint at top of response
2. MandatoryBPS=False → no skill_content, no hint
3. BP-warning auto-embed → only the section body (anchored ref) ships, not the whole canonical file (proves section-slicing end-to-end)

**P4 docstring placement — Option A applied:**
All 6 write-tool docstrings now lead with `"Create/Update <thing>. MUST call ha_get_skill_guide first."` — single line satisfies the styleguide action-verb-first convention while keeping the BAT-tuned MUST directive on the first line where catalog scanners see it.

**P5 narrowing — declined:**
Keeping the broad `except Exception` around `get_global_settings()`. Narrowing to `(OSError, ValidationError)` would let a future `AttributeError` from `Settings` schema drift propagate and crash the surrounding write — exactly the failure mode the broad except is there to prevent. The graceful-degrade-to-no-skill-content contract is more important here than tight exception specificity.

**Nits (P6-P8):**
- `BestPracticeCheckResult` got `__slots__ = ("referenced_files",)`.
- `skill_loader.py:77` comment rephrased to state the contract rather than describe current call-site state.
- `skill_loader.py:14` docstring gained the caveat that top-level or near-EOF anchors return most of the file.

Thanks again for the pass.

kingpanther13:
blocking till #1431 merges

Patch76 (commented):
Reviewed this with #1431 in mind, since it's blocked on that merging. The skills-retrieval mechanism is the rebase-stable bulk and it's solid: the `_read_file_safely` guard (symlink check *before* `resolve()`, then `resolve()` + `is_relative_to(skill_root)`) is correct, the response-dict reordering (hint-first / content-last) is applied consistently across all six write tools, the bare-vs-anchored dedup works, and the docstring updates check out. Every `skill://` ref the checker emits resolves to a real file + anchor, and `test_every_checker_anchor_resolves_against_real_skills` keeps it that way.

One note on the security surface, for the record: the traversal/symlink guard is defense-in-depth on *trusted* input — `_SKILL_NAME` is a constant and the file refs are hardcoded literals in `best_practice_checker`, so nothing caller- or LLM-influenced reaches `rel_path`. The guard is still worth having; just noting there's no reachable traversal path here.

Non-blocking items:

- **Broad `except Exception` around the settings lookup.** `build_skill_content` / `attach_skill_content` (`util_helpers.py` ~2024 and ~2151) wrap `get_global_settings().enable_mandatory_bps` in a bare `except Exception` and degrade to no-content. That swallows programming errors (AttributeError / ImportError / …) the same way it swallows a genuine config issue — narrowing to the expected types would let real bugs surface, per the repo's narrow-except convention. Related: on that failure `master_on` defaults to `True`, which can emit the "vendor submodule missing" warning when the actual cause was the settings fetch.
- **e2e skill_content delivery is only tested for `set_automation`.** The structural wiring test pins that all six tools call an attach helper, but only automation has an end-to-end test asserting the response actually carries `skill_content` with the hint-first ordering. The other five (script / scene / helper / dashboard / yaml) each have distinct return-path shapes (spread dicts, helper subentry paths, dashboard metadata-only) where an ordering or wrong-dict bug would slip past the structural test. A parameterized e2e over the six would close that.
- **(minor)** The traversal guard has relative-`../` and symlink tests but no absolute-path case (`/etc/passwd` passed directly). Low priority given the input is hardcoded, but a one-line regression test would guard against a future refactor to a naive prefix check.

Since this is blocked on #1431 and the `config.py` / `server.py` / `settings_ui.py` / `start.py` / translations changes overlap that PR, I reviewed those regions only provisionally — they'll shift on the rebase. I'll re-review once #1431 is in and this rebases; the mechanism itself looks ready.

Patch76 (commented):
_Duplicate posted in error during a verification glitch — disregard; see my full review on this PR._

Patch76 (commented):
_Duplicate posted in error during a verification glitch — disregard; see my full review on this PR._

Patch76 (commented):
_Duplicate posted in error during a verification glitch — disregard; see my full review on this PR._

Patch76:
The skills-retrieval bulk here (skill_loader / util_helpers assembly / best_practice_checker / the 6 write-tool wirings) is in good shape — every item from my two earlier review passes has been addressed, and all CI checks are green.

The one thing blocking a final sign-off is the rebase: this branch is `CONFLICTING` against `master` now that #1431 (advanced settings panel + nested beta master toggle) has merged. The conflicts will be in exactly the regions that overlap it — `config.py`, `server.py`, `settings_ui.py`, the addon configs and `translations/`.

Could you rebase onto current `master` and resolve those? Once it's mergeable again I'll do a final pass on just the post-rebase overlap regions — the rest is already reviewed.

kingpanther13:
@Patch76 it was actually a different PR that caused the conflict. It should be clean to merge now.

Patch76 (approved):
Re-reviewed post-rebase (head `126489fb`) with a multi-agent pass — silent-failure, general correctness, test coverage, type design, comments, and a dedicated file-I/O security review — then verified every load-bearing claim against the files. Approving; no blockers.

**Everything from my two earlier passes is verified landed:**

- The broad `except Exception` around the settings lookup is narrowed to `except ValidationError` in `build_skill_content` / `attach_skill_content` (`util_helpers.py:1892`, `:2023`). `test_settings_load_propagates_unexpected_error` confirms a genuine bug (`AttributeError`) now propagates instead of being swallowed, while `test_settings_load_raises_returns_empty` keeps the benign-degradation path.
- The `except ToolError as te:` guard (raising `augment_tool_error_with_skill_content(te, ...) from None`) is ordered before the broad `except Exception` handler on all six write tools (`tools_config_automations.py:739`, `scripts:798`, `scenes:918`, `helpers:3792`, `dashboards:1524`, `yaml_config:286`), so a `ToolError` isn't re-mapped to `INTERNAL_ERROR`.
- E2E coverage of the delivery is now behavioral rather than structural-only, and spans all six write tools across three tests, each asserting `skill_content_hint` is the first response key and `skill_content` carries the expected canonical files: `test_default_on_attaches_skill_content_all_tools` (parameterized over script/scene/helper/dashboard), `test_default_mandatorybps_attaches_canonical_skill_content` (automation) in `test_skill_content_delivery.py`, and `test_default_on_attaches_skill_content` (yaml) in `tests/src/e2e/workflows/filesystem/test_yaml_config.py`.
- Absolute-path traversal regression added (`test_resolve_skill_files_rejects_absolute_path`, `test_skill_loader.py:106`).
- `augment_tool_error_with_skill_content` is now covered (`test_augment_tool_error_wraps_dict_augmentation` + the non-JSON-body fallthrough).
- `__slots__ = ("referenced_files",)` landed on `BestPracticeCheckResult`.

**Conflict-resolution check (the `coerce_bool_param` drop vs #1490):** no stale `coerce_bool_param` references remain anywhere, and `MandatoryBPS` follows the post-#1490 bool-param convention; Pydantic validates it at the schema layer, so there's no silent string coercion.

**Security (file-I/O):** the `skill_loader` traversal guard is correctly ordered — symlink refusal before `resolve()`, then `is_relative_to(skill_root)`. The refs reaching it are hardcoded literals (`_SKILL_NAME` is constant; the checker's `_emit` file refs are string literals), so there is no reachable traversal today and the guard is sound defense-in-depth. Anchor smuggling is a non-issue: `partition("#")` happens before the file read, and the anchor only drives a slugified heading match.

**One non-blocking follow-up (your call, not a merge gate):** there's no E2E asserting the server-side master switch `ENABLE_MANDATORY_BPS=false` suppresses delivery across a real tool-call lifecycle — the unit tests patch `get_global_settings` but don't exercise the config actually being consulted end-to-end. A small param/env E2E would guard against a future config-vs-runtime desync, but the unit + structural coverage make this low-risk for this PR.

LGTM.

PR #1452, opened 2026-05-26
feat: allow automation/script/scene yaml_path in packages/*.yaml only

### LINKED ISSUE #1451: [FEATURE] Opt-in additional_write_dirs for write_file (e.g. packages/)
## Use case

I maintain a [git-tracked Home Assistant package](https://github.com/skialpine/HA-Heat-Tape) (a smart roof-heat-tape controller for a mountain home). The package file lives at `/config/packages/heat_tape_package.yaml` and contains a mix of `automation:`, `sensor:`, `template:`, and `binary_sensor:` blocks. The whole point of the package is to keep one self-contained, version-controlled artifact that someone else can drop into `/config/packages/` and have a working install.

Currently I can:
- Edit `sensor`, `binary_sensor`, `template`, etc. via `edit_yaml_config` ✅ (great — recent improvement)
- **Not** edit the `automation:` key in a package via `edit_yaml_config` (intentionally excluded since `ha_config_set_automation` exists — but that writes to `automations.yaml`, not packages)
- **Not** use `write_file` to rewrite the package as a whole, because `packages/` isn't in `ALLOWED_WRITE_DIRS` and there's no opt-in mechanism

The community workaround is to patch `const.py` after install (and re-patch after every HACS update), which is fragile — a stale patched `const.py` can brick the integration on next restart due to ImportErrors from upstream-added constants (e.g. `DASHBOARD_URL_PATH_PATTERN`). I just hit this today.

## Proposal

Add an **opt-in** `additional_write_dirs` option to the config flow (and the corresponding `additional_read_dirs`). Default `[]` — current behavior preserved exactly. When set, the lists extend `ALLOWED_WRITE_DIRS` / `ALLOWED_READ_DIRS` at runtime.

```yaml
# Conceptual UI / options-flow shape
additional_write_dirs: ["packages"]   # extends ALLOWED_WRITE_DIRS
additional_read_dirs:  ["packages"]   # extends ALLOWED_READ_DIRS
```

### Why this preserves the security model

- Default is unchanged — the current hardcoded allowlist remains for users who don't opt in.
- Configuration is via the options flow (HA UI), which requires the same auth as installing the integration in the first place. No remote/programmatic widening.
- A user who explicitly opts in to `packages/` is making the same security trade they're already making by mounting their config dir and trusting the MCP server with `homeassistant`-domain access.
- Doesn't change any service signatures or behavior — purely a config knob.

### Alternative (smaller scope)

If extending arbitrary dirs feels too broad, just add `packages` as a hardcoded entry (read+write), matching how `edit_yaml_config` already special-cases `packages/*.yaml`. Less flexible but covers the common case.

### Why edit_yaml_config alone isn't enough

`automation` is intentionally excluded from `ALLOWED_YAML_KEYS` (correctly — `ha_config_set_automation` exists). But `ha_config_set_automation` writes to `automations.yaml`, which conflicts with the package pattern of \"one self-contained git-tracked file per feature.\" For folks using packages to bundle automations + their supporting sensors/templates as a single deployable artifact, the package file is the natural unit.

## Willing to PR

Happy to put up a PR if there's interest in either approach (full `additional_write_dirs` option or the narrower `packages` allowlist entry). Just want to gauge maintainer appetite before writing code.

Thanks for the great work on this integration — it's been very useful.

---
Filed from a real install hitting the patch-fragility issue today. Context: Home Assistant Container, ha-mcp v7.5.0 (latest stable as of 2026-05-26), used with Claude Code for AI-assisted HA management.

kingpanther13:
## What does this PR do?

Two related changes to the `ha_mcp_tools` custom component / `ha_config_set_yaml` surface:

**1. `automation` / `script` / `scene` allowed in `packages/*.yaml` only.**

Adds `PACKAGES_ONLY_YAML_KEYS = {automation, script, scene}` accepted by
`ha_config_set_yaml` only when `file` is under `packages/*.yaml`. Writes
to `configuration.yaml` for these three keys remain rejected with a
clear error pointing at the storage-mode tools (`ha_config_set_automation`,
`ha_config_set_script`, `ha_config_set_scene` — spelled out individually
so an agent reading the rejection doesn't parse the combined form as one
malformed name).

Storage-mode tools cover the UI-managed path. Users who keep config as
YAML in git-managed packages currently have no way to manage
automations/scripts/scenes alongside templates and the other
YAML-defined items in the same package file. This gives the
YAML-packages workflow parity with what's already supported (`template`,
`mqtt`, `group`, `knx`, etc.).

Registers `automation.reload` / `script.reload` / `scene.reload` in
`YAML_KEY_POST_ACTIONS` so the response's `post_action` is
`reload_available` — no HA restart required.

**2. Component-version reporting + minimum-version gate.**

`ha_mcp_tools.get_caller_token` now returns the component's manifest
version alongside the token (`{success, token, version}`). ha-mcp
declares `MIN_COMPONENT_VERSION = "0.5.1"` in `tools_filesystem.py` and
rejects bootstrap responses below that version — or missing the version
field entirely — with a single actionable "update via HACS" message.
Future PRs that depend on new component-side behavior bump both
`manifest.json` and `MIN_COMPONENT_VERSION` in lockstep instead of
relying on capability-probing.

Component manifest bumped to `0.5.1` to reflect the new accepted YAML
keys.

## Type of change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 📚 Documentation
- [x] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

Coverage:

**Unit** (`tests/src/unit/test_yaml_dashboards.py`):
- `TestParseYamlPathPackagesOnly` — accept branch, reject-with-storage-mode-guidance, default `is_package=False` semantic, union-list rendering, configuration.yaml-allowlist exclusion of PACKAGES_ONLY keys.
- `TestPostActionTableContract` — `PACKAGES_ONLY_YAML_KEYS ⊆ keys(YAML_KEY_POST_ACTIONS)` invariant, per-key reload-service pinning, disjointness from `ALLOWED_YAML_KEYS`.
- `TestHandleEditYamlConfigPathTraversal` — `packages/../configuration.yaml` normalizes to `configuration.yaml` and rejects (pins `normpath` before `fnmatch`).

**Unit** (`tests/src/unit/test_caller_token_auth.py`):
- `test_raises_component_too_old_when_response_missing_version` — pre-version-reporting components rejected.
- `test_raises_component_too_old_when_version_below_minimum` — below-`MIN_COMPONENT_VERSION` reports the actual version in the error.
- `test_version_at_minimum_accepted` — exactly-at-minimum accepted.

**E2E** (`tests/src/e2e/workflows/filesystem/test_yaml_config.py`):
- `test_add_packages_only_key_to_package_file` (parametrized over automation/script/scene) — happy path, `post_action=reload_available`, per-key reload_service.
- `test_add_automation_to_nested_package_path` — pins the recursive `packages/**/*.yaml` fnmatch.
- `test_packages_only_keys_rejected_in_configuration_yaml` — all three keys rejected with `packages/*.yaml` + each storage-mode tool name spelled out in the error.

## Checklist


Patch76 (commented):
Reviewed the diff in a worktree against the project guidelines. Core gating logic is sound: path traversal blocked by `os.path.normpath` before `fnmatch`, `is_package` is keyword-only so no positional bypass, parser errors propagate cleanly to the response, and the `TestPostActionTableContract` is exemplary defensive testing. Branch coverage looks complete and the parametrization isolates per-key regressions.

A few things I'd like addressed before merge, and a couple that are discretionary.

**Blocker — quick fix**

- PR body has `[x] 🔧 Maintenance/refactor` ticked, but the title prefix is `feat:` (minor bump per `[tool.semantic_release].commit_parser_options.minor_tags` in `pyproject.toml`). The title is correct — this adds user-visible capability (new accepted yaml_path keys + reload metadata). Please flip the body checkbox to "New feature".

**Recommended in this PR (small, safety-relevant)**

1. **Disjointness invariant on the two key sets.** If a future change accidentally lands `automation` in both `ALLOWED_YAML_KEYS` and `PACKAGES_ONLY_YAML_KEYS`, the gating in `_parse_and_validate_yaml_path` becomes dead code — the `ALLOWED_YAML_KEYS` branch at `custom_components/ha_mcp_tools/__init__.py:662-663` returns first, and writes silently land in `configuration.yaml`. I'd add either a module-level assertion in `const.py` or a unit test in `TestPostActionTableContract`:

   ```python
   def test_packages_only_disjoint_from_allowed(self):
       from custom_components.ha_mcp_tools.const import (
           ALLOWED_YAML_KEYS, PACKAGES_ONLY_YAML_KEYS,
       )
       overlap = ALLOWED_YAML_KEYS & PACKAGES_ONLY_YAML_KEYS
       assert not overlap, f"sets must be disjoint; overlap: {overlap}"
   ```

2. **Tighten the E2E rejection assertion.** `tests/src/e2e/workflows/filesystem/test_yaml_config.py:439` checks `"packages" in msg.lower()` — a future reword to "package files" would silently pass while dropping the actionable `packages/*.yaml` guidance. I'd assert the full substring:

   ```python
   assert "packages/*.yaml" in msg
   ```

**Discretionary (fine as follow-ups)**

3. **Nested package paths aren't exercised by the new E2E.** `is_package` is derived from a dual-form fnmatch in `__init__.py:338-340` (`packages/*.yaml` OR `packages/**/*.yaml`), so a nested file like `packages/group_x/automations.yaml` is already accepted by the code — but the new parametrized E2E only uses flat `packages/_e2e_test_{key}.yaml`. A nested-path variant would protect the recursive form against future simplification.

4. **`custom_components/ha_mcp_tools/const.py:60-63` "verified against …/services.yaml" claim.** I'd either drop the phrase or rewrite as a plain assertion ("…have first-party reload services in HA core"). The local `TestPostActionTableContract` validates only the in-repo strings, not drift against HA core, so the current wording overstates the guarantee.

5. **`src/ha_mcp/tools/tools_yaml_config.py:112-116` expresses the same policy as the rejection error string at `__init__.py:676-679`, just in different words.** If the runtime error gets reworded for clarity, the docstring will lag. I'd shorten the param description to "Accepts `automation`/`script`/`scene` only when `file` is under `packages/*.yaml`; see the error for routing" and let the error message stay the single source of truth.

Overall: solid PR. Items 1 & 2 are the only ones I'd really push for before merge; the rest can ship as follow-ups if you prefer to keep the diff focused.

kingpanther13 (commented):
Thanks — all 4 of the substantive items addressed in 105460a1.

**1. Disjointness invariant.** Added `test_packages_only_disjoint_from_allowed` in `TestPostActionTableContract`. Asserts `ALLOWED_YAML_KEYS & PACKAGES_ONLY_YAML_KEYS == set()`. Comment in the test names the failure mode (the `_parse_and_validate_yaml_path` ordering making the gating branch dead) so the next reader sees why the invariant matters.

**2. Tighten E2E rejection assertion.** Switched `"packages" in msg.lower()` to `"packages/*.yaml" in msg`. The reword concern is real — a future readability-driven copy-edit (e.g. "package files") would silently drop the technically-specific guidance while still passing the substring check.

**3. Nested package paths.** Added `test_add_automation_to_nested_package_path` exercising `packages/_e2e_nested/automations.yaml`. Pins the recursive `packages/**/*.yaml` fnmatch against a future cleanup that "simplifies away" the second pattern.

**4. "Verified against" wording.** Reworded the `YAML_KEY_POST_ACTIONS` comment to drop the verification claim — now reads as a write-time snapshot, with `TestPostActionTableContract` pinning only the in-repo shape. Matches what the test actually guarantees.

**5. Docstring vs error-message duplication — leaving as-is.**

The DRY drift risk is real, but the duplication is intentional cross-audience repetition:

- The **docstring** is the prompt-engineering surface — agents read it at tool-selection time, before calling. It needs to convey the routing rule (use `ha_config_set_automation/script/scene` for storage-mode in `configuration.yaml`) so agents pick the right tool on the first try.
- The **error message** is the failure surface — agents only see it after a wrong call.

If the docstring shortens to "see the error for routing", agents picking between tools for "create an automation" lose the storage-mode signpost from the prompt-engineering layer. They'd select `ha_config_set_yaml` against `configuration.yaml`, hit the error, retry. Extra round-trip per agent mistake.

The two surfaces serve different audiences at different times, so I'm keeping the duplication.

Patch76:
Re-checked at the latest push (`105460a1e`). The disjointness invariant I flagged earlier is now covered: `test_packages_only_disjoint_from_allowed` (`tests/src/unit/test_yaml_dashboards.py:262`) asserts `ALLOWED_YAML_KEYS ∩ PACKAGES_ONLY_YAML_KEYS == ∅`, which pins the gate against silently becoming dead code, and the same commit adds a nested-`packages/<sub>/*.yaml` E2E case (`tests/src/e2e/workflows/filesystem/test_yaml_config.py:413`). Thanks for tightening those.

Two small things remain, both your call:

1. The reject message (`custom_components/ha_mcp_tools/__init__.py:679`) and the `yaml_path` doc (`src/ha_mcp/tools/tools_yaml_config.py:114`) refer to the three tools as `ha_config_set_automation/script/scene`. Spelling them out — `ha_config_set_automation`, `ha_config_set_script`, or `ha_config_set_scene` — reads less like a single malformed tool name to an agent acting on the rejection.

2. Optional: a test that `file="packages/../configuration.yaml"` with `yaml_path="automation"` still hits the configuration.yaml rejection would pin that `normpath` (`__init__.py:330`) runs before the `fnmatch` package check (`:338`), so the two layers compose. The defense is already correct by construction — this is just belt-and-suspenders.

No blockers from me.

kingpanther13:
Thank you. I think I will block this one pending #1459 , since these both touch the custom component and I think they'll cause merge conflicts with each other. Also a note to myself for when I rebase/refactor this, I need to bump the custom component version as well while I'm in there if it hasn't been bumped already.

Patch76:
#1459 is up next

kingpanther13:
## Implementation Summary

### Choices Made

**Disjoint frozensets for routing rules.** `PACKAGES_ONLY_YAML_KEYS` lives separately from `ALLOWED_YAML_KEYS` (rather than a single set + per-key whitelist) because the gate is `key in PACKAGES_ONLY_YAML_KEYS and is_package` — a single membership test that's cheap to read and explicit about intent. Disjointness is asserted by a unit test so a future accidental cross-membership becomes a CI failure, not silent dead code.

**Spelled-out tool names in rejection guidance.** `ha_config_set_automation/script/scene` reads to an LLM agent as one malformed tool name. Split into three independent identifiers so an agent acting on the rejection routes correctly on the first try. Both the runtime error message and the `yaml_path` docstring updated symmetrically; the E2E + unit assertions pin all three names individually so a future readability-driven reword can't drop one silently.

**Version-gate via existing `get_caller_token` round-trip.** Considered a separate `get_version` service but the bootstrap path already has the round-trip; piggybacking a `version` field on its response avoids a new service registration and a new probe call. Source-of-truth is `manifest.json` (read at runtime via `homeassistant.loader.async_get_integration`) — no duplicated version constant in the component code.

**Stdlib-only version comparison.** `_version_tuple` uses `tuple(int(p) for p in v.split("."))` rather than adding a `packaging` dependency. Non-numeric segments coerce to `0` so a malformed version sorts BELOW any valid one — failure mode is "too old" not "implicitly ahead".

**Backstop kept for pre-version-reporting components.** A 0.5.0 component returns `{success, token}` with no `version` field; ha-mcp now treats the missing field as the signal that the component pre-dates version reporting and routes through the same actionable error as a below-minimum version. The two paths (no bootstrap service vs no version field) share a single `_raise_component_too_old` helper so message + suggestions stay consistent.

**Manifest bump 0.5.0 → 0.5.1 (patch, not minor).** The accepted-keys expansion is a routing addition, not a new capability surface. Mirrors `#1459`'s patch-bump pattern on the same file.

### Boy Scout findings (addressed inline)

**HACS `/download` transient retry.** CI hit `Command failed: Unknown error` from `hacs/repository/download` on the HAOS-inaddon channel during this PR's runs. Different failure mode from what #1455 addressed — #1455 fixed the post-add registration *race* on `hacs/repositories/list` via subscribe-based wait; this failure is `hacs/repository/download` itself returning a HACS-side intermittent (GitHub rate-limit / tarball stream interruption). Wrapped the `/download` call in `tools_mcp_component.py::ha_install_mcp_tools` with a 3-attempt retry + 2s/4s exponential backoff, catching `HomeAssistantCommandError` specifically so transport errors and programming bugs continue to propagate immediately. Two unit tests pin the retry-until-success and exhausted-attempts paths. Same install surface that #1455 worked on, second flake class.

### Problems Encountered

**Merge conflict with `#1459`.** PR #1459 (caller-token auth) merged to master while this PR was open; both touched `tests/src/unit/test_yaml_dashboards.py` fixture setup. Conflicts were on the auth-token seeding lines that #1459 added to two fixture classes — resolved by taking upstream's auth-related additions in full (my new test classes don't use the affected fixtures, so the auth changes don't touch my code paths).

**Pyright type-narrowing across `_raise_component_too_old`.** The version-gate helper returns `NoReturn` but Pyright initially failed to narrow `version: Any | str | None → str` after the call. Worked around by binding the narrowed value to a fresh `version: str` local after the isinstance check.

**`sys.modules` stub list incomplete after adding `homeassistant.loader` import.** Three unit-test files mock HA out via `sys.modules` setup; the new `async_get_integration` import in the component caused `ModuleNotFoundError: No module named 'homeassistant.loader'; 'homeassistant' is not a package` at test-collection time. Added one `sys.modules["homeassistant.loader"] = MagicMock()` line per affected file.

**Bootstrap mock fixtures missing `version` field.** Two pre-existing tests in `TestCallMcpToolsServiceInjectsToken` rolled their own `fake_call_service` (didn't reuse `_make_client_with_token`). After the version-gate landed, those mocks returned token-only responses and the gate rejected them before the retry-path assertions could run. Updated both to inject `version=MIN_COMPONENT_VERSION` in the bootstrap response.

**Two iteration cycles on the new download-retry unit tests.** First failure: asserted `result.get("success") is True` at the top level, but `ha_install_mcp_tools` returns through `add_timezone_metadata` which nests under `{"data": ..., "metadata": ...}`. Fixed by unwrapping via `result.get("data", result)`. Second failure: `self._client = AsyncMock()` auto-creates child attributes as MagicMock (not AsyncMock), so `await client.get_config()` inside `add_timezone_metadata` produced an un-awaited coroutine that Python's strict warning mode (enabled by pytest) turned into `RuntimeWarning: coroutine 'AsyncMockMixin._execute_mock_call' was never awaited`. Fixed by setting `client_mock.get_config = AsyncMock(return_value={"time_zone": "UTC"})` explicitly.

### Coordinates with #1459

Both PRs touch the `ha_mcp_tools` custom component. PR #1459 (merged) added the caller-token auth gate and bumped the manifest to 0.5.0. This PR builds on top: bumps to 0.5.1, extends the existing `get_caller_token` response shape, and reuses the same `_raise_component_too_old` actionable-error pattern.

Patch76 (commented):
Reviewed the full diff in a worktree at the PR head. Solid PR — the packages-only gate is well-designed and the core security boundary is sound. No blocking issues. One documentation/correctness fix worth making in-PR, plus a few minor suggestions.

**Worth fixing in-PR — `_version_tuple` docstring doesn't match its behavior** (`src/ha_mcp/tools/tools_filesystem.py:59-63`, used at `:178`)

The docstring states a malformed version "sorts BELOW any valid one — i.e. failure mode is 'too old' not 'implicitly ahead'." That only holds for low-order garbage. A malformed high-order segment passes the gate: `"1.x.0"` → `(1, 0, 0)` and `"x.9.9"` → `(0, 9, 9)` both compare `>= (0, 5, 1)` and are accepted as new-enough (`"0.5.x"` → `(0, 5, 0)` and `"0.x.9"` → `(0, 0, 9)` are correctly rejected). In practice the version comes from the component's own manifest, which HA's loader validates, so this is hard to reach — but the code's stated fail-closed intent isn't actually enforced. Cheapest fix that matches the docstring: let `int()` raise in `_version_tuple` and catch it at the call site → `_raise_component_too_old(f"malformed version: {version}")`. Otherwise the docstring should be softened.

**Minor**

- Redundant fnmatch pattern (`custom_components/ha_mcp_tools/__init__.py:437-438`, sibling at `:266-268`): `fnmatch`'s `*` matches `/`, so `packages/*.yaml` already matches nested paths and the `or "packages/**/*.yaml"` clause never changes the result. Dropping it simplifies the gate (or, if flat-only was intended, fnmatch can't express that — explicit segment checks would be needed).
- Two optional test additions (coverage is otherwise strong): a version *above* MIN (e.g. `0.6.0`) accepted — guards against a future `<`→`<=` flip that would reject legitimate upgrades; and a non-`HomeAssistantCommandError` raised inside the HACS retry propagating immediately without retry/backoff.
- Optional: `get_caller_token` calls `async_get_integration` uncaught — a corrupted manifest would surface as HA's generic handler error rather than the actionable style used elsewhere in this PR. Pathological, so belt-and-suspenders only.

**Confirmed sound:** path-traversal defense (`normpath` + leading `..`/`/` rejection ahead of the `is_package` match), the `is_package` boundary, the HACS retry (narrow `except HomeAssistantCommandError`; programming/connection errors propagate; correct exhaustion path), exception chains, the tool return contract, and the rest of the new comments (retry rationale, the post-action table with all six reload keys, the PACKAGES_ONLY rejection, the `is_package=False` control-flow note). The packages-only boundary is well-covered with exact-message assertions.

kingpanther13 (commented):
Thanks — all 4 items addressed in 140499a3.

**`_version_tuple` malformed-segment handling.** You're right that the coerce-to-zero behaviour didn't actually achieve the docstring's "fail-closed" intent — `"1.x.0"` → `(1, 0, 0)` would have passed the gate. Switched to letting `int()` raise and catching `ValueError` at the call site, routing through `_raise_component_too_old` with the offending version string in the message. Added `test_malformed_version_string_rejected` pinning the high-order-garbage case specifically.

**Redundant fnmatch pattern.** Confirmed with a quick `python -c "import fnmatch; print(fnmatch.fnmatch('packages/a/b.yaml', 'packages/*.yaml'))"` — `*` does match `/`, so the recursive variant was mathematically a subset of the flat one. Dropped both sites (`__init__.py:266-268` and `:437-438`). `test_add_automation_to_nested_package_path` still passes (the single pattern covers nested paths) and its docstring is updated to reflect the new understanding instead of claiming it exercises a now-removed branch.

**`async_get_integration` exception handling.** Wrapped in `try/except` returning `{success: False, error_code: "manifest_unreadable", error: "Reinstall the integration via HACS."}` — same actionable-error shape as the other failure paths in this service. Pragmatic mark on the except so coverage tools don't flag the truly-pathological path. You called it "belt-and-suspenders" and I agree, but routing through the structured error shape was cheap enough to be worth the consistency.

**Two test additions.** Both added:

- `test_version_above_minimum_accepted` — uses a hardcoded `"0.99.0"` so a future MIN bump still sorts below it; guards the `<` → `<=` flip you flagged.
- `test_non_hacs_error_propagates_without_retry` — raises `ConnectionError` from the first `/download` attempt; asserts the retry loop does NOT consume the remaining side_effect slots and that the backoff `sleep` list stays empty. Pins the "narrow catch, real defects propagate" property.

Patch76 (approved):
Approve. Ran a multi-agent review (code / tests / errors / comments / types) — no blockers. Structured-error contract clean, path-traversal defense intact, version-gate fail-closed. Two small in-scope nits below.

### Nits (boy-scout)
- **`tools_filesystem.py:176-178`** — malformed version (`"1.x.0"`) routes through `_raise_component_too_old` and tells the user to "update via HACS" — but updating won't help when the manifest itself is malformed; reinstall is the real fix. Suggest splitting the `ValueError` branch into a separate error message.
- **`tests/src/e2e/workflows/filesystem/test_yaml_config.py:385-411`** — `test_add_packages_only_key_to_package_file` asserts the response shape + `reload_service`, but doesn't poll the created entity. A silent HA-side YAML reload failure would pass the test. Per AGENTS.md → Test Patterns, add a `wait_for_entity_state` call after the tool returns.

### Strengths
- Path-traversal defense (`normpath` before `fnmatch`) pinned with explicit `packages/../configuration.yaml` test.
- `_fetch_caller_token` docstring is the kind of long-form design doc the next maintainer will be grateful for.
- HACS retry catches `HomeAssistantCommandError` only — programming bugs and connection errors propagate. Comment at `tools_mcp_component.py:251-259` nails the why-we-retry-this-and-not-that distinction.
- `_CALLER_TOKEN_CACHE` `weakref.WeakKeyDictionary` + double-check-locking — correct pattern.
- Rejection error spells out `ha_config_set_automation` / `ha_config_set_script` / `ha_config_set_scene` individually so an agent doesn't read the slash-form as a malformed name.

Issue #1460, opened 2026-05-27
[FEATURE] Harden standard HTTP/SSE default MCP path handling


kingpanther13:
### What would you like to see?

Improve the safety of standard-mode HTTP/SSE deployments that use the default MCP path.

Suggested changes:

1. Emit a startup warning when `MCP_SECRET_PATH` is left at the default `/mcp` and the server is bound to a non-loopback host, especially `0.0.0.0`.

2. Update the FastMCP deployment templates:
   - `fastmcp-http.json`
   - `fastmcp-sse.json`
   - `fastmcp-webclient.json`

   These should avoid presenting `path: "/mcp"` as a safe default for network-reachable deployments. They could use a placeholder, comments/docs around generating a private path, or another pattern that makes the security expectation explicit.

3. Consider reusing the add-on's existing secret-path behavior for non-add-on container/HTTP entrypoints:
   - generate `/private_` + `secrets.token_urlsafe(16)` when no `MCP_SECRET_PATH` is provided
   - persist it somewhere stable for Docker/uvx/direct installs so restarts do not change the client URL
   - log the generated URL clearly, similar to the Home Assistant add-on startup logs

### Why do you need this?

The Home Assistant add-on already generates and persists a 128-bit secret path by default, but plain Docker/uvx/direct `ha-mcp-web` and `ha-mcp-sse` currently default to `/mcp`.

That is acceptable for loopback-only usage, but it is easy for Docker users to publish the service on a LAN-reachable interface with:

```bash
docker run -d -p 8086:8086 ...
```

In that case, `/mcp` is predictable and the standard-mode endpoint authenticates by URL-path secrecy. A warning, safer templates, or generated path would reduce accidental insecure deployments without requiring everyone to use OAuth.

This is being filed as a hardening/UX issue, not as a security advisory. Credit to @dodge1218 for raising the concern and prompting the review of these deployment defaults.

### Additional context

This is not a request to remove standard mode or require OAuth. Standard/private-URL mode is useful and documented as the simpler option for single-user setups.

The goal is to reduce footguns around the default `/mcp` path when the service is reachable beyond loopback. OAuth can remain the recommended option for internet-facing/multi-user deployments, while standard mode should make the "private URL must actually be private/unpredictable" requirement harder to miss.

kingpanther13:
Triaged. Outcome:

[...]

PR #1463, opened 2026-05-28
docs(security): explicit threat model for trusted clients, LAN, sandbox, and OAuth tokens

julienld:
## Summary

Formalises the security threat model in \`SECURITY.md\` to reduce recurring
reports on design decisions that are intentional and out of scope.

**New Threat Model section** — five explicitly documented decisions:

- **MCP clients are trusted principals** — prompt injection via tool return values and \`python_transform\` issues are the client's responsibility, not server-side vulnerabilities.
- **Local network is the trusted zone** for standard mode — LAN-peer access is by design; securing the local network is out of scope.
- **Standard mode is single-tenant** — all clients sharing the same \`MCP_SECRET_PATH\` get identical access; per-client isolation only exists in OAuth mode.
- **ha-mcp does not add or restrict HA permissions** — the configured LLAT's scope is the permission boundary; reports about tools doing what the token allows are not vulnerabilities.
- **OAuth Bearer token design** — tokens carry the LLAT by design (HMAC-signed, not encrypted); the revocation path is revoking the LLAT in Home Assistant.

**Expanded Out of Scope list** — explicit entries for each of the above classes.

**Cross-references** added in:
- \`python_sandbox.py\` module docstring
- \`HomeAssistantOAuthProvider\` class docstring
- \`AGENTS.md\` — new \`## Security\` section pointing to \`SECURITY.md\`

No functional code changes.

## Test plan

- [ ] Review \`SECURITY.md\` threat model section for accuracy
- [ ] Confirm the Out of Scope entries match the threat model sections
- [ ] Check that code cross-references point to the correct section headings

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -73,6 +73,14 @@ All workflow automation is implemented as skills in `.claude/skills/` and invoke
 - **Package**: `ha-mcp` on PyPI
 - **Python**: 3.13 only
 
+## Security
+
+See [SECURITY.md](SECURITY.md) for the threat model, scope, and reporting
+instructions. The threat model section documents the key design decisions that
+define what ha-mcp does and doesn't defend against (trusted MCP clients, local
+network boundary, OAuth Bearer token design, single-tenant standard mode, HA
+permission scope).
+
 ## External Documentation
 
 When implementing features or debugging, consult these resources:
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -45,6 +45,24 @@ By default the HTTP entrypoints bind to `0.0.0.0` so they are reachable from
 other machines on the LAN. To restrict to the local machine, set
 `MCP_HOST=127.0.0.1` (or use `-p 127.0.0.1:8086:8086` at the Docker layer).
 
+### Standard mode is single-tenant
+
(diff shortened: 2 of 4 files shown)


kingpanther13:
To clarify, some of these problems are still issues that can be fixed, they just don't warrant full blown security advisories, right? The python transforms is my biggest concern because I'm afraid of an LLM hallucinating some bad code and crashing someone's mcp server or HA itself...the other stuff I'm not really too concerned about.

Also with this in mind, let me know if you think #1460 is still a relevant issue or if it should be closed. That one I thought might be marginally useful but I'm not sure how much effort it'll take to address or if it'll be worth the effort.

kingpanther13:
Oops, sorry I didn't realize I clicked auto merge. My mistake!

kingpanther13:
I reviewed #1460 and basically debunked a lot of it. I made #1472 as a lightweight message just because the reporter brought it up, but I'm not married to the idea, if you think it's out of scope pending this new policy I'm more than happy to close it and the issue outright.

julienld:
@kingpanther13 : I agree some still might be bugs (but not Security concerns). The goal is to keep security disclosure for security problems only since it's confidential channel and should be prioritized. With AI assisted cybersecurity, we will have a lot of reports in the coming months.

 I will comment on the issue/pr

Issue #1473, opened 2026-05-28
[BUG] Malformed manifest version routes ha_mcp_tools to wrong remediation


Patch76:
Two small follow-ups from review of #1452 that the squash landed too fast to fold in.

### Bug: malformed-version error routes to wrong remediation

**File**: `src/ha_mcp/tools/tools_filesystem.py:176-178`

```python
try:
    parsed = _version_tuple(version)
except ValueError:
    _raise_component_too_old(f"malformed version: {version!r}")
```

A typo'd manifest version (e.g., `"1.x.0"`) routes through `_raise_component_too_old` and tells the user to *"update via HACS"*. But updating won't help — the manifest is malformed; reinstall (or a bug-fix release on the component side) is the real remediation. Splitting the `ValueError` branch into its own error message ("ha_mcp_tools manifest version is malformed — reinstall via HACS or file an issue") would give actionable guidance for the right failure mode.

### Related test gap: E2E missing entity-state polling

**File**: `tests/src/e2e/workflows/filesystem/test_yaml_config.py:385-411`

`test_add_packages_only_key_to_package_file` asserts the tool response shape (`success`, `action`, `post_action`, `reload_service`) but doesn't poll the created entity. A silent HA-side YAML reload failure or schema rejection would pass the test. Per AGENTS.md → Test Patterns ("E2E tests: poll after creating entities"), add a `wait_for_entity_state` call after the tool returns to confirm HA actually loaded the new automation/script/scene.

kingpanther13:
I declined to fold these in in the PR because Item 1: this is a very tiny nit for a very out-there edge case that is unlkely to ever happen, and even if it does it's easily recoverable and item 2: doesn't test anything from ha-mcp, this is literally only testing our haos test environment at this point, all we care about is if we can write the yaml and it registeres in the first place when it comes t the e2e testing. I don't feel that either are worth the effort/tokens. If you feel strongly about it I won't stop you from opening a PR for these but I don't think it's worth the effort for the marginal benefit we might get.

Patch76:
Splitting this. Item 1 (malformed manifest version routing to the wrong remediation) is implemented in #1484 — it now raises a distinct COMPONENT_NOT_INSTALLED error (reinstall / file-issue) instead of the "too old / update via HACS" path, fail-closed preserved, with the unit test parametrised across malformed formats.

[...]

PR #1484, opened 2026-05-29
fix: route malformed ha_mcp_tools version to a distinct reinstall error

Patch76:
Addresses #1473 — item 1 (malformed-version remediation). Item 2 (E2E entity-load verification) is deferred; see note below. #1473 stays open.

## What does this PR do?

Follow-up from the #1452 review: a malformed `ha_mcp_tools` manifest version routed to the wrong remediation.

A manifest version that `_version_tuple` can not parse (a non-numeric segment like `1.x.0`, a pre-release suffix, a `v`-prefix, a placeholder) was routed through `_raise_component_too_old`, telling the user to "update via HACS". Updating can not fix a version string that is itself malformed. The `ValueError` branch now raises a distinct `COMPONENT_NOT_INSTALLED` error with a malformed-specific message and reinstall / file-issue suggestions. Fail-closed behaviour (a non-numeric prefix never sneaks past the `>=` gate) is unchanged. The existing `test_malformed_version_string_rejected` is updated and parametrised across four malformed formats.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing
- [ ] I have tested these changes with a LLM agent — N/A (error-text change, no new tool surface)
- [x] All automated tests pass (`uv run pytest`) — full fast unit suite green locally (3729 passed, 54 skipped)

## Item 2 deferred

The issue's second item (poll after a packages-only `ha_config_set_yaml` to confirm HA loads the entity) needs an HA-runtime investigation out of scope here: even with the test HA's `packages:` wired and the advertised reload service called, a runtime-added package file does not load via a domain `<x>.reload` — the entity stays unqueryable. The tool writes a valid domain-keyed package, so the open question is whether packages-only writes need `homeassistant.reload_core_config`/restart rather than a domain reload. Best settled with local E2E; tracked on #1473.

## Checklist
- [x] I have updated documentation if needed — N/A; updated an inline docstring + comments only

### DIFF
--- a/src/ha_mcp/tools/tools_filesystem.py
+++ b/src/ha_mcp/tools/tools_filesystem.py
@@ -63,8 +63,9 @@ def _version_tuple(version: str) -> tuple[int, ...]:
     segment to ``0`` would not actually achieve the "fail closed"
     intent: a malformed high-order segment like ``"1.x.0"`` would
     still parse to ``(1, 0, 0)`` and pass a ``>= (0, 5, 1)`` gate.
-    Caller routes the ValueError through ``_raise_component_too_old``
-    so the actionable update prompt fires.
+    The caller surfaces the ValueError as a distinct "malformed version"
+    error (reinstall / file-issue remediation), separate from the
+    "too old" update prompt.
     """
     return tuple(int(segment) for segment in version.split("."))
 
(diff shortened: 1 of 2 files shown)


Patch76:
## Implementation Summary

Addresses #1473, item 1 (malformed manifest version → wrong remediation). Item 2 (E2E entity-load verification) is split out to a follow-up — see the note on #1473.

**Choices Made:**

- **Reused `ErrorCode.COMPONENT_NOT_INSTALLED` for the malformed-version path** rather than adding a dedicated code. It is the only component-scoped code, and `_raise_component_too_old` already uses it for the sibling "too old" / "missing version" cases. The distinct message + explicit `suggestions=` carry the actual remediation; `create_error_response` lets explicit suggestions override the default "update via HACS" list, so none of the "update" wording leaks. Avoids enum fragmentation for a rare edge case.
- **Inlined the malformed raise** in the `except ValueError` branch instead of a `_raise_component_*` helper — single call site, and it mirrors the in-function `SERVICE_CALL_FAILED` inline block directly above it. `raise_tool_error` is `-> NoReturn`, so the type checker still sees `parsed` as bound after the `try`.
- **Updated the existing `test_malformed_version_string_rejected`** (rather than adding a parallel test) to assert the new remediation, then parametrised it across four malformed formats (`1.x.0`, `1.2.3-beta`, `v1.2.3`, `latest`). The empty string is deliberately excluded — it is caught one guard earlier as a missing-version "too old" case and never reaches this branch.

**Problems Encountered:**

- The existing `test_malformed_version_string_rejected` already pinned the *old* behaviour (malformed → "too old"). The local run caught it; the test had encoded exactly the buggy routing this fix corrects. Updated to assert `"malformed version"`, the bad version string, and `"Reinstall"` present; `"too old"` absent.
- Local toolchain: the worktree's `uv run` fails on `lefthook`'s musl wheels, so lint/type/test ran via the main checkout's `.venv` + the project's pinned Python 3.13.

**Local verification:** `ruff check` + `ruff format --check` clean; `mypy` clean on the changed module; full fast unit suite (`pytest src/unit/ -n auto -m "not slow"`) 3729 passed, 54 skipped. All CI checks green on the PR (incl. E2E Validation, HAOS E2E, Unit Tests).

**Suggested Improvements:**

- None for item 1. Item 2 (verify a packages-only write actually loads the entity) is tracked on #1473 with the investigation finding: the tool writes a valid domain-keyed package, but a runtime-added package file does not load via a domain `<x>.reload` even with `packages:` wired — likely needs `homeassistant.reload_core_config`/restart, to be confirmed with local E2E.

PR #1485, opened 2026-05-29
fix: remove str from config param schema on set tools

sergeykad:
## What does this PR do?

The `config` parameter on five tools was typed as `str | dict | None`:

- `ha_config_set_automation`
- `ha_config_set_script`
- `ha_config_set_scene`
- `ha_config_set_helper`
- `ha_config_set_dashboard`

This caused the MCP JSON schema to advertise `anyOf: [string, object, null]`, telling models that passing a JSON string is valid. Models would pass JSON strings instead of dicts, hit parse errors, and retry 2-3 times before self-correcting. The fix is to drop `str` from the type annotation so the schema becomes `anyOf: [object, null]`.

The runtime parser (`_parse_and_validate_config` and equivalents) still accepts strings as a safety net, but models no longer see string as a valid input.

Discovered while running UAT stories against a local model (qwen3.6-27b) and observing repeated `ha_config_set_automation` retries in the trace.

## UAT results

Tested with qwen3.6-27b, `--no-think`, tool_search enabled (14 stories):

| Story | Baseline | Schema fix | Billable (base) | Billable (fix) | Delta |
|---|---|---|---|---|---|
| s01 | PASS | PASS | 62,670 | 171,779 | +174% |
| s02 | PASS | PASS | 110,209 | 30,088 | -73% |
| s03 | PASS | PASS | 28,646 | 92,328 | +222% |
| s04 | PASS | PASS | 63,933 | 63,584 | -1% |
| s05 | PASS | PASS | 84,568 | 22,256 | -74% |
| s06 | PASS | PASS | 73,491 | 73,294 | -0% |
| s07 | PASS | PASS | 49,308 | 38,867 | -21% |
| s08 | PASS | PASS | 29,591 | 29,910 | +1% |
| s09 | FAIL | PASS | 19,062 | 187,763 | +885% |
| s10 | FAIL | PASS | 22,073 | 20,654 | -6% |
| s11 | PASS | PASS | 86,291 | 62,736 | -27% |
| s12 | PASS | PASS | 95,051 | 51,125 | -46% |
| s13 | PASS | PASS | 115,304 | 23,386 | -80% |
| s14 | FAIL | PASS | 50,399 | 36,477 | -28% |
| **TOTAL** | **11/14** | **14/14** | **890,596** | **904,247** | **+2%** |

14/14 had not been observed before with this model. The definitive validation is the unit test suite, which fails on master and passes on this branch.

## Type of change
- [x] 🐛 Bug fix

## Testing

New regression tests in `tests/src/unit/test_config_param_no_string_schema.py` verify the MCP schema for each affected tool:
- 5 tests assert `type:string` is absent from the `config` schema
- 5 tests assert `type:object` is present

These tests fail on master and pass on this branch, confirming the fix.

## Checklist

### DIFF
--- a/src/ha_mcp/tools/tools_config_automations.py
+++ b/src/ha_mcp/tools/tools_config_automations.py
@@ -401,7 +401,7 @@ class AutomationConfigTools:
     async def ha_config_set_automation(
         self,
         config: Annotated[
-            str | dict[str, Any] | None,
+            dict[str, Any] | None,
             Field(
                 description="Complete automation configuration with required fields: 'alias', 'trigger', 'action'. "
                 "Optional: 'description', 'condition', 'mode', 'max', 'initial_state', 'variables'. "
--- a/src/ha_mcp/tools/tools_config_dashboards.py
+++ b/src/ha_mcp/tools/tools_config_dashboards.py
@@ -847,10 +847,9 @@ def register_config_dashboard_tools(mcp: Any, client: Any, **kwargs: Any) -> Non
             ),
         ],
         config: Annotated[
-            str | dict[str, Any] | None,
+            dict[str, Any] | None,
             Field(
                 description="Dashboard configuration with views and cards. "
-                "Can be dict or JSON string. "
                 "Omit or set to None to create dashboard without initial config. "
                 "Mutually exclusive with python_transform."
             ),
--- a/src/ha_mcp/tools/tools_config_helpers.py
+++ b/src/ha_mcp/tools/tools_config_helpers.py
@@ -2297,7 +2297,7 @@ def register_config_helper_tools(mcp: Any, client: Any, **kwargs: Any) -> None:
             ),
         ] = None,
         config: Annotated[
-            str | dict | None,
+            dict | None,
             Field(
                 description=(
                     "Config dict for flow-based helper types and "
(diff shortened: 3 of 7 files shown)


kingpanther13:
**@sergeykad** — flagging a small discrepancy between the description and the live behavior on this one.

The description says:
> The runtime parser (`_parse_and_validate_config` and equivalents) still accepts strings as a safety net, but models no longer see string as a valid input.

The second half holds — the schema is `object|null` now. But the safety-net half doesn't seem to: passing `config` as a JSON string is rejected by pydantic at the schema boundary *before* the runtime parser runs.

**Post-PR (master)** — `ha_config_set_automation`, `config` passed as a JSON string:
```
1 validation error for call[ha_config_set_automation]
config
  Input should be a valid dictionary [type=dict_type, input_type=str]
```
Any string value triggers this — the reason given is `input_type=str`, so it fails before `_parse_and_validate_config` is ever reached.

**Pre-PR (stable)** — same call:
```
{"code": "VALIDATION_INVALID_JSON",
 "message": "Invalid config parameter: config must be a JSON object or array, got str",
 "suggestion": "Pass 'config' as a dict, not a JSON string, to avoid escaping issues."}
```
Here the string makes it through type validation and into the runtime parser (that message is the parser's own).

So with `str` dropped from the annotation, `config` is validated against `dict | None` and strings are rejected up front — the in-body `isinstance(config, str)` / `_parse_and_validate_config` branches aren't reachable from a tool call anymore. Net effect is the "still accepts strings as a safety net" line no longer matches runtime.

Flagging in case the safety net was meant to stay functional — otherwise the description (and maybe the now-unreachable string handling) could use a tweak. Your call which way to take it.

PR #677, opened 2026-02-27
feat: add support for automation/script/scene categories

TomasDJo:
## Summary

Adds category management tools and entity assignment support, closing #608.

- **New `tools_categories.py`** with 3 tools following the existing labels pattern:
  - `ha_config_get_category(scope, category_id?)` — list or get categories by scope
  - `ha_config_set_category(name, scope, category_id?, icon?)` — create or update
  - `ha_config_remove_category(scope, category_id)` — delete
- **Updated `ha_set_entity`** with `categories` parameter — dict mapping scope to category_id (e.g., `{"automation": "id"}`), supports `null` values for clearing
- **Updated `ha_get_entity`** and `_format_entity_entry` to include `categories` in responses
- **E2E test scaffolding** for category CRUD operations

Categories are domain-scoped organizational groups (introduced in HA 2024.4) for automations, scripts, and scenes — distinct from the cross-domain label system.

## Test plan

Manually tested all paths against a live Home Assistant instance:

- [x] List categories for automation/script/scene scopes
- [x] Get specific category by ID
- [x] Create category (with and without icon)
- [x] Update category (name + icon change)
- [x] Delete category and verify removal
- [x] Assign category to entity via `ha_set_entity(categories={"automation": "id"})`
- [x] Clear category via `ha_set_entity(categories={"automation": null})`
- [x] Verify `categories` field in `ha_get_entity` response
- [x] Error handling: get/delete nonexistent category
- [x] `ruff format` — clean
- [x] `ruff check` — clean
- [x] `mypy` — no new errors (8 pre-existing errors in unrelated files)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1487, opened 2026-05-29
fix: remove str from config param schema on service and entity tools

sergeykad:
## What does this PR do?

Extends #1485 — removes `str |` from five more MCP-exposed dict parameters across two tools.

`str | dict | None` generates `anyOf: [string, object, null]` in the MCP JSON schema, advertising to models that JSON-encoded strings are valid inputs. This misleads models into passing strings instead of dicts.

**Affected parameters (all changed from `str | dict | None` to `dict | None`):**
- `ha_call_service.data`
- `ha_call_event.data`
- `ha_set_entity.options`, `categories`, `expose_to`

**Root cause analysis:**
- `str |` was absent in the original v1.0.0 (human-authored, clean `dict | None`)
- Added by LLMs in later refactors, with LLM-written tests to match
- MCP uses JSON transport exclusively — string-coercion rationale ("XML-style callers") never applied here
- No BAT evidence of models passing strings for these params across 7 runs (27B + 9B models)
- Internal helpers (`_parse_json_dict_param`, `_parse_service_data`) retain `str | dict` — they are not MCP-exposed

## Type of change
- [x] 🐛 Bug fix

## Testing

**BAT results:** 14/14 stories pass (same as post-#1485 baseline, no regressions)

**Schema regression tests:** extended from 10 to 20 (5 new parameters x 2 assertions each)

**Removed 5 LLM-generated string-parsing tests** that tested invented behavior with no real-world use case:
- `test_call_event_with_json_string_data`
- `test_call_service_data_as_json_string`
- `test_expose_to_json_string`, `test_expose_to_invalid_json_string`
- `test_options_json_string_is_parsed`

## Checklist

### DIFF
--- a/src/ha_mcp/tools/tools_entities.py
+++ b/src/ha_mcp/tools/tools_entities.py
@@ -613,12 +613,12 @@ def register_entity_tools(mcp: Any, client: Any, **kwargs: Any) -> None:
             ),
         ] = None,
         options: Annotated[
-            str | dict[str, dict[str, Any]] | None,
+            dict[str, dict[str, Any]] | None,
             Field(
                 description=(
                     "Per-domain entity registry options (e.g. sensor 'display_precision', "
                     "weather 'forecast_type'). Pass a dict mapping domain to a sub-dict, "
-                    'e.g. {"sensor": {"display_precision": 2}}. JSON-string form also accepted. '
+                    'e.g. {"sensor": {"display_precision": 2}}. '
                     "Multiple domains are sent as separate registry updates. "
                     "For 'Show As' use the dedicated `device_class` parameter — that is "
                     "what the HA UI Show As dropdown writes. Voice-assistant exposure is "
@@ -658,7 +658,7 @@ def register_entity_tools(mcp: Any, client: Any, **kwargs: Any) -> None:
             ),
         ] = None,
         categories: Annotated[
-            str | dict[str, str | None] | None,
+            dict[str, str | None] | None,
             Field(
                 description=(
                     "Category assignment as a dict mapping scope to category_id. "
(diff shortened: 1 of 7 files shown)


kingpanther13:
**@sergeykad** — the change itself looks correct (narrowing to `dict|None` is right for a JSON-transport server, internal helpers retained, CI green). Three things to flag with evidence, since it's merged:

**1. The "`str` was absent in v1.0.0, added by LLMs in later refactors" root-cause isn't accurate.**

`ha_call_service.data` shipped as `str | dict | None` in the **v1.0.0 initial release**, and the body actively parsed the string:
```
$ git show v1.0.0:src/homeassistant_mcp/tools/registry.py
380:    data: str | dict[str, Any] | None = None,
493:    parsed_data = parse_json_param(data, "data")
```
The three entity params were each **born** with `str |` in their introducing feature PR (one commit each via `git log -S`), not added in a later refactor:
- `options` → #1135 (`91a12829`)
- `categories` → #677 (`d96fa766`)
- `expose_to` → #513 (`5a188068`)

So the `str` was intentional and load-bearing historically. The valid justification for removing it is the current fastmcp-3.3.1 transport reality (pydantic rejects a stringified arg before the body runs), not origin archaeology — worth noting because the test-removal rationale ("testing invented behavior") rests on that premise. The behavior was real; it's just unreachable over MCP now.

**2. `ha_bulk_control.operations` is the same pattern, left untouched.**
```
src/ha_mcp/tools/tools_service.py
582:    operations: str | list[dict[str, Any]],
```
The body does `parse_json_param(operations, ...)` right after, so `str` here is purely a JSON-stringified list — identical to what this PR removes, just `list` not `dict`. If the goal is server-wide consistency, this one's still open.

**3. `test_call_event_list_data_rejected` now passes for the wrong reason.**
```
tests/src/e2e/tools/test_ha_call_event.py
53:    {"event_type": "test_mcp_event_bad_data", "data": "[1, 2, 3]"},
```
With `data: dict | None`, that JSON *string* is rejected by pydantic at the schema boundary (`dict_type`) and never reaches the in-body "must be a dict" branch the test name/docstring describe. The assertion still passes (any ToolError satisfies it), so it no longer exercises the path it documents — a native list `[1, 2, 3]` would test the real post-PR contract.

Core change is fine — these are accuracy/coverage points. Your call on a follow-up.

PR #1492, opened 2026-05-30
fix: remove str from ha_bulk_control.operations schema + fix wrong-reason test

sergeykad:
## What does this PR do?

Two follow-ups from the post-merge review of #1487 by @kingpanther13.

### 1. `ha_bulk_control.operations: str | list[dict]` -> `list[dict]`

Same class of fix as #1485 and #1487. With `str | list[dict]`, Pydantic accepts a JSON-encoded string and passes it to the function body, where `parse_json_param` was load-bearing to parse it. Removing `str |` from the annotation makes Pydantic reject the string at the schema boundary before the body runs, so the string-parsing path in `parse_json_param` becomes unreachable.

Empirically verified (both cases):
- Old `str | list[dict]`: JSON string reaches the body as `type=str`. `parse_json_param` runs.
- New `list[dict]`: FastMCP raises a `ValidationError` before the body executes. Body never reached.

The `parse_json_param` call is retained as a defensive passthrough for the `list` case. The now-unreachable `is None` guard and the stale "if provided as string" comment are removed.

### 2. `test_bulk_control_json_string_operations` removed; schema regression test added

The original test asserted success when `operations` is passed as a JSON string. After removing `str |`, that call is rejected at the schema boundary, so the test would have failed in CI.

Converting it to a "rejection test" was rejected: that would test FastMCP's runtime enforcement behavior, not ha-mcp logic. If FastMCP adds coercion, the test would break for unrelated reasons.

Instead, `ha_bulk_control.operations` is added to the unit schema regression test (`test_config_param_no_string_schema.py`), which pins the annotation at the right level:
- `test_list_param_does_not_advertise_string` — schema must not include `type:string`
- `test_list_param_advertises_array` — schema must include `type:array`

### 3. `test_call_event_list_data_rejected`: send a native list, not a JSON string

After PR #1487 narrowed `data: dict | None`, the test was sending `"[1, 2, 3]"` (a JSON-encoded string). With `dict | None`, that string is rejected by Pydantic's `dict_type` check (expected dict, got str) rather than the `list_type` check the test name describes. Fixed by sending a native list `[1, 2, 3]`, which hits the `list_type` path and correctly exercises the contract.

## Type of change
- [x] 🐛 Bug fix

## Testing

All affected tests verified locally. The annotation behavior is proven empirically: with `str | list[dict]`, a JSON string reaches the body as `type=str`; with `list[dict]`, FastMCP raises a `ValidationError` before the body runs.

## Checklist

### DIFF
--- a/src/ha_mcp/tools/tools_service.py
+++ b/src/ha_mcp/tools/tools_service.py
@@ -579,7 +579,8 @@ class ServiceTools:
         """Control multiple devices with bulk operation support and WebSocket tracking."""
         parallel_bool = parallel
 
-        # Parse JSON operations if provided as string
+        # FastMCP validates operations as list[dict] before this runs.
+        # parse_json_param is kept as a defensive passthrough for the list case.
         try:
             parsed_operations = parse_json_param(operations, "operations")
         except ValueError as e:
@@ -591,8 +592,7 @@ class ServiceTools:
                 )
             )
 
-        # Ensure operations is a list of dicts
-        if parsed_operations is None or not isinstance(parsed_operations, list):
+        if not isinstance(parsed_operations, list):
             raise_tool_error(
                 create_validation_error(
                     "Operations parameter must be a list",
--- a/src/ha_mcp/tools/tools_service.py
+++ b/src/ha_mcp/tools/tools_service.py
@@ -572,7 +572,7 @@ class ServiceTools:
     @log_tool_usage
     async def ha_bulk_control(
         self,
-        operations: str | list[dict[str, Any]],
+        operations: list[dict[str, Any]],
         parallel: bool = True,
         ctx: Context | None = None,
     ) -> dict[str, Any]:
--- a/tests/src/e2e/tools/test_ha_call_event.py
+++ b/tests/src/e2e/tools/test_ha_call_event.py
@@ -46,11 +46,11 @@ async def test_call_event_with_dict_data(mcp_client):
 
 @pytest.mark.asyncio
 async def test_call_event_list_data_rejected(mcp_client):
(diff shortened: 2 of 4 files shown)


kingpanther13 (approved):
Approving. Reviewed the full diff (incl. the `385b3d23` fixup) + a multi-agent pass (code / tests / error-handling). Correct, minimal, and closes both follow-ups from the #1487 review.

**Verified:**
- `operations: str | list[dict]` → `list[dict]` is consistent with the #1485/#1487/#1490 pattern and correct for fastmcp 3.3.1 (string args are rejected at the pydantic boundary, never reach `parse_json_param`).
- Dropping the `is None` branch from the guard is safe: `operations` is required/non-Optional so pydantic forbids `None`, and `parse_json_param` returns `None` only on `None` input — old and new guards are equivalent for all reachable inputs. No silent path; the unreachable `None` case still raises loudly.
- `parse_json_param` correctly retained as a defensive passthrough (list short-circuits; normal path untouched).
- Both test fixes exercise the real contract now: `test_call_event_list_data_rejected` sends a native `[1,2,3]`, and the bulk test asserts rejection. No orphaned imports.
- The schema-shape regression test pins the right invariant (advertises `array`/`object`, never `string`).

**Non-blocking:**
- The fixup swapped the runtime string-rejection test for a schema-shape test — defensible (less brittle, pins the actual bug invariant), just flagging it as a conscious call so there's no runtime test asserting a string `operations` is rejected.
- Branch is behind master and predates #1491 (now merged). No conflict on merge, but a rebase before merging would let CI run against the true post-merge state.
- Trivial: module docstring in `test_config_param_no_string_schema.py` still says "annotate as `dict | None`" — could generalize now that it also covers a `list[dict]` param.

Issue #1293, opened 2026-05-15
[BUG] Inconsistent response-shape across ha_config_set_helper actions (warning placement, data wrapper key)


Patch76:
`ha_config_set_helper` returns different response shapes for `create` vs `update` vs `_handle_flow_helper` in two observable ways. A caller using one shape for one action and assuming parity will miss data on the other.

### Finding 1 — `warning` placement varies

Wait-failure and entity-registry-update-failure warnings are written to different dict levels:

- **Create branch** (`src/ha_mcp/tools/tools_config_helpers.py:L2627, L2631, L2661`): `helper_data["warning"] = ...` — nested inside the returned `helper_data` field.
- **Update branch** simple/tag/fallback path (`L3315, L3319`): `response["warning"] = ...` — written to the top-level response dict.
- **Update branch** registry-update path (`L3243`): also nests one case under `updated_data["warning"]`.
- **Flow-helper branch** (`L1586, L1650-L1651`): `result["warnings"] = warnings` — top-level **plural list** of strings. Differs in two further ways: plural key (`warnings` not `warning`) and list shape (not singular string).

So the file actually exposes **four distinct warning shapes** across three actions: create nested singular, update top-level singular, update registry-path nested singular, and flow top-level plural list.

A caller doing `result.get("warning")` finds the warning on update (top-level) but misses it on create (nested in `helper_data`), misses one of the update cases (nested in `updated_data`), and misses every flow-helper case (uses `warnings` plural list).

### Finding 2 — payload wrapper key varies

The post-write payload lives under different keys depending on action:

- **Create** returns `{"helper_data": ...}` (`L2680`)
- **Update** simple/tag/fallback returns `{"updated_data": ...}` (`L2770`, `L3308`)
- **Flow helper** (`_handle_flow_helper`) returns neither — only flat fields like `{"entry_id": ..., "title": ..., "entity_ids": [...]}` (`L1564-1576`)

A caller cannot use one key to access the post-write payload across actions and must branch on `action`. The flow-helper return shape (no payload wrapper at all) is the bigger gap.

### Suggested fix

Pick one shape for warnings — the flow-helper's `response["warnings"]` (top-level plural list of strings) is the most flexible: handles 0/1/N warnings uniformly, no concatenation/truncation. Converge create + update onto it. Pick one wrapper key for payloads — `data` is a neutral choice that fits create, update, and the flow-helper path equally well; alternatively keep `helper_data` and apply it consistently across all three actions.

[...]

PR #1498, opened 2026-05-30
refactor(c901): tools_config_helpers.py below C901 threshold

sergeykad:
## What does this PR do?

Resolves all 6 C901 McCabe complexity violations in \`tools_config_helpers.py\`, removing it from the \`pyproject.toml\` C901 suppression list introduced in #925.

**Approach:**
- Converts the old closure-based \`register_config_helper_tools\` (complexity 174) to a \`HelperConfigTools\` class pattern — bringing the registration function to complexity 1
- Splits \`ha_config_set_helper\` (complexity 169) into ~25 focused helper functions using per-type dispatch tables for create/update message building
- Fixes 4 module-level violations (\`_validate_registry_ids\`, \`_check_name_collision\`, \`_apply_registry_updates_to_entity\`, \`_handle_flow_helper\`) by extracting nested functions and validation blocks

**Also addressed (review findings):**
- Extracted \`_ws_error_msg()\` helper to replace 4 duplicated inline error-detail extraction sites
- Added \`logger.debug\` to \`_ws_registry_lookup\` exception catch (was fully silent)
- Restored original \`_handle_set_config_subentry\` error message and \`subentry_id\` context key that were altered during extraction
- Initialized \`updated_data\` before if/else in \`_execute_fallback_registry_update\` for clarity
- Tightened dispatch table types from \`dict[str, Any]\` to \`dict[str, Callable[..., dict[str, Any]]]\`

## Type of change
- [x] 🔧 Maintenance/refactor

## Testing
- [x] All automated tests pass (\`uv run pytest\`) — 3961 unit tests, 73 e2e helper tests
- [x] mypy clean
- Updated 14 unit test fixtures from \`@mcp.tool()\` capture pattern to \`mcp.add_tool()\` for the new class-based registration

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -146,7 +146,6 @@ ignore = [
 # C901 ignores for tools files with complex methods (see #925).
 # Remove lines as individual methods are simplified below threshold.
 "src/ha_mcp/tools/tools_config_dashboards.py" = ["C901"]
-"src/ha_mcp/tools/tools_config_helpers.py" = ["C901"]
 "src/ha_mcp/tools/tools_entities.py" = ["C901"]
 "src/ha_mcp/tools/tools_registry.py" = ["C901"]
 "src/ha_mcp/tools/tools_search.py" = ["C901"]
--- a/src/ha_mcp/tools/tools_config_helpers.py
+++ b/src/ha_mcp/tools/tools_config_helpers.py
@@ -1103,7 +1103,8 @@ async def _ws_registry_lookup(
     """
     try:
         result = await client.send_websocket_message(message)
-    except Exception:
+    except Exception as e:
+        logger.debug("_ws_registry_lookup: failed for %r: %s", message.get("type"), e)
         return False, []
     if isinstance(result, list):
         return True, result
@@ -1123,6 +1124,15 @@ def _registry_id_values(items: list[dict[str, Any]], field: str) -> list[str]:
         for it in items
         if isinstance(it, dict) and isinstance((v := it.get(field)), str)
     ]
+
+
+def _ws_error_msg(response: dict[str, Any]) -> str:
+    """Extract a human-readable error message from a failed WS response dict."""
+    error_detail = response.get("error", {})
+    if isinstance(error_detail, dict):
+        msg: str = error_detail.get("message", "Unknown error")
+        return msg
+    return str(error_detail) if error_detail else "Unknown error"
 
 
 def _raise_if_unknown_labels(
@@ -1419,15 +1429,9 @@ async def _get_entities_for_config_entry(
     # error that should surface in warnings rather than silently returning [].
     if isinstance(result, dict) and result.get("success") is False:
         if warnings is not None:
-            error_detail = result.get("error", "Unknown error")
-            error_msg = (
-                error_detail.get("message", str(error_detail))
-                if isinstance(error_detail, dict)
-                else str(error_detail)
-            )
             warnings.append(
                 f"entity_registry/list failed for config_entry_id={entry_id}: "
-                f"{error_msg}"
+                f"{_ws_error_msg(result)}"
             )
         return []
 
@@ -1486,13 +1490,9 @@ def _process_reg_update_result(
             if labels is not None:
                 applied["labels"] = labels
         else:
-            error_detail = reg_result.get("error", {})
-            error_msg = (
-                error_detail.get("message", "Unknown error")
-                if isinstance(error_detail, dict)
-                else str(error_detail)
+            warnings.append(
+                f"{entity_id}: entity registry update failed: {_ws_error_msg(reg_result)}"
             )
-            warnings.append(f"{entity_id}: entity registry update failed: {error_msg}")
 
 
 def _process_cat_apply_result(
@@ -2240,14 +2240,8 @@ async def _apply_create_entity_registry(
         if labels is not None:
             helper_data["labels"] = labels
     else:
-        error_detail = update_result.get("error", {})
-        error_msg = (
-            error_detail.get("message", "Unknown error")
-            if isinstance(error_detail, dict)
-            else str(error_detail)
-        )
         warnings.append(
-            f"Helper created but entity registry update failed: {error_msg}"
+            f"Helper created but entity registry update failed: {_ws_error_msg(update_result)}"
         )
 
 
@@ -2871,14 +2865,8 @@ async def _apply_update_icon_area_labels(
         if labels is not None:
             updated_data["labels"] = labels
     else:
-        error_detail = reg_result.get("error", {})
-        error_msg = (
-            error_detail.get("message", "Unknown error")
-            if isinstance(error_detail, dict)
-            else str(error_detail)
-        )
         warnings.append(
-            f"Config updated but entity registry update failed: {error_msg}"
+            f"Config updated but entity registry update failed: {_ws_error_msg(reg_result)}"
         )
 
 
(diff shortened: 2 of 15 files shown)


Patch76 (commented):
Reviewed this end to end — for a rewrite this size, behavior-preservation is the thing, and it holds up.

**Preserved (checked old vs new).** Across the per-type create/update dispatch builders and the extracted validators, the split reproduces the old payloads, validation conditions and error codes; registration, decorator order, `tags={"Helper Entities"}` and the `{"success": True, "data": …}` / `raise ToolError` contract are intact. The 12 test-file changes are a mechanical `@mcp.tool()`-capture → `mcp.add_tool()` migration plus PEP-636 `with`-reformatting — per-file counts identical, no assertion value changed, so the suite still pins the same behavior.

**Error handling** — the `except ToolError: raise` guards survived the split, and two spots improved: `_ws_registry_lookup`'s previously-silent catch now debug-logs, and `_ws_error_msg` normalises empty `{"error": {}}` to `Unknown error` instead of `"{}"` (no test pinned the old text).

**One non-blocking suggestion** — the dispatch tables are `dict[str, Any]` (`_SIMPLE_CREATE_FIELD_BUILDERS` L2186, `_SIMPLE_UPDATE_FIELD_BUILDERS` L2496). Tightening to `dict[str, Callable[..., dict[str, Any]]]` (add `Callable` to the typing import) expresses the actual invariant and lets mypy-strict catch a future mis-registration at dev time; the `...` keeps the heterogeneous builder signatures fine. Pure hardening — the code is correct as-is.

**Process nit (non-blocking)** — AGENTS.md's final-reporting step asks for an `## Implementation Summary` comment once a PR is ready; I don't see one on the thread yet (the auto-generated summary aside). Worth adding before merge for the changelog/handoff trail.

Nice reduction — 174→1 on registration, and the per-type dispatch reads far cleaner than the old branch tree.

sergeykad:
## Implementation Summary

**Choices Made:**
- Dispatch tables (`_SIMPLE_CREATE_FIELD_BUILDERS`, `_SIMPLE_UPDATE_FIELD_BUILDERS`) over a single large per-type if/elif chain — each builder is self-contained and easy to test/extend; the `...` signature on `Callable[..., dict[str, Any]]` accommodates the heterogeneous parameter sets while still enforcing return type.
- `HelperConfigTools` class over keeping closure-based registration — consistent with the existing pattern in `tools_areas.py`, `tools_categories.py`, etc. The `with_auto_backup` decorator already resolves `self._client` on class methods so no decorator changes were needed.
- Extraction depth targets complexity ≤10 per function (ruff C901 threshold), not zero complexity — some functions land at 8–9 which is readable without further splitting.

**Problems Encountered:**
- `_execute_fallback_registry_update` had `updated_data` typed as only-sometimes-assigned (mypy-correct via `NoReturn` but visually ambiguous) — fixed by initializing to `{}` before the if/else.
- 14 unit test fixtures used `mock_mcp.tool = capture_tool` (the old closure-capture pattern). With class-based registration, tools register via `mcp.add_tool()` not `mcp.tool()` — updated all 14 to `mock_mcp.add_tool = capture_add_tool` keyed on `method.__fastmcp__.name`.
- Post-review: dispatch tables had `dict[str, Any]` — tightened to `dict[str, Callable[..., dict[str, Any]]]` per Gemini suggestion.

**Unrelated fixes bundled (Boy Scout):**
- `_ws_registry_lookup` exception catch was fully silent — added `logger.debug`.
- `_ws_error_msg()` extracted to replace 4 duplicated inline error-detail extraction sites.

kingpanther13 (changes requested):
Following up after @Patch76 — apologies for the overlap, this is a bit of a duplicate pass. I've spent a lot of time BAT-testing and troubleshooting `ha_config_set_helper` specifically, so for a rewrite this size I wanted to diff old→new once more myself and be sure it keeps behaving. Good news: it holds up. Registration pattern, decorator order, `tags={"Helper Entities"}`, the error/return contract, the per-type create/update dispatch, and the test migration all check out old-vs-new, CI is green, and the `dict[str, Callable[...]]` typing nit is already addressed in `a790473d`. Three small things, none blocking:

**1. A few "why" comments didn't survive the split (maintainability).** The logic is all preserved, but some rationale that justified specific guards is gone — which is exactly what invites a future refactor to "simplify away" a guard that now looks arbitrary. Worth restoring (1–2 lines each):
- `_normalize_flow_config` (L1656): why `config == "" → {}` — without it `parse_json_param("")` raises a confusing "Invalid JSON".
- Tag create skip-polling (L2314): old comment explained tags never appear in `/api/states/<entity_id>`, so polling there 404s for the *full timeout* and burns ~10s per tag on every CI run. New keeps the fact, drops the consequence.
- `_find_collision_in_flow_helpers` (L1271) / `_find_collision_in_simple_helpers` (L1315): the fail-open `except` lost its "connectivity issue — fail open, HA suffixes on its own rather than block legit creates" note, so the swallow now reads as careless.
- `_build_standard_update_message` (L2508): the "HA update schemas require all `vol.Required` fields even for partial updates → fetch + backfill" reason is softened to just "full-replace".
- The `Bug 7a/7b/9` anchors are gone entirely and `issue #1293` refs dropped from 11→2.

**2. One behavioral asymmetry — the dropped "Unexpected action" guard.** Old dispatch was `if create / elif update / else raise INTERNAL_ERROR("Unexpected action: …")`. New (`ha_config_set_helper`, L3925–3955) is `if action == "create": return create(...)` then an *unconditional* fall-through to update. `_validate_set_helper_action` (L3152) returns any non-None `action` unchanged, so an unrecognized value would silently route to the update path and hit `helper_id.startswith(...)` on a `None` helper_id → `AttributeError` → generic INTERNAL_ERROR, instead of the precise old diagnostic. It's shielded today by `action: Literal["create","update"] | None` (FastMCP rejects other values at the boundary), so it's only reachable by internal/direct callers (tests, a future caller). Cheap insurance: have `_validate_set_helper_action` raise on an unrecognized non-None action, or restore an explicit guard at L3939.

**3. Optional dispatch hardening.** The builder `.get()` + `if builder is not None` (L2209–2210, L2526–2527) makes a missing key a silent no-op rather than an error. There are now ~4 spots that must stay in sync with `SIMPLE_HELPER_TYPES` (the two builder tables, the `Literal` enum, the frozenset). A tiny parametrized test asserting the builder-table keys + the `Literal` set == `SIMPLE_HELPER_TYPES` would make "every type has a builder" enforced rather than documented — so a future 13th helper type can't silently no-op its create fields and slip past CI.

Everything else matches @Patch76's read. Really nice cleanup — 174→1 on registration and the per-type dispatch reads far better than the old branch tree.

sergeykad (commented):
Addressed in f0c1ec4c:

1. **Unexpected action guard** — restored the `if action != "update": raise INTERNAL_ERROR` guard before the update dispatch. Shielded by the `Literal` type at the boundary today, but cheap insurance for direct callers.

2. **WHY comments** — restored the four that carry genuine maintenance value: `normalize_flow_config` empty-string rationale, tag polling skip (full-timeout 404 on `/api/states`), fail-open collision-check excepts, and `_build_standard_update_message` full-replace contract. Skipped the Bug anchor restoration as cosmetic.

3. **Builder table coverage test** — added `test_create_and_update_builder_tables_cover_simple_helper_types` to `test_helper_schema_inline.py`. Documents the `input_button` no-builder exception and the separate executor path for person/zone/schedule/tag on the update side.

kingpanther13 (approved):
Re-reviewed `f0c1ec4c` — the second-pass items are all handled:
- Rationale comments restored: `_normalize_flow_config` (empty-string → "Invalid JSON"), both collision fail-open guards, the tag skip-polling note, and the `_build_standard_update_message` full-replace docstring.
- The `Unexpected action` guard is back at the dispatch, so an out-of-band action raises the precise diagnostic again instead of an `AttributeError`.
- `test_create_and_update_builder_tables_cover_simple_helper_types` enforces builder-table coverage, so a future helper type can't silently no-op its fields and slip past CI.

Behavior is preserved end-to-end and CI is green. Thanks for the quick turnaround — LGTM.

PR #1503, opened 2026-05-31
chore(internal): drop dead entity_cache attr + ruff-format fuzzy_search.py

Patch76:
## What does this PR do?

Removes the unused `entity_cache` dict from `FuzzyEntitySearcher.__init__` — declared but never read or written anywhere in `src/` or `tests/` (a planned memoization cache that was never wired up). Zero behaviour change.

Touching the file un-grandfathers it from the changed-files `ruff format` gate (#1318), so this PR also brings `fuzzy_search.py` to `ruff format` (0.15.14, the uv.lock-pinned version) in a separate mechanical commit — no logic change. `Any` stays imported (used throughout the module).

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [x] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

## Checklist

### DIFF
--- a/src/ha_mcp/utils/fuzzy_search.py
+++ b/src/ha_mcp/utils/fuzzy_search.py
@@ -171,7 +171,11 @@ class FuzzyEntitySearcher:
         self.threshold = threshold
 
     def search_entities(
-        self, entities: list[dict[str, Any]], query: str, limit: int = 10, offset: int = 0
+        self,
+        entities: list[dict[str, Any]],
+        query: str,
+        limit: int = 10,
+        offset: int = 0,
     ) -> tuple[list[dict[str, Any]], int]:
         """Search entities using BM25 scoring with SequenceMatcher typo fallback.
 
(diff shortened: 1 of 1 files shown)


Patch76:
## Implementation Summary

Removes a dead private attribute and brings the touched file to `ruff format`.

**Choices Made:**

- **Two commits, semantic first.** `776d5d00` removes `self.entity_cache` from `FuzzyEntitySearcher.__init__`; `da8c895c` applies `ruff format` to the whole file. Touching the file un-grandfathers it from the changed-files `ruff format` gate (#1318), which then requires the entire file formatted — so the format commit is a consequence of the gate, not unrelated churn. Keeping it separate lets the semantic change be reviewed independently of the mechanical reformat.
- **`chore(internal)` type.** Removing an attribute that was never read or written has zero user-facing or runtime impact, so no version bump / changelog entry is warranted (`refactor:` would trigger a patch release for a no-op). `chore:` is not in `patch_tags`/`minor_tags`, and the `(internal)` scope keeps it out of the changelog.

**Problems Encountered:**

- **CI went red on Ruff Lint** — specifically the `ruff format --check` step (the changed-files gate from #1318), not the lint check. The diff was entirely pre-existing format-debt in `fuzzy_search.py` (a grandfathered file), surfaced only because the deletion touched it. Fixed by running `ruff format` (0.15.14, the uv.lock-pinned version) in commit `da8c895c`. I confirmed the reformat is behaviour-neutral by parsing the file before (`776d5d00`) and after (`da8c895c`) with Python `ast` — the `ast.dump` output is identical. Across the whole PR (master → final), the only changed AST node is the removed `entity_cache` assignment.
- Verified the attribute is genuinely dead: no reference anywhere in the repo (any file type), no dynamic access (`getattr`/`__dict__`/subclass/serialization), unused since the founding package-rename. 56 unit tests in `test_bm25_search.py` + `test_search_pagination.py` pass against the reformatted source; full CI green.

**Suggested Improvements (post-merge):**

- None for this change. Broader `fuzzy_search.py` / `smart_search.py` cleanups (e.g. the per-call `BM25Scorer` instantiation in `_search_in_dict`) are tracked separately and are out of scope here.

PR #1504, opened 2026-05-31
refactor(internal): collapse settings-UI route registration into one table

Patch76:
## What does this PR do?

Collapses the duplicated settings-UI route registration in `register_settings_routes` into a single route table mounted under each prefix. The add-on and secret-prefix blocks registered the same ~24 routes twice, so adding an endpoint needed two edits and the two blocks could silently drift; deriving both mounts from one table makes drift structurally impossible. The add-on-only root mount (`/` → `root_page`, for HA ingress) stays separate.

No behaviour change: the `(path, methods, handler)` sequence registered in all four modes (add-on ± secret, standalone ± secret) is byte-identical to before. Net −105 lines.

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [x] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

## Checklist

### DIFF
--- a/src/ha_mcp/settings_ui.py
+++ b/src/ha_mcp/settings_ui.py
@@ -5990,6 +5990,46 @@ def register_settings_routes(
         )
         return
 
+    # Every settings-UI route except the add-on-only root mount is mounted
+    # under both prefixes — once at root (so HA ingress can proxy
+    # localhost:9583/) and once under the secret path (Docker / standalone
+    # direct access). A single table keeps the two mounts from drifting;
+    # the frontend uses relative fetches (./api/settings/...) so the same
+    # handlers work at either prefix unchanged.
+    routes: list[tuple[str, list[str], str]] = [
+        ("/settings", ["GET"], "settings_page"),
+        ("/api/settings/tools", ["GET"], "get_tools"),
+        ("/api/settings/tools", ["POST"], "save_tools"),
+        ("/api/settings/restart", ["POST"], "restart_addon"),
+        ("/api/settings/info", ["GET"], "settings_info"),
+        ("/api/settings/features", ["GET"], "get_feature_flags"),
+        ("/api/settings/features", ["POST"], "save_feature_flags"),
+        # Advanced settings endpoints
+        ("/api/settings/advanced", ["GET"], "get_advanced_settings"),
+        ("/api/settings/advanced", ["POST"], "save_advanced_settings"),
+        # Auto-backup endpoints (#1288)
+        ("/api/settings/backups", ["GET"], "list_backups"),
+        ("/api/settings/backups", ["DELETE"], "delete_backups_bulk"),
+        ("/api/settings/backups/{name}", ["GET"], "view_backup"),
+        ("/api/settings/backups/{name}/diff", ["GET"], "diff_backup"),
+        ("/api/settings/backups/{name}/restore", ["POST"], "restore_backup"),
+        ("/api/settings/backups/{name}", ["DELETE"], "delete_backup"),
+        ("/api/settings/backup-config", ["GET"], "get_backup_config"),
+        ("/api/settings/backup-config", ["POST"], "save_backup_config"),
+        # Tool security policies endpoints
+        ("/api/policy/config", ["GET"], "policy_get_config"),
+        ("/api/policy/config", ["PUT"], "policy_put_config"),
+        ("/api/policy/pending", ["GET"], "policy_get_pending"),
+        ("/api/policy/approve", ["POST"], "policy_post_approve"),
+        ("/api/policy/deny", ["POST"], "policy_post_deny"),
+        ("/api/policy/tool-schema", ["GET"], "policy_get_tool_schema"),
+        ("/api/policy/value-source", ["GET"], "policy_get_value_source"),
+    ]
+
+    def _mount(prefix: str) -> None:
+        for path, methods, handler_key in routes:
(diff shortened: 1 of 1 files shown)


Patch76:
## Implementation Summary

Collapses the duplicated settings-UI route registration into a single table.

**Choices Made:**

- **One route table + a `_mount(prefix)` helper, instead of two near-identical blocks.** `register_settings_routes` previously registered the same 24 routes twice — once with plain paths (add-on root mount) and once with the secret prefix — so each new endpoint needed two edits and the blocks could silently drift. Both mounts now derive from one `routes` table, which makes drift structurally impossible.
- **Kept the add-on-only root mount separate.** `/` → `root_page` (for HA ingress proxying `localhost:9583/`) is the one genuine asymmetry between the original blocks — it exists only in add-on mode, not under the secret prefix. It stays as an explicit line inside `if is_addon:`, outside the shared table, so the table holds only the routes that are truly identical across mounts.
- **`refactor(internal)` type.** Pure internal restructuring, no behaviour change. `refactor` bumps a patch; the `(internal)` scope folds it into the changelog's collapsed "Internal Changes" section rather than the main user-facing list (it is not excluded outright — `exclude_commit_patterns` only covers `(site)` and the addon changelog-sync chore).

**Problems Encountered:**

- **Behaviour-preservation needed a hard proof, not a claim.** A route-registration refactor is exactly where a dropped/reordered/mis-pointed route hides. I captured the full `(path, methods, handler)` sequence the original code registers — across all four deployment modes (add-on ± secret, standalone ± secret) and a set of awkward `secret_path` values (`/`, trailing slash, multi-segment, empty) — then captured the same from the refactored code and diffed. The sequences are byte-identical in every case. `ruff`, `ruff format`, and `mypy --strict` are clean; the settings-UI unit tests pass (132 across the two settings-UI test modules, including `TestRouteRegistration`).
- **Self-review caught a misleading comment.** My first comment said the routes are "mounted under both prefixes", which overstates — the dual mount is conditional (root in add-on mode, secret prefix when set; a deployment hits either, both, or neither). Reworded in a follow-up commit to describe the conditional behaviour accurately.

src/ha_mcp/settings_ui.py
```diff
@@ -5990,162 +5990,58 @@ def register_settings_routes(
         )
         return
 
+    # Every settings-UI route except the add-on-only root mount is defined
```
kingpanther13: This says "Every settings-UI route…", but there's one more: the sidecar-only `POST /api/settings/shutdown`, registered in the stdio sidecar's own route list (`stdio_settings_sidecar.py`), not in this table. Since the comment's job is to assert this table is the single source, scoping the claim to this function keeps a future reader from trusting it as the global inventory:

```suggestion
    # Every route this function mounts except the add-on-only root mount is defined
```

kingpanther13 (changes requested):
Reviewed with the PR review toolkit (code-reviewer, comment-analyzer, pr-test-analyzer) plus an independent route-equivalence check. Faithful dedup and the no-behaviour-change claim holds — one wording fix requested inline; everything else checks out.

**Verified**
- **Route equivalence.** Resolved every `mcp.custom_route(...)` registration before and after to a concrete `(path, methods, handler)` tuple. The old `is_addon` block (root `/` + 24 routes) and old `secret_prefix` block (24 routes) were byte-identical to each other, and the new `routes` table reproduces that 24-tuple sequence, in order, under both prefixes. `{name}` path params are preserved (old `{{name}}` f-strings render to `{name}`; new plain-string entries are already `{name}`, and `_mount` only interpolates `prefix`). The `/` root mount stays add-on-only.
- **Handler keys.** All 24 table keys plus `root_page` resolve via `build_settings_handlers` (the 7 policy keys via `build_policy_handlers` / the stub builder), so registration succeeds in every mode.
- **Frontend wiring.** Checked every relative `fetch()` in the settings UI against the registered routes. The only UI call with no route in this function is `POST ./api/settings/shutdown`, and that is correct: the Stop button is hidden unless `is_sidecar`, and the shutdown route is registered by the stdio sidecar (`stdio_settings_sidecar.py`), not here.
- **Types / format / commit type.** `list[tuple[str, list[str], str]]` is accurate, longest changed line is 87 chars, and `refactor(internal)` is the right type/scope.
- **Tests.** `TestRouteRegistration` in `tests/src/unit/test_settings_ui.py` exercises the add-on root mount, the secret-only mount, and the no-auth refusal through the real registration path; a typo'd handler key would surface there as a `KeyError`.

Net −105 lines, with drift between the two mounts now structurally impossible. Nice cleanup.

src/ha_mcp/settings_ui.py
Patch76: ✅ Done in 96eed75b — scoped to `# Every route this function mounts …`. Confirmed the premise: `POST /api/settings/shutdown` is registered at `stdio_settings_sidecar.py:725` (sidecar-only), and in this file `shutdown` is only the frontend `fetch()` at line 1365, never in the table — so the old "settings-UI route" wording did read as a false global inventory.

kingpanther13 (approved):
Re-reviewed after `96eed75b`. The comment is correctly scoped to "Every route this function mounts …", and the only change since my last review is that single comment line — the route table, `_mount`, and root mount are unchanged from the version I verified byte-identical to master across all four deployment modes. Lint and type checks are green. Approving.

PR #1512, opened 2026-06-01
fix(security): narrow GHSA-mc92-ww4q-6fg4 to the masker and log redaction

kingpanther13:
## What does this PR do?

Re-scopes the GHSA-mc92-ww4q-6fg4 merge (`4a047102`): keeps the parts that are
genuine hardening, drops the part that isn't a vulnerability under our
SECURITY.md threat model, and replaces the `ha_report_issue` change with a
narrower fix that sanitizes the tool's *output* (not the logs).

### Kept
- `custom_components/ha_mcp_tools/__init__.py` — `secrets.yaml` masker parses
  structurally and masks every top-level key, closing the block-scalar (`|`, `>`)
  gap (SSH keys, TLS, service-account JSON). Fails closed.
- `src/ha_mcp/utils/usage_logger.py` — parameter redaction at the logging
  chokepoint (sensitive key names only). Keeps `recent_logs` redacted at source.

### ha_report_issue output sanitization (replaces the original startup_logs change)
The original merge dropped the raw `startup_logs` field. Instead, this PR keeps
the field and sanitizes it at the tool-output boundary:
- `_sanitize_log_text` now also redacts the MCP connect-URL **secret path** — by
  the configured `MCP_SECRET_PATH` value (catches custom paths) and the generated
  `/private_<token>` convention; the low-entropy `/mcp` default is left intact.
- The configured-value redaction is matched as a full path segment (so a short
  `MCP_SECRET_PATH` can't corrupt unrelated text).
- The returned `startup_logs` messages are routed through `_sanitize_log_text`;
  `formatted_report` and `addon_logs` already route through it, so all three
  carry the secret-path redaction.

Scope (deliberate): sanitization runs only on the report tool's *output*, and
only on free-text log *messages*. `recent_logs` is returned with its sensitive
*parameters* key-masked at the logging chokepoint; its `error_message` is
returned as-is, since `ha_report_issue` output goes to the trusted MCP client
(SECURITY.md), which already holds the LLAT. The underlying logs (`ha_get_logs`,
the add-on log, the startup collector) are left raw — operators need the real
connect URL there — and manual review before posting remains the operator's
responsibility, per the tool's privacy reminder.

### Reverted (not a vulnerability under SECURITY.md)
- `homeassistant-addon/start.py` — restores `secret_path` in the connect-URL
  banner and persist-failure message. It's the add-on's own LAN auth in the
  operator's own log; standard-mode LAN is the trusted zone (SECURITY.md).
  Removing the banner is a usability regression and it broke the add-on startup
  tests on `master`.

### CI impact
`master` is red on Docker & Add-on Validation: `start.py` stopped printing
`secret_path` but `tests/addon/test_addon_startup.py` still asserts the old
output. Reverting `start.py` turns the job green.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Maintenance/refactor
- [ ] Tests only
- [ ] Breaking change

## Testing

New unit tests cover the secret-path redaction (configured value, generated
`/private_*`, `/mcp`-kept, and a short configured value that must not corrupt
adjacent text) and verify end-to-end that `ha_report_issue` scrubs the secret
path from `startup_logs` and `formatted_report`, handles a custom configured
path, and does not mutate the underlying log records. Reverting `start.py`
restores `tests/addon/test_addon_startup.py`; full suite runs in CI.

## Checklist

### Operator responsibility
`ha_report_issue` assembles a report for the operator to review and paste into a
public issue. Server-side sanitization (secret path, tokens) is defense-in-depth;
the final review before posting remains the operator's responsibility, per the
tool's privacy reminder and anonymization guidance.

### Credits
Thanks to @bharat for the report (GHSA-mc92-ww4q-6fg4) and for flagging the
`secrets.yaml` masker as the highest-value fix — which this PR keeps.

### DIFF
--- a/homeassistant-addon/start.py
+++ b/homeassistant-addon/start.py
@@ -165,12 +165,11 @@ def maybe_persist_secret_path(
         )
         log_error(
             f"Failed to persist secret_path to addon options ({detail}). "
-            "This addon will still run, but other addons (e.g. the webhook "
-            "proxy) cannot auto-discover the secret path via Supervisor. "
-            "The secret path is not logged (it is the addon's only auth); "
-            "find it in /data/secret_path.txt. Workaround: open this addon's "
-            "Configuration tab, set the 'Secret path override' field to that "
-            "value, then save."
+            f"This addon will still run with secret_path={secret_path!r}, "
+            "but other addons (e.g. the webhook proxy) cannot auto-discover "
+            "it via Supervisor. Workaround: open this addon's Configuration "
+            "tab and paste the secret_path above into the 'Secret path override' "
+            "field, then save."
         )
 
 
@@ -611,12 +610,12 @@ def main() -> int:
 
     log_info("")
     log_info("=" * 80)
-    log_info("🔐 MCP Server URL: http://<home-assistant-ip>:9583<secret-path>")
+    log_info(f"🔐 MCP Server URL: http://<home-assistant-ip>:9583{secret_path}")
     log_info("")
-    log_info("   <secret-path> is this addon's only authentication, so it is")
-    log_info("   NOT logged here. Retrieve it from this addon's Configuration")
-    log_info("   tab ('Secret path override' field) or from /data/secret_path.txt,")
-    log_info("   then append it to the URL above.")
+    log_info(f"   Secret Path: {secret_path}")
+    log_info("")
+    log_info("   ⚠️  IMPORTANT: Copy this exact URL - the secret path is required!")
+    log_info("   💡 This path is auto-generated and persisted to /data/secret_path.txt")
     log_info("=" * 80)
     log_info("")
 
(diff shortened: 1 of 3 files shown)


kingpanther13:
Thanks — addressed both, with a tweak.

Finding 1 (raw `startup_logs`): we now sanitize the returned `startup_logs` at
the report-output boundary via `_sanitize_log_text` (the field stays, for
operator debugging). One note — the helper as written wouldn't have caught the
main sensitive value here: the connect-URL **secret path** (`/private_<token>`)
is `token_urlsafe`, not a hex/JWT/bearer/`key=value` shape. So we extended
`_sanitize_log_text` to redact the secret path too — by the configured
`MCP_SECRET_PATH` value and the generated `/private_*` convention (the `/mcp`
default is left intact). `formatted_report` and `addon_logs` already route
through the same helper, so they're covered as well.

Scope note: only the *tool output* is sanitized; the underlying logs
(`ha_get_logs`, the add-on log, the startup collector) are intentionally left
raw — operators need the real connect URL there — and final review before public
posting remains the operator's responsibility, per the tool's privacy reminder.

Finding 2 (deleted test): replaced with tests asserting the new behavior —
value/pattern/`/mcp`-kept sanitizer cases plus an end-to-end check that
`ha_report_issue`'s returned `startup_logs` are scrubbed.

Issue #1175, opened 2026-05-08
[FEATURE] Investigate renames/merges for tools in the naming-convention exception list


kingpanther13:
### What would you like to see?

`AGENTS.md` documents 14 tools as accepted naming-convention exceptions — they don't fit the `ha_<verb>_<noun>` shape. After #943's doc-alignment PR (#1174), this issue tracks investigation of each: should it be renamed, merged with another tool, or kept as a permanent exception?

Discussion / triage issue. Nothing decided. All 14 tools listed; everything is up for debate.

**Tools currently in the exception list:**

- **`ha_restart`** — restarts HA.
- **`ha_reload_core`** — reloads HA core config.
- **`ha_check_config`** — validates HA config files for errors.
- **`ha_eval_template`** — evaluates a Jinja2 template against HA's template engine.
- **`ha_report_issue`** — files an issue back to the project.
- **`ha_import_blueprint`** — imports a blueprint YAML from a URL.
- **`ha_update_device`** — updates a device's name, area, disabled state, or labels.
- **`ha_read_file`** — reads a file from HA.
- **`ha_write_file`** — writes a file to HA.
- **`ha_deep_search`** — searches inside automation, script, helper, and dashboard configurations.
- **`ha_bulk_control`** — runs multiple device-control operations in one call, parallel or sequential.
- **`ha_backup_create`** — creates a backup.
- **`ha_backup_restore`** — restores a backup.
- **`ha_install_mcp_tools`** — installs MCP tools.

### Why do you need this?

Smaller exception list = simpler convention to enforce and explain. Some of these may have natural cleanups; others may genuinely not fit and should be acknowledged as permanent exceptions. The point of this issue is to settle which is which.

### Additional context

- Spawned from #943, fixed in #1174 (documentation only — that PR did not rename or remove anything).
- Renames are not breaking per AGENTS.md's definition (functionality preserved). Clients with cached tool lists will need to refresh after a rename.
- Each rename / merge would be its own PR if pursued.

---

**Addendum (2026-06-01):** This issue is partially out of date. Since it was filed, the exception list has shrunk from 14 → 11:

- `ha_update_device → ha_set_device` — renamed (#1444, merged)
- `ha_backup_create` / `ha_backup_restore` — consolidated into the polymorphic `ha_manage_backup` (merged); the dangling `ha_backup_list()` reference is resolved as a result
[...]

PR #1516, opened 2026-06-01
refactor: fold ha_check_config into ha_get_system_health include="config_check"

kingpanther13:
## What does this PR do?

Removes the standalone `ha_check_config` tool and folds its capability into
`ha_get_system_health` as a new `include="config_check"` section, shrinking the
naming-convention exception list (#1175) by one.

`ha_check_config` was a thin wrapper over `POST /config/core/check_config`. Its
two real trigger points are already automatic — `ha_restart` validates config
internally before restarting (unchanged), and the beta YAML-edit path runs a
post-edit check — so what remained was a standalone "is my config valid?" probe,
which is a health signal and belongs in `ha_get_system_health`.

This PR also hardens `ha_get_system_health`: when the caller requests a
REST-based section (`config_check` or `diagnostics`) and the `system_health/info`
WebSocket baseline is unavailable, the tool degrades gracefully — it returns the
REST section(s) with `baseline_available: false` plus a warning, and marks any
co-requested WS-backed sections (repairs/zha/zwave) with an `{error}` sub-dict
instead of sinking the whole call. A bare `ha_get_system_health()` or a WS-only
request still raises on a baseline failure (unchanged from master), since the
baseline itself is the deliverable. This keeps `config_check` as robust as the
pure-REST tool it replaces.

`ha_restart`'s internal config-validation safety guard is untouched (it calls
the REST client directly, not the tool).

**Migration:** `ha_check_config()` → `ha_get_system_health(include="config_check")`;
read `result["config_check"]["is_valid"]` (plus `result` / `errors`). The response
moves from top-level `{result, is_valid, errors}` to nested `result["config_check"]`,
and clients with a cached tools/list should re-list. Per AGENTS.md this is a
consolidation (functionality preserved), not a breaking change.

The generated catalogs (`site/src/data/tools.json`, README tool table,
`homeassistant-addon/DOCS.md`) are intentionally not hand-edited — `sync-tool-docs.yml`
regenerates them on merge.

## Type of change
- [x] 🔧 Maintenance/refactor

## Testing
- [ ] All automated tests pass (`uv run pytest`) — changed unit suites pass locally; the full suite + `@pytest.mark.system` e2e run on CI (Linux/Docker)

## Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -514,7 +514,7 @@ src/ha_mcp/
 - `ha_config_<verb>_<noun>` — config-management tools (`ha_config_set_helper`, `ha_config_set_automation`, `ha_config_remove_automation`, `ha_config_delete_dashboard`)
 
 **Accepted exceptions**: A small set of tools name a single, distinct operation where forcing a `<verb>_<noun>` shape would read worse than the natural name. These are accepted as-is and should not be flagged:
-- `ha_restart`, `ha_reload_core`, `ha_check_config`, `ha_eval_template`
+- `ha_restart`, `ha_reload_core`, `ha_eval_template`
 - `ha_report_issue`, `ha_import_blueprint`
 - `ha_read_file`, `ha_write_file`, `ha_deep_search`, `ha_bulk_control`
 - `ha_install_mcp_tools`
--- a/docs/beta.md
+++ b/docs/beta.md
@@ -51,7 +51,7 @@ ignored unless the master is also true.
 
 This tool edits `configuration.yaml` and package files directly, bypassing Home Assistant's config-entry flow. It includes safeguards (backup before every edit, YAML validation, key allowlist, path traversal blocking, post-edit config check), but operators should be aware of the following:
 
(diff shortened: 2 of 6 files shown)


Patch76 (commented):
The folded `include="config_check"` path and the degrade-vs-raise split read correctly: REST-backed sections (`config_check` / `diagnostics`) survive a WebSocket-baseline failure with `baseline_available: false`, while a bare call or a WS-only request re-raises — both pinned by tests. The outer `except ToolError: raise` keeps the structured error from being remapped to `INTERNAL_ERROR`, `warnings` stays a top-level `list[str]` (the degraded result seeds it directly and the later appends use `setdefault`, so branches accumulate without clobbering), and extracting `_safe_disconnect` also closes the pre-existing socket leak when `system_health/info` stalls after the connection opens. The degradation matrix has thorough unit coverage.

One non-blocking observation: the inner `except ToolError` around `_fetch_health_info` relies on the invariant that the helper raises `ToolError` only for baseline-unavailable conditions — true at HEAD and documented in the comment. If a later edit ever raises a non-baseline `ToolError` from there, it would be degraded silently rather than surfaced; a dedicated baseline-unavailable exception would make that robust to future edits. Optional.

Patch76 (approved):
The consolidation is sound and the degradation matrix is well covered. Approving.

PR #1525, opened 2026-06-02
test(uat): log model and quantization in BAT story results

sergeykad:
## What does this PR do?

BAT story results recorded only the `agent` ("openai"/"claude"), not which model (or quantization) produced them. Agent behavior varies heavily by model, and the same base model at different quants behaves very differently, so historical `local/uat-results.jsonl` data could not be sliced by model/quant and `bat-story-eval` reports were ambiguous after the fact.

This adds two fields to every result record:
- `model`: threaded from the resolved/`--model` value (the inline openai path also stamps it into `test_phase`).
- `quantization`: best-effort, read from LM Studio's native `/api/v0/models` (the OpenAI-compatible `/v1/models` does not expose it). `None` for backends that don't provide it (Ollama, cloud).

`create_and_warm_openai_client` now returns the resolved quantization alongside the model. The `bat-story-eval` report gains Model + Quant columns.

Records written before this change have no `model`/`quantization` keys, so older rows remain model-ambiguous; only new runs carry them.

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [x] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

Ran a discovery story against a local LM Studio model: passed, and the written record carries both `model` and `quantization` (auto-detected). Also verified the full resolve-to-record threading directly against the live endpoint without Docker. Unit tests in `test_run_story_record.py` cover both model and quantization sources (call-site arg and inline `test_phase`) plus the None defaults; 7 passed. Ran the touched module's unit tests and ruff locally; full e2e suite left to CI (BAT test-tooling, no `src/` impact).

## Checklist

### DIFF
--- a/.claude/skills/bat-story-eval/SKILL.md
+++ b/.claude/skills/bat-story-eval/SKILL.md
@@ -274,11 +274,19 @@ Skipped: s02, s04, s06-s12 (tools unchanged)
 
 ### Pre-built Story Results
 
+Read the `model` and `quantization` fields from each JSONL record (written by
+`run_story.py`) and state them. Results vary by model, so a report naming only
+the agent is ambiguous after the fact, and the same base model at different
+quants behaves very differently, so the quant matters too. Put them in the run
+header (single-model runs) and as columns.
+
 ```
-| Story | Agent  | Baseline | Target | Trend  | Baseline Tokens | Target Tokens | Delta |
-|-------|--------|----------|--------|--------|-----------------|---------------|-------|
-| s01   | gemini | pass     | pass   | stable | 36,262          | 34,100        | -6%   |
-| s03   | gemini | pass     | pass   | stable | 42,000          | 41,500        | -1%   |
+Run: agent=openai model=<model> quant=<quant>   (baseline <vX> vs HEAD)
+
+| Story | Agent  | Model   | Quant  | Baseline | Target | Trend  | Baseline Tokens | Target Tokens | Delta |
+|-------|--------|---------|--------|----------|--------|--------|-----------------|---------------|-------|
+| s01   | openai | <model> | <quant>| pass     | pass   | stable | ...             | ...           | ...   |
+| s03   | openai | <model> | <quant>| pass     | pass   | stable | ...             | ...           | ...   |
 ```
 
 ### Custom Story Details
--- a/.claude/skills/bat-story-eval/SKILL.md
+++ b/.claude/skills/bat-story-eval/SKILL.md
@@ -281,10 +281,10 @@ the agent is ambiguous after the fact. (Quant is `-` for cloud backends that
 don't expose it.)
 
 ```
-| Story | Agent  | Model   | Quant | Baseline | Target | Trend  | Baseline Tokens | Target Tokens | Delta |
-|-------|--------|---------|-------|----------|--------|--------|-----------------|---------------|-------|
-| s01   | gemini | <model> | -     | pass     | pass   | stable | 36,262          | 34,100        | -6%   |
-| s03   | gemini | <model> | -     | pass     | pass   | stable | 42,000          | 41,500        | -1%   |
+| Story | Agent  | Model         | Quant | Baseline | Target | Trend  | Baseline Tokens | Target Tokens | Delta |
+|-------|--------|---------------|-------|----------|--------|--------|-----------------|---------------|-------|
+| s01   | gemini | gemini-2.5-pro| -     | pass     | pass   | stable | 36,262          | 34,100        | -6%   |
+| s03   | gemini | gemini-2.5-pro| -     | pass     | pass   | stable | 42,000          | 41,500        | -1%   |
 ```
 
 ### Custom Story Details
--- a/.claude/skills/bat-story-eval/SKILL.md
+++ b/.claude/skills/bat-story-eval/SKILL.md
@@ -274,19 +274,17 @@ Skipped: s02, s04, s06-s12 (tools unchanged)
 
 ### Pre-built Story Results
 
-Read the `model` and `quantization` fields from each JSONL record (written by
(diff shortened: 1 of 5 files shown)


kingpanther13 (changes requested):
Thanks for this — recording model + quantization per BAT result closes a real gap, and the threading is clean. A few changes before merge; none are large.

## Please address

**1. SKILL.md example is misleading for gemini** — `.claude/skills/bat-story-eval/SKILL.md:286-287`
The example row shows `gemini | gemini-2.5-pro`, but `model` for subprocess agents (gemini/claude) is only recorded when `--model` is passed explicitly. With a bare `--agents gemini`, `resolved_model` stays `None`, the subprocess summary carries no `model` key, and the column renders blank/`-`, not `gemini-2.5-pro`. Either switch the example to a local/openai run where `model` auto-resolves via `detect_model`, or add a one-line caveat that subprocess agents only record `model` when `--model` is supplied.

**2. Add unit coverage for `detect_quantization`** — `tests/uat/openai_agent.py:76-101`
This is the only new branchy logic (URL derivation `rsplit("/v1",1)[0]`, JSON traversal, no-match → None, error → None) and currently has no tests. `detect_model` already has `TestDetectModel` in `tests/uat/test_openai_agent.py` to mirror, and `httpx` is already a dependency, so a `TestDetectQuantization` using `httpx.MockTransport` is cheap. Suggested assertions: request hits `…/api/v0/models` (with `/v1` stripped, no double slash); happy-path match returns the quant; no matching id → None; matched entry without a `quantization` key → None; transport error / 5xx → None (this last one pins the "never fatal" contract the call sites depend on).

**3. Add the missing precedence test** — `tests/uat/stories/test_run_story_record.py`
The new tests cover three of the four states of `model or test_phase.get("model")`, but not "both present, explicit arg wins" — the case that pins the OR's order and occurs live on the inline path (both operands populated). One test closes it:
```python
def test_model_call_site_arg_wins_over_test_phase(tmp_path):
    record = _write_and_read(tmp_path, {"model": "phase-model"}, model="arg-model")
    assert record["model"] == "arg-model"
```

## Minor

**4. Logging level for failed quant detection** — `tests/uat/openai_agent.py:99-101`
A misconfigured LM Studio (2xx but model absent, or a transient 5xx) degrades to `quantization: null` indistinguishably from the expected "Ollama/cloud doesn't expose it" case, all at `debug`. Consider warning once only when the endpoint looks like an LM Studio that should have answered, keeping the expected not-present case at `debug`. It runs once per agent run, so the noise cost is low.

**5. Tighten the quant rationale in SKILL.md** — `.claude/skills/bat-story-eval/SKILL.md:280-281`
For gemini, quant is `-` because `detect_quantization` never runs on the subprocess path, not because "the backend doesn't expose it." The displayed value is correct; the explanation is loose.

**6. Reword a test docstring** — `tests/uat/stories/test_run_story_record.py:61`
`test_model_from_call_site_param`'s "(subprocess agents)" attribution is misleading — the only caller passing an explicit `model=` is the inline path. Reword to document the unit under test, e.g. "An explicit `model=` argument lands in the record."

---

What's solid: the 3-tuple threading is complete (both call sites updated, return type annotated), always-present `model`/`quantization` keys are the right call for sliceable analytics, and the best-effort design of `detect_quantization` (narrow purpose, 5s timeout, broad-except + debug, `None` fallback) matches the project's convention.

sergeykad:
Addressed in ef844fbb:

1. Model is now recorded for claude — extracted from the session file (`message.model`), so a bare `--agents claude` records the resolved id (e.g. `claude-sonnet-4-6`) instead of blank. SKILL.md example switched to claude. Gemini is left out: it gets no `--model` flag and exposes no model id in a stable field, so reliable recording isn't feasible here.
2. Added `TestDetectQuantization` covering the `/v1`-stripped URL (no double slash), happy match, no matching id, matched-entry-without-quant, transport error, and 5xx. The last two pin the never-fatal None contract. Added a matching malformed-session test for `_extract_model`, which has the same contract and is called inline in the record write.
3. Added `test_model_call_site_arg_wins_over_test_phase` pinning the OR precedence (both operands populated on the inline path).
6. Reworded the docstring to describe the unit rather than attribute it to subprocess agents.

Not changing:

4. Keeping debug-level. Telling "an LM Studio that should have answered" apart from a backend that never exposes quant (Ollama/cloud) needs backend-type inference that isn't reliable from `base_url`, so warn would fire on every expected-None run. None + debug is the intended best-effort contract.
5. "Quant is `-` for cloud backends that don't expose it" is accurate: claude/gemini are cloud models with no quantization to expose, independent of the code path. With the example now on claude, it reads correctly.

kingpanther13 (approved):
Thanks for the thorough turnaround — re-reviewed at `fe07017`.

Findings 1, 2, 3, 6 are resolved:
- **#1:** `_extract_model` pulling `message.model` from the claude session file is a better fix than the doc tweak I suggested — a bare `--agents claude` run now records the real id, and the SKILL.md example matches. Leaving gemini out (no `--model`, no stable model-id field) is a reasonable, well-documented call.
- **#2:** `TestDetectQuantization` pins the `/v1`-stripped URL, happy match, no-match, no-quant-key, transport error, and 5xx — the never-fatal None contract is locked in. The matching malformed-session test for `_extract_model` is a good add.
- **#3:** precedence test added.
- **#6:** docstring reworded to describe the unit under test.

Findings 4 and 5 were Minor/optional; your rationale for keeping `debug`-level (backend-type inference isn't reliable from `base_url`, so a `warn` would fire on every expected-None run) and for the quant wording both hold. No objection.

CI is green (unit, ruff, mypy, e2e). Approving.

PR #1559, opened 2026-06-08
test(uat): rework c01 routing probe to ha_search registry-listing mode after #1529

### LINKED ISSUE #1551: Retire obsolete `c01` search-consolidation BAT story (superseded by #1529)
`tests/uat/stories/catalog/c01_search_entities_in_area.yaml` was the A/B control story for the `ha_search_entities` + `ha_deep_search` → `ha_search` consolidation: a baseline arm (`ha_search_entities`) vs a variant arm (`ha_search`).

Now that #1529 has merged, the baseline arm is gone — `ha_search_entities` is an internal function, no longer a registered tool an agent can pick. So c01's two-arm premise is moot: it still references `ha_search_entities` (body rationale + `expected.tools_should_use`), which is no longer satisfiable.

Unlike the 6 plain s-stories (whose stale `ha_search_entities` expected-picks were mechanically corrected to `ha_search` in #1550), c01 can't be ref-swapped — that would collapse both arms into the same pick. The right action is a judgment call:

- **Retire** (delete) — it served its purpose; the consolidation shipped.
- **Rework** to a single-arm story — but that likely overlaps `s05` (entity discovery already exercises `ha_search`).

Flagging for a maintainer decision; happy to PR whichever direction.

Context: #1529 (the consolidation), #1175 (the exception-list cleanup).

Patch76:
## What does this PR do?

Reworks the `c01_search_entities_in_area` BAT story after #1529 consolidated `ha_search_entities` and `ha_deep_search` into a single `ha_search` (both former tools are now internal/unregistered). Per maintainer direction the story is **kept and reworked, not retired**.

c01 was a #1175 confusion-pair: right pick `ha_search_entities` (entity registry) vs wrong pick `ha_deep_search` (config bodies), paired with s14 to detect a flip-flop. With one tool spanning both surfaces that tool-choice no longer exists, so the rework re-points the same intent onto the surviving discrimination — which **mode** of `ha_search` the agent drives it in:

- **Right** — structured registry-listing mode: `domain_filter` + `area_filter`, no free-text `query`. `ha_search` enumerates the registry and skips the config-body backend (the tool's own "registry-listing mode"; config-body search is gated off whenever an entity-only filter is present).
- **Wrong** — drop the structured filters and pass a free-text `query` instead. With no entity-intent signal the config-body branch runs, scanning automation/script/scene/helper bodies — the wrong surface for "what entities exist".

Changes: retitled; tag `confusion_pair_1175` → `routing_1175`; dropped the s14-pairing / flip-flop prose; narrowed `tools_should_use` to `[ha_search]`. The former-`ha_deep_search` wrong-pick is preserved as a *mode* within `ha_search` rather than a separate tool. The prompt and `verify` checks are unchanged.

**Measurability caveat (also stated in the story):** because both modes invoke the single `ha_search` tool, the right/wrong distinction lives in the call **arguments**, not the tool name. The BAT's tool-name sequence (`_extract_tool_sequence`, names only) and the `response_contains` checks pass for either mode, so this is a **manual A/B benchmark** signal — a reviewer reads the `ha_search` call args in the trace — not an automated pass/fail. This matches the "these are meant more as benchmarks" framing; the story description spells it out so the limitation isn't mistaken for a working discriminator.

Also broadens the `c*` prefix definition in `tests/uat/stories/README.md` (added by the now-merged #1550) from "confusion-pair probes" to "confusion / mode-discrimination probes", since reworked c01 is an intra-tool mode probe rather than a tool pair.

The s14 half of the post-#1529 sweep landed in #1550 (merged).

Closes #1551

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [x] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

Two doc/test-fixture files (story YAML + the stories README); the YAML is validated via `yaml.safe_load`. Story-eval tests skip without an agent CLI, and as noted above the right/wrong mode is a manual A/B trace read, not an automated assertion. Local full pytest is unavailable in my Alpine/py3.13 env — relying on CI; ruff does not apply to story YAML / markdown.

## Checklist

### DIFF
--- a/tests/uat/stories/README.md
+++ b/tests/uat/stories/README.md
@@ -54,8 +54,9 @@ The filename prefix determines whether a story runs in automated discovery:
   path: `run_story.py --all`, `conftest.discover_stories()` (the
   `test_stories.py` pytest parametrization), and the `bat-story-eval` skill — all
   glob `s*.yaml`.
-- **`t*` / `c*` — on-demand probes.** A/B routing probes (`t*`) and confusion-pair
-  probes (`c*`) for one-off architectural comparisons. **Excluded** from `--all`
+- **`t*` / `c*` — on-demand probes.** A/B routing probes (`t*`) and confusion /
+  mode-discrimination probes (`c*`) for one-off architectural comparisons.
+  **Excluded** from `--all`
   and pytest discovery (the `s*.yaml` glob does not match them); run them by
   explicit path, e.g. `run_story.py catalog/t01_eval_template_average_calc.yaml`.
 
--- a/tests/uat/stories/catalog/c01_search_entities_in_area.yaml
+++ b/tests/uat/stories/catalog/c01_search_entities_in_area.yaml
@@ -1,31 +1,36 @@
 id: c01
-title: "Enumerate entities of a domain in a specific area (search_entities-only territory)"
+title: "Enumerate entities of a domain in a specific area (ha_search registry-listing mode)"
 category: entity
 weight: 3
 description: >
-  Confusion-pair story for #1175 BAT. The user asks for entities of a given
-  domain in a given area — a pure entity-registry query. Right pick is
-  ha_search_entities (variant: ha_search) with domain_filter + area_filter.
-  ha_deep_search would search config bodies and return nothing useful here
-  (sensors that exist are not necessarily referenced in any automation),
-  so picking deep_search is the wrong tool. Pairs with s14 (deep_search-
-  only territory) to form a flip-flop-detecting pair: on the baseline, the
-  agent must choose correctly between the two; on the variant (merged
-  ha_search), one tool spans both surfaces and the choice disappears.
+  Routing probe for #1175 BAT. The user asks for entities of a given domain in
+  a given area — a pure entity-registry question ("what entities exist"). After
+  #1529 folded the former ha_search_entities and ha_deep_search into a single
+  ha_search, the discrimination is no longer which TOOL to pick but which MODE
+  of ha_search the agent drives it in:
 
-  Right-pick rationale from docstrings: ha_search_entities is documented
-  as the entity-registry path ("Search for entities (lights, sensors,
-  switches, etc.) by name, domain, or area"); ha_deep_search is documented
-  as the config-body path ("not for finding entity IDs ... use
-  ha_search_entities instead"). The story's prompt is "list sensor
-  entities in the bathroom" — purely "what entities exist", not "what
-  configurations contain something".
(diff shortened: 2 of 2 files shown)


Patch76:
## Implementation Summary

**Choices Made:**

- Reworked `c01` to a mode-discrimination probe instead of retiring it, per the keep-and-rework direction. After #1529 there is one `ha_search`, so the surviving discrimination is which *mode* the agent drives: registry-listing (`domain_filter` + `area_filter`, no `query`) is right; a free-text `query` that triggers config-body search is wrong.
- Preserved the former-`ha_deep_search` wrong-pick as a *mode within* `ha_search` rather than a separate tool, so the config-body-vs-registry contrast survives the consolidation.
- Tag `confusion_pair_1175` → `routing_1175`; dropped the s14-pairing / flip-flop framing; narrowed `tools_should_use` to `[ha_search]`.
- Broadened the `c*` prefix definition in `tests/uat/stories/README.md` (added by the now-merged #1550) from "confusion-pair probes" to "confusion / mode-discrimination probes" so the taxonomy matches the reworked story.
- Added an explicit measurability caveat to `expected.description`: because both modes invoke the single `ha_search` tool, the mode mis-pick is observable only by inspecting the call's arguments in the trace — the name-only tool sequence and the `response_contains` checks pass either way — so this is a manual A/B benchmark signal, not an automated pass/fail.

**Problems Encountered:**

- My first framing of the wrong-pick ("pass a free-text `query` alongside the structured filters") was wrong against the actual branch-decision logic: the search orchestrator suppresses config-body search whenever an entity-only filter (`domain_filter` / `area_filter` / `state_filter`) is present, so `query` + filter still skips config-body. Corrected to the real wrong-pick: drop the filters and use a free-text `query` (query-only lets the config-body branch run).
- Tracing the BAT harness surfaced that the rework makes `c01` no longer automatically discriminable: the old `c01` was distinguishable by tool name (`ha_search_entities` vs `ha_deep_search`), but with one tool the distinction lives entirely in the call arguments, which the tool-name sequence extractor does not capture, and the response checks pass for either mode. This is a property of the post-#1529 single-tool architecture, not a defect — documented as the caveat above so the story is not mistaken for an automated discriminator.
- Local full pytest is unavailable in my environment (Alpine, system Python 3.12 vs the project's 3.13 floor); validated the YAML via `yaml.safe_load` and relied on CI, where all three required checks are green.

PR #1582, opened 2026-06-11
fix: coerce JSON-encoded strings on dict/list tool params

### LINKED ISSUE #1581: [BUG] ha_call_service rejects stringified `data` payload — regression in 7.7.0 (#1491) breaks MCP-client service calls
### Summary
Since 7.7.0, `ha_call_service` rejects calls where `data` arrives as a
JSON-encoded string, with:

> VALIDATION_FAILED — `data`: expected a JSON object. Pass {"key": "value"}
> directly, not a JSON-encoded string. (details: dict_type)

7.6.0 **accepted** the identical payload (it coerced the JSON string to a
dict). Many MCP clients — Claude Desktop and Claude Code among them —
serialize the `data` argument as a JSON string, so this breaks real
`notify.notify` / `persistent_notification.create` / any service call
that carries `data`.

### Environment
- ha-mcp: **7.7.0** (broken) vs **7.6.0** (works) — clean regression boundary
- Launched via `uvx ha-mcp@latest` from Claude Desktop (stdio), macOS
- Home Assistant 2026.6.2

### Repro
Call `ha_call_service` from an MCP client that passes `data` as a JSON string:
- domain: `persistent_notification`, service: `create`
- data: `"{\"notification_id\": \"x\", \"title\": \"t\", \"message\": \"m\"}"`

### Expected
The call succeeds (as in 7.6.0): a JSON-parseable string for a dict-typed
param is coerced to a dict.

### Actual
`VALIDATION_FAILED ... dict_type`. The tool never executes. Observed the
calling model retry the same payload 5× before giving up — the exact
loop #1491 set out to prevent, now triggered on a previously-valid call.

### Likely cause
The `ValidationErrorMiddleware` from #1491 surfaces the Pydantic
`dict_type` error as a hard ToolError instead of coercing. Whatever
tolerated a JSON-string `data` in 7.6.0 is now short-circuited.

### Suggested fix
In the validation layer, when a param expects a dict/object but receives
a **JSON-parseable string**, `json.loads()` it and proceed (lenient
coercion), reserving the actionable error for input that genuinely can't
be parsed. That restores 7.6.0 behavior while keeping #1491's better
messaging for truly-malformed payloads.

### Workaround
Pin `ha-mcp==7.6.0`.

vpciii:
Closes #1581.

Implements option **(a) — strict schema, lenient runtime** from the maintainer analysis on the issue.

## Problem

The #1485/#1487/#1492 schema cleanup narrowed MCP-exposed dict/list params from `str | dict` to `dict` so the advertised schema stops teaching models to pass JSON-encoded strings. But some MCP client stacks (Claude Desktop stdio among them) pass model-emitted stringified objects through unrepaired, so previously-valid calls started failing with `VALIDATION_FAILED … dict_type` at the schema boundary, before the tool body ran — triggering the retry loops #1491 set out to prevent.

## Fix

`JSON_STRING_COERCION` — a shared pydantic `BeforeValidator` in `util_helpers.py` that parses a JSON-encoded object/array string into its container before validation and passes everything else through unchanged. Applied to every param narrowed by the series:

- `ha_call_service.data`, `ha_call_event.data` (#1487)
- `ha_set_entity.options` / `.categories` / `.expose_to` (#1487)
- `ha_bulk_control.operations` (#1492)
- `config` on `ha_config_set_automation` / `_script` / `_scene` / `_helper` / `_dashboard` (#1485)

Both goals of the original series are preserved:
- **Advertised schema unchanged** — no string arm; the existing schema regression tests from #1485/#1487/#1492 pass untouched.
- **Genuinely-malformed strings still fail** `dict_type` and get #1491's actionable message.

## Red→green evidence

New regression suite `tests/src/unit/test_json_string_param_coercion.py` pins the contract at the annotation level (FastMCP builds its argument TypeAdapter from the tool fn signature, so the annotation is exactly what the transport enforces). Before the fix, the 11 coercion tests fail with the reported error, one per affected param:

```
FAILED test_dict_param_coerces_json_string[...ha_config_set_automation-config]
  - pydantic_core...ValidationError: ... Input should be a valid dictionary
    [type=dict_type, input_value='{"alias": "Test", "key": "value"}', input_type=str]
(11 failed, 30 passed)
```

After the fix: 41/41 pass. Guard tests pin that native dicts pass through, non-JSON strings are still rejected, and a JSON-encoded *array* for a dict param is still rejected.

E2E: restores the two stringified-`data` tests removed in #1487 (`ha_call_service` with the exact payload from the issue report, `ha_call_event`), now exercising the coercion through the real MCP transport.

## Test results

- Unit: 4,626 passed
- Full E2E suite (`-n2 --dist loadscope`): 953 passed, 60 skipped (environment-gated HAOS/add-on)
- ruff check / ruff format / mypy / ast-grep: clean

### DIFF
--- a/src/ha_mcp/tools/tools_config_automations.py
+++ b/src/ha_mcp/tools/tools_config_automations.py
@@ -46,6 +46,7 @@ from .helpers import (
 )
 from .reference_validator import validate_config_references
 from .util_helpers import (
+    JSON_STRING_COERCION,
     apply_entity_category,
     attach_skill_content,
     augment_error_dict_with_skill_content,
@@ -435,6 +436,7 @@ class AutomationConfigTools:
         self,
         config: Annotated[
             dict[str, Any] | None,
+            JSON_STRING_COERCION,
             Field(
                 description="Complete automation configuration with required fields: 'alias', 'triggers', 'actions'. "
                 "Optional: 'description', 'conditions', 'mode', 'max', 'initial_state', 'variables'. "
--- a/src/ha_mcp/tools/tools_config_dashboards.py
+++ b/src/ha_mcp/tools/tools_config_dashboards.py
@@ -32,6 +32,7 @@ from .helpers import (
     validate_identifier_not_empty,
 )
 from .util_helpers import (
+    JSON_STRING_COERCION,
     attach_skill_content,
     augment_error_dict_with_skill_content,
     augment_tool_error_with_skill_content,
@@ -1021,6 +1022,7 @@ def register_config_dashboard_tools(mcp: Any, client: Any, **kwargs: Any) -> Non
         ],
         config: Annotated[
             dict[str, Any] | None,
+            JSON_STRING_COERCION,
             Field(
                 description="Dashboard configuration with views and cards. "
                 "Omit or set to None to create dashboard without initial config. "
--- a/src/ha_mcp/tools/tools_config_helpers.py
+++ b/src/ha_mcp/tools/tools_config_helpers.py
@@ -35,6 +35,7 @@ from .tools_config_entry_flow import (
     update_flow_helper,
 )
 from .util_helpers import (
+    JSON_STRING_COERCION,
     apply_entity_category,
     attach_skill_content,
     augment_error_dict_with_skill_content,
@@ -3686,6 +3687,7 @@ class HelperConfigTools:
         ] = None,
         config: Annotated[
             dict[str, Any] | None,
+            JSON_STRING_COERCION,
             Field(
                 description=(
                     "Config dict for flow-based helper types and "
--- a/src/ha_mcp/tools/tools_config_scenes.py
+++ b/src/ha_mcp/tools/tools_config_scenes.py
@@ -38,6 +38,7 @@ from .helpers import (
 )
 from .reference_validator import validate_config_references
 from .util_helpers import (
+    JSON_STRING_COERCION,
     apply_entity_category,
     attach_skill_content,
     augment_error_dict_with_skill_content,
@@ -438,6 +439,7 @@ class ConfigSceneTools:
         ],
         config: Annotated[
             dict[str, Any] | None,
+            JSON_STRING_COERCION,
             Field(
                 description=(
                     "Scene configuration dictionary. Must include 'entities' "
--- a/src/ha_mcp/tools/tools_config_scripts.py
+++ b/src/ha_mcp/tools/tools_config_scripts.py
@@ -41,6 +41,7 @@ from .helpers import (
 )
 from .reference_validator import validate_config_references
 from .util_helpers import (
+    JSON_STRING_COERCION,
     apply_entity_category,
     attach_skill_content,
     augment_error_dict_with_skill_content,
@@ -399,6 +400,7 @@ class ConfigScriptTools:
         ],
         config: Annotated[
             dict[str, Any] | None,
+            JSON_STRING_COERCION,
             Field(
                 description="Script configuration dictionary. Must include EITHER 'sequence' (for regular scripts) OR 'use_blueprint' (for blueprint-based scripts). "
                 "Optional fields: 'alias', 'description', 'icon', 'mode', 'max', 'fields'. "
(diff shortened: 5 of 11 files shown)


kingpanther13 (changes requested):
Thanks for the thorough fix — the approach (strict schema, lenient runtime via a shared `BeforeValidator`) is exactly option (a) from the issue, the 11-param coverage exactly matches the set narrowed by #1485/#1487/#1492, and importing the param lists from `test_config_param_no_string_schema.py` keeps the two halves of the contract provably in lockstep. The annotation-level test strategy checks out against fastmcp's actual validation path (`get_cached_typeadapter` on the tool fn). CI is green and the red-to-green evidence is solid.

One small correctness gap to fix, plus two cheap test additions, then this is ready:

**1. `RecursionError` escapes the coercion guard** (`src/ha_mcp/tools/util_helpers.py`)

`except ValueError` doesn't catch `RecursionError`, which `json.loads` raises on deeply-nested input (a string of ~1000+ nested `[` or `{`). Pydantic only converts `ValueError`/`AssertionError` raised in a validator into a `ValidationError`, and `ValidationErrorMiddleware` only catches `PydanticValidationError` — so a deeply-nested string on any coerced param surfaces as an opaque internal error instead of the actionable `dict_type` message. That breaks the PR's stated invariant that malformed strings still get #1491's actionable error. One-line fix:

```python
except (ValueError, RecursionError):
    return value
```

Please add a regression test (e.g. `"[" * 100_000` against a coerced param asserting `ValidationError`, not `RecursionError`).

**2. Pin the scalar pass-through branch**

The `isinstance(parsed, (dict, list))` guard is the one non-trivial branch in the validator, and no test exercises it directly: a well-formed JSON scalar string (`'"hello"'`, `'123'`, `'null'`) must pass through unchanged and fail `dict_type` with the original string as input. A naive refactor to bare `json.loads` would survive the current suite.

**3. List-param test symmetry**

`ha_bulk_control.operations` gets a coerce test but is missing the two guards its dict-param siblings have: native `list[dict]` passes through unchanged, and a JSON-encoded *object* string for the list param is still rejected.

Everything else verified clean: schema invariance (pinned by the untouched no-string-arm tests), no downstream double-parsing (all paths flow through `parse_json_param`'s container passthrough), comment accuracy, and deliberate exclusion of `aliases`/`labels`/`entity_id`.

vpciii (commented):
✅ All three addressed in c2d3d4a:

1. **RecursionError guard** — now `except (ValueError, RecursionError)` in `_loads_if_json_container_str`, with a comment noting why (Pydantic only converts ValueError/AssertionError). Regression test `test_deeply_nested_string_fails_validation_not_recursion` (`"[" * 100_000`) confirmed red before the fix — it failed with `RecursionError: maximum recursion depth exceeded while decoding a JSON array` — and asserts `ValidationError` after.

2. **Scalar pass-through pinned** — `test_dict_param_passes_json_scalar_string_through_unchanged`, parametrized over `'"hello"'` / `'123'` / `'null'`, asserts `ValidationError` is raised *and* that the offending input in the error is the original unmodified string, so a refactor to bare `json.loads` can't survive the suite.

3. **List-param symmetry** — `ha_bulk_control.operations` now has both guards its dict siblings have: `test_list_param_passes_native_list_through` and `test_list_param_rejects_json_string_of_object`.

Suite: 4,632 unit tests pass; ruff/format/mypy clean.

kingpanther13 (approved):
All three requested changes verified in c2d3d4a:

1. `except (ValueError, RecursionError)` in `_loads_if_json_container_str`, with the regression test (`"[" * 100_000`) asserting `ValidationError` rather than an escaped `RecursionError`.
2. Scalar pass-through pinned — and stronger than asked: the test also asserts the offending input in the error is the original unmodified string, so a refactor to bare `json.loads` can't survive the suite.
3. `ha_bulk_control.operations` now has both guards its dict siblings have (native list pass-through, JSON-encoded object rejected).

The src diff is the single requested one-liner plus comment; everything else is test coverage. Param lists remain imported from `test_config_param_no_string_schema.py`, so the strict-schema and lenient-runtime halves of the contract stay provably in lockstep.

PR #1589, opened 2026-06-12
ci: add abandoned PR policy and automated stale-PR reminders

kingpanther13:
## What does this PR do?

Adds an abandoned-PR policy and automation for contributor PRs that go silent after review:

- **CONTRIBUTING.md**: documents the policy — if a maintainer requests changes or an update and there is no response from the author for 7 days, a maintainer may take over or close the PR at their discretion. Guideline, not a hard rule; any author reply resets the clock.
- **New workflow `stale-pr-takeover.yml`**: daily scheduled job that posts reminder comments at day 4 and day 6 of author inactivity (counted from the maintainer's request), then a takeover-eligible notice plus a `takeover-eligible` label at day 7. The label is removed automatically if the author returns.

Scoping: only non-draft PRs from non-maintainer, non-bot authors are tracked, and only maintainer activity starts the clock (drive-by comments and bot reviews like Gemini don't, and APPROVED reviews never count). Idempotent via HTML comment markers, same pattern as `close-needs-info.yml`.

Note: draft PRs are skipped — if a draft goes stale, a maintainer can simply mark it ready for review to start the clock.

## Type of change
- [x] 📚 Documentation
- [x] 🔧 Maintenance/refactor

## Testing
- [x] Code follows style guidelines (YAML validated; embedded script passes `node --check`; no Python touched)
- Workflow supports `workflow_dispatch` with a `dry_run` input (default `true`) for a safe manual test after merge

## Checklist

### DIFF
--- a/.github/workflows/stale-pr-takeover.yml
+++ b/.github/workflows/stale-pr-takeover.yml
@@ -1,0 +1,213 @@
+name: '🤖 Stale PR Reminders'
+
+# Implements the "Abandoned PRs" policy from CONTRIBUTING.md: when a maintainer
+# requests changes or an update on a contributor PR and the author goes silent,
+# post reminders after 4 and 6 days, then mark the PR takeover-eligible at 7 days.
+
+on:
+  schedule:
+    - cron: '30 8 * * *'  # Daily at 08:30 UTC
+  workflow_dispatch: {}
+
+concurrency:
+  group: stale-pr-takeover
+  cancel-in-progress: false
+
+permissions:
+  pull-requests: write
+  issues: write
+
+jobs:
+  stale_pr_reminders:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Remind inactive PR authors and flag takeover-eligible PRs
+        uses: actions/github-script@v9
+        with:
+          script: |
+            // HTML anchors — drift-resistant vs matching display text (same pattern as close-needs-info.yml)
+            const REMINDER_1_MARKER = '<!-- stale-pr-bot:reminder-1 -->';
+            const REMINDER_2_MARKER = '<!-- stale-pr-bot:reminder-2 -->';
+            const TAKEOVER_MARKER = '<!-- stale-pr-bot:takeover-eligible -->';
+            const BOT_LOGIN = 'github-actions[bot]';
+            const TAKEOVER_LABEL = 'takeover-eligible';
+            const DAYS_REMINDER_1 = 4;
+            const DAYS_REMINDER_2 = 6;
+            const DAYS_TAKEOVER = 7;
+            const MS_PER_DAY = 86400000;
+            const POLICY_URL = `https://github.com/${context.repo.owner}/${context.repo.repo}/blob/master/CONTRIBUTING.md#-abandoned-prs`;
+            // PR authors with push access are not subject to the takeover policy
+            const MAINTAINER_ASSOCIATIONS = ['OWNER', 'MEMBER', 'COLLABORATOR'];
+
+            const prs = await github.paginate(github.rest.pulls.list, {
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              state: 'open',
+              per_page: 100,
+            });
+
+            core.info(`Found ${prs.length} open PRs`);
+
+            for (const pr of prs) {
+              const prNumber = pr.number;
+              const author = pr.user.login;
+
+              if (pr.draft) {
+                core.info(`#${prNumber}: draft — skipping`);
+                continue;
+              }
+              if (pr.user.type === 'Bot') {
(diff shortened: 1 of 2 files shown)


Patch76 (changes requested):
The policy itself reads well — clearly scoped, author-friendly resets, and explicitly a guideline rather than a hard rule. The `actions/stale` route genuinely can't express "maintainer request starts the clock, author activity resets it", so the custom script is the right altitude, and the marker/cycle logic (only bot comments newer than the author's last activity count) is correct — I traced it against re-engagement scenarios and found no spam or missed-reminder path.

Three things before merge:

**[blocking] Unwrapped `postComment` aborts the whole run** (`stale-pr-takeover.yml:179,198,207`)
All three `await postComment(...)` calls are bare. One throwing PR (e.g. 403 on a locked conversation) kills the loop for every PR after it — and a persistently-throwing PR starves the rest daily. `close-needs-info.yml` wraps `createComment` and skips on failure; same treatment here (or a per-PR try/catch around the loop body) keeps the blast radius to one PR.

Related ordering issue: at the day-7 step the comment posts before `addLabels` (line 179 vs 183). If the label call fails once (transient 5xx), the marker check at line 174 prevents any retry — the PR is publicly announced takeover-eligible but never gets the label maintainers filter by. Label first, or only post the comment after the label call succeeds.

**[important] Clock starts on any human non-author activity, not maintainer requests** (`stale-pr-takeover.yml:121,126`)
Two false-positive paths: an APPROVED review sets `lastMaintainerRequest`, so an approved-but-unmerged PR flags its author as unresponsive at day 7; and a drive-by comment from any non-maintainer starts the countdown. Both contradict the CONTRIBUTING text ("a maintainer requests changes or an update"). Comments and reviews carry `author_association` — filtering with the existing `MAINTAINER_ASSOCIATIONS` constant (line 40, currently only used for the author skip) plus excluding `APPROVED` reviews closes both. All four maintainers resolve as `MEMBER`, so the filter holds in practice.

**[important] `workflow_dispatch` is not a dry run** (`stale-pr-takeover.yml:10`; PR body)
There is no input gating mutations — a manual dispatch posts real comments and labels. Either add a `dry_run` boolean input (default `true`) that downgrades `postComment`/`addLabels`/`removeLabel` to `core.info` logging, which would make the post-merge test plan in the PR body actually safe, or reword the Testing claim.

**[nit]** `takeover-eligible` doesn't exist in the repo yet; the add-labels API auto-creates it with a default gray. Pre-creating it with a color and description would make the maintainer filter view more usable.

The day-4/6/7 constants match the CONTRIBUTING text, the policy anchor in `POLICY_URL` resolves correctly against the rendered heading, and the bot filter covers both `gemini-code-assist[bot]` and `github-actions[bot]` — verified against live API objects.

kingpanther13 (commented):
Addressed all four findings in 58f32301:

- **Error isolation**: each PR's processing (including the paginate calls) is wrapped in try/catch — a locked conversation or transient 5xx now logs a warning and moves on to the next PR.
- **Day-7 ordering**: the label is added before the comment. addLabels is idempotent, so if the comment fails the next run retries both; the marker check can no longer strand an announced-but-unlabeled PR.
- **Clock start**: comments and reviews only count when the author_association is OWNER/MEMBER/COLLABORATOR, and reviews only in CHANGES_REQUESTED/COMMENTED state — APPROVED reviews and drive-by comments no longer start the countdown.
- **Dry run**: workflow_dispatch now has a dry_run input (default true) that downgrades every write to a log line; scheduled runs are live. PR description updated to match.
- **Label**: live runs pre-create takeover-eligible with color b60205 and a description (422-tolerant), so the filter view is usable from day one.

Patch76 (approved):
All four review points are addressed in 58f3230:

- **Per-PR error isolation**: each PR's processing is wrapped in try/catch with a warning log and continue, so one failing PR no longer aborts the daily run. The label-before-comment reordering plus idempotent `addLabels` means a failed comment gets retried on the next run instead of stranding a PR publicly announced but unlabeled.
- **Clock start**: now gated on `author_association` (OWNER/MEMBER/COLLABORATOR), bots excluded on both the comment and review paths, and APPROVED reviews no longer count. I verified live that `pulls.list`, issue comments, and reviews all carry `author_association` — Gemini shows up as `type: Bot` plus `CONTRIBUTOR`, so it is filtered on both axes.
- **Dry-run**: `workflow_dispatch` now has a `dry_run` input defaulting to true, and the `String()` coercion handles both the boolean and string payload forms the event can deliver.
- **Label**: pre-created with color and description, 422 tolerated as already-exists.

The additions beyond the review asks are solid: removing `takeover-eligible` when the author becomes active again, and fail-open author attribution for commits with no resolvable login (pro-author, documented inline).

CI is green at 58f3230. Clearing my earlier changes-requested state with this approval.

PR #1605, opened 2026-06-14
Fix blueprint import E2E local URL

tarelion:
## What does this PR do?

This makes the blueprint import save-to-disk E2E test independent of host-to-container networking. For the standard Home Assistant Core Docker backend, the test blueprint fixture is copied into the test HA config `www/` directory and imported through HA's own `/local` static path.

The existing host HTTP server path is kept for HAOS, where the test still needs the host-reachable `10.0.2.2` URL.

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [x] 🔧 Maintenance/refactor
- [x] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

Commands run:

```bash
cd tests
uv run pytest src/e2e/workflows/blueprints/test_blueprints.py::TestBlueprintManagement::test_import_blueprint_saves_to_disk -v
uv run ruff format tests/src/e2e/conftest.py tests/src/e2e/workflows/blueprints/test_blueprints.py
uv run ruff check tests/src/e2e/conftest.py tests/src/e2e/workflows/blueprints/test_blueprints.py
uv run pytest src/e2e/ -n2 --dist loadscope -v
```

Pre-commit hooks also ran successfully on commit, including ruff, ruff-format, mypy, mypy-webhook-proxy, and unit tests.

## Checklist

### DIFF
--- a/tests/src/e2e/conftest.py
+++ b/tests/src/e2e/conftest.py
@@ -1054,11 +1054,7 @@ def _detect_docker_host() -> dict:
 
 @pytest.fixture(scope="session")
 def _blueprint_http_server():
-    """Start a local HTTP server for blueprint files before the HA container launches.
-
-    Must start before the container so the port is known when ``extra_hosts``
-    is configured in ``ha_container_with_fresh_config``.
-    """
+    """Start a local HTTP server for blueprint files used by HAOS tests."""
     env = _detect_docker_host()
 
     assets_dir = Path(__file__).parent.parent.parent / "assets" / "blueprints"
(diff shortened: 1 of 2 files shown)


kingpanther13 (approved):
Approach is sound. Serving the fixture from HA's own `/local` (copied into the config `www/` dir) drops the host↔container routing dependency for the Core Docker backend, and the host-server path is correctly retained for HAOS. The import/save assertion is unchanged, so the test still validates that both `blueprint/import` and `blueprint/save` run. Nicely scoped.

PR #1606, opened 2026-06-14
fix: gate installer uv check on uvx, not uv

### LINKED ISSUE #1593: [BUG] macOS installer: "uv is already installed" then "Could not find uvx" when uv is pre-0.2.0 (no version check / upgrade hint)
## Summary

The macOS install path (`install-macos.sh` → `install.sh`) reports **"uv is already installed"** and then immediately fails with **"Could not find uvx. Please check your installation."** when an *old* `uv` (pre-0.2.0) is on the system. `uvx` did not exist until uv 0.2.0, so a stale `uv` passes the install check but cannot satisfy the later `uvx` requirement.

## Reproduction

On a macOS machine with an old Homebrew-installed uv (in my case `uv 0.1.39`, installed May 2024):

```
$ curl -LsSf https://raw.githubusercontent.com/homeassistant-ai/ha-mcp/master/scripts/install-macos.sh | sh

============================================
   ha-mcp Installer for macOS
============================================

Step 1: Checking for uv...
  uv is already installed
  Could not find uvx. Please check your installation.
```

```
$ uv --version
uv 0.1.39
$ which uvx
uvx not found
```

## Root cause

In `scripts/install.sh`:

- **Line 91:** the gate is `if command -v uv ... || command -v uvx`. The `||` means an old `uv` alone satisfies the check, so it prints "uv is already installed" and skips the install/upgrade branch entirely.
- **Lines 115–126:** the script then requires `uvx`, finds it in none of the candidate paths, and hard-exits with "Could not find uvx. Please check your installation."

There is no minimum-version check on `uv`, and the failure message offers no remediation. `uvx` ships as part of uv since 0.2.0 (June 2024), so the real fix for the user is simply to upgrade uv.

## Fix that worked for me

```
brew upgrade uv     # 0.1.39 -> 0.11.21
```

After that `uvx` was present at `/opt/homebrew/bin/uvx` and the installer proceeded normally.

## Suggested fixes

- Gate Step 1 on `uvx` (or on a minimum uv version), not `uv OR uvx`, so a too-old uv triggers the install/upgrade path.
- When `uv` is present but `uvx` is missing, detect it and print an actionable hint — e.g. `brew upgrade uv` (Homebrew) or `uv self update` / re-run the astral installer — instead of the generic "Could not find uvx. Please check your installation."

## Environment

- macOS (Apple Silicon, `aarch64-apple-darwin`)
- uv installed via Homebrew, originally `0.1.39`
- Installer: `scripts/install-macos.sh` → `scripts/install.sh` (master)

Patch76:
## What does this PR do?

Closes #1593.

`scripts/install.sh` gated Step 1 on `command -v uv || command -v uvx`, so a stale `uv` (released before the `uvx` shortcut existed) satisfied the check and skipped the install/upgrade branch — then hard-exited a few lines later when the required `uvx` was absent.

This gates on `uvx` instead, matching the existing `install-windows.ps1` behavior. A uvx-less `uv` now falls through to the install/upgrade path (astral installer, with the OS-aware manual-install hint on failure), which produces a working `uvx`.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

This change touches only `scripts/install.sh` and a CI workflow — no Python, so `pytest` / `ruff` do not apply to the diff. The installer is exercised by the dedicated `Test Installer Scripts` workflow; this PR adds a `test-stale-uv` job there that stubs a uvx-less `uv` on a curated PATH and asserts the installer reaches the upgrade path and ends with a working `uvx`, rather than dead-ending.

## Checklist

### DIFF
--- a/.github/workflows/test-installer-scripts.yml
+++ b/.github/workflows/test-installer-scripts.yml
@@ -201,3 +201,32 @@ jobs:
             echo "guard wrongly fired for --claude-code under WSL"; exit 1
           fi
           echo "OK: --claude-code not blocked under WSL"
+
+  test-stale-uv:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Stale uv without uvx falls through to the upgrade path (issue #1593)
+        run: |
+          set -e
+          # Reproduce a pre-uvx uv: a `uv` binary on PATH with no `uvx` beside
+          # it. The old `uv || uvx` gate accepted this stale uv and then
+          # hard-exited at the uvx lookup; the gate now keys on uvx, so this
+          # must instead reach the install/upgrade path and end with a uvx.
+          STALE="$(mktemp -d)"
+          printf '#!/bin/sh\necho "uv 0.1.39"\n' > "$STALE/uv"
+          chmod +x "$STALE/uv"
+          # Curated PATH: the stale uv stub plus the system dirs the installer
+          # needs (sh, curl, python3), and crucially no real uvx. A real uvx
+          # leaking onto PATH would mask the bug, so fail loudly if present.
+          export PATH="$STALE:/usr/bin:/bin"
+          command -v uv >/dev/null || { echo "precondition: uv stub not on PATH"; exit 1; }
+          if command -v uvx >/dev/null; then
+            echo "precondition failed: uvx unexpectedly on PATH"; exit 1
+          fi
+          sh scripts/install.sh
+          # The upgrade path must have produced a working uvx.
+          export PATH="$HOME/.local/bin:$PATH"
+          command -v uvx >/dev/null || { echo "uvx missing after upgrade path"; exit 1; }
+          echo "OK: stale uv triggered the upgrade path, uvx now available"
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -88,7 +88,11 @@ printf "\n"
 
 # Step 1: Check/install uv
 printf "${YELLOW}Step 1: Checking for uv...${NC}\n"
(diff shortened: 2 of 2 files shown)


Patch76:
**Choices Made:**

- Gated Step 1 on `uvx` rather than parsing a minimum `uv` version. The presence of `uvx` is itself the "uv new enough to matter" signal, and this mirrors the existing `install-windows.ps1` check (`Get-Command uvx`); a version parse would be more brittle and unnecessary.
- Did not modify the later "Could not find uvx" exit branch. With the gate keyed on `uvx`, a uvx-less `uv` now flows into the existing install/upgrade branch — which already prints OS-aware manual-install hints on failure — so that exit is effectively unreachable in the reported scenario. Rewriting it would be unrelated churn.
- Kept the explanatory comment version-agnostic ("ships only with newer uv releases") instead of citing a version number, so it cannot rot.
- Added a dedicated `test-stale-uv` CI job rather than extending an existing installer job: it needs a curated PATH (a stale `uv` stub with no real `uvx`) that would interfere with the other scenarios' environments. It asserts both preconditions and fails loudly if a real `uvx` leaks onto PATH, so it cannot silently pass.

**Problems Encountered:**

- The report's root cause was correct, but its version detail ("uvx ships since uv 0.2.0") is inaccurate: `uvx` was added as a `uv tool run` alias in 0.2.19 and the interface stabilized in 0.3.0. The fix comment therefore avoids pinning a version.

PR #1610, opened 2026-06-14
fix(dashboards): recurse into nested cards in find_card + usable python_path

### LINKED ISSUE #1599: ha_config_get_dashboard card_type search misses custom:bubble-card while dashboard body search finds it
## Bug Description

`ha_config_get_dashboard` finds no cards for `card_type="custom:bubble-card"`, even though the same dashboard configuration contains `custom:bubble-card` and `ha_search(query="custom:bubble-card", search_types=["dashboard"])` finds dashboard-body matches.

## Steps to Reproduce

1. Run:
   `ha_config_get_dashboard(url_path="lovelace", card_type="custom:bubble-card", include_config=false)`
2. Run:
   `ha_config_get_dashboard(url_path="default", card_type="custom:bubble-card", include_config=false)`
3. Run:
   `ha_search(query="custom:bubble-card", search_types=["dashboard"], include_config=false)`

## Expected

`ha_config_get_dashboard(..., card_type="custom:bubble-card")` should return matching Bubble Card instances, or document that custom card type matching is unsupported or limited.

## Actual

`ha_config_get_dashboard` returns `match_count: 0` / no matches for `custom:bubble-card`.

`ha_search` confirms the dashboard config body contains `custom:bubble-card` in both `default` and `lovelace`.

## Observed Tool Calls

```text
ha_config_get_dashboard(url_path="lovelace", card_type="custom:bubble-card", include_config=false)
-> success=true, action=find_card, matches=[], match_count=0

ha_config_get_dashboard(url_path="default", card_type="custom:bubble-card", include_config=false)
-> success=true, action=find_card, matches=[], match_count=0

ha_search(query="custom:bubble-card", search_types=["dashboard"], include_config=false)
-> success=true, config_total_matches=2, dashboards include default and lovelace, match_in_config=true
```

## Environment

- ha-mcp: 7.7.0.dev421
- Installation: Home Assistant add-on
- Home Assistant: 2026.6.2
- Platform: HAOS on aarch64
- Relevant toggles: `enable_dashboard_partial_tools=true`, `enable_tool_search=true`

## Additional Context

The dashboard resources include Bubble Card as a Lovelace resource, and dashboard body search can see the literal `custom:bubble-card`. The mismatch appears specific to the structured `card_type` lookup path in `ha_config_get_dashboard`.

Patch76:
## What does this PR do?

Closes #1599.

`ha_config_get_dashboard` search mode (`_find_cards_in_config`) only walked top-level cards in views and sections, so a card nested inside a `vertical-stack`, `horizontal-stack`, `grid`, or `conditional` card was never matched by `card_type` / `entity_id` / `heading` — while `ha_search` (a literal config-body scan) still found it. The reporter saw this with `custom:bubble-card`; it is in fact card-type-agnostic (any nested card, built-in or custom).

This makes the card walk recursive (`_walk_card`): it descends into `cards` (list: stacks/grids) and `card` (dict: conditional and wrapper cards like `custom:auto-entities`), verified against the HA frontend docs. `picture-elements`' `elements` is a different shape and out of scope. A depth bound guards pathological configs and logs a warning rather than silently truncating.

Each match now also carries `python_path` — a Python subscript chain (`['views'][0]['cards'][1]['cards'][0]`) usable directly in `ha_config_set_dashboard(python_transform=...)`. The existing `jq_path` (jq dot-notation) is kept for back-compat but never worked spliced into `python_transform` (attribute access on a dict raises `AttributeError`); the docstring example, the search hint, the tool description, and the IndexError/KeyError retry suggestion now point at `python_path`. Flat `*_index` fields stay top-level-only; the path strings are the authoritative locator at any depth.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

Ran the relevant unit suite locally (`tests/src/unit/test_find_cards_badges_headers.py`, 24 pass) plus `ruff check` and `mypy src/` (clean); the full pytest/e2e suite runs in CI. New unit tests cover nested and doubly-nested matches, `card`-dict (conditional) nesting, nested-heading and header-nested search, flat-index semantics, the depth bound, and an end-to-end check that `python_path` splices into a working `safe_execute` expression. The two `test_python_transform.py` search-hint assertions were updated `jq_path` → `python_path`.

## Checklist


Patch76:
**Choices Made:**

- Made the card walk recurse type-agnostically — descending into any `cards` list and any `card` dict — rather than whitelisting container types. This covers built-in stacks/grids/conditional cards and custom wrappers (`custom:auto-entities`, stack-in-card, etc.) without a brittle type list. The two nesting keys were confirmed against the HA frontend docs; `picture-elements`' `elements` is a distinct shape (elements, not cards) and is left out of scope.
- Added `python_path` (a Python subscript chain) as the usable locator and kept `jq_path` (jq dot-notation) for back-compat. The existing `jq_path` never actually worked spliced into `python_transform` — `config.views[0]...` is attribute access on a plain dict and raises `AttributeError` — so the documented find→edit workflow was broken before this PR. The docstring example, the search hint, the tool description, and the IndexError/KeyError retry suggestion now point at `python_path`.
- Kept the flat `view_index` / `section_index` / `card_index` fields top-level-only. They cannot express arbitrary nesting depth, and repurposing `card_index` would break existing consumers and tests; the path strings are the authoritative locator at any depth.
- Added a recursion depth bound that logs a warning when exceeded, so truncation on a pathological/malformed config is visible rather than silent. It sits far above any real dashboard nesting.

**Problems Encountered:**

- The report framed this as specific to custom card-type matching (`custom:bubble-card`). It is actually card-type-agnostic: any nested card, including a built-in `gauge`, was missed, because the traversal only ever visited top-level cards. Reproduced with a nested built-in card to confirm.
- Fixing the broken `python_transform` locator was coupled into this PR rather than split out, because a nested match the agent cannot act on is only half the fix — finding the card is useful only if the returned path can edit it.

Patch76:
It's some small fix ... but have a look, please.

kingpanther13 (changes requested):
Reviewed the branch against real nested dashboard configs (running the PR's own `_walk_card` standalone) plus a multi-agent pass over code/tests/error-handling/comments. The core fix is correct **for the reported case** — cards nested in `vertical-stack`/`horizontal-stack`/`grid` (`cards`) and `conditional`/wrapper cards (`card`) are now found, the recursion wiring into flat views, sections, and the header card is right, the depth bound has no off-by-one, and the recursion masks no exceptions. The `jq_path`→`python_path` discovery is a genuine second bug fix (the old `jq_path` was never usable spliced into `python_transform` — `config.views[0]` is attribute access on a dict → `AttributeError`); that swap is complete and clean across docstrings, the hint, `server.py`, and the retry suggestion, and the EXAMPLES block is runnable (the new unit test proves it).

But the PR's stated thesis — "it is in fact card-type-agnostic (any nested card, built-in or custom)" — is not yet met, there are two silent-failure issues against this repo's own warn-don't-truncate contract, and the new tests have a structural blind spot. None of the below is a follow-up; it all lands here.

## High

**H1 — `custom:button-card`'s `custom_fields.<name>.card` nesting is still missed; the "any nested card, built-in or custom" claim doesn't hold.** `_walk_card` descends only a top-level `cards` list or top-level `card` dict (`tools_config_dashboards.py:192-193`, `265-289`). `custom:button-card` — one of the most-installed HACS cards — embeds full sub-cards under `custom_fields.<fieldname>.card`, which is neither key on the button-card itself, so the descent stops at the wrapper. Running the PR's own `_walk_card` against a button-card-with-`custom_fields` config returns: `card_type="tile"` → 0, nested `custom:button-card` → only the top wrapper, `entity_id` → 0. This is not hypothetical: a common pattern wraps an entire view in one `custom:button-card` and puts the whole UI under `custom_fields.content.card` → a `vertical-stack` of grids/tiles — every tile, picture-entity, and nested button-card stays invisible after this PR. Fix: in `_walk_card`, also descend each `custom_fields[*]` value that is a dict carrying a `card` (dict) or `cards` (list) key, extending the `jq_path`/`python_path` prefixes through `['custom_fields']['<name>']['card']`.

**H2 — Depth-bound truncation is silent at the response boundary.** When `depth > _MAX_CARD_DEPTH`, `_walk_card` logs a server-side `logger.warning` and drops every card below that point (`tools_config_dashboards.py:218-236`), but the `find_card` response (`:986-1004`) carries no `warnings[]` entry — `match_count` reads as complete. The comment at `:222-229` claims this makes truncation "visible rather than silently dropping," yet it's only visible in server logs the agent never sees. This file already has the idiom (`_note_screenshot_ignored` → `result.setdefault("warnings", []).append(...)`, `:656-661`) and AGENTS.md § Return Values mandates a top-level `warnings: list[str]` for degraded results. Fix: thread a `truncated` flag out of `_walk_card` (return `(matches, truncated)` or pass a small accumulator) and append a `warnings` entry in the response builder when it fires.

**H3 — `match_count: 0` with `"Try broader search criteria"` is a confident false-negative on un-traversed surfaces.** For a card that demonstrably exists under a key the walker doesn't descend (`custom_fields`, picture-elements `elements`), `find_card` returns `success: true, match_count: 0` and a hint that broadening criteria will help — it won't; that surface was never searched. This is precisely the #1599 symptom, just relocated to the remaining axes. Even where you choose not to traverse a surface, disclose it: on a 0-match result over a non-trivial config, add a `warnings` entry stating the search covered standard `cards`/`card` containers only and point at get-mode for the full config, and drop the misleading sole-remedy hint at `:1002`.

## Medium

**M1 — Recursion + `_card_matches` can false-match non-card dicts and widens heading/title matches.** With `card_type` omitted, `_card_matches` skips the type check entirely, so any dict reached under a key literally named `card`/`cards` that carries an `entity`/`entities`/`title`/`heading` field now matches and emits a structurally valid `python_path` that would edit a non-card; `heading=` likewise matches a container's `title` at every depth, widening false positives versus the pre-PR top-level-only behavior. Low probability on standard dashboards, higher on heavily customized ones — and it's a precision regression in a feature whose whole value is path accuracy. Fix: require `"type" in node` before treating a descended node as a card (or have `_card_matches` require a `type` key to match on `entity`/`heading` alone). Confirm whether the `heading`→`title` widening is intended.

**M2 — picture-elements `elements[].card` is declared out of scope but is a real nesting shape.** The comment at `:186-191` scopes it out deliberately; per the no-defer norm, either descend it here too, or at minimum it must be covered by H3's disclosure so a 0-match on a picture-elements dashboard doesn't read as authoritative.

**M3 — `custom:state-switch`'s `states:` (name→card map) is another uncovered nesting key** — same treatment as M2.

**M4 (tests) — `test_deep_nesting_bounded` is near-degenerate.** It nests `_MAX_CARD_DEPTH + 5` and asserts `matches == []`, which passes for any bound from ~1 to `MAX+4`, for a flipped `>=` guard, or even for fully broken recursion — it pins neither the bound's value nor its off-by-one, never asserts a leaf *just inside* the bound is still found, and never asserts the warning fired. Add three: (a) boundary-inside — a leaf at `depth == _MAX_CARD_DEPTH` is returned with the correct `python_path`; (b) boundary-outside — `depth == _MAX_CARD_DEPTH + 1` is dropped; (c) a `caplog` assertion that the depth-bound warning was emitted (this also locks H2).

**M5 (tests) — `python_path` round-trip is proven for the flat-view prefix only.** The diff emits three distinct prefixes (flat, sections `['sections'][i]['cards'][j]`, header `['header']['card']`); only the flat one is executed through `safe_execute` (`test_python_path_is_usable_in_safe_execute`). The sections/header prefixes are checked by string-equality only — a typo there passes the test but raises `KeyError`/`IndexError` on a real edit. Add `safe_execute` round-trips for a sections-nested match and a header-nested match.

**M6 (tests) — no multi-sibling / both-levels-match coverage, and mixed `cards`+`card` on one node is untested.** Every nested test asserts exactly one match. Add: two matching tiles in one stack → `len == 2` with distinct `...['cards'][0]` / `['cards'][1]`; a container that matches AND has a matching descendant → both returned (pins the match-self-then-recurse order); a single node carrying both `cards` and `card` with a match under each.

**M7 (tests) — characterization tests for the known gaps** so the boundary is explicit and a future fix flips a documented assertion: button-card `custom_fields` (H1), `elements` (M2), state-switch (M3). If H1/M2/M3 are fixed here (preferred), these become positive assertions instead.

**M8 (tests) — no e2e exercises the actual user promise end-to-end:** find a nested card via `ha_config_get_dashboard`, feed its `python_path` into `ha_config_set_dashboard(python_transform=...)` against a live dashboard, re-fetch, assert the nested card changed. The e2e file was only touched to swap two hint assertions. `ha_config_get_dashboard` is already e2e-covered so a new file isn't strictly mandated, but this is the gap between "unit-proven logic" and "the feature shipped" — recommend adding one.

## Low / docs

**D1 — the `_NESTED_*` comment (`:186-191`) understates the gap.** It names picture-elements as the *only* out-of-scope shape and claims "any custom wrapper that follows the same `cards`/`card` convention is covered" — true as far as it goes, but it omits the common `custom:button-card` `custom_fields` and state-switch conventions that don't follow it, so the net message reads as broader coverage than exists. Reword to state the rule positively and name the gap class.

**D2 — "at any nesting depth" over-promises in three docstrings** (`:201-205` `_walk_card`, `:306-307` `_find_cards_in_config`, `:836` the tool docstring) — literally false past the depth bound and for unsupported keys. Reword to bounded, accurate phrasing ("for nested as well as top-level cards, descending `cards`/`card` up to a depth bound").

**D3 — `python_path` is a bare subscript suffix usable only when appended after `config`.** The MODE 2 description (`:834-835`) calls it "a Python subscript chain for `python_transform`" without stating the `config` prefix requirement; an agent reading only that summary could emit `python_transform=python_path + ...` with no `config`. The EXAMPLES block is correct — just make the prefix requirement explicit in the field description.

**D4 — `include_config=True` duplicates nested subtrees.** A matched container's `card_config` contains descendants that are themselves separate matches with their own `card_config`, so deep stacks multiply payload super-linearly. The default path (`include_config=False`) strips `card_config` (`:980-982`), so it doesn't bite by default. Document it in the `include_config` param, or omit nested `cards`/`card` from a matched container's emitted `card_config`.

**D5 — badge matches now advertise a `python_path` that fails for bare-string badges.** The badge branch emits `['views'][i]['badges'][j]` (`:339`), and the universal hint now says "use python_path for targeted updates." When the badge is a bare entity-id string (the common form), splicing `config[...]['badges'][j]['icon'] = ...` raises `TypeError` (a string isn't subscript-assignable). Note that string-form badges must be converted to dict form first, or don't emit `python_path` for string badges.

**D6 (awareness, not blocking) — the module is now ~1967 lines** (AGENTS.md flags ~1000). This PR isn't the cause and splitting is a separate concern; noting it for the maintainer.

Happy to hand over the `custom_fields` traversal and the test additions as concrete diffs if that's useful.

Patch76:
I don't get it ... I use this many skills'n stuff ... but you alway find such big gaps. I’m a rational, logical thinker, and my questions for Claude shouldn’t be that wrong or incomplete. But still, there are major gaps. This is VERY exhausting for me. I’ve spent days verifying my own workflows – involving Gates, research, and the use of skills and tools (multi-agent pass over code/tests/error-handling/comments) – over and over again. And yet ... suckz, bro!

kingpanther13:
I was actually already working on this one that's part of it lol. You can always try opening up a fresh context Claude and ask if to run pr review toolkit on your own PR, or just run pr review toolkit within the same session can help too, since the agents don't have as much context they dig up more stuff. Of course sometimes they find things that aren't really there because all they know is that they must find something.

kingpanther13:
Oh and sometimes I use Codex to do a review, that almost always finds something that Claude missed. I didn't do that here but I've been using it lately and it finds things that usually check out.

Patch76:
Thanks for running `_walk_card` against real configs — the `custom_fields` gap and the warn-don't-truncate points were exactly right. Everything below landed in `603619f6` (on top of the two earlier branch commits); nothing deferred.

## High

**H1 — button-card `custom_fields` nesting.** Fixed. `_walk_card` now descends each `custom_fields.<name>` value as a node, so `custom:button-card` sub-cards under `custom_fields.<name>.card` / `.cards` are found, with the prefixes extended through `['custom_fields']['<name>']`. The whole-view-in-one-button-card pattern (UI under `custom_fields.content.card`) resolves. `test_finds_card_nested_in_button_card_custom_fields` asserts the full `python_path`.

**H2 — depth-bound truncation now disclosed at the boundary.** Fixed. `_walk_card` threads a `truncation` accumulator; the find_card response appends a top-level `warnings[]` entry when the bound was hit (the `logger.warning` stays for server logs). Matches the `warnings: list[str]` contract in AGENTS.md → Return Values. Boundary-inside (found at `depth == _MAX_CARD_DEPTH`), boundary-outside (dropped, accumulator populated, `caplog` asserts the warning) are tested.

**H3 — false-negative on un-traversed surfaces.** Fixed. On a 0-match over a populated dashboard (`_config_has_searchable_cards`), the response adds a `warnings[]` entry stating only the standard `cards` / `card` / `custom_fields` containers were searched and pointing at get-mode; the misleading sole-remedy hint is replaced with a two-part one.

## Medium

**M1 — non-card false-matches.** Fixed the precision half: a descended dict must carry a `type` key before it can match (`"type" in card and _card_matches(...)`), so action targets / entity rows / style blocks reached under `card` / `cards` / `custom_fields` no longer match or emit a path that edits a non-card. `test_typeless_dict_under_cards_not_matched` covers it. On the `heading` → `title` widening: that fallback is pre-existing `_card_matches` behaviour, now gated behind the `type` requirement. I kept it rather than restricting to `type == "heading"`, since a heading/markdown card legitimately carries a `title` — I can narrow it to `type == "heading"` if you prefer the tighter match.

**M2 — picture-elements.** Per the HA frontend docs, picture-elements nests *elements*, not cards — no element type carries a nested `card` (the `conditional` element nests further `elements`). So there is no card-bearing key to descend; I covered it through H3's disclosure plus a characterization test (`test_characterization_picture_elements_not_traversed`) pinning the boundary. If you have an `elements[].card` shape I'm not seeing, point me at it and I'll add the descent.

**M3 — state-switch `states`.** Same treatment: disclosed via H3 and pinned by `test_characterization_state_switch_states_not_traversed`. Descending the `states` name→card map is a clean follow-on if we want state-switch as a first-class container; the disclosure keeps a 0-match honest until then.

**M4 — depth test strengthened.** Replaced the degenerate test with boundary-inside / boundary-outside / `caplog`-warning assertions (see H2).

**M5 — python_path round-trips.** Added `safe_execute` round-trips for the sections prefix and the header prefix (not just the flat one) — each splices the emitted `python_path` after `config` and asserts the nested card mutated.

**M6 — sibling / both-levels / mixed coverage.** Added: two tiles in one stack → two matches with distinct `['cards'][0]` / `[1]`; a container that matches and has a matching descendant → both returned (pins match-self-then-recurse); a node carrying both `cards` and `card` → both descended.

**M7 — characterization tests.** button-card `custom_fields` is now a positive assertion (H1); state-switch and picture-elements are characterization tests pinning the documented gaps, ready to flip to positive when traversed.

**M8 — full e2e.** Not added: `ha_config_get_dashboard` is already e2e-covered, and the M5 unit round-trips execute all three `python_path` prefixes through `safe_execute`, which is the locator-correctness guarantee the e2e would add. I can add the find→transform→re-fetch e2e if you want the websocket-layer integration pinned too.

## Low / docs

D1 reworded to name the gap class (state-switch / picture-elements) and the deliberate no-generic-walk rationale (tile `features` and view `conditions` also carry `type` and would false-match). D2 "at any nesting depth" → bounded phrasing across the three docstrings. D3 states the `config`-prefix requirement for `python_path` in the MODE 2 description. D4 documents the `include_config=True` nested-subtree duplication on the param. D5 stops emitting `python_path` for bare-string badges (not subscript-assignable; `jq_path` still emitted). D6 (module size) left for a separate split as you noted.

kingpanther13 (changes requested):
Verified `603619f6` against the prior review: H1 (custom_fields descent — empirically `tile` 0→2, correct splice-able path), H2 (truncation accumulator → `warnings[]`), M1 (type-gate — empirically drops a typeless dict), M4 (boundary inside/outside + caplog), M5 (sections + header round-trips), M6, M7, and D1–D5 all landed. And you were right on M2 — picture-elements nests *elements*, not cards; my "elements[].card" finding was wrong. Conceded.

I then exercised the new tool live against a real HA instance (a throwaway copy of a button-card-wrapped dashboard, seeded with the edge shapes below). The custom_fields recursion works well in practice — a dashboard whose entire UI sits under one `custom:button-card`'s `custom_fields.content.card` now resolves all its nested tiles / picture-entities / button-cards with correct, splice-able `python_path`s, and a `python_transform` round-trip through a clean `python_path` lands and re-reads correctly. The items below are what that live pass surfaced; each now has a concrete reproduction. All land here.

## Required

**1. The H3 disclosure is the wrong shape — it's both suppressible and cries wolf.** It's gated on `not matches and _config_has_searchable_cards(config)`, and `_config_has_searchable_cards` is true for essentially any populated dashboard. Two consequences, both reproduced live:
- **Suppressible.** An un-walkable container (state-switch `states`, picture-elements `elements`, an unknown `custom:*` wrapper) that *itself* matches the search makes `matches` non-empty, so the warning is suppressed and its un-traversed children are silently dropped. Live: a `custom:state-switch` card matched a `card_type` search → no warning — yet `include_config=true` shows that same match's `card_config` literally contains a tile under `states.on` that was never surfaced and never disclosed. So the #1599 false-negative class is only half-closed.
- **Cry-wolf.** On a genuine true-negative 0-match the warning fires anyway. Live: a dashboard containing *no* state-switch and *no* picture-elements returned the warning naming both on two different absent-card searches. Following the hint's own advice then dead-ends — "try other criteria" yields the same false warning again; "fetch the full config" works but is the heavy op `find_card` exists to avoid.

One fix for both: detect un-coverable shapes *during the walk* (a `type`-bearing node carrying a non-traversed child-bearing key) and emit the warning iff one is present, independent of `match_count`.

**2. Traverse `custom:state-switch` `states` (don't just disclose it).** Unlike picture-elements, `states` is a card-bearing `name → card` map — a clean ~8-line mirror of the `custom_fields` descent you just added. Live: a tile under `states.on` is invisible to both `card_type=tile` and an `entity_id` search for that tile's entity. After traversing, flip `test_characterization_state_switch_states_not_traversed` to a positive assertion and drop `states` from both the disclosure text and the `_NESTED_*` comment (three coupled sites — update in lockstep).

**3. Make the `python_path` / `jq_path` key interpolation quote/dot-safe (prerequisite for #2).** Segments are built with raw interpolation: `['{name}']`. Reproduced live end-to-end: a `custom_fields` key `o'brien` yields `python_path: ...['o'brien']['card']`, and pasting that into `python_transform` exactly as the tool's hint instructs returns `Syntax error: unterminated string literal` — so a card under an apostrophe key is **findable but uneditable** (its `jq_path` is broken too). A key with a `.` (`weird.dotkey`) breaks only `jq_path` (`.custom_fields.weird.dotkey.card` reads the dot as nesting); its `python_path` survives and edited fine. Build `python_path` segments with `repr(name)` (→ `["o'brien"]`) and quote non-identifier `jq_path` keys; apply to both the custom_fields and the new states descent. This is a live bug for button-card field names now and a more-likely one for state names (`on'hold`) once #2 lands.

**4. Breadcrumb the malformed-slot skips.** The depth branch now records into `truncation`, but the `not isinstance(card, dict)` early return (and the non-dict `card`/`custom_fields` no-ops) skip structurally-present data with no record — inconsistent with the mechanism this commit added. Record non-None malformed slots (in `truncation` or a `skipped` list), or at minimum `logger.debug`.

**5. Add the websocket-layer e2e round-trip.** Find a nested card → feed its `python_path` into `ha_config_set_dashboard(python_transform=...)` → re-fetch → assert the nested card mutated. The clean-key round-trip works live (verified), so this pins it at the integration layer; include a special-character-key case (e.g. a `custom_fields` key with an apostrophe) so the e2e also catches the #3 regression, not just the happy path. `test_lifecycle.py` (Docker/CI).

**6. Cover the disclosure layer — it currently has zero tests.** `_config_has_searchable_cards` and the response-level `warnings[]` are unasserted, so an over/under-warn regression ships CI-green. Add (the helper is pure `dict→bool`; the response uses the existing harness in `test_config_get_dashboard_search_error.py`): warning fires when an un-coverable shape is present; silent on empty/strategy; silent on a true-negative with no such shape; the response carries a top-level `warnings: list[str]`. Plus the untested `custom_fields` edge cases — a field that is itself a card with `type` (matches at `custom_fields.<name>`), a template-string field (skipped, no raise), multiple fields with distinct paths, doubly-nested custom_fields — and the badge `python_path` omission (string badge omits, dict badge includes).

## Doc nit

The picture-elements comment asserts an absolute ("no element carries a nested `card`") the tests can't guarantee; describe the code's actual contract instead ("`elements` is not a traversed key").

All in this PR.

Patch76:
All six items addressed in `685428b3`, plus a regression an own-author review pass caught (below).

## Required

**Item 1 — disclosure shape.** Reworked so it keys off the *presence* of a non-traversed child-bearing key during the walk, not off `match_count`. `_walk_card` collects any walked card carrying a key in `_UNTRAVERSED_NESTED_KEYS` (`("elements",)`, `tools_config_dashboards.py:212`) into an `uncovered` accumulator (`:334`), and the search emits the warning iff that list is non-empty (`:1224`). This closes both halves: a matching un-walkable container no longer suppresses the warning (collection runs regardless of whether anything matched), and a true-negative 0-match over a fully-coverable dashboard no longer fires it. The old `_config_has_searchable_cards` gate is removed (dead after this change).

**Item 2 — traverse `custom:state-switch` `states`.** Added an `_NESTED_STATES_KEY` (`:206`) descent block (`:437`) mirroring the `custom_fields` descent: each `name -> card` value is walked directly. Flipped `test_characterization_state_switch_states_not_traversed` to a positive `test_state_switch_states_traversed`, dropped `states` from the disclosure text and the descent-rule comment, and updated the tool docstring — the three coupled sites.

**Item 3 — quote/dot-safe key interpolation.** `python_path` segments are now built with `repr()` via `_py_key` (`:218`); `jq_path` segments quote non-identifier keys via `_jq_key` (`:228` — dot notation for identifier-safe keys, else `[json.dumps(key)]`). Applied to both the `custom_fields` and the new `states` descent. A `custom_fields` key `o'brien` now yields `...["o'brien"]['card']` (was an unterminated literal) and a dotted key is bracketed in `jq_path` so the dot is inert. Round-trip-verified through `safe_execute` in the unit tests.

**Item 4 — breadcrumb malformed-slot skips.** The non-dict early return now logs a debug breadcrumb when the slot is structurally present (non-`None`), and the non-dict `card` / `cards` / `custom_fields` / `states` no-ops each log via `logger.debug` instead of dropping silently. I kept this at `debug` per your "at minimum logger.debug" — a malformed config slot is a data condition, not a search-completeness condition like the depth-bound warning.

**Item 5 — websocket-layer e2e round-trip.** `test_find_nested_card_python_path_round_trip` in `test_lifecycle.py` (`TestFindCard`) seeds a button-card wrapping a tile under a `custom_fields` key with an apostrophe, finds it, splices the returned `python_path` into `ha_config_set_dashboard(python_transform=...)`, re-fetches, and asserts the nested card mutated — plus that the parent button-card is untouched and the card count is unchanged, so a `python_path` resolving to the wrong node would fail rather than pass.

**Item 6 — cover the disclosure layer.** Added `TestFindCardDisclosureWarnings` in `test_config_get_dashboard_search_error.py`: warning fires on a present un-coverable shape even when the container itself matches; silent on a true-negative with no such shape; silent on empty; `warnings` is a top-level `list[str]`. Added the `custom_fields` edge cases (field-is-itself-a-card, template-string field skipped without raise, multiple fields with distinct paths, doubly-nested `custom_fields`) and the badge `python_path` presence/omission pair.

## Doc nit

The picture-elements comment now states the code contract (`elements` is not a traversed key, so a node carrying it is disclosed at the boundary) rather than the unguaranteeable absolute.

## One regression the review pass caught

Item 3's key rendering introduced a crash path the old raw interpolation did not have: a non-string mapping key (e.g. an int key from a corrupted config) would raise `TypeError` in `_jq_key`'s `re.fullmatch` and abort the whole search. Guarded with an `isinstance(name, str)` skip + breadcrumb (`_log_non_str_key`, `:240`) on both the `custom_fields` and `states` loops, with a regression test.

One deliberate non-change: the `states` descent is ungated (any card carrying a `states` dict is traversed), matching the ungated `custom_fields` descent. The `type`-gate in `_walk_card` means a `states` map of non-card dicts contributes no false matches; the only cost is traversal work, which the depth bound caps. Pinned by `test_non_card_state_value_skipped`.

kingpanther13 (approved):
Approving. The recursion now meets the "any nested card" thesis across `cards`/`card`/`custom_fields`/`states`; disclosure is presence-based, so a matching un-walkable container no longer suppresses the warning and a true negative no longer cries wolf; and locator keys are quote/dot-safe so special-character field/state names stay findable and editable. Verified live (a button-card `custom_fields`-wrapped dashboard resolves all its nested cards, and a `python_path` round-trip lands) and against the new unit + e2e coverage — including the apostrophe-key e2e and the self-caught non-string-key guard.

Bounding `_UNTRAVERSED_NESTED_KEYS` to `elements` is the right call: unknown custom keys are inherently undetectable without the generic `type`-walk that would false-match `features`/`conditions`.

Thanks for the iterations.

kingpanther13:
Disregard the code scanning thing, it's a feature I turned on that I probably shouldn't have ,this is safe to merge. I'm going to go ahead and just merge it now

PR #1626, opened 2026-06-15
feat: allow filesystem tools to access HAOS sibling volumes

### LINKED ISSUE #1586: [FEATURE] Allow filesystem tools to access HAOS sibling volumes (/share, /media, /ssl, /backup)

The custom filesystem directories feature added in v7.7.0 (#1568) works great for extending read/write access to subdirectories of /config. However, it only accepts paths relative to the config directory and blocks absolute paths and ../ traversal. This means the other HAOS volume mounts that the add-on container already has access to (/share, /media, /ssl, /backup) are completely unreachable from the filesystem tools.

I'd like the ability to grant the filesystem tools access to these HAOS volumes, either through a separate "Allowed HAOS volumes" setting in the web Settings UI, or by extending the custom directories field to accept the fixed set of known HAOS volume names (share, media, ssl, backup) which resolve to their absolute mount points internally.

Ideally each volume would have independent read/write control. /share and /media are natural read/write candidates, while /ssl and /backup would typically only need read access.

### Why do you need this?

On HAOS, /share is the standard cross-add-on exchange directory. I use it to pass files between add-ons (EMHASS output, Node-RED exports, custom scripts). Right now the only way to read or write anything there from an AI assistant is through SSH or the File Editor UI, neither of which is reachable through the MCP tool layer.

I tried proxying through both the Advanced SSH & Web Terminal add-on and the File Editor add-on using ha_manage_addon, but SSH has no REST API for command execution (it's a ttyd web terminal) and File Editor returns HTTP 420 on every API call via ingress (session auth required).

/media is similarly useful for managing TTS output, camera snapshots, and dashboard images. /backup would be handy for verifying a backup exists before a risky operation (ha_manage_backup handles creation and restore, but there's no way to browse backup contents or confirm a specific backup file landed). /ssl is a minor convenience for checking certificate expiry.

The add-on container already maps all four volumes (the config.yaml map field lists share, media, ssl, backup). The filesystem is accessible at the process level; only the tool-level allowlist prevents it.

### Additional context

Environment: HAOS on RPi, ha-mcp v7.7.0 stable add-on, ha_mcp_tools v7.7.0 custom component installed via HACS, beta filesystem tools enabled via web Settings UI.

The custom directories feature (#1568) is the natural extension point. The security model just needs a way to accept these specific volume paths without opening arbitrary absolute paths. A simple checkbox list per volume in the web Settings UI would be the cleanest approach.

Workarounds attempted and failed:
- ha_manage_addon proxy to Advanced SSH (a0d7b954_ssh): no REST API, ttyd only
- ha_manage_addon proxy to File Editor (core_configurator): HTTP 420 on all API calls
- Custom directories with ../share: blocked by path traversal check
- Custom directories with /share: blocked by absolute path check

kingpanther13:
## What does this PR do?

Closes #1586.

**Sibling-volume access (#1586).** The custom-directory allowlist (#1568)
rejected every absolute path, so the HAOS volumes the container already mounts
(`/share`, `/media`, `/ssl`, `/backup`) were unreachable. Users can now add one
of those roots — or a subdirectory — in the existing "Custom filesystem
directories" settings editor; the `ha_mcp_tools` component enforces access
against the volume root exactly as it does config-relative dirs (read+write; the
`.storage`/`secrets.yaml` deny floor still applies). Component → 0.9.0, with
`MIN_COMPONENT_VERSION` raised to match.

**`ha_manage_backup` snapshot list.** Adds `(snapshot, list)`, exposing HA's
`backup/info` inventory so a caller can discover backup IDs / confirm a backup
landed (read-only, exempt in read-only mode) — the maintainer-requested fold-in
from #1586.

**Security hardening.** Permission was checked on the lexically-normalized path
(which erases an intermediate symlink), but the file was opened from the raw
path the kernel resolves physically — so a `<allowed>/<symlink>/..` could escape
the allowlist to arbitrary files (read/write/delete). Enforcement now resolves
the RAW path (symlinks then `..`, as `open(2)` does) and re-checks containment,
closing it for the new volume paths and the existing config-relative allowlist.
The fix is entirely in the custom component (the server does no file I/O), so it
ships with the component update.

**CI + docs.** `custom_components/` is now covered by CI's ruff and mypy (it was
`src/`-only); latent mypy findings fixed. Corrected stale beta/README docs
(per-edit backups *are* restorable via `ha_manage_backup`; writable dirs include
`dashboards/` plus configured custom dirs/volumes).

## Type of change
- [x] 🐛 Bug fix
- [x] ✨ New feature
- [x] 📚 Documentation
- [x] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

## Checklist

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -148,7 +148,7 @@ jobs:
       run: uv sync --dev
 
     - name: Run ruff check
-      run: uv run ruff check src/ tests/ homeassistant-addon/ homeassistant-addon-webhook-proxy/ scripts/
+      run: uv run ruff check src/ tests/ custom_components/ homeassistant-addon/ homeassistant-addon-webhook-proxy/ scripts/
 
     - name: Run ruff format check on changed Python files
       run: |
@@ -210,7 +210,7 @@ jobs:
 
     - name: Run mypy
       run: |
-        uv run mypy src/ homeassistant-addon/ scripts/
+        uv run mypy src/ custom_components/ homeassistant-addon/ scripts/
         uv run mypy homeassistant-addon-webhook-proxy/start.py
 
   # Fast unit tests (no Docker, no HA instance needed)
--- a/README.md
+++ b/README.md
@@ -245,8 +245,8 @@ Some tools require a companion custom component installed in Home Assistant. Sta
 | Tool | Description |
 |------|-------------|
 | `ha_config_set_yaml` *(beta)* | Safely add, replace, or remove top-level YAML keys in `configuration.yaml` and package files (automatic backup, validation, and config check) |
-| `ha_list_files` *(beta)* | List files in allowed directories (www/, themes/, custom_templates/) |
-| `ha_read_file` *(beta)* | Read files from allowed paths (config YAML, logs, www/, themes/, custom_templates/, custom_components/) |
+| `ha_list_files` *(beta)* | List files in allowed directories |
+| `ha_read_file` *(beta)* | Read files from allowed paths (config YAML, logs, and allowed directories) |
 | `ha_write_file` *(beta)* | Write files to allowed directories |
 | `ha_delete_file` *(beta)* | Delete files from allowed directories |
 
--- a/custom_components/ha_mcp_tools/__init__.py
+++ b/custom_components/ha_mcp_tools/__init__.py
@@ -492,7 +492,14 @@ def _is_volume_path_allowed(abs_path: str, extra_dirs: list[str] | None) -> bool
         return False
     if _violates_volume_deny_floor(normalized):
         return False
-    if not _matches_extra_dir(normalized, extra_dirs):
+    # POSIX-explicit allowlist match (not _matches_extra_dir, which joins on
+    # os.sep): a volume path is inherently "/"-separated, so match on a "/"
+    # boundary so a configured "/share" grants "/share" and "/share/..." but
+    # never "/shared". Config-relative entries in the mixed list never match an
+    # absolute path here.
+    if not extra_dirs or not any(
+        normalized == d or normalized.startswith(d + "/") for d in extra_dirs
+    ):
         return False
     return _resolves_within(Path(root), abs_path)
 
--- a/custom_components/ha_mcp_tools/__init__.py
+++ b/custom_components/ha_mcp_tools/__init__.py
@@ -10,11 +10,13 @@ import errno
 import fnmatch
 import logging
 import os
+import posixpath
 import secrets
 import shutil
+from collections.abc import Awaitable, Callable
 from datetime import datetime
 from io import StringIO
-from pathlib import Path
+from pathlib import Path, PurePosixPath
 from typing import Any
 
 import voluptuous as vol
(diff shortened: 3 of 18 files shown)


Patch76 (approved):
Verified against `e745f5b`:

- **Symlink-safe containment.** `_resolves_within` at `custom_components/ha_mcp_tools/__init__.py:413` resolves the raw input via `Path.resolve()` (matching `open(2)` semantics) and checks `is_relative_to(base.resolve())`, fail-closed on `(OSError, ValueError)`. `_violates_volume_deny_floor` at L435 applies the `.storage` / `secrets.yaml` floor against both the requested segments and the resolved target case-insensitively, so a renamed symlink pointing at `.storage/auth` or `secrets.yaml` is denied even when the lexical path looks innocuous. `TestVolumeSymlinkEscape` and `TestResolvesWithin` pin the regression for both the `<volume>/<symlink>/..` lexical-erase escape and the resolved-target floor.
- **Volume gate (`_is_volume_path_allowed`, L478-504).** Three-pass: `_volume_root_for` boundary-matches on `/` so `/shared` cannot ride on `/share`, deny floor on the normpath'd path, allowlist match on a POSIX `/` boundary (not `os.sep` — the volumes are inherently POSIX and the comment at L495-499 calls that out), then `_resolves_within`. `_normalize_extra_dir` at L507 makes `extra_dirs` canonical upstream, so Gemini's normpath suggestion at L504 would be redundant against the contract that hits this function.
- **`list_backups` (`src/ha_mcp/tools/backup.py:769`).** WS error tuple handled, undated entries sink via `_date_floor = datetime.min.replace(tzinfo=UTC)` so a mix of `…Z` and `…+00:00` cannot misorder lexically, `total` surfaced separately from `count` so truncation is observable. `_summarize_backup` (L734) takes `max` size across agents, coerces float→int and excludes `bool`. The `logger.error` at L828 that Gemini flagged matches the pre-existing sibling pattern at L514 and L712 — a separate cross-cutting refactor PR is the right place for that, not this one.
- **`MIN_COMPONENT_VERSION` bump to 0.9.0.** Cross-repo gate at `src/ha_mcp/tools/tools_filesystem.py:64` plus the version check that fails before any volume-path operation reaches an old component, so a stale install surfaces the actionable upgrade prompt rather than a silent path mismatch.
- **Coverage.** 1324 lines in `tests/src/unit/test_custom_component_filesystem.py` — `TestResolvesWithin`, `TestVolumePathEnforcement`, `TestVolumeSymlinkEscape`, `TestNormalizeVolumeDir`, `TestListFilesSyncVolumePaths`, `TestDenyFloor`'s symlink-target branches — plus `TestSummarizeBackup` + undated-sort pin in `tests/src/unit/test_backup_agent_lookup.py`. Pipeline including HAOS E2E (inaddon + jenkins) green.

Two non-blocking observations:

- `get_connected_ws_client` returns `(client | None, dict | None)`; the `error or create_error_response(...)` fallback at backup.py:785-790 is unreachable behind the `if error:` guard. Belt-and-suspenders, harmless.
- A direct test injecting `OSError` / `ValueError` on `Path.resolve()` to pin the False return of `_resolves_within` would close the only branch the escape tests cover indirectly.

PR #1642, opened 2026-06-16
feat: add advanced setting to pin the stdio settings-UI sidecar port

### LINKED ISSUE #1587: [FEATURE] Advanced setting to pin the stdio settings-UI sidecar port

Add an advanced setting (settings UI Advanced tab + env var) to pin the stdio settings-UI sidecar to a fixed port instead of picking a random free port at every spawn (`stdio_settings_sidecar.py`).

### Why do you need this?

The sidecar's random port means a new browser origin on every ha-mcp restart in stdio mode. Anything origin-scoped breaks across sessions: bookmarks to the settings page go stale, and browser-side localStorage (e.g. the theme/accessibility preferences from PR #1574, until those move server-side) resets every session. A pinned port gives stdio users a stable URL and stable origin, matching the docker and addon experience.

### Additional context

Split out of the PR #1574 review (request-changes item 1): theme/accessibility prefs are being persisted server-side in that PR, but a stable sidecar port is independently useful. Default should remain the current random-port behavior (no port conflicts out of the box); the pin is opt-in. If the pinned port is busy at spawn, log a clear error and fall back to random rather than failing the sidecar.

SealKan:
## What does this PR do?

Adds an advanced setting to pin the stdio settings-UI sidecar to a fixed port instead of a random free port at every spawn (#1587).

`sidecar_pin_port` (env `HA_MCP_SIDECAR_PORT`, Advanced settings tab): `0` (default) keeps today's ephemeral-port behaviour; `1024-65535` pins the port so the settings URL and browser origin stay stable across restarts (stops bookmarks from going stale and stops browser-side localStorage from resetting each session). If the pinned port is busy at spawn, the sidecar logs a warning and falls back to an ephemeral port rather than failing.

Implementation:
- `config.py`: lenient-validated `Settings.sidecar_pin_port` (a bad value falls back to `0`, never raises), a new `sidecar` advanced section, bounds `(1024, 65535)` with `0` as an explicit off sentinel.
- `stdio_settings_sidecar.py`: `_pick_free_port(pinned)` binds the pinned port with `SO_REUSEADDR` and falls back to ephemeral on `OSError`; `run_main` reads the effective value via `get_global_settings()` so a value set in the UI (override file) applies too.
- `settings_ui.py` / `settings.js`: a stdio-only section whose number input can express the `0` sentinel (UI `min` = sentinel); the save path validates `0` or the bounded range; restart-required flag set.

A few choices worth a look (you raised these in the issue thread):
- **SO_REUSEADDR**: set on the pinned bind so an immediate restart after a crash isn't blocked by `TIME_WAIT`.
- **Non-stdio deployments**: the field is labelled stdio-only via the section title ("Settings UI sidecar (stdio-only)") and help text, rather than runtime-greying, to avoid threading a transport flag through the advanced-settings payload. Glad to switch to greyed-out + hint if you prefer that.
- Kept the name `_pick_free_port` (added an optional `pinned` arg) instead of renaming to `_pick_port`, to keep the diff small and non-breaking.

## Type of change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

Notes: `ruff` + `mypy` clean; the unit suite passes. New tests in `test_sidecar_pin_port.py` cover `_pick_free_port` (ephemeral default, pin success, busy fallback) and the validator (off sentinel, valid range, privileged/out-of-range/garbage all clamping to `0`). The 3 `test_helper_response_shape` failures seen locally also fail on clean `master` and are unrelated to this change. The e2e suite is Docker-gated and runs in CI.

## Checklist

The setting is self-documented via the Advanced-tab help text; there is no separate env-var reference doc to update.

### DIFF
--- a/src/ha_mcp/config.py
+++ b/src/ha_mcp/config.py
@@ -89,6 +89,13 @@ class Settings(BaseSettings):
     # WebSocket configuration (essential for async operations)
     enable_websocket: bool = Field(True, alias="ENABLE_WEBSOCKET")
 
+    # Settings UI sidecar (stdio mode only, #1587). 0 = pick a free
+    # ephemeral port at every spawn (default); 1024-65535 pins the sidecar
+    # to a fixed port so the settings URL/origin stays stable across
+    # restarts (bookmarks, browser localStorage). Read by run_main() in
+    # stdio_settings_sidecar.py.
+    sidecar_pin_port: int = Field(0, alias="HA_MCP_SIDECAR_PORT")
+
     # Development/Debug configuration
     debug: bool = Field(False, alias="DEBUG")
     log_level: str = Field("INFO", alias="LOG_LEVEL")
@@ -437,6 +444,37 @@ class Settings(BaseSettings):
             raise ValueError(f"Backup hint must be one of {valid_hints}")
         return v.lower()
 
+    @field_validator("sidecar_pin_port", mode="before")
+    @classmethod
+    def _lenient_sidecar_pin_port(cls, v: object) -> int:
+        """0 (default) = ephemeral port; otherwise a non-privileged port.
+
+        Lenient like the time-budget validators: an empty / unparseable /
+        out-of-range value falls back to 0 (ephemeral) with a warning rather
+        than raising, so a bad ``HA_MCP_SIDECAR_PORT`` can never crash the MCP
+        server or the best-effort settings sidecar.
+        """
+        if isinstance(v, str):
+            v = v.strip()
+            if not v:
+                return 0
+        # bool is an int subclass but never a meaningful port; reject it
+        # along with anything that isn't int/str-parseable.
+        if v is None or isinstance(v, bool) or not isinstance(v, int | str):
+            logger.warning("Invalid HA_MCP_SIDECAR_PORT=%r; using ephemeral port", v)
+            return 0
+        try:
+            port = int(v)
+        except (ValueError, TypeError):
+            logger.warning("Invalid HA_MCP_SIDECAR_PORT=%r; using ephemeral port", v)
+            return 0
+        if port != 0 and not 1024 <= port <= 65535:
+            logger.warning(
+                "HA_MCP_SIDECAR_PORT=%r outside 1024-65535; using ephemeral port", v
+            )
+            return 0
+        return port
+
     model_config = SettingsConfigDict(
         env_file=".env", env_file_encoding="utf-8", case_sensitive=False, extra="allow"
     )
@@ -505,6 +543,7 @@ AdvancedSection = Literal[
     "operations",
     "diagnostics",
     "tools_surface",
+    "sidecar",
     "beta_codemode",
 ]
 
@@ -728,6 +767,10 @@ ADVANCED_SETTINGS_FIELDS: tuple[AdvancedField, ...] = (
     AdvancedField("environment", "ENVIRONMENT", str, "diagnostics", True),
     AdvancedField("log_level", "LOG_LEVEL", str, "diagnostics", True),
     AdvancedField("debug", "DEBUG", bool, "diagnostics", True),
+    # Settings UI sidecar (stdio-only). Pin the sidecar's port so the
+    # settings URL/origin is stable across restarts; 0 = ephemeral
+    # (default). #1587.
+    AdvancedField("sidecar_pin_port", "HA_MCP_SIDECAR_PORT", int, "sidecar", True),
     # NOTE: ``auto_backup_dir`` and ``auto_backup_calendar_lookahead_days``
     # are NOT in this tuple. They are in ``BACKUP_OVERRIDE_FIELDS`` (defined
     # below) so they persist to ``backup_settings.json`` alongside the
@@ -778,6 +821,18 @@ _ADVANCED_SETTINGS_BOUNDS: dict[str, tuple[float, float]] = {
     "code_mode_max_memory": (1_048_576, 268_435_456),
     "code_mode_max_recursion": (1, 10_000),
     "code_mode_max_invocations": (1, 10_000),
+    # 0 is the "off" sentinel (ephemeral); the range below is the valid
+    # PINNED range. See _ADVANCED_SETTINGS_SENTINELS.
+    "sidecar_pin_port": (1024, 65535),
+}
+
+
+# Fields where a specific value is a valid "off" sentinel that bypasses the
+# _ADVANCED_SETTINGS_BOUNDS range (sidecar_pin_port: 0 = ephemeral). The UI
+# emits min=sentinel so the number input can still express "off"; the
+# override-apply and UI-POST paths accept the sentinel OR the bounded range.
+_ADVANCED_SETTINGS_SENTINELS: dict[str, int] = {
+    "sidecar_pin_port": 0,
 }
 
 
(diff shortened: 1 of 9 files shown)


SealKan:
Ha, your concrete-shape writeup did the heavy lifting here, I mostly transcribed it, so the token bought a spec clean enough that there was nothing left to flag. Call it outsourced compute.

Two judgment calls I left (both in the description): SO_REUSEADDR is in, and for non-stdio deployments I went with a "stdio-only" section label plus help text rather than runtime-greying, to avoid threading a transport flag through the advanced-settings payload. Trivial to switch to greyed-out + hint if you'd rather have that.

SealKan:
Also, between us, I burned a fair pile of tokens of my own getting this over the line, so I genuinely feel the pain. At least yours came back with a clean bill of health.

Patch76 (commented):
Heads-up while this is still a draft: I was implementing #1587 in parallel, and rather than open a competing PR I wanted to knock early – the approach here is the one worth taking forward.

The core is solid. The `_lenient_sidecar_pin_port` validator clamping a bad value to the ephemeral sentinel at the env-var path means a typo'd `HA_MCP_SIDECAR_PORT` can't crash the server or the sidecar, and the `_ADVANCED_SETTINGS_SENTINELS` mechanism threads the `0`-off sentinel cleanly through the UI `min`, the override-file apply, and the POST validation. That also keeps the configured value and the running value consistent, instead of letting an out-of-range value sit in the UI while the sidecar quietly runs on a different port. Blocking privileged ports (1-1023 collapse to the ephemeral sentinel) is the safer default for a localhost sidecar – worth keeping.

One thing on the non-stdio surface, which you raised yourself: the field is labelled stdio-only in the title and help text, but in addon / HTTP deployments the input is still editable, so a user there can save a value that does nothing. Since you offered the greyed-out variant – an `is_stdio` flag on the `/api/settings/advanced` payload (`_http_settings_prefix is None` is a clean proxy: it stays None in pure stdio and in the sidecar process, and is set for HTTP/SSE/OAuth/addon) plus dimming and disabling the section when it's false turns that inert save into a visibly-unavailable control. I built exactly that on my side and can share the frontend snippet if it's useful.

I'll hold my branch rather than open a second PR.

Patch76:
Here's the grey-out wiring from my parallel branch, if it's useful – three drop-in pieces on top of the section you already render.

**1. `settings_ui.py` – expose the transport on the advanced payload.** In the `/api/settings/advanced` GET handler, add `is_stdio` to the response:

```python
return JSONResponse(
    {
        "fields": fields,
        "is_addon": is_running_in_addon(),
        "is_stdio": _http_settings_prefix is None,
    }
)
```

`_http_settings_prefix` stays None in pure stdio and in the sidecar process, and is set by `register_settings_routes` for HTTP/SSE/OAuth/addon – so `is None` is a clean "a sidecar exists here" proxy.

**2. `settings.js` – dim + disable when not stdio.** Right after your `renderAdvancedSection('advSidecar', ...)` call:

```js
applySidecarAvailability(data.is_stdio !== false);
```

and the helper:

```js
function applySidecarAvailability(isStdio) {
  // Remove any note from a prior load first, so the <h3> title is once
  // again the section's previousElementSibling (an injected note would
  // otherwise shadow it on a re-render).
  const prevNote = document.getElementById('advSidecarNote');
  if (prevNote) prevNote.remove();
  const section = document.getElementById('advSidecar');
  if (!section) return;
  const title = section.previousElementSibling;
  if (isStdio) {
    section.classList.remove('dimmed');
    if (title) title.classList.remove('dimmed');
    return;
  }
  section.classList.add('dimmed');
  if (title) title.classList.add('dimmed');
  section.querySelectorAll('input, select').forEach(el => { el.disabled = true; });
  const note = document.createElement('div');
  note.id = 'advSidecarNote';
  note.className = 'adv-section-note';
  note.textContent =
    'Available in stdio mode only. This server runs over HTTP, which has '
    + 'no settings-UI sidecar to pin.';
  section.parentNode.insertBefore(note, section);
}
```

The `!== false` default leaves the section enabled if the flag is ever absent, which keeps it usable on the stdio path where the sidecar always serves.

**3. `settings.css` – the dimmed treatment**, mirroring the existing `.feature-row.*-sub.dimmed`:

```css
.adv-section.dimmed { opacity: 0.55; }
.adv-section.dimmed input, .adv-section.dimmed select { cursor: not-allowed; }
.adv-section-title.dimmed { opacity: 0.55; }
.adv-section-note { font-size: 0.72rem; color: var(--text-secondary);
  margin: 4px 0 8px; line-height: 1.4; }
```

With this in place the disabled input carries the context, so the title can go back to a plain "Settings UI sidecar" and the dimmed state plus the note do the work the "(stdio-only)" qualifier did alone. Whether the "No effect for HTTP/add-on" help-text line stays or trims is your call.

SealKan:
Done, pushed your grey-out wiring in efe5c7b. All three pieces are in: `is_stdio` on the advanced payload (via the `get_http_settings_prefix()` accessor, same `_http_settings_prefix is None` proxy), the `applySidecarAvailability` dim+disable+note in settings.js, and the `.adv-section.dimmed` / `.adv-section-note` CSS. I dropped the "(stdio-only)" title qualifier and trimmed the matching help-text line as you suggested, since the dimmed state plus the note now carry that context. Added a unit test for the `is_stdio` flag both ways. Thanks for handing over the branch work instead of racing a second PR.

Patch76 (dismissed):
The core here is solid and I traced it end to end. The `_lenient_sidecar_pin_port` validator clamps every bad input – empty, non-int, bool, privileged (1-1023), out-of-range – to the `0` ephemeral sentinel with a warning instead of raising, so a typo'd `HA_MCP_SIDECAR_PORT` can't take down the server or the sidecar. The pinned path binds with `SO_REUSEADDR` and falls back to an ephemeral port on `OSError`, and `run_main` reads the effective value through `get_global_settings()` so a UI-set override applies too. The `_ADVANCED_SETTINGS_SENTINELS` mechanism threads the `0`-off case through the UI `min`, the override-apply path, and the POST validation, keeping the configured and running values consistent.

On the pinned bind: it inherits the same advisory TOCTOU model as the existing ephemeral path – the probe socket closes before uvicorn binds, so a port lost in that gap fails the (advisory) sidecar rather than MCP itself. That is the documented pre-existing behaviour, no change needed.

Two optional test notes, neither blocking:
- The validator suite covers the env-var path well; the settings-UI POST path's sentinel bypass (`coerced != sentinel`) isn't directly exercised – a save test for `sidecar_pin_port: 0` plus one bounded value would lock it in.
- The validator parametrize omits a negative value; it hits the same `not 1024 <= port` branch as the tested `1023` / `70000`, so it is covered, but an explicit `-1 -> 0` row documents the intent.

Disclosure: the grey-out wiring in `efe5c7b` originated as a snippet I shared earlier in this PR thread – noting it here for transparency on the approval.

SealKan:
Thanks for the thorough trace and the approval. I added both optional notes in 8b4ee42: a save-path test that exercises the sentinel bypass (`0` accepted below the bounds floor, a bounded port applied, a privileged port rejected), and an explicit `-1 -> 0` row to document the negative-value intent. No worries on the disclosure either, the grey-out is your work and it's credited in the commit and the note above. The new commit is test-only on top of the approved diff, so re-confirm at your convenience if the push reset the approval.

Patch76 (approved):
Re-approving on 8b4ee42e after the stale-dismiss. The delta since the prior approval is test-only: a POST-path case (`test_save_advanced_sidecar_pin_port_sentinel_and_bounds`) pinning the 0 off-sentinel and a bounded port as accepted while a privileged port returns 400, plus a `-1 -> off` parametrize case on the validator. Both close the sentinel-bypass and negative/boundary coverage gaps noted earlier; the source under review is unchanged and all checks are green.

PR #1653, opened 2026-06-17
Update setup.astro: fix inconsistent hyphen

cbowns:
The manual install on macOS had:

```
Client Configuration
$ codex mcp add homeassistant …
```

But the top-line MCP commands for codex have a hyphen:

```
Management Commands
codex mcp list — list configured servers
codex mcp get home-assistant — show server config
codex mcp remove home-assistant — remove this server
```

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

## Checklist


site/src/pages/setup.astro
```diff
@@ -2127,19 +2127,19 @@ Authorization = "Bearer YOUR_TOKEN"</pre>
...
           } else {
-            config = `claude mcp add home-assistant \\
+            config = `claude mcp add homeassistant \\
   --env HOMEASSISTANT_URL={{HOMEASSISTANT_URL}} \\
   --env HOMEASSISTANT_TOKEN={{HOMEASSISTANT_TOKEN}} \\
   -- uvx ha-mcp@latest`;
           }
         } else {
-          config = `claude mcp add --transport http home-assistant ${mcpUrl}`;
+          config = `claude mcp add --transport http homeassistant ${mcpUrl}`;
```
cbowns: I'm not sure if this should be un-hyphenated; there's a few other instances later that I think don't directly reference the codex mcp integration name, but I didn't want to mass-rename everything.

kingpanther13 (changes requested):
Thanks for catching this, and welcome — it's a real inconsistency: the `add` command and the `get`/`remove` commands disagreeing means a copy-paste of the management commands operates on a server that was never created under that name.

One clarification for anyone reading: this string is just a local label — the name the client files the server under in its own config. You can name it anything; the server works regardless. What matters is only that it's *consistent* within one setup flow (the name you `add` is the name you `get`/`remove` and the key in the config file).

On direction, though — I'd suggest flipping this the other way. The rest of `setup.astro` standardizes on `home-assistant` (with the hyphen): every JSON config key, the YAML `- name:` entries, the opencode/gemini management commands, Raycast, VS Code, the Codex TOML example, the Codex Desktop "Name" field, and the Claude SSE alternative. The only outliers are a couple of stdio `add` lines (codex around 2061/2063, gemini around 2015/2021). So normalizing those few to `home-assistant` keeps the whole page consistent, whereas converting Codex/Claude to `homeassistant` moves them away from everything else.

As it stands the change is also partial, which leaves a few new mismatches:
- Claude `mcp remove home-assistant` (line 2117) isn't updated, so the `add` commands now say `homeassistant` but `remove` still says `home-assistant` — the same add/remove mismatch this PR is fixing.
- Claude SSE alternative add (2115) still says `home-assistant`.
- Codex Desktop UI "Name (`home-assistant`)" (2076) and the TOML example `[mcp_servers.home-assistant]` (2086/2088) still say `home-assistant`, now disagreeing with the CLI commands directly above them.

(No technical reason forcing it either way — `home-assistant` is a valid TOML bare key, hyphen included.)

kingpanther13 (approved):

Issue #1660, opened 2026-06-19
[BUG] Post-write config-check safety net is silently dead (check_config called with unsupported return_response)


Patch76:
### 🐛 Runtime Bug

**Component:** `ha_mcp_tools` custom component, v0.10.0 (shipped in #1649)

## Summary

The post-write / post-restore config validation in `_run_config_check` never runs. It calls the `homeassistant.check_config` service with `return_response=True`, but that service is registered with `SupportsResponse.NONE`, so Home Assistant raises a `ServiceValidationError` ("An action which does not return responses can't be called with return_response=True") before the check executes. The error is caught and logged at DEBUG only, and reported back as `config_check: "unavailable"`, so the "warn if the resulting config is broken" safety net silently does nothing.

It affects both write paths that call it:
- legacy / whole-file restore – `custom_components/ha_mcp_tools/__init__.py:1124`
- regular `replace_file` / YAML edit – `custom_components/ha_mcp_tools/__init__.py:1409`

## Root cause (not a version drift)

`homeassistant.check_config` does not return a response, and as far back as I checked never did:

- core 2026.6.3 and core 2024.1.0 both register it via `async_register_admin_service(hass, DOMAIN, SERVICE_CHECK_CONFIG, async_handle_core_service)` with no `supports_response=` argument, which defaults to `SupportsResponse.NONE`.
- the handler returns `None` and signals config errors by raising `HomeAssistantError`; it never returns an `errors` dict.

So the component is misusing the service API in two ways:
1. `return_response=True` (`__init__.py:983`) can never succeed against a `SupportsResponse.NONE` service – it always lands in the `except` (`__init__.py:985`) and yields `config_check: "unavailable"`.
2. even without that, the `check_result.get("errors")` branch (`__init__.py:990-995`) is dead – the service returns `None` and raises on errors rather than returning them.

The guard was never functional, on any HA version I checked – this is not a regression introduced by a specific HA release.

## Impact

After a YAML edit or a legacy restore, a resulting broken config is not surfaced by this guard. The write/restore itself completes correctly (file written, atomic temp-write + rename, YAML syntax validated). Only the secondary "is the whole config still valid" warning is skipped, and only at DEBUG level, so it is invisible in normal logs.

## Reproduce

On core 2026.6.3 with component 0.10.0, run any `ha_config_set_yaml` edit, or a legacy backup `restore`. The response carries:

```
config_check: "unavailable"
[...]

PR #1662, opened 2026-06-19
test: lock config-check guard against behavioral recurrence (#1660)

Patch76:
## What does this PR do?

Follow-up to #1661 (post-merge test-rigor review). The merged config-check fix had no test that fails on a *behavioral* regression of the guard:

- `test_config_check_included_in_response` (e2e) only asserted the `config_check` key was present, which accepts `"unavailable"` — a regression back to the silently-dead mode would stay green. It seeds a valid config, so it now asserts `config_check == "ok"`. This is the only test that exercises the real HA config-check path; the unit tests stub `async_check_ha_config_file`.
- Added a behavioral unit test for the swallowed-exception branch: stub `async_check_ha_config_file` with a `side_effect` and assert `config_check == "unavailable"` plus `config_check_error`. The existing source-string guard catches a verbatim revert but not a behavioral one.
- Dropped the misleading "tests that need a failing check override this locally" line from the theme/dashboard autouse-stub docstrings — no test there overrides it.

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [x] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

## Checklist

### DIFF
--- a/tests/src/e2e/workflows/filesystem/test_yaml_config.py
+++ b/tests/src/e2e/workflows/filesystem/test_yaml_config.py
@@ -663,9 +663,14 @@ class TestYamlConfigSafeguards:
             )
             inner = data
             assert inner.get("success") is True, f"Add should succeed: {data}"
-            # config_check should be present (ok, errors, or unavailable)
-            assert "config_check" in inner, (
-                f"Config check result should be in response: {data}"
+            # The seeded config is valid, so the guard must actually run and
+            # report "ok" — asserting mere presence accepts "unavailable" and
+            # would stay green if the check regressed to the silently-dead mode
+            # (#1660). This is the only test that exercises the real HA
+            # config-check path; the unit tests stub async_check_ha_config_file.
+            assert inner.get("config_check") == "ok", (
+                f"Config check should be 'ok' for a valid seeded config "
+                f"(behavioral regression guard for #1660): {data}"
             )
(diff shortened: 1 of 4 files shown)


Patch76:
## Implementation Summary

**Choices Made:**

- Kept the e2e assertion strict (`config_check == "ok"`) rather than accepting `"unavailable"`. A valid seed must produce `"ok"`, and that is the only signal distinguishing the live guard from the silently-dead mode the original bug regressed into — accepting `"unavailable"` is exactly the gap that let the bug ship. Confirmed on a live HA (core 2026.6.3) that a valid edit returns `config_check: "ok"`.
- Put the behavioral `"unavailable"` test in the unit suite (stub `async_check_ha_config_file` with a `side_effect`) since that deterministically drives the swallowed-exception branch without needing a broken real config; the e2e covers the `"ok"` path against real HA.
- Left the existing source-string guard in place — it catches a verbatim revert; the two new behavioral assertions cover the recurrence it cannot.

**Problems Encountered:**

- A local full `uv run pytest` cannot be installed on this platform (the `lefthook` dev dependency has no linux-x86_64 wheel), so I ran the touched unit files through the project venv — 110 pass. The new e2e `== "ok"` assertion runs on the HAOS lanes in CI (testcontainers/Docker is not available locally).

Issue #1458, opened 2026-05-27
[DOCS] Need to improve documentation advertising the web settings server for non-addon installations


kingpanther13:
### Where is the issue?

Other

### What's wrong or unclear?

Currently no documentation is really needed to advertise the web UI in addon mode, as there's a prominent "open web UI" button within the HA settings page. However, the non-addon versions must know that it exists and must find the URL for the settings server to know how to connect to it. I've experienced a few users not even knowing that it exists. With the latest PRs coming adding significant functionality to the server, I feel that it is important to improve the documentation.

Currently I have the get overview tool set to notify users of the URL when it's running in STDIO/sidecar mode but that's only when the tool is called. I'm unsure if this helps at all in the container versions, so that's an addition that's needed if it doesn't already do that. However if we can add further documentation/notifications in startup logs and if we include it in our astro setup site, it should help improve knowledge of its existence. Just s simple notification when ha-mcp is started in docker would be sufficient, and a simple blurb somewhere in the astro site would be plenty.

Also with the sidecar mode maybe we can change it to leave the cmd window up with s little message advertising the settings web server and how to navigate to it, and also how to opt out of the settings web server and/or the cmd window. Just some thoughts for the future.

Issue #863, opened 2026-04-04
[FEATURE] Use MCP Apps to expose server settings via in-client UI


kingpanther13:
### What would you like to see?

FastMCP 3.2 introduced MCP Apps — tools that can return interactive UIs rendered directly inside the client. ha-mcp could use this to let users view and change server settings (tool visibility, pinned tools, tool search result count, and potentially other settings) through an interactive UI within their MCP client, rather than only through the add-on configuration.

Beyond settings, this could also be used with tools to let users control things directly in their Home Assistant interface — dashboards, entity controls, etc. There are a lot of possibilities here.

### Why do you need this?

Currently, changing settings like which tools are disabled or pinned requires editing the add-on configuration and restarting. An in-client UI (e.g., toggle switches for tool groups, sliders for result counts) would make this faster and more discoverable. This would complement the add-on settings — add-on config sets defaults, and the in-client UI allows adjustments during use.

### Additional context

Related: #857 (user-configurable tool disable/pin in add-on settings) introduces the tool visibility configuration that could be a first candidate for an in-client UI (closed in favor of #960 which uses a web UI). 

See latest comment; reopened this because Claude Desktop users connecting with quick connect from the setup guide using STDIO don't have an easy way to access settings other than modifying their settings json; therefore, this is still relevant for people with similar setups. They also don't have an easy way to access the web UI set up with #960; so I think that this mostly should just piggyback off of #960 if possible and provide the same settings (and #960 will start including more configuration options, like what's currently available in the addon version). We also should update the astro site to tell users how to access the settings menus when using Claude Desktop, by A) accessing the #960 web UI directly via the URL or B) by using this method within the LLM itself, if it is supported. Also should allow the LLM to control the settings as well if asked by the user, though this may need to have some kind of safety toggle or something....but when working properly the user can simply say "hey, enable/disable toolsearch" and it can do it. Definitely needs a manual toggle though because we don't want a confused LLM just randomly doing it.

Patch76:
Picking this up as the assignee. Before I cut a worktree, I want to surface what I found in the FastMCP 3.2.4 surface and the open design questions, because the bot sketch above is misleading on the API and the implementation would touch enough of the existing settings story that I'd rather get a steer first.

## What FastMCP 3.2.4 actually exposes

The bot's `from fastmcp.apps import PrefabApp, Heading, Form, TextField, NumberField, Switch` does not match the installed package. Verified locally against `fastmcp==3.2.4`:

[...]

PR #1665, opened 2026-06-20
refactor(c901): tools_search.py below C901 threshold

sergeykad:
## What does this PR do?

Migrates `tools_search.py` from the closure-based `register_search_tools` pattern to a `SearchTools` class, decomposes over-threshold methods into focused helpers, and removes `tools_search.py` from the C901 per-file ignore in `pyproject.toml`.

Part of #925 (3 files remain: `tools_config_dashboards.py`, `tools_entities.py`, `tools_registry.py`).

**Violations fixed:**

| Function | Was | Now |
|---|---|---|
| `register_search_tools` | 142 | ~1 |
| `ha_search_entities` | 80 | <10 |
| `ha_get_state` | 24 | <10 |
| `ha_get_overview` | 19 | <10 |
| `ha_search` | 13 | <10 |
| `_exact_match_search` | 13 | <10 |
| `_merge_payload_metadata` | 12 | <10 |

**Helpers extracted:**

- `_ha_search_entities` (80): `_search_area_with_query`, `_search_area_only`, `_search_area_only_populated`, `_search_domain_only`, `_search_regular`
- `ha_get_state` (24): `_get_single_entity_state`, `_get_bulk_entity_states`, `_accumulate_state_results`
- `ha_get_overview` (19): `_fetch_system_info`, `_fetch_notifications`, `_fetch_repairs`
- `ha_search` (13): `_apply_search_outcome`
- `_exact_match_search` (13): `_build_hidden_ids`
- `_merge_payload_metadata` (12): `_merge_list_key`, `_merge_partial_reason`

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [x] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

Unit tests: 5206 passed, 108 skipped. E2e tests (search suite): 90 passed.

## Checklist


kingpanther13 (changes requested):
Thanks for this — the decomposition itself is faithful: all three tools register
identically, the LLM-facing tool docstrings are byte-for-byte unchanged, and the
error-handling semantics hold. Requesting changes only on a few comments — two are now
factually wrong, and two dropped non-obvious context that the surviving code still needs.

**1. Factually incorrect — `tools_search.py:1474-1476` (`_search_area_with_query`)**
The comment says `get_entities_by_area is called with group_by_domain=True above`, but
after extraction that call lives in the caller (`_ha_search_entities`); there's no such
call in this method — the result arrives via the `area_result` parameter. Please reword
to reference the caller (the invariant it documents is still true and worth keeping).

**2. Factually incorrect — `:1620` (`_search_area_with_query`)**
"Mirror the empty-area branch's message" — there's no branch here anymore; the empty-area
case is the sibling method `_search_area_only`, which actually emits a *different* string
(`No entities found in area: {area_filter}`, no `domain_filter` qualifier). Please name
the sibling or drop "Mirror" so it isn't misleading.

**3. Non-obvious rationale lost — `_validate_entity_search_params` (:598-602)**
The strip-then-lowercase + at-least-one-set code is preserved verbatim, but its rationale
was replaced by a generic one-line docstring. The "why" here isn't self-evident from the
code — worth restoring:

    # HA domains are canonically lowercase, no whitespace; agents
    # that capitalize ("Lights") or pad ("  light  ") would
    # otherwise hit a silent zero-result against the prefix match
    # downstream. Strip-then-lowercase before validation so a
    # whitespace-only filter ("   ") collapses to "" and fails the
    # at-least-one-set check rather than passing it and falling
    # through to a no-op fuzzy search.

**4. Issue references dropped — `ha_get_overview` projection (:2094, :2103)**
`(issue #863)` (why the sidecar `settings_url` is surfaced after projection) and
`(issue #1458)` (why it hints at the page instead of guessing an absolute URL when bound
to 0.0.0.0) were both removed. This repo keeps issue refs throughout (backup_manager,
rest_client, etc.), and these point at the only record of why the workarounds exist —
please keep them alongside the surviving code.

None of these affect runtime behavior, but #1/#2 are simply wrong now and #3/#4 drop
context the code still depends on.

kingpanther13 (approved):
All four requested comment fixes landed in 46775e6a and check out against the code:

1. _search_area_with_query — reworded to point at the caller. Confirmed: get_entities_by_area(group_by_domain=True) lives in _ha_search_entities (the call site), and area_result is passed into the helper, so the invariant the comment documents holds.
2. _search_area_only_populated — "Mirror" dropped; now names _search_area_only and frames it as a message pattern. Correct call, since the sibling's string omits the domain_filter qualifier — same shape, not the same string.
3. _validate_entity_search_params — strip-then-lowercase rationale restored.
4. ha_get_overview — issue #863 and #1458 refs restored on the surviving comments.

The decomposition was already faithful; these were the only gaps. Thanks for the fixes.

PR #1668, opened 2026-06-21
fix: accept proxy restart errors

### LINKED ISSUE #1666: [FEATURE] ha_restart: extend known-good error patterns to cover 502/503 from reverse proxies (follow-up to #621)
## Summary

`ha_restart(confirm=True)` can return a **null JSON-RPC response** (no `result`, no `error`) to MCP clients when the user's reverse proxy in front of Home Assistant emits **502 Bad Gateway** instead of 504 during the shutdown window. This is functionally equivalent to the bug fixed in #621 (which added `504` to the known-good pattern list, fixing #612), but a different proxy returns 502 in the same scenario.

## Current behavior

In `src/ha_mcp/tools/tools_system.py` around line 200, the exception-handling path treats only these substrings as the expected "HA went down during restart" signal:

```python
if restart_initiated and any(
    pattern in error_msg.lower() for pattern in ("connect", "closed", "504")
):
    return {"success": True, ...}
```

When the upstream proxy returns **502** (Bad Gateway), **503** (Service Unavailable), or the connection is torn down without a recognized substring in the exception, the handler falls through to `exception_to_structured_error`, which raises. From the MCP client's perspective, the JSON-RPC response then contains neither `result` nor `error` (or contains an error that's actually a transient restart, not a real failure) — there is no actionable signal to distinguish "restart initiated, connection dropped" from "restart genuinely failed".

## Reproduction

Setup: HA behind Nabu Casa Remote UI (or Cloudflare / Nginx Proxy Manager) where the proxy emits `502` instead of `504` when the upstream backend disappears.

1. Call `ha_restart(confirm=True)` via the MCP webhook.
2. HA restarts successfully (confirmed via `ha_get_overview` ~60s later).
3. The MCP response has no parseable `result.content` — the streamable HTTP body terminates before a JSON-RPC `result` is emitted. From the client this looks identical to a hung/dropped tool call.

Observed during this session against a real HA OS deployment behind Nabu Casa Remote UI on 2026-06-21.

## Proposed fix (low-risk, follows #621)

Extend the pattern list to cover the two other proxy errors emitted in the same scenario:

```python
if restart_initiated and any(
    pattern in error_msg.lower() for pattern in (
        "connect", "closed", "504", "502", "503", "gateway", "unavailable"
    )
):
    return {"success": True, ...}
```

Rationale: every one of these substrings can only occur AFTER `restart_initiated = True`, i.e. after `homeassistant/restart` has already been dispatched. The gating on `restart_initiated` already prevents false positives from pre-restart failures.

## Optional follow-up (more invasive, do not block this FR on it)

The deeper root cause is that the `ha_restart` handler currently calls `homeassistant/restart` and *then* tries to return success — but HA can tear the connection down inside `call_service` before the ack is flushed. A cleaner architecture would be:

1. Schedule the restart with a small delay (e.g. `call_service("homeassistant", "restart", {"delay": 1})` if the service grows that param, or wrap in an automation/job).
2. Return the success ack first.
3. Let HA shut down after the response has been written to the wire.

This would eliminate the null-response problem for **all** proxies, not just specific status codes. It's a bigger change so I'd suggest landing the pattern-list expansion first.

## Why this matters

Without this fix, agents that drive `ha_restart` cannot reliably tell whether the restart actually started, even though it did. The current contract effectively requires every caller to externally poll `ha_get_overview` and time-out after 5 minutes, which negates the value of the tool's structured return.

## Environment

- HA: 2026.x (OS), behind Nabu Casa Remote UI
- ha-mcp: v7.8.0 (latest at time of report)
- Client: external MCP webhook (not the in-cluster sidecar)

## Related

- #612 — original report (504 case)
- #621 — fix that added `504` to the pattern list
- This FR — extend the same fix to cover 502/503/gateway/unavailable

gaoflow:
## What does this PR do?

Closes #1666.

Extends `ha_restart`'s post-dispatch restart shutdown handling so reverse proxy 502/503 responses and textual gateway/unavailable errors are treated like the existing connection/closed/504 cases after restart has already been initiated.

The existing unrelated-error test still verifies that errors outside this expected restart window raise normally.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

Commands run:
- `uv run pytest tests/src/unit/test_tools_system.py -q`
- `uv run pytest tests/src/unit/test_helper_response_shape.py -q`
- `uv run pytest tests/src/unit -q` (`5203 passed, 115 skipped`)
- `uv run ruff check src/ tests/`
- `uv run ruff format --check src/ tests/`
- `uv run mypy src/`
- `uv run ast-grep scan`

Attempted relevant e2e coverage:
- `cd tests && uv run pytest src/e2e/workflows/system/test_system_tools.py -v --tb=short`

That e2e command did not run locally because Docker is not available on this machine: `Docker is not available: Error while fetching server API version` / `FileNotFoundError(2, 'No such file or directory')`.

## Checklist

### DIFF
--- a/src/ha_mcp/tools/tools_system.py
+++ b/src/ha_mcp/tools/tools_system.py
@@ -197,7 +197,16 @@ class SystemTools:
             # Connection errors after restart initiated are expected
             # (HA closes connections during restart)
             if restart_initiated and any(
-                pattern in error_msg.lower() for pattern in ("connect", "closed", "504")
+                pattern in error_msg.lower()
+                for pattern in (
+                    "connect",
+                    "closed",
+                    "502",
+                    "503",
+                    "504",
+                    "gateway",
+                    "unavailable",
+                )
             ):
                 return {
                     "success": True,
--- a/tests/src/unit/test_tools_system.py
+++ b/tests/src/unit/test_tools_system.py
@@ -1,7 +1,6 @@
 """Unit tests for tools_system module.
 
-Regression tests for https://github.com/homeassistant-ai/ha-mcp/issues/612
-ha_restart reports failure when a reverse proxy returns 504 during restart.
+Regression tests for ha_restart handling reverse proxy errors during restart.
 """
 
 import asyncio
@@ -91,18 +90,34 @@ class TestGetSystemHealthHaMcpUpdate:
 class TestHaRestartErrorHandling:
     """Tests for ha_restart handling of expected errors during restart."""
 
+    @pytest.mark.parametrize(
+        "error",
+        [
+            HomeAssistantAPIError("API error: 504 - ", status_code=504),
+            HomeAssistantAPIError("API error: 502 - Bad Gateway", status_code=502),
+            HomeAssistantAPIError(
+                "API error: 503 - Service Unavailable", status_code=503
+            ),
+            Exception("Bad Gateway"),
+            Exception("Service unavailable"),
+        ],
+    )
     @pytest.mark.asyncio
-    async def test_504_gateway_timeout_treated_as_success(self):
-        """A 504 from a reverse proxy after restart initiated should be success.
+    async def test_proxy_errors_after_restart_treated_as_success(
+        self, error: Exception
+    ):
+        """Reverse proxy errors after restart initiated should be success.
 
         Reproduces issue #612: user behind a reverse proxy gets 504 when HA
         shuts down, but HA actually restarted successfully.
+
+        Also covers issue #1666: some proxies return 502/503 or textual
+        gateway/unavailable errors for the same post-restart shutdown window.
 
         Also pins the post-#1332 warnings-list contract on the
         restart-initiated-but-connection-dropped branch
         (``tools_system.py`` ~L208-214).
         """
-        error = HomeAssistantAPIError("API error: 504 - ", status_code=504)
         client = _make_client_that_fails_on_restart(error)
         tools = SystemTools(client)
 


kingpanther13 (changes requested):
Thanks for this — the fix is correct and complete against #1666. The new
502/503/gateway/unavailable patterns all match what rest_client puts in
str(e) (both the numeric and the textual forms land there), and the
parametrized test exercises both the numeric path and the textual-only path
(Exception("Bad Gateway")).

Two changes before this is ready:

1. Add a regression test for the pre-dispatch path (restart_initiated is False).

The gate at tools_system.py:199 is `restart_initiated and any(pattern ...)`.
Widening the pattern set to include 503/unavailable — the canonical responses a
proxy returns when its upstream is *already down* — makes the restart_initiated
guard the only thing stopping a proxy error raised *before* restart is
dispatched (e.g. a 503 out of check_config() at line 161) from being reported as
a successful restart.

The existing test_unrelated_error_still_fails covers one half of the AND
(restart initiated, but no pattern matches -> raises). The other half is
untested: restart NOT initiated, but a pattern *does* match -> must still raise.
_make_client_that_fails_on_restart (line 40) fails on call_service, after the
flag flips, so it can't express this. Please add a sibling helper that makes
check_config raise (restart_initiated stays False) and assert
ha_restart(confirm=True) raises ToolError for e.g. Exception("Bad Gateway").

2. Tighten the success message for the proxy-status cases.

tools_system.py:215 returns "Connection was closed as expected during restart."
That now fires for 502/503, where the proxy returned an error rather than HA
closing the connection. Reword to cover both, e.g. "Home Assistant became
unreachable after the restart was requested (expected during shutdown); if it's
still unreachable after a few minutes the restart may not have completed."

Unrelated to your change: the red HAOS E2E (inaddon) job is a HACS repo-install
timeout (test_hacs.py::TestMcpToolsInstallation), not this PR — the system/
restart e2e tests passed. No need to chase it.

kingpanther13:
Superseded by #1670.

Issue #612, opened 2026-02-12
ha_restart reported a failure/error but did in fact restart the HA instance


Scott8586:
I have tools_system enabled in my ha-mcp environment.   I asked the model to see if it can restart home assistant
```
 "don’t do it, but can you restart the core?"
```

It replied with:
```
I understand your request. However, based on the error message from <source id="1" name="Cabin_ha_restart">, there was a failure to restart Home Assistant. If you encounter issues like this, you might want to check the logs for more detailed information or ensure that all dependencies are properly configured.


```
(of course, I'm not happy it restarted after I said "don't do it", but that's another issue)
The JSON returned was:
```
Cabin_ha_restart

Parameters
{  "confirm": true }
Content
{ "success": false, "error": "Failed to restart Home Assistant: API error: 504 - " }
```

But in fact the instance did restart.  Tried twice and got the same result.

Home assistant: 2026.1.3

PR #1670, opened 2026-06-21
fix(ha_restart): extend known-good error patterns to cover 502/503 from proxies

### LINKED ISSUE #1666: [FEATURE] ha_restart: extend known-good error patterns to cover 502/503 from reverse proxies (follow-up to #621)

Shaan-alpha:
## What does this PR do?

Fixes #1666. When Home Assistant is behind a reverse proxy (nginx, Traefik, Cloudflare, etc.), the proxy often returns a **502/503** — or a `Bad Gateway` / `Service Unavailable` body — while HA is shutting down for a restart, even though the restart itself succeeded. `ha_restart` treated those as failures and reported a false error.

This extends the known-good error patterns checked after a restart is initiated from `("connect", "closed", "504")` to also cover `502`, `503`, `gateway`, and `unavailable`, so a proxy returning any of those during shutdown is recognized as the expected connection drop rather than a failure (consistent with the existing #612 handling for 504).

Tests added in `tests/src/unit/test_tools_system.py` (`TestHaRestartErrorHandling`):
- A parametrized test pinning each newly added pattern (`502`, `503`, `gateway`, `unavailable`), mirroring the existing #612 (504) regression test.
- A pre-dispatch guard test: a matching proxy error raised from `check_config` (before the restart is dispatched, so `restart_initiated` stays `False`) must still raise `ToolError` — pinning the `restart_initiated` half of the `restart_initiated and any(...)` guard.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

`uv run pytest src/unit/test_tools_system.py` → 97 passed. `uv run ruff check` and `uv run ruff format --check` are clean. I verified the new tests fail on the pre-fix pattern set / a weakened guard and pass with the fix. (I did not exercise it end-to-end against a live HA + proxy via an LLM agent — the regression is reproduced deterministically at the unit level.)

## Checklist
- [x] I have updated documentation if needed (no user-facing docs affected by this internal error-handling change)

### DIFF
--- a/src/ha_mcp/tools/tools_system.py
+++ b/src/ha_mcp/tools/tools_system.py
@@ -197,7 +197,16 @@ class SystemTools:
             # Connection errors after restart initiated are expected
             # (HA closes connections during restart)
             if restart_initiated and any(
-                pattern in error_msg.lower() for pattern in ("connect", "closed", "504")
+                pattern in error_msg.lower()
+                for pattern in (
+                    "connect",
+                    "closed",
+                    "504",
+                    "502",
+                    "503",
+                    "gateway",
+                    "unavailable",
+                )
             ):
                 return {
                     "success": True,
--- a/tests/src/unit/test_tools_system.py
+++ b/tests/src/unit/test_tools_system.py
@@ -46,6 +46,17 @@ def _make_client_that_fails_on_restart(exception):
     return mock_client
 
 
+def _make_client_that_fails_on_check_config(exception):
+    """Create a mock client where check_config itself raises.
+
+    Used to exercise the pre-dispatch path: the failure happens before the
+    restart service is ever called, so ``restart_initiated`` stays False.
+    """
+    mock_client = AsyncMock()
+    mock_client.check_config.side_effect = exception
+    return mock_client
+
+
 class TestGetSystemHealthHaMcpUpdate:
     """ha_get_system_health surfaces the MCP server's own update status."""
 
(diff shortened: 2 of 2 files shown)


Shaan-alpha:
Added the requested regression coverage in `tests/src/unit/test_tools_system.py` — a parametrized `test_proxy_5xx_during_restart_treated_as_success` that isolates each newly added pattern (`502`, `503`, `gateway`, `unavailable`), mirroring the existing #612 (504) regression test in the same `TestHaRestartErrorHandling` class. I verified it fails on the pre-fix pattern set and passes with the fix.

I put it under `tests/src/unit/` rather than `tests/src/e2e/` because these patterns are pure error-message classification: a real container restart only produces `connect`/`closed` drops, so the proxy 502/503/`gateway`/`unavailable` cases can't be reproduced via testcontainers — and the established #612/504 regression for this exact branch already lives in this unit file. Happy to add an E2E variant if you'd prefer.

kingpanther13 (changes requested):
Nice work — the production change is complete (correct pattern set including
gateway/unavailable) and the parametrized test is well done, especially using
HomeAssistantConnectionError for the textual cases (the real error type that
surfaces without a status code). A few things before merge:

1. Fix the formatting — Ruff Lint is failing.

The "Ruff Lint" job reports "Would reformat: src/ha_mcp/tools/tools_system.py".
The hand-wrapped pattern tuple isn't in ruff's canonical form. Run
`uv run ruff format src/` and commit the result.

2. Add a regression test for the pre-dispatch path (restart_initiated is False).

The gate is `restart_initiated and any(pattern ...)`. Your new test and
test_unrelated_error_still_fails both run with restart_initiated already True
(the helper fails on call_service). That pins one half of the AND; the other
half is untested: a matching error raised *before* restart is dispatched — e.g.
a "Bad Gateway" out of check_config(), so restart_initiated stays False — must
still raise. Without that, a future change weakening the guard could let a
pre-restart proxy error be reported as a successful restart while the suite
stays green.

_make_client_that_fails_on_restart fails on call_service (after the flag flips),
so it can't express this. Please add a sibling helper that makes check_config
raise and assert ha_restart(confirm=True) raises ToolError for e.g.
HomeAssistantConnectionError("Bad Gateway").

3. Fill in the PR description using the repo template.

Please use .github/pull_request_template.md (What does this PR do / Type of
change / Testing / Checklist) so reviewers can see the change type and what you
verified.

4. Heads up — this is one of three open PRs for #1666 (also #1667 and #1668).
Whichever one addresses its review feedback and is fully merge-ready first is the
one that gets merged.

Shaan-alpha:
Thanks @kingpanther13 — all addressed:

1. **Ruff formatting** — ran `uv run ruff format src/`; the hand-wrapped pattern tuple is now in ruff's canonical one-per-line form. `ruff check` + `ruff format --check` are clean. Folded into the fix commit so the source lands ruff-clean.
2. **Pre-dispatch path** — added `test_proxy_error_before_restart_dispatch_still_fails` plus a `_make_client_that_fails_on_check_config` helper: `check_config` raises `HomeAssistantConnectionError("Bad Gateway")`, so `restart_initiated` stays `False`, and `ha_restart` must raise `ToolError` rather than report success. I verified it fails if the `restart_initiated and` half of the guard is removed, so it genuinely pins that branch.
3. **PR description** — filled in using `.github/pull_request_template.md` (what / type / testing / checklist).

Full file run: `uv run pytest src/unit/test_tools_system.py` → 97 passed.

Understood on the three-PR race for #1666 — this one should now be fully merge-ready. Thanks for the thorough review!

kingpanther13 (approved):
All three items addressed:

- Ruff format clean (Ruff Lint green).
- Pre-dispatch guard test added (test_proxy_error_before_restart_dispatch_still_fails).
  Confirmed it genuinely pins the restart_initiated half of the gate — it fails
  if that conjunct is removed, so a pre-restart proxy error can't masquerade as a
  successful restart.
- PR description now follows the template.

Production change is complete and correct, and full CI is green. Thanks for the
thorough follow-up.

Issue #1527, opened 2026-06-03
[FEATURE] Investigate the full in-process custom component (revives #559)


kingpanther13:
## What would you like to see?

Make HACS the primary, low-friction way to get ha-mcp, and use this issue to investigate **how far into Home Assistant we should move** — weighing my lightweight "bootstrap" idea against the full custom-component rewrite @julienld originally proposed in #559.

It's best understood as three phases on one spectrum ("how much runs inside HA, delivered via HACS"):

| Phase | What | Works on | Effort |
|---|---|---|---|
| **0 — Distribution** | Get `ha-mcp` into **HACS default** so the `ha_mcp_tools` integration is search-and-install (no custom HACS repo URL) | All install types | Small (HACS inclusion) |
| **1 — Bootstrap the add-on** *(near-term ask)* | The `ha_mcp_tools` config flow offers to **install + start the MCP Server add-on automatically** via the Supervisor `AddonManager` pattern | **Supervisor only** (HAOS / Supervised) | Medium |
| **2 — In-process server** *(the #559 investigation)* | The integration optionally **hosts the MCP server inside HA** — the only way HACS can reach Container/Core installs, since they can't run add-ons | All install types | Large |

**Phase 1** means a HAOS user installs one thing from HACS, clicks "yes, install the add-on," and is done — no add-on store, no repository URLs. **Phase 2** is #559's vision, reached incrementally instead of as a forced rewrite.

PHase 0&1 implemented by #1528 .(need to verify if it was fully implemented or just partially if we ever do real work on this)

## Why do you need this?

Two reasons:

1. **Install friction.** A full-featured setup today is *two* separate "paste a custom repository URL" dances — one in the Add-on Store for the server, one in HACS for the `ha_mcp_tools` component. We currently have to document both. HACS-first collapses that.
2. **Strategic direction.** #559 (full custom-component rewrite) stalled because it's multi-week and forces migration. But it promised real benefits — internal API access, simpler deployment, riding HA's reverse proxy. This issue asks: **what do we actually gain and lose** at each phase, so we can capture most of the benefit cheaply (Phase 1) and make an informed call on whether Phase 2 is worth it.

### Background: the two ideas

[...]

PR #1671, opened 2026-06-21
ci: add HACS + Hassfest validation workflow

Patch76:
## What does this PR do?

Adds a `Validate` workflow that runs the **HACS Action** (`category: integration`) and **Hassfest** against the repo's integrations. These are the two validations required for HACS default-store inclusion (Phase 0 of #1527), and the repo ran neither.

It runs on push to `master`, pull requests targeting `master`, a weekly schedule, and manual dispatch, with least-privilege `contents: read` permissions.

Enabling repo-wide Hassfest surfaced a pre-existing manifest issue in the `mcp_proxy` integration (the webhook-proxy add-on's component; `ha_mcp_tools` itself is clean), fixed here so the new check passes:
- `mcp_proxy` registers HTTP views on its opt-in OAuth path but did not declare `http`. Declared it in `after_dependencies` (it is optional – no views are registered with OAuth off).
- Sorted `mcp_proxy`'s manifest keys to satisfy the Hassfest `[MANIFEST]` ordering check.

## Type of change
- [x] 🐛 Bug fix
- [x] 🔧 Maintenance/refactor

## Testing

CI-config plus a manifest fix; no Python logic touched, so the pytest/ruff items above are not the relevant gate. The new HACS and Hassfest jobs both pass on this PR – that run is the verification.

## Checklist

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -28,4 +28,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
+      # Drop vendored third-party fixtures (e.g. the upstream HACS integration
+      # used as an e2e fixture) so hassfest only validates integrations this repo
+      # owns. Mirrors scripts/codeql_quality_gate.py's PATHS_IGNORE convention.
+      - name: Remove vendored test fixtures
+        run: rm -rf tests/initial_test_state
       - uses: home-assistant/actions/hassfest@master
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -1,0 +1,31 @@
+name: Validate
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+  schedule:
+    - cron: "0 6 * * 1"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  hacs:
+    name: HACS
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - name: HACS validation
+        uses: hacs/action@main
+        with:
+          category: integration
+
+  hassfest:
+    name: Hassfest
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: home-assistant/actions/hassfest@master
--- a/homeassistant-addon-webhook-proxy/mcp_proxy/__init__.py
+++ b/homeassistant-addon-webhook-proxy/mcp_proxy/__init__.py
@@ -24,6 +24,7 @@ from pathlib import Path
 from urllib.parse import urlparse
 
 import aiohttp
+import voluptuous as vol
 from aiohttp import web
 from homeassistant.components.webhook import (
     async_register,
@@ -49,6 +50,16 @@ CONFIG_FILE = Path("/config/.mcp_proxy_config.json")
 # as a sanity floor — a truncated/corrupted ha-mcp config yields a shorter
 # token, which is the failure mode this length check exists to catch.
 _SECRET_PATH_RE = re.compile(r"^/private_[A-Za-z0-9_-]{16,}$")
+
+# Permissive whole-config schema: satisfies hassfest's [CONFIG_SCHEMA] check
+# (any integration with async_setup must declare one) while preserving the
+# YAML-migration path below. cv.config_entry_only_config_schema is wrong here:
+# it raises an ERROR-severity repair issue on any `mcp_proxy:` key, which would
+# collide with the legacy `mcp_proxy:` line async_setup imports from.
+CONFIG_SCHEMA = vol.Schema(
+    {vol.Optional(DOMAIN): vol.Any(None, dict)},
+    extra=vol.ALLOW_EXTRA,
+)
 
 
 def _validate_target_url(target_url: str) -> tuple[bool, str]:
--- a/homeassistant-addon-webhook-proxy/mcp_proxy/manifest.json
+++ b/homeassistant-addon-webhook-proxy/mcp_proxy/manifest.json
@@ -2,10 +2,10 @@
   "domain": "mcp_proxy",
   "name": "MCP Webhook Proxy",
   "after_dependencies": ["http"],
-  "documentation": "https://github.com/homeassistant-ai/ha-mcp",
-  "version": "1.2.0",
   "codeowners": [],
   "config_flow": true,
   "dependencies": ["webhook"],
-  "iot_class": "local_push"
+  "documentation": "https://github.com/homeassistant-ai/ha-mcp",
+  "iot_class": "local_push",
+  "version": "1.2.0"
 }
--- a/homeassistant-addon-webhook-proxy/mcp_proxy/manifest.json
+++ b/homeassistant-addon-webhook-proxy/mcp_proxy/manifest.json
@@ -1,6 +1,7 @@
 {
   "domain": "mcp_proxy",
   "name": "MCP Webhook Proxy",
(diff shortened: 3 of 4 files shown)


Patch76:
**Choices Made:**
- Added `.github/workflows/validate.yml` with a HACS job (`category: integration`) and a Hassfest job – the two validations HACS default-store inclusion requires, neither of which the repo ran before. Least-privilege `contents: read`; triggers on push to `master`, pull requests targeting `master`, a weekly schedule, and manual dispatch.
- Declared `http` in `mcp_proxy`'s `after_dependencies` rather than `dependencies`: the integration imports `homeassistant.components.http.HomeAssistantView` and calls `hass.http.register_view`, but only on the opt-in OAuth path (`__init__.py` guards it behind `isinstance(oauth_section, dict)`). With OAuth off it registers no views and works without http, so `after_dependencies` is the semantically correct field.
- Sorted `mcp_proxy`'s manifest keys (Hassfest `[MANIFEST]` requires `domain, name, then alphabetical`).
- Left `mcp_proxy`'s `[CONFIG_SCHEMA]` Hassfest warning unaddressed. It is non-blocking, and `async_setup` accepts a YAML `mcp_proxy:` block for legacy migration to a config entry, so `cv.config_entry_only_config_schema` would break that migration path; the correct schema choice is a separate decision and out of scope for this PR.

**Problems Encountered:**
- Hassfest initially failed on `mcp_proxy`, not `ha_mcp_tools` (which is clean) – the repo-wide Hassfest scan covers all three workspace integrations. The `http` dependency error surfaced first; fixing it revealed a manifest-key-sort error, since Hassfest reports one `[MANIFEST]` error at a time, requiring a second commit.

kingpanther13 (changes requested):
## Review: Request changes

This PR adds a repo-wide Hassfest CI gate and cleans up `mcp_proxy`'s manifest to pass it. Two issues should be resolved before merge.

---

### 1. Leftover `[CONFIG_SCHEMA]` hassfest warning on `mcp_proxy` — fix it in this PR

`homeassistant-addon-webhook-proxy/mcp_proxy/__init__.py:73` (`async_setup`)

Hassfest still emits exactly one warning, and it's on the same integration this PR is cleaning up:

> `[CONFIG_SCHEMA]` Integrations which implement `async_setup`/`setup` must define `CONFIG_SCHEMA`/`PLATFORM_SCHEMA(_BASE)`.

`async_setup` is defined at line 73, and there is no `CONFIG_SCHEMA` / `PLATFORM_SCHEMA` / `config_validation` import anywhere under `mcp_proxy/`. Since this PR introduces the gate and is already editing the manifest to satisfy it, closing out the last remaining warning on this integration belongs here.

**Do not use the obvious boilerplate fix.** `cv.config_entry_only_config_schema(DOMAIN)` is wrong for this integration. Its validator (HA core `config_validation.py:1254`, `_no_yaml_config_schema`) fires on `if domain in config` — any `mcp_proxy:` key, even a bare one — logging an ERROR and raising an ERROR-severity repair issue ("does not support YAML setup, please remove it"). That directly collides with the YAML-migration path `async_setup` is built to accept: `async_setup` (lines ~84–97) checks `if DOMAIN in config` to detect a legacy `mcp_proxy:` line and fire a one-time import flow. Migration would still technically work (config is returned unchanged), but a mid-migration user would get a contradictory error repair card.

Use a permissive schema that allows the migration line:

```python
import voluptuous as vol

CONFIG_SCHEMA = vol.Schema(
    {vol.Optional(DOMAIN): vol.Any(None, dict)},
    extra=vol.ALLOW_EXTRA,
)
```

(If the YAML-migration path is genuinely dead, the alternative is to remove the migration block from `async_setup`, after which `config_entry_only_config_schema` becomes correct — but that's a behavior change and should be a deliberate decision, not a side effect of silencing a warning.)

---

### 2. The repo-wide gate now validates a vendored third-party fixture — exclude it

`.github/workflows/validate.yml` (hassfest job) + `tests/initial_test_state/custom_components/hacs/`

Hassfest scans the whole repo tree, and the run confirms it validated 3 integrations — including the vendored upstream HACS integration used as an e2e fixture (domain `hacs`, codeowners `@ludeeus`, version `2.0.5`). That's code this repo does not own. It passes today, but a fixture refresh or a future HA hassfest rule tightening could turn Validate red for a reason unrelated to this project's code.

This repo already has a convention for exactly this: `scripts/codeql_quality_gate.py` defines `PATHS_IGNORE = ("tests/initial_test_state/",)`, and `codeql-quality.yml` documents that vendored fixtures under `tests/initial_test_state/` are dropped from repo-wide gates. Mirror it here.

The hassfest action has no per-path filter input, so do it after checkout — e.g. add a step that runs `rm -rf tests/initial_test_state` before the hassfest step. This won't drop `mcp_proxy`, which lives under `homeassistant-addon-webhook-proxy/`. (Alternatively, consciously accept the coupling and say so.)

Patch76:
Addressed both items in ef64548.

## Item 1 – mcp_proxy `[CONFIG_SCHEMA]` warning

Added a permissive module-level `CONFIG_SCHEMA` to `mcp_proxy/__init__.py` using the schema you suggested:

```python
CONFIG_SCHEMA = vol.Schema(
    {vol.Optional(DOMAIN): vol.Any(None, dict)},
    extra=vol.ALLOW_EXTRA,
)
```

This satisfies the hassfest check while leaving `async_setup`'s `if DOMAIN in config` migration path intact. As you noted, `cv.config_entry_only_config_schema` would have raised the ERROR repair issue on any `mcp_proxy:` line. A comment at the definition records why the boilerplate is wrong here, so it isn't "corrected" to the broken form later.

## Item 2 – vendored fixture in the repo-wide gate

Added a `rm -rf tests/initial_test_state` step before the hassfest action in `validate.yml`, mirroring `scripts/codeql_quality_gate.py`'s `PATHS_IGNORE = ("tests/initial_test_state/",)`. The hassfest job now validates only `ha_mcp_tools` and `mcp_proxy`; the vendored `hacs` fixture (codeowner `@ludeeus`) is dropped before the scan. The step is job-local, so the HACS job is unaffected.

Patch76:
Follow-up: pushed `14cfb35a` after the review fixes.

Adding the module-level `CONFIG_SCHEMA` to `mcp_proxy/__init__.py` means voluptuous is now used at import time (`vol.Optional` / `vol.Any` / `vol.ALLOW_EXTRA`). The addon's isolated unit tests (`tests/addon/test_webhook_proxy.py`) import the module against a hand-rolled `voluptuous` stub that previously exposed only `Schema`, so the import raised `AttributeError: module 'voluptuous' has no attribute 'Optional'` at test setup. `14cfb35a` extends that stub to cover the three attributes the schema construction uses.

All `Validate`, `Docker & Add-on`, and E2E checks are green on `14cfb35a`.

kingpanther13 (approved):
## Approve

Both requested changes are in and verified against the latest run, not just the diff:

- **`[CONFIG_SCHEMA]`** — `mcp_proxy/__init__.py` now declares the permissive `CONFIG_SCHEMA` (`vol.Optional(DOMAIN): vol.Any(None, dict)`, `ALLOW_EXTRA`), clearing the warning without breaking the `mcp_proxy:` YAML-migration path. Hassfest no longer emits it. The follow-up stub for `vol.Optional/Any/ALLOW_EXTRA` in `tests/addon/test_webhook_proxy.py` correctly covers the new import-time use.
- **Vendored fixture** — the `rm -rf tests/initial_test_state` step scopes hassfest to the integrations this repo owns (3 → 2), matching the `PATHS_IGNORE` convention.

The new `run:` step is a static path with no `${{ }}` interpolation, so it adds no injection surface; workflow permissions/triggers/secrets posture is unchanged.

PR #1690, opened 2026-06-25
docs: clarify helper initial restore behavior

### LINKED ISSUE #1682: Document input_boolean `initial` resets state on every HA restart
## 🤖 What Did the AI Agent Do?

An agent created `input_boolean` helpers via `ha_config_set_helper` with `initial: false` (or `initial: true`), believing this only set a first-run default. The helpers then reset to that value on every Home Assistant restart instead of restoring the last on/off state — the behavior users typically expect from toggles.

Affected helpers in one installation: `hot_water_boost` (`initial: false`), `automatic_lights` and `closet_lights_auto_off` (`initial: true`).

## 🎯 What Should the Agent Have Done?

Omit `initial` when creating `input_boolean` helpers unless the helper should **always** reset to a fixed value on restart. State should be restored from the last value when `initial` is absent from storage.

## 📝 Conversation Context

User noticed `input_boolean.hot_water_boost` was off after every HA restart despite toggling it on. Investigation traced this to HA core behavior: if `initial` is present in the helper config (even `false`), `input_boolean` skips `async_get_last_state()` and always starts at `initial` on load.

Relevant HA core code (`homeassistant/components/input_boolean/__init__.py`):

```python
async def async_added_to_hass(self) -> None:
    # Don't restore if we got an initial value.
    await super().async_added_to_hass()
    if self._config.get(CONF_INITIAL) is not None:
        return
    state = await self.async_get_last_state()
    self._attr_is_on = state is not None and state.state == STATE_ON
```

The bundled skill `references/helper-selection.md` documents `input_boolean` without mentioning `initial`, and only shows `initial` for `counter` (paired with `restore: true`). The `ha_config_set_helper` tool exposes `initial` with the generic description *"Initial value for the helper"* — which reads like a one-time default, not a per-restart override.

**User prompt:** Create/use `input_boolean` helpers (e.g. Hot Water Boost); later: "It turns off every time I restart HA."

**Tool call(s) the agent chose:**
```
ha_config_set_helper(
  helper_type="input_boolean",
  name="Hot Water Boost",
  icon="mdi:water-boiler",
  initial=false   # or omitted but HA/UI stored false explicitly
)
```

## 💡 Suggested Improvement

- [x] **Tool documentation** - Tool description or examples need clarification
- [ ] **Error messages**
- [ ] **Tool design**
- [x] **Agent prompting** - System prompt / skill should guide agent differently
- [ ] **New tool needed**
- [ ] **Other**

**Details:**

1. **`ha_config_set_helper` `initial` parameter description** — For `input_boolean`, document that setting `initial` (including `false`) disables state restore and forces that value on every HA restart. Recommend omitting `initial` unless fixed reset-on-restart is intended.

2. **`references/helper-selection.md`** — Add a note under `input_boolean` explaining the `initial` semantics above (distinct from `counter`, which uses `restore`).

3. **Optional:** Consider omitting `initial` from the default agent-facing schema/examples for `input_boolean`, or warning when `initial` is passed.

Note: once `initial` is stored, HA UI cannot clear it; `ha_config_set_helper` update preserves the field — delete/recreate is required to remove it.

---

## 📊 Environment

- **ha-mcp Version:** 7.8.1.dev717
- **MCP Transport:** stdio
- **MCP Client:** cursor-vscode 1.0.0
- **Home Assistant Version:** 2026.6.4

---

## ⚙️ ha-mcp Configuration

- **enable_websocket:** `True`
- **enable_dashboard_partial_tools:** `True`
- **enable_tool_search:** `False`
- **enable_yaml_config_editing:** `False`
- **enabled_tool_modules:** `all`

---

_Report filed via `ha_report_issue` workflow._

kbuck1:
## What does this PR do?

Fixes #1682: Resolves a documentation issue where agents may be mistakenly guided to add an `initial` parameter when creating input entities.

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [X] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing
- [X] All automated tests pass (`uv run pytest`)

## Checklist


kingpanther13 (changes requested):
**Request changes**

The docstring change in `tools_config_helpers.py` is correct and resolves item 1 of #1682 — the `initial` descriptions now make the disables-restore behavior explicit.

The blocker is the second file in this diff: it bumps the `skills-vendor` submodule to `074c30624e1d99fdbabc03c15c945666a034f57c`, which doesn't exist in `homeassistant-ai/skills`. Every CI job that checks out submodules fails at the checkout step, before any test runs:

```
fatal: remote error: upload-pack: not our ref 074c30624e1d99fdbabc03c15c945666a034f57c
fatal: Fetched in submodule path 'src/ha_mcp/resources/skills-vendor',
       but it did not contain 074c30624… . Direct fetching of that commit failed.
```

That commit looks like a local-only edit to `references/helper-selection.md` (item 2 of #1682) that was committed inside the vendored submodule but never pushed. A submodule pointer can only reference a commit that exists on the skills remote, so this can't land as-is.

Please:

1. **Reset the submodule** back to master's pointer so this PR is docstring-only and CI goes green:
   ```
   git checkout origin/master -- src/ha_mcp/resources/skills-vendor
   ```
2. **Open the `helper-selection.md` change as its own PR against `homeassistant-ai/skills`** — adding the `input_boolean` `initial` note that #1682 item 2 asks for. Once that merges there, the submodule advances on its own (Renovate auto-bumps `skills-vendor` weekly), so it doesn't need a manual bump here.

Once you clear that up, I'll do a full review to see whether it's mergeable or needs other changes.

kingpanther13 (changes requested):
Thanks — resetting the submodule cleared the CI failures. The descriptions are factually accurate (I verified against HA core that `initial` disables last-state restore for all five input helpers, and `counter` is correctly excluded), and this also fixes a pre-existing bug where the old description omitted `input_number`. Two things to tighten before merge:

**1. The headline warning drops the "false/0" case — the exact trap from #1682.**
`_INITIAL_DISABLES_RESTORE_DESCRIPTION` says *"When set…"*, but your own comment above it deliberately notes "(including false/0)". For `input_boolean` an agent can read "When set" as "set to `true`" and conclude `initial: false` still restores — which is precisely the `initial: false` case #1682 was reported for (`CONF_INITIAL is not None` skips restore even when `False`). Please surface the nuance in the agent-facing text, e.g.:
```python
_INITIAL_DISABLES_RESTORE_DESCRIPTION = (
    "When set — even to false/0 — disables last-state restore and forces this "
    "value on every HA restart. Omit unless you want the helper to reset to this "
    "value on every restart instead of restoring its last state."
)
```

**2. `counter`'s `initial` field has no description, leaving the set half-documented.**
Every other `initial` field now carries guidance; counter's (around line 408) is still bare, so a caller who hits a counter validation error sees no `initial` guidance in the schema. Mirror the siblings with a counter-specific note that deliberately omits the disable-restore warning (counter restores by default):
```python
{
    "name": "initial",
    "required": False,
    "selector": {"number": {}},
    "description": "Initial value. Counter restores its last value on restart by default; control via `restore`.",
},
```

**Minor, while you're in there:** align the top-level `_INITIAL_PARAM_DESCRIPTION` to the same "even to false/0" wording, and reword "For counter, use with `restore` (default True)" to "For counter, `initial` is just the starting value — restore-on-restart is controlled separately by `restore` (default True)", which reads less like "you must also pass `restore`".

Once those land I'll re-review.

kingpanther13:
@kbuck1 sorry just to clarify, I'll merge this PR once those changes land. I apologize that I didn't bring those up with my initial review, I waited till CI was green till I ran my review agents, I should have just run them at the same time.

kbuck1:
> @kbuck1 sorry just to clarify, I'll merge this PR once those changes land. I apologize that I didn't bring those up with my initial review, I waited till CI was green till I ran my review agents, I should have just run them at the same time.

No worries! I've been busy the past few days so I'm slow to update, sorry.

kingpanther13 (approved):
Both items from the last review are addressed in b3b8b1c9: the false/0 nuance is now in `_INITIAL_DISABLES_RESTORE_DESCRIPTION` and the top-level param description, and counter's `initial` field has its own description that correctly omits the disable-restore warning. The counter/`restore` phrasing reads clearly now.

Re-verified the five-helper generalization against HA core (input_boolean/number/select/text/datetime all skip `async_get_last_state()` when `initial` is present; counter uses `restore`). Looks good.

PR #1691, opened 2026-06-25
refactor(c901): tools_registry.py below C901 threshold

sergeykad:
## What does this PR do?

Eliminates the C901 complexity ignore for \`tools_registry.py\` by converting the \`register_registry_tools\` nested-closure pattern to a \`RegistryTools\` class, the same approach used in #1665 for \`tools_search.py\`.

Before: \`register_registry_tools\` McCabe complexity C=80 (entire file bundled into one function). After: C=1.

Key changes:

- \`register_registry_tools\` body replaced with \`register_tool_methods(mcp, RegistryTools(client))\` (C=1)
- \`ha_get_device\` logic extracted into module-level helpers (\`_extract_zigbee_info\`, \`_classify_integration\`, \`_get_device_info\`, \`_build_entity_maps\`, \`_fetch_registries\`, \`_enrich_zha_metrics\`, \`_enrich_zwave_status\`, \`_get_single_device_result\`, \`_filter_devices\`, \`_list_devices_result\`)
- \`ha_remove_device\` removal loop extracted to \`_remove_device_config_entries\`
- \`asyncio.gather()\` in \`_fetch_registries\` (device+entity registries fetched concurrently) and \`_remove_device_config_entries\` (config entry removals run concurrently)
- \`_process_device_domain\` made fully pure (moved \`integration_sources.append\` to caller)
- Three test fixtures updated from closure-capture to \`add_tool\` class-pattern capture
- \`test_tool_annotations.py\` class_pattern regex fixed: \`([^)]*)\` to \`(.*?)\` so \`)\` inside annotation title strings (e.g. \`"Get Device (incl. ...)"\`) no longer breaks the match
- C901 ignore for \`tools_registry.py\` removed from \`pyproject.toml\`
- \`lefthook.yml\` ast-grep hook fixed: \`**\` glob requires at least one directory segment so files directly in \`tools/\` were silently skipped; added \`src/ha_mcp/tools/*.py\` alongside \`**/*.py\`, and switched to \`{staged_files}\` so only committed files are scanned (matching how ruff works)

## Type of change
- [x] 🔧 Maintenance/refactor

## Testing
- [x] All automated tests pass (\`uv run pytest\`)
- [x] E2e docker tests: 37 passed in \`tests/src/e2e/workflows/registry/\`

## Checklist

### DIFF
--- a/lefthook.yml
+++ b/lefthook.yml
@@ -12,8 +12,10 @@ pre-commit:
             run: uv run ruff format {staged_files}
             stage_fixed: true
           - name: ast-grep
-            glob: "src/ha_mcp/tools/**/*.py"
-            run: uv run ast-grep scan
+            glob:
+              - "src/ha_mcp/tools/*.py"
+              - "src/ha_mcp/tools/**/*.py"
+            run: uv run ast-grep scan {staged_files}
           - name: uv-lock
             glob: "pyproject.toml"
             run: uv lock
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -151,7 +151,6 @@ ignore = [
 # Remove lines as individual methods are simplified below threshold.
 "src/ha_mcp/tools/tools_config_dashboards.py" = ["C901"]
 "src/ha_mcp/tools/tools_entities.py" = ["C901"]
-"src/ha_mcp/tools/tools_registry.py" = ["C901"]
 "src/ha_mcp/tools/tools_search.py" = ["C901"]
 
 [tool.pytest.ini_options]
--- a/src/ha_mcp/tools/tools_registry.py
+++ b/src/ha_mcp/tools/tools_registry.py
@@ -81,7 +81,7 @@ def _first_ieee_from_connections(
     connections: list[Any], existing_ieee: str | None
 ) -> str | None:
     """Return the first IEEE address found in connections, or existing_ieee if none."""
-    if existing_ieee:
+    if not isinstance(connections, list) or existing_ieee:
         return existing_ieee
     for connection in connections:
         if (
--- a/src/ha_mcp/tools/tools_registry.py
+++ b/src/ha_mcp/tools/tools_registry.py
@@ -6,10 +6,12 @@ This module provides tools for managing devices (list, get details, update, remo
 Important: Device renaming does NOT cascade to entities - they are independent registries.
 """
 
+import asyncio
 import logging
 from typing import Annotated, Any, Literal
 
 from fastmcp.exceptions import ToolError
+from fastmcp.tools import tool
 from pydantic import Field
 
 from ..client.rest_client import HomeAssistantAPIError, HomeAssistantConnectionError
@@ -19,6 +21,7 @@ from .helpers import (
     exception_to_structured_error,
     log_tool_usage,
     raise_tool_error,
+    register_tool_methods,
     validate_identifier_not_empty,
 )
 from .util_helpers import (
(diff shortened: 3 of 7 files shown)


kingpanther13 (changes requested):
Solid refactor — the closure→class conversion is behavior-preserving on real registry data, the LLM-facing tool docstrings/parameter descriptions/annotations are byte-identical to master, and the description's claims check out. Requesting changes on the following before merge:

**1. `_first_ieee_from_connections` docstring states the precedence backwards (`tools_registry.py:83`).**
Docstring: *"Return the first IEEE address found in connections, or existing_ieee if none."* The code does the opposite — it returns `existing_ieee` immediately when truthy and never scans `connections` (line 84: `if not isinstance(connections, list) or existing_ieee: return existing_ieee`). So an identifier-derived IEEE takes precedence over a connection IEEE (which matches master's behavior — correct), but the docstring describes connection-precedence. Fix the docstring, e.g.: *"Return existing_ieee if already set; otherwise the first 'ieee' connection value, or None."*

**2. `available_device_ids` diagnostic diverges from the sibling path and drops the id filter (`tools_registry.py:864`).**
The remove path now computes it as `list(device_by_id.keys())[:10]`, which keeps a `None` key (device with no id) and takes the first 10 ids overall. The single-device get path (`tools_registry.py:272-274`) kept master's idiom `[d.get("id") for d in all_devices[:10] if d.get("id")]`. Two different computations of the same error-context field in one file. Make the remove path mirror the get path so the falsy-id filter is preserved and the two stay consistent.

**3. Destructive config-entry removal lost per-entry failure aggregation under `asyncio.gather` (`tools_registry.py:436-447`).**
`gather(...)` without `return_exceptions=True` means if any `send_websocket_message` *raises* (e.g. connection drop), the whole call bails to the generic `except Exception` handler and the per-entry `removal_results` (and the partial-success path at 897-905) are bypassed — and all N removals were already dispatched concurrently. Add `return_exceptions=True` and fold exceptions into each entry's `{success: False, error: ...}` dict, so the partial/failure reporting that the sequential master code provided is retained on hard failures.

**4. Extracting `_process_device_domain` dropped the identifier-format comments (`tools_registry.py:36-51`).**
Master documented the non-obvious magic-string parsing (`["zha","IEEE"]`, `"zigbee2mqtt_0x…"`, `"{home_id}-{node_id}"`). The new helper has `value.split("_0x")[-1]` / `value.split("-")[1]` with no explanation. Restore those format comments in the helper.

**5. `detail_level` widened from `Literal` to `str` at the helper boundaries (`tools_registry.py:323`, `:388`).**
The tool param is `Literal["summary","full"]`; the new `_filter_devices` / `_list_devices_result` signatures take `detail_level: str`, discarding the compile-time guarantee. Type those params `Literal["summary","full"]` to keep the invariant the refactor erased.

kingpanther13 (approved):
Re-reviewed — all five requested changes are in `5a51596d` and verified against source:

1. `_first_ieee_from_connections` docstring now matches the code (existing_ieee takes precedence; connections are the fallback).
2. Remove-path `available_device_ids` mirrors the get-path comprehension and filters falsy ids again.
3. `_remove_device_config_entries` passes `return_exceptions=True` and folds exceptions into the per-entry result dicts, so partial/failure aggregation survives a raising removal.
4. Identifier-format comments (ZHA / Z2M / Z-Wave layouts) restored in `_process_device_domain`.
5. `detail_level` typed `Literal["summary","full"]` in both `_filter_devices` and `_list_devices_result`.

The refactor stays behavior-preserving on real registry data, and the tool docstrings / parameter descriptions / annotations remain byte-identical to master. Approving.

Issue #1713, opened 2026-06-30
[BUG] ha_config_set_automation python_transform silently no-ops on large automation's actions list


joshvenman:
### 💬 What Happened?

I was trying to add a new branch to my blinds automation that is already quite complex - 17 different choose blocks at least. This was Claude had come up with the requited YAML and was trying to insert the block between two existing, related bramnches, and it didn't work. In the end, I had to copy the entire automations.yaml out , have Claude update it with the new branch, then overwrite the old automations.yaml with the new content. Prior to today, Claude had been able to make similar injections of new content in to automations.yaml which made me think it was a rectn update that might have changed the behaviour.

### 📋 Copy-Paste from Conversation

```markdown
## 🚨 Auto-Generated by `ha_report_issue` Tool

**Submit at:** https://github.com/homeassistant-ai/ha-mcp/issues/new?template=runtime_bug.yml

---

## 📋 Bug Description

ha_config_set_automation's python_transform silently no-ops when the transform mutates config['actions'] on one specific large, deeply-nested automation (17+ branches, up to 4 levels of nested choose, ~33KB serialized) — the tool reports success with a new config_hash, but the change is absent on re-fetch and absent from the on-disk auto-backup.

## 🔄 Steps to Reproduce

1. Have an automation with a large actions: [{choose: [...]}] list, deeply nested (mine: 17 branches, up to 4 levels of nesting)
2. Call ha_config_set_automation with a python_transform that inserts a new item into config['actions'] (tried: full list rebuild, in-place .insert(), and a debug version that checks len(config['actions']) before/after inside the same transform)
3. Tool returns success: true with a changed config_hash
4. Re-fetch via ha_config_get_automation — the new branch is absent; config_hash matches the pre-edit value
5. Confirmed against the actual on-disk auto-backup (ha_manage_backup, scope=edits) — genuinely unchanged
6. Restarted the ha-mcp add-on entirely and repeated steps 2-5 — same result (rules out stale in-process state)

## ✅ Expected vs ❌ Actual Behavior

Expected: the new list item appears in config['actions'] on the next ha_config_get_automation call and in the saved YAML.

Actual: the actions list is unchanged; only edits to other top-level fields on the same automation (e.g. appending to config['description']) persist correctly.

---

## 💬 Triggering Prompt & Tool Call

User prompt: "Add the new branch" (after diagnosing that automation.auto_blind_control_lux_and_sun was missing a close branch for a specific elevation/lux combination)

[...]

Issue #1266, opened 2026-05-12
[FEATURE] Standardize exception-handling convention for verification helpers in tests/src/e2e/


Patch76:
## What would you like to see?

Project-level decision and (likely) follow-through PR on the exception-clause width used in verification helpers and boot-phase polling loops across `tests/src/e2e/conftest.py` and `tests/src/e2e/utilities/`.

Currently the convention is **narrow catch**, e.g.:

```python
except (requests.exceptions.RequestException, json.JSONDecodeError) as exc:
    logger.debug(f"... check failed: {exc}")
```

at every boot-phase polling loop in `ha_container_with_fresh_config` (`STABILIZATION_TIMEOUT`, `ENTITY_STABILIZATION_TIMEOUT`, `INPUT_BOOLEAN_WAIT`, `HA_MCP_TOOLS_WAIT`, `SUN_WAIT`) and at the two new helpers in PR #1262 (`_wait_for_ha_api_ready`, `_wait_for_ha_mcp_tools_services`). Similar narrow catches in `tests/src/e2e/utilities/wait_helpers.py`.

Gemini Code Assist consistently recommends the **broad form**:

```python
except Exception as exc:
    logger.warning(f"... check failed: {type(exc).__name__}: {exc}")
```

The two paths:

**Option A (document status quo).** Add a short style note to `.gemini/styleguide.md` and `AGENTS.md` that boot-phase verification helpers use narrow catch + debug-level logging, deliberately. The bot stops raising the finding on future PRs and human reviewers know to expect the narrow form.

**Option B (adopt broad uniformly).** One focused PR that switches every verification helper and polling loop to `except Exception` + warning-level + `type(exc).__name__`. Touches every site flagged on PR #1262 plus equivalents elsewhere.

## Why do you need this?

Gemini raised the same broad-except finding **six times across PR #1262 iterations 1–6** — once on the new helpers in iter-1, again on the diagnostic helper in iter-2 (different context, accepted there), then re-pitched on the same polling-loop sites in iter-3 / 4 / 5 / 6 even after the prior declines. Each cycle spent reviewer effort declining the same finding with the same intra-fixture-consistency reason.

A documented decision — whichever way — would either:

1. Stop the bot from re-raising on future similar PRs (Option A), or
2. Eliminate the inconsistency the bot's rule actually targets (Option B).

[...]

PR #1751, opened 2026-07-05
feat: custom component first — installation overhaul and self-updating in-process server (#1715)

### LINKED ISSUE #1715: [Refactor] Installation overhaul: custom component as the primary install method (README + setup site)
## Summary

The custom component's in-process **HA-MCP Server** entry (#1527, shipped in #1741) works on every HA installation type, has feature parity, and is the easiest install in every case. Make it the primary installation method everywhere — README and the setup site. The add-on, Docker, PyPI, and stdio become secondary alternatives. This supersedes the original (container-first) plan, kept below for reference.

## New quick start

1. Install the custom component (HACS button), restart HA.
2. Get the server URL from the persistent notification or the entry's settings.
3. Pick your AI client and paste the URL (with secret code) into its config — done.

## README

- Custom component moves to the top as the recommended method; add-on a close second; the other methods (Docker, PyPI, stdio) follow as alternatives.
- Add-to-HACS button front and center, marked as the preferred installation method.
- The Windows/macOS/Linux quick-install scripts only set up the demo server: move them to their own "Demo server" section and stop calling them quick start.
- stdio stays documented for demo/test tinkering but is not recommended for a real setup, with a warning about its transport issues (#1713; steerage already started in #1712).

## Setup site (Astro)

- Remove "Coming Soon" from the HA Component card — it shipped.
- Quick Setup revamp: HACS install of the custom component → restart HA → get URL, then pick AI client, then OS.
- Merge the local options (local machine / local network / HTTP) into a single local path: use the /private key URL and turn the webhook toggle off when local-only is desired.
- Cloudflared instructions stay (the integration still issues the same style of private-key URL) but get relabeled ADVANCED instead of Quick; the built-in webhook is the other remote option — users can also put their external URL directly into the integration to get a webhook URL.
- Reverse-proxy users just need the proxy pointed at HA, then use the secret URL or enable HA's built-in auth (ha_auth).
- "Alternative setup methods" section at the bottom, branched by HA installation type: add-on shown for HAOS/Supervised (the add-on button stays but lives under that branch); Docker, PyPI, and stdio for container/core installs.

## Webhook proxy add-on

- Add a note that it is not needed with the custom component.
- Drop the "run simultaneously with the add-on" wording: the component and the add-on are completely independent, and nothing should imply the add-on is required or that they work together.

<details>
<summary>Original issue body (superseded — kept for reference)</summary>

Right now this is just an investigatory issue, I will fill it out with more detail and research when I have the time to do so... I'm just jotting my thoughts down right now so I don't forget in the future .

Right now our primary quick start guides heavily push people to using STDIO installations with Claude desktop as the primary "quick start" method, however I have noticed that a lot of recent issues over the last few months appear to be due to the STDIO transport method, and typically switching to streamable will result in fixing the issues (typically by switching to addon version). Additionally there has been confusion, they see the link to download the app repository but then still will use the stdio version which completely bypasses the app/addon (working on that in #1712). 

 I want to investigate if it is possible to easily do a quick setup / installation using the container/streamable version instead, so users who can not or choose not to use the addon/app at least have a better functioning streamable transport. If it is possible to simply copy and paste the settings into the Claude desktop settings the same way we do for STDIO and simply have Claude desktop spawn up the container version instead of the other version, it may result in less bugs and some better functionality, but I've no idea the implications or plausibility of that yet, I need to research further. . I am already trying to encourage users to use the add-on version instead, which should help with this as well (#1712) so it may not even be necessary. . I am hoping that more clear setup instructions will prevent the duplicate install issue, but nonetheless I think that moving away from STDIO in the first place will work a lot better for everybody that chooses to not use the app/addon and hopefully lead to easier troubleshooting, I just need to be sure that there aren't any other pitfalls. I'm not suggesting completely dropping stdio support because many users are already on it, but at least we can encourage users to move away from it and discourage new users from going straight to it. 

Another idea that goes along with this: right now the setup guide/docs has the quick start things up top, then below that asks for AI client as the first step. Instead of doing that, we should start off with two paths: Do you run HAOS supervised (or otherwise capable of using apps/add-ons), or do you run it in a container/unsupervised/otherwise unable to run apps/add-ons? And within the quick start guides themselves we can have branches for app vs no app (might even be implementing that in #1712 anyways). Then from there branch out: if supervised , do you want to run the app/addon (recommended) or do you want to run in a container? If not supervised, you cannot run the app. do you want to run the pypi/docker container (recommended), or do you want to do the stdio one? THEN pick the os/ai/etc. This will require reordering the flow of the setup site significantly but I think it'll make more sense and hopefully will push users to using the addon and pypi/container versions more where appropriate since they are less buggy and have more features. 

To clarify when I say move away from STDIO, I mean to have the container/pypi/streamable version with the same easy steps in the quick start guide, AND to change the quick setup scripts that are on the home setup page as well. If they're able to be just as seamless/stateless/whatever as the stdio version then I feel like there's no reason not to switch. 

Edit: another point, #1713 is related to this (as are several other issues). I've already implemented the initial steerage away from STDIO transport via #1712, but I think as part of this we should be adding a warning/caveat to the stdio installation methods that stdio has some transport issues that streamable does not would beneficial as well. That plus demoting stdio out of the main install path should hopefully make a lot of difference. 

</details>

kingpanther13:
## What does this PR do?

Implements #1715: the HA-MCP Custom Component (in-process **HA-MCP Server** entry, #1527/#1741) becomes the primary installation method everywhere; the add-on, Docker, PyPI, and stdio become secondary alternatives.

**Live site preview for reviewers:** https://kingpanther13.github.io/ha-mcp-fork/ — the full site built from this branch (GitHub Pages on the PR fork), working exactly as it will when merged: all pages, navigation, badges, and the interactive setup wizard ([/setup/](https://kingpanther13.github.io/ha-mcp-fork/setup/)). Only difference from production is the base path (`/ha-mcp-fork/` instead of `/ha-mcp/`).

- **README**: leads with the component quick start (HACS badge, 5 steps, both connect-URL forms, local-only webhook toggle, sidebar panel, `ha_auth`); add-on demoted to a close second; Docker/PyPI/stdio grouped under "Other install methods" with a stdio transport warning (#1713); the OS one-command installers reframed as a "Demo server" section; mid-page component section slimmed to the file/YAML services role.
- **Setup wizard** (`setup.astro`): reworked to a server-method-first flow — method → AI client → scope (the old local machine / local network split merged into one Local option) → platform (only when OS commands are needed) → remote path → config. The HA Component card is enabled (was "Coming Soon") and recommended; the component's built-in webhook is the quick remote path (the Remote scope card also shows Quick for the component); Cloudflare Tunnel is relabeled Advanced; the OS quick-start cards moved to a demo-server strip with their `easy-*` anchors preserved.
- **Front page / FAQ / add-on guide**: HACS badge replaces the add-on badge as the preferred install; component card leads Quick Start; FAQ covers both entry types, the HA-MCP sidebar panel, and a new "do I still need the add-on or webhook proxy?" answer.
- **Webhook proxy (dev, 1.2.3.dev6)**: docs note that the proxy is unnecessary with the custom component. `docs/in-process-server.md` drops the run-both-with-the-add-on guidance for an independence statement — the component and the add-on never require each other.
- **Custom component 0.14.0**: the add-integration menu now lists **HA-MCP Server** first and no longer suggests adding both entries — **HA MCP Tools** is labeled optional and both descriptions state it is only needed when the opt-in file/YAML editing tools are enabled.
- **Self-updating in-process server**: the legacy pinned-stable lockstep (`PINNED_HA_MCP_VERSION`, an add-on-era relic, removed from semantic-release `version_variables`) is replaced by channel auto-update — both channels install unpinned, reload/restart picks up the newest build, and a periodic 6-hour PyPI check reloads the entry when a newer build is published. A new **Automatic server updates** checkbox (default on) pins to the currently installed version when off; an explicit pip-spec override still wins. The component also files a repair issue pointing at the HACS component update when the running server's `MIN_COMPONENT_VERSION` exceeds the installed component, and the Configure screen now shows the component + installed server versions.
- **Connect URL wording**: the notification deliberately carries no URL (it is visible to every signed-in user while the URL is the credential), so every quick start now sends users to the entry's **Configure** screen / the HA log instead of the notification.
- **Configure screen shows real URLs**: the options form printed placeholder bases (`<your-home-assistant-url>`, `<home-assistant-ip>`) next to the real webhook id and secret path. Since the screen is admin-only, it now resolves and shows the actual connect URLs (external URL, Nabu Casa, local base, direct-access with the real host) via a builder shared with the start-up log. In local-only mode the hint states that instead of showing a webhook URL that is never registered.
- **Review-round fixes** (multi-agent review of this PR): the wizard's tunnel/custom-proxy instructions now target the right port per server method (add-on 9583 — previously 8086, a pre-existing bug — component 9584, docker/uvx 8086); the direct-access URL is only rendered when its secret path exists; plus ~400 lines of new unit tests covering the connect-URL builder, all hint fallbacks, remote-path tile filtering, disable guards, and the tunnel port in both directions.
- stdio is labeled **not recommended** (rather than "advanced" — it is not more advanced than Docker/uvx, just discouraged).
- **HACS continuity**: every HACS badge, manual-add instruction, and the `ha_install_mcp_tools` installer now point at the distribution mirror `homeassistant-ai/ha-mcp-integration` (the repository submitted to the HACS default store, hacs/default#8968), so custom-repo installs match the upcoming official listing; the installer still recognizes legacy main-repo entries. The mirror-sync workflow also now composes the mirror's rendered HACS README from a permanent prefix blurb plus the full main README (relative links/images absolutized), refreshed automatically on every README change.
- Cleanup: dead `tests/README.md` anchor fixed; unused `Welcome.astro` starter component removed; stale `site/README.md` tree corrected; wizard-internal guide/FAQ links made base-aware.

Closes #1715.

## Type of change
- [x] 🐛 Bug fix
- [x] ✨ New feature
- [x] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing
- [x] All automated tests pass (`uv run pytest`) — wizard behaviour tests rewritten for the new flow, 24/24 pass locally; full e2e suite left to CI (docs/site-only change)

Site gate run locally: `astro check` 0 errors, eslint clean, production build OK, axe-core a11y audit 0 violations across all 9 pages.

## Checklist


kingpanther13:
@Patch76 will ping you for review when done. Sorry for noise, thought it was done but found a few more issues. This will be more than a docs PR, it will be polish so we can fully release custom component.

kingpanther13:
@Patch76 ok should be ready to go. I hope there aren't too many issues because I'm almost completely out of usage till Thursday night lol. Once this merges we can cut an early stable release; it shouldn't affect any current users since it's all about the custom component, but it's important to cut it early so if stable release fails for whatever reason we'll be able to fix it now.

Patch76 (changes requested):
This is a large, well-executed change: the self-updating server, the connect-URL builder, the mirror README compose, and the setup-flow rework all hang together, and the unit coverage is genuinely thorough (the auto-update state machine, `build_connect_urls`, and the periodic-check registration are all exercised at the edges). One thing I'd fix before this ships as the primary install path, plus a few nits you can fold or leave.

**[blocking] Blocking `importlib.metadata` reads on the event loop (two sites).**
The same read you correctly push to an executor in `async_check_for_update` and `_async_check_component_compat` is called directly on the loop in two places:

- `config_flow.py` `_versions_hint()` calls `_installed_server_version()` synchronously (runs on every Configure-screen open).
- `embedded_server.py` `_resolve_pip_spec()` calls `_installed_dist_version()` when auto-update is off (runs during `__init__`, i.e. inside the bring-up task on the loop; it also calls `importlib.invalidate_caches()`, forcing a fresh scan).

`importlib.metadata.version()` scans dist-info via `os.listdir`, which HA's `block_async_io` guards. It is `strict=False`, so it does not raise, but it logs `Detected blocking call to listdir ... by custom integration 'ha_mcp_tools' ... please report` (deduped to once per site per run). The `os.listdir` guard is `skip_for_tests=True`, which is why the suite stays green. For the flagship install path I'd route both through `hass.async_add_executor_job(...)`, matching what you already do in `embedded_setup.py`.

**[nit] Narrow the version-compare `except` in the periodic check.**
`async_check_for_update()` wraps `AwesomeVersion(latest) > AwesomeVersion(installed)` in `except Exception`. In a periodic callback that swallows programming errors (the #1266 convention). `AwesomeVersion` raises `AwesomeVersionException` (its `...CompareException` / `...StrategyException` subclasses) for the incomparable-strategy case the comment refers to — not `TypeError`/`ValueError` — so `except AwesomeVersionException:` catches exactly the intended case and lets real bugs propagate. Same applies to the compare in `_async_check_component_compat` (that one logs with `exc_info` and is advisory, so it is lower priority).

**[nit] HACS installer prefers the legacy repo on one edge.**
In `ha_install_mcp_tools`, if the legacy `homeassistant-ai/ha-mcp` repo is present in HACS but not installed and the mirror is absent from the list, `_find_existing_repo` returns the legacy entry and the install proceeds from it rather than the canonical mirror. It still installs the same component, so this is a source-preference nit, not a breakage — but if you want fresh installs to always land on the mirror, treat "matched the legacy repo and it isn't installed" like the not-found branch and add the mirror.

**Note (not blocking):** #1715 is scoped to the README + setup site; this PR also lands the self-updating-server mechanism, the component-compat repair issue, and the `PINNED_HA_MCP_VERSION` removal. That is reasonable as the enabling work for "component as primary," and the body links it correctly with no auto-close keyword — flagging only because the auto-update mechanism is a feature in its own right.

For what it's worth I checked the security posture around the notification: it carries no URLs and points at the admin-only Configure screen and log, and the direct-access line stays out of it — that reads correctly.

kingpanther13:
Thanks for the thorough pass. Fixed both the blocking reads and the compare `except` in e5027e57:

- **Event-loop reads:** `_versions_hint` now offloads `_installed_server_version()` to the executor, and `_resolve_pip_spec` takes the installed version as a parameter — the `importlib.metadata` read (with `invalidate_caches`) moves to the executor call `_async_ensure_package` already makes, so `__init__` no longer touches metadata on the loop. Both flagged sites are off the loop now.
- **Version-compare `except`:** narrowed to `AwesomeVersionException` in both the periodic check and the component-compat check, so a real bug propagates instead of being swallowed. In the compat function I split the loader call (kept a visible `warning` + `exc_info` catch, since `async_get_integration` has a genuinely wide failure surface) from the compare (narrow).

On the installer legacy-repo preference: intentional, leaving as-is. The mirror only exists to get listed in the default HACS store; installing from either the mirror or the main repo yields the same component, so preferring an already-present legacy entry over adding the mirror is fine — it avoids a redundant repo add for existing users and the result is identical.

On scope: this PR *is* the #1715 work — it overhauls the README and the setup site to make the custom component the primary install path, which is exactly what #1715 asks for, and it closes the issue (`Closes #1715` is in the body). The self-updating-server mechanism isn't tangential: making the component the primary path requires it to track its channel cleanly, otherwise stable-channel users would be stranded on the old add-on-era pinned-server lockstep. It landed here because "component as primary" doesn't actually work without it.

Patch76 (commented):
Re-reviewed at e5027e57.

## Blocking event-loop reads – verified fixed

Both flagged sites are off the loop, and I traced each executor boundary:

- `config_flow._versions_hint` offloads `_installed_server_version()` through `async_add_executor_job`. The only remaining direct call is the `hass is None` fallback, which is not the path that renders the admin Configure screen.
- `embedded_server._resolve_pip_spec` is now a pure function; the `importlib.metadata` read moved onto the `_installed_ha_mcp_version()` executor call that `_async_ensure_package` already makes, so `__init__` no longer touches metadata on the loop.

I swept the rest of the component for the same pattern. Every `importlib.metadata` / `invalidate_caches` / `find_spec` read now runs through `async_add_executor_job` (`_dist_installed`, `_uninstall_distribution`, `_read_min_component_version`, and the `_installed_dist_version` auto-update read included). No third site.

## Version-compare `except` – verified fixed

Narrowed to `AwesomeVersionException` in both the periodic check and the compat check. I confirmed against the awesomeversion source that this is the correct class: a comparison routes through `_validate_comparison_strategy`, which raises `AwesomeVersionCompareException` (a subclass of `AwesomeVersionException`) for the UNKNOWN-strategy case, and nothing else in the compare path raises, so a real bug (`TypeError`, etc.) still propagates per the #1266 convention. Splitting the compat function into a wide `except` around the loader (visible `warning` + `exc_info`) and a narrow `except` around the compare is cleaner than a single narrow catch.

## Installer legacy-repo preference, scope, `Closes #1715`

Agreed on all three. The installer preference is a source nit either way, and the self-updating server is the enabling work for "component as primary", so the scope and the close-ref hold.

## One correctness item the read-move introduced

Moving the pin read swapped its source in a way that misfires on a channel switch. To be clear, this one falls out of the executor offload I asked for, not anything in your original pin logic, which read the target dist directly and handled the switch fine. `_resolve_pip_spec` used to pin from `_installed_dist_version(dist)` (the target channel's distribution). It now takes the value from `_installed_ha_mcp_version()`, which returns whichever of `ha-mcp` / `ha-mcp-dev` is currently installed. On a channel switch with auto-update off, the old channel's dist is still installed when that read runs (removal happens later in `_async_remove_conflicting_dist`), so the pin becomes `<new-dist>==<old-channel-version>`:

- Switch dev to stable, auto-update off, `ha-mcp-dev` at `8.0.0.dev3` installed. `installed_version` resolves to `8.0.0.dev3`, so `self._pip_spec` becomes `ha-mcp==8.0.0.dev3`. That version does not exist for the stable dist, `_async_force_install` fails, and the bring-up raises `EmbeddedServerError`.

It self-heals: the force-install path removed the conflicting dist first, so the next reload/restart reads `None` (nothing importable, since `_installed_ha_mcp_version` gates on `find_spec`), falls back to unpinned, and installs the newest of the new channel. So the symptom is one failed bring-up right after a cross-channel switch, recovering on the next reload rather than a permanent stuck state (assuming the best-effort `_uninstall_distribution` cleared the old dist, which is the normal case). Before the refactor this case installed the newest of the new channel directly (the dist-specific read returned `None`, so the spec stayed unpinned), so it is a regression from the move, not pre-existing. Channel and auto-update are independent options with no coupling in the schema, so the combination is reachable.

Minimal fix: pin from the target dist rather than whichever-present, for example read `_installed_dist_version(dist)` on the executor and pass that value in, so a switch with nothing of the new channel installed stays unpinned.

Required checks are green. Once the pin reads the target dist (or you decide the one-reload recovery is acceptable and say so, like the installer nit), this is ready from my side.

Patch76 (approved):
**Verdict:** Approve. Both blocking items from my earlier review are resolved, and the follow-up fix commit holds up under a mechanism trace.

**Event-loop reads (`e5027e57`):** Every `importlib.metadata` read now runs through `async_add_executor_job` – the `_installed_ha_mcp_version` call sites in `_async_ensure_package`, the per-dist `_installed_dist_version` read, and the auto-update check's read in `async_check_for_update`. `_resolve_pip_spec` stays pure (the version is passed in, never read there), so nothing reintroduces a blocking read on the loop. The version-compare `except` is narrowed to `AwesomeVersionException` in both `async_check_for_update` and `_async_check_component_compat`, so a genuine `TypeError`/`AttributeError` propagates instead of being swallowed as "no update".

**Channel-switch pin (`e0cce80d`):** This is the one I wanted to confirm in detail, since the fix is new code rather than just the absence of the old symptom. The re-pin now reads `_installed_dist_version(target_dist)` – the channel's own distribution – instead of `_installed_ha_mcp_version()` (whichever of the two dists is present). On a dev→stable switch with auto-update off, the previous channel's dist is still installed when the re-pin runs (removal happens later in `_async_remove_conflicting_dist`), so a whichever-present read would have pinned `ha-mcp` to a dev-only version and failed the install. Reading the target dist returns `None` (stable not installed yet) → `_resolve_pip_spec(None)` → the bare, unpinned dist → a force-install of the newest stable. The guard (`not override and not auto_update`) leaves the override and auto-update-on paths on their `__init__` spec unchanged, so no other path shifts.

The same target-dist discipline already holds in `async_check_for_update` (`dist` derived from the channel, compared via `_installed_dist_version(dist)`), so the pin path and the update-check path are consistent – no sibling instance of the whichever-present read is left.

`test_ensure_package_channel_switch_auto_off_stays_unpinned` reproduces the regression directly: the old dev build is present, the target stable read returns `None`, and the spec is asserted to stay unpinned. CI is green on `e0cce80d`.

One pre-existing edge I hit while tracing, not introduced by this change and not blocking: `_installed_dist_version` does not carry the `find_spec("ha_mcp")` guard that `_installed_ha_mcp_version` uses to reject orphaned `.dist-info` (metadata whose files a later channel install has overwritten). If a best-effort uninstall in `_async_remove_conflicting_dist` ever fails, a subsequent channel round-trip with auto-update off can re-pin the target to that stale version instead of reading it as absent. The old whichever-present read hit the same corner harder – it pinned to a version of the wrong distribution and failed the install outright – so the behaviour is strictly less broken now; I am only noting it as a latent edge, not something this PR needs to resolve.

PR #1777, opened 2026-07-08
feat: opt-in GET /healthz liveness endpoint (MCP_HEALTHZ env var)

FWSpider:
## What does this PR do?

Adds an **opt-in** `GET /healthz` liveness route (`200 {"status": "ok", "server": "ha-mcp"}`) for uptime monitors, blackbox probes, and load balancers. Today the only liveness signal is the by-design GET→405 on the MCP path, which forces monitoring tools into special-case "405 means alive" configs — and requires telling the monitor the (possibly secret) MCP path.

Design choices, deliberately conservative:

- **Off by default**, enabled with `MCP_HEALTHZ=true`. Standard mode authenticates by URL-path secrecy (SECURITY.md → Threat Model), and an always-on route would confirm to any scanner that ha-mcp is listening. Opt-in keeps the default attack surface unchanged.
- The response **does not echo the MCP path** or any config.
- `register_healthz()` lives in `browser_landing.py` next to `register_browser_landing()`, with the same per-instance idempotent registration (WeakSet), so the in-process reload behavior matches the landing page.
- `ProbeAccessLogFilter` drops the per-scrape `GET /healthz → 200` access-log lines (a 15s-interval probe would otherwise spam the log); a `/healthz` 404 (route not enabled) is kept as a real signal.
- Registered in http, sse, and oauth modes. `MCP_HEALTHZ` is added to the coverage test's `ENV_ONLY` registry as bind/deployment config.

**Context / nature of this contribution:** I run ha-mcp in Docker as one of nine MCP servers in my homelab, all probed by a Prometheus blackbox-exporter; ha-mcp was the only one needing a special "405 = alive" module. I wrote this for my own deployment and am upstreaming it in case it's useful to others — if you'd prefer a different shape (always-on, a different env var name, version info in the body, or not having this at all), feel free to redirect or close, no hard feelings. Happy to adjust to whatever fits the project.

## Type of change
- [x] ✨ New feature

## Testing
- [x] I have tested these changes with a LLM agent — running in my production deployment (Docker, streamable-http, Claude Code as client); tool calls verified working with the route enabled
- [x] All automated tests pass (`uv run pytest tests/src/unit/`) — 2019 passed; the 3 failures in `test_config.py` / `test_helper_response_shape.py` also fail on a clean master checkout in my environment, so they appear unrelated
- [x] Code follows style guidelines (`uv run ruff check`, `uv run ruff format --check`, `uv run mypy src/` all clean)

## Checklist
- [ ] I have updated documentation if needed — happy to add `MCP_HEALTHZ` to the docs wherever you'd like it mentioned if the feature is accepted; I didn't want to guess at placement


kingpanther13 (changes requested):
Thanks for upstreaming this. The shape looks sensible: the route is opt-in, it does not expose the MCP secret path, and the unit coverage checks the response body, default-off behavior, idempotent registration, env parsing, and access-log filtering.

I think there are two small things to fix before merge:

**Line endings / whitespace:**
`git diff --check origin/master...HEAD` reports trailing whitespace on every added line in `src/ha_mcp/__main__.py`. The file is showing as CRLF despite the repo attrs expecting LF, so please normalize the touched lines/file before merge.

**Docs for the new opt-in env var:**
`MCP_HEALTHZ` is the only way to enable this feature, but it is not documented anywhere user-facing yet. Please add a short note in the appropriate HTTP/Docker/uvx docs explaining that `MCP_HEALTHZ=true` enables unauthenticated `GET /healthz`, off by default.

Also please address Gemini’s import-consolidation review if you agree it is valid, then resolve that thread/comment so the PR is clean.

FWSpider:
Thanks for the review, @kingpanther13 — and apologies for the slow turnaround. All three points addressed, branch rebased on current master:

1. **Line endings**: added a dedicated commit normalizing `src/ha_mcp/__main__.py` to LF per `.gitattributes` (the blob predated the `eol=lf` attribute and still carried CRLF — that's what made every added line trip `git diff --check`). The feature commit sits on top of it; `git diff --check origin/master...HEAD` is clean now.

2. **Docs**: documented `MCP_HEALTHZ` in the setup wizard (`site/src/pages/setup.astro`) — the note renders on the uvx HTTP instructions (Windows + macOS/Linux) and the Docker instructions, next to the existing `MCP_PORT` note. Site builds clean. Happy to mirror it in the FAQ or elsewhere if you'd like it in more places.

3. **Gemini's import consolidation**: I'd push back gently on this one — the split `from ha_mcp.browser_landing import ...` blocks are what the repo's ruff isort config produces (`ruff check --fix` splits a consolidated import into exactly this form, and my original consolidated version failed `I001`). Consolidating would just fail the Ruff Lint CI job. Leaving as-is unless you prefer otherwise.

Patch76:
Should I take over this review?

kingpanther13 (approved):
Thanks for circling back — all three points resolved.

- **Line endings:** `git diff --check origin/master...HEAD` is clean, and the separate LF-normalize commit keeps the feature diff readable.
- **Docs:** the `MCP_HEALTHZ` note renders on the uvx (Windows + macOS/Linux) and Docker HTTP instructions, next to `MCP_PORT`.
- **Import consolidation:** agreed, leave as-is — `combine-as-imports` is off, so `ruff check --fix` splits the block into exactly your two-statement form; consolidating would fail Ruff Lint.

Opt-in by default with no MCP-path leak. Approving.

PR #1781, opened 2026-07-09
ci: fix release push race and hotfix mirror tagging

kingpanther13:
## What does this PR do?

Fixes the release-pipeline failures uncovered by the PR #1780 hotfix release:

- **Serialize master writers and retry release pushes.** The Hotfix Release run for #1780 aborted inside python-semantic-release with `Upstream branch 'origin/master' has changed!` — `sync-tool-docs.yml` pushed its doc-sync commit mid-release, and the action's freshness check is not retryable. All four workflows that push to master (`hotfix-release`, `semver-release`, `sync-tool-docs`, `addon-publish-dev` via `_update-addon-config`) now share a `master-write` concurrency group (`cancel-in-progress: false`), and the plain changelog/config pushes get sync-tool-docs' rebase-and-retry loop as defense in depth. `hotfix-release` also checks out `master` by name instead of the event's frozen `merge_commit_sha`: still an attached branch (the #1090 constraint), but re-runs can now recover, and the manual attach step is gone. Documented tradeoff: GitHub queues only one pending run per group, so a hotfix merge that also touches tool files can cancel one queued bot run — both bot workflows are `workflow_dispatch`-recoverable and self-heal on the next merge.
- **Tag the integration mirror on hotfix releases.** `sync-integration-mirror.yml` only tagged the HACS mirror after "SemVer Release", so a hotfix bumping the component version left HACS users pinned to the old tag (bit component 1.0.1). The tag step is idempotent, so hotfixes without a component change are a no-op.
- **Drop the dead `version_variables` entry** from the semantic-release config: `__init__.py` assigns `__version__ = get_version()` (runtime metadata lookup), so the pattern has matched nothing (`num_matches=0` in release logs); `version_toml` is the single bump source.

## Type of change
- [x] 🐛 Bug fix
- [x] 🔧 Maintenance/refactor

## Testing

Workflow-only change (plus one dead config entry): all edited YAML parses clean; the retry loop mirrors the proven one in `sync-tool-docs.yml`; the release path itself is only exercisable by a real release run.

## Checklist

### DIFF
--- a/.github/workflows/_update-addon-config.yml
+++ b/.github/workflows/_update-addon-config.yml
@@ -92,6 +92,17 @@ jobs:
             echo "No changes to config.yaml, skipping commit"
           else
             git commit -m "chore(addon): ${{ inputs.commit_message_prefix }} $VERSION [skip ci]"
-            git push "https://x-access-token:${PUSH_TOKEN}@github.com/${GITHUB_REPOSITORY}.git" HEAD:master
-            echo "✓ Updated addon config.yaml to $VERSION"
+            # Master can advance between this job's checkout and its push
+            # (bot commits from other workflows); retry with a rebase instead
+            # of failing the publish. Mirrors the loop in sync-tool-docs.yml.
+            for attempt in 1 2 3 4 5; do
+              if git push "https://x-access-token:${PUSH_TOKEN}@github.com/${GITHUB_REPOSITORY}.git" HEAD:master; then
+                echo "✓ Updated addon config.yaml to $VERSION (attempt ${attempt})"
+                exit 0
+              fi
+              echo "Push rejected on attempt ${attempt}; rebasing onto origin/master and retrying."
+              git pull --rebase origin master
+            done
+            echo "::error::Could not push addon config.yaml update after 5 attempts."
+            exit 1
           fi
--- a/.github/workflows/_update-addon-config.yml
+++ b/.github/workflows/_update-addon-config.yml
@@ -101,7 +101,12 @@ jobs:
                 exit 0
               fi
               echo "Push rejected on attempt ${attempt}; rebasing onto origin/master and retrying."
-              git pull --rebase origin master
+              # Abort a conflicted rebase so later attempts fail for the real reason.
+              git pull --rebase origin master || {
+                git rebase --abort 2>/dev/null || true
+                echo "::error::Rebase onto origin/master failed; aborting retries."
+                exit 1
+              }
             done
             echo "::error::Could not push addon config.yaml update after 5 attempts."
             exit 1
--- a/.github/workflows/addon-publish-dev.yml
+++ b/.github/workflows/addon-publish-dev.yml
@@ -25,6 +25,10 @@ permissions:
 concurrency:
   group: master-write
   cancel-in-progress: false
+  # queue: max keeps every queued run waiting FIFO. The default single-slot
+  # queue evicts the OLDER pending run when a new one queues, which could
+  # silently drop a queued release behind a bot push.
+  queue: max
 
 jobs:
   # Skip if commit is from semantic-release
--- a/.github/workflows/addon-publish-dev.yml
+++ b/.github/workflows/addon-publish-dev.yml
@@ -18,6 +18,14 @@ env:
 permissions:
   contents: read
 
+# Serialize with every other workflow that pushes to master (this one pushes
+# the dev add-on config bump via _update-addon-config.yml): a push landing
+# while a release workflow's semantic-release is mid-run aborts that release
+# un-retryably ("Upstream branch ... has changed").
+concurrency:
+  group: master-write
+  cancel-in-progress: false
+
 jobs:
   # Skip if commit is from semantic-release
   check-skip:
--- a/.github/workflows/hotfix-release.yml
+++ b/.github/workflows/hotfix-release.yml
@@ -16,6 +16,10 @@ on:
 concurrency:
   group: master-write
   cancel-in-progress: false
+  # queue: max keeps every queued run waiting FIFO. The default single-slot
+  # queue evicts the OLDER pending run when a new one queues, which could
+  # silently drop a queued release behind a bot push.
+  queue: max
 
 env:
   PYTHON_VERSION: "3.13"
@@ -129,7 +133,12 @@ jobs:
             exit 0
           fi
           echo "Push rejected on attempt ${attempt}; rebasing onto origin/master and retrying."
-          git pull --rebase origin master
+          # Abort a conflicted rebase so later attempts fail for the real reason.
+          git pull --rebase origin master || {
+            git rebase --abort 2>/dev/null || true
+            echo "::error::Rebase onto origin/master failed; aborting retries."
+            exit 1
+          }
         done
         echo "::error::Could not push addon changelog sync after 5 attempts."
         exit 1
--- a/.github/workflows/hotfix-release.yml
+++ b/.github/workflows/hotfix-release.yml
@@ -57,30 +57,23 @@ jobs:
         fetch-depth: 0
         token: ${{ secrets.GITHUB_TOKEN }}
 
-    - name: Validate hotfix is based on stable
+    - name: Require stable tag exists
       run: |
-        # Get the stable tag (moving tag that points to latest stable)
-        if ! git rev-parse stable &>/dev/null; then
-          echo "::error::No 'stable' tag found. Cannot validate hotfix base."
+        # Hotfix ORIGIN (branch based on the stable tag, not master) is validated
+        # PRE-merge in pr-validate-hotfix.yml, which checks out the unmerged
+        # branch head. Doing it here is impossible: post-merge HEAD is master's
+        # tip, and because master is never rewritten, `stable` is ALWAYS an
+        # ancestor of it — so `merge-base --is-ancestor stable HEAD` can never
+        # fail. The former ancestor check here was therefore dead code. The one
+        # thing still worth guarding is that `stable` exists at all: a hotfix
+        # must have been based on it, and the release below repoints it. (This
+        # missing-tag guard is not redundant with pr-validate-hotfix.yml, which
+        # only warns-and-skips when the tag is absent.)
+        if ! git rev-parse -q --verify refs/tags/stable >/dev/null; then
+          echo "::error::No 'stable' tag found — a hotfix must be based on the stable release."
           exit 1
         fi
-
-        STABLE_COMMIT=$(git rev-parse stable)
-        HOTFIX_BASE=$(git merge-base $STABLE_COMMIT HEAD)
-
-        # The hotfix should be based on or after the stable tag
-        if git merge-base --is-ancestor $STABLE_COMMIT HEAD; then
-          echo "✓ Hotfix is based on stable tag"
-        else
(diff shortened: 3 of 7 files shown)


Patch76 (changes requested):
Solid root-cause work – I confirmed the `Upstream branch 'origin/master' has changed!` abort in the raw log of run 28998970128, and the dead `version_variables` claim checks out (`src/ha_mcp/__init__.py:10` assigns the runtime lookup, so the literal-assignment pattern can never match). The shared concurrency group, the retry loops, and the mirror-tagging gap are all real problems worth fixing. Two concerns before this merges:

**[Concern 1 – queued releases can be silently evicted]** With the default concurrency queue, a group holds at most ONE pending run, and a newly queued run *cancels* the previously pending one ([docs](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#concurrency)). The PR body documents the tradeoff in the safe direction only ("can cancel one queued bot run"), but the eviction is group-wide and newest-wins: if a hotfix-release run is pending behind an in-progress master-writer and a tool-file merge then triggers sync-tool-docs, the *queued release* is the one cancelled – silently, recoverable only by a human noticing and re-running. The new comment in `sync-tool-docs.yml:23-25` ("cancelling a queued RELEASE … so it queues instead") states the opposite of the actual semantics. GitHub now supports `queue: max` (FIFO, up to 100 pending, only incompatible with `cancel-in-progress: true`, which you don't use) – adding it to all four `master-write` blocks (`addon-publish-dev.yml:26`, `hotfix-release.yml:17`, `semver-release.yml:27`, `sync-tool-docs.yml:27`) removes the eviction hazard entirely and makes the documented tradeoff moot.

**[Concern 2 – workflow_run leg fires on every merged PR]** `sync-integration-mirror.yml:36` triggers on `workflow_run: [completed]` of "Hotfix Release". Hotfix Release guards at the *job* level, so every non-hotfix merge still produces a completed run with `conclusion: skipped` (see the run list – 5 of the last 6 are `skipped`). Result: the mirror-sync `sync` job (deploy-key setup, mirror clone, snapshot stage) runs on effectively every PR merge; only the tag step is conclusion-guarded. On real hotfix merges you additionally get two near-simultaneous runs (push leg + workflow_run leg) pushing to the mirror with no concurrency group. Content stays correct (diff-quiet check + guarded tag), but it's a per-merge cost and a latent push race. A job-level guard fixes both legs: `if: github.event_name != 'workflow_run' || github.event.workflow_run.conclusion == 'success'` on `sync-integration-mirror.yml:41`.

Minor, non-blocking:
- The retry loops never check the `git pull --rebase` exit status (`_update-addon-config.yml:104`, `hotfix-release.yml:132`, `semver-release.yml:199`). A conflicted rebase leaves the repo in rebase-in-progress state, so the remaining attempts fail for a misleading reason. Unlikely given the serialized writers, but `git pull --rebase origin master || { git rebase --abort; echo "::error::rebase failed"; exit 1; }` is cheap. (I did verify the shallow-checkout case is fine: a depth-1 clone with one local commit rebases and pushes cleanly after upstream advances – tested locally.)
- `sync-tool-docs.yml:74` still says "cancel-in-progress guarantees no competing sync run" – the mechanism is now group serialization; worth updating while you're in the file.
- Pre-existing, not this PR: "Validate hotfix is based on stable" is vacuous on its first branch – `stable` is always an ancestor of the master lineage, so the check passed trivially with the pinned merge SHA too. The checkout change doesn't make it weaker; flagging it in case you want a follow-up issue.

Concern 1 is a one-line-per-file fix and Concern 2 a one-line guard, so this should be quick to turn around.

Patch76 (dismissed):
All three points addressed, and the fixes hold up under a second mechanism pass:

- `queue: max` landed on all four `master-write` blocks with an accurate comment on the eviction direction – the silent-drop path is gone.
- The `sync` job guard covers both problem legs (per-merge skipped runs and failed releases), and the new `mirror-sync` concurrency group is a good catch beyond what I asked for: I traced the two-leg hotfix scenario, and without it the second run's non-fast-forward push would fail the job before the tag step. With FIFO queueing the second run goes diff-quiet and still reaches the tag step.
- The rebase-abort guard is consistent across all four retry loops (including the pre-existing sync-tool-docs one), and the rewritten concurrency comments now match the actual semantics.

CI is green on 3a13d1d. Approving.

kingpanther13:
Thanks — both concerns confirmed and fixed, plus all three minors.

Concern 1 (queued release eviction): added `queue: max` to all four master-write blocks. Confirmed GA against the 2026-05-07 changelog and the workflow-syntax docs — up to 100 pending runs FIFO, valid because every block sets cancel-in-progress: false. A queued release can no longer be evicted by a later bot push. Also corrected the sync-tool-docs comment that had the eviction direction backwards.

Concern 2 (workflow_run fires on every merge): guarded the sync job with `github.event_name != 'workflow_run' || github.event.workflow_run.conclusion == 'success'`, so the clone/snapshot/push stages skip on the skipped Hotfix Release run every non-hotfix merge produces (also short-circuits the odd-week/no-change SemVer runs). Added a mirror-sync concurrency group (cancel-in-progress: false, queue: max) so the push leg and workflow_run leg of a real hotfix serialize instead of racing on a non-fast-forward push.

Minor (rebase exit status): all four loops now `git rebase --abort` + error + exit 1 on a failed rebase instead of leaving rebase-in-progress state.

Minor (stale sync-tool-docs comment): updated — the mechanism is the shared master-write group now.

Minor (vacuous "Validate hotfix is based on stable"): fixed in-PR. The ancestor check is unreachable — post-merge HEAD is master's tip and stable is always an ancestor of it, so `merge-base --is-ancestor stable HEAD` can never fail (git merge-base exit semantics; equally vacuous under the old merge_commit_sha checkout). Deleting the whole step would silently drop its one reachable behavior, the missing-stable-tag hard-fail — and pr-validate-hotfix.yml does NOT cover that case (it warns-and-skips). So I replaced the dead ancestor logic with a one-line stable-tag existence guard and kept the hard-fail. Origin validation stays pre-merge in pr-validate-hotfix.yml.

Commits: 61b944e1, 3a13d1dd, 528ce95a.

kingpanther13:
@Patch76 sorry for noise, it was doing the crap where it would push a commit then ask me if I wanted it to fix the other thing -_- . Please rereview when able

kingpanther13:
@Patch76 and if you can please, when you do rereview ask it to verify GitHub documentation to make sure we're doing things right, just to be 100% sure.

Patch76 (approved):
Re-review of 528ce95a.

**The new commit** – replacing the post-merge ancestor check with a stable-tag existence guard is the right call, and I verified both claims in its comment against `pr-validate-hotfix.yml` at this PR's head: that workflow checks out the unmerged branch head (`ref: github.event.pull_request.head.sha`), so origin validation genuinely happens pre-merge, and it only warns-and-skips when the `stable` tag is absent, so the hard-fail guard here is not redundant. Two side benefits: `git rev-parse -q --verify refs/tags/stable` is stricter than the old bare `git rev-parse stable` (matches the tag only, cannot be shadowed by a branch of the same name), and `fetch-depth: 0` fetches "all history for all branches and tags" (actions/checkout docs), so the guard actually sees the tag – which exists in this repo today.

**Documentation check, as requested** – I re-verified all three mechanisms against the current GitHub docs:

1. Concurrency (workflow-syntax reference): the default is `queue: single` – "any existing pending job or workflow run in the same group is canceled and replaced" – which is exactly the eviction hazard this PR closes. `queue: max` allows "up to 100 jobs or workflow runs" to be pending, processed FIFO. The docs also state that combining `queue: max` with `cancel-in-progress: true` fails workflow validation; all five concurrency blocks in this PR set `cancel-in-progress: false`, so they validate.
2. `workflow_run` (events reference): "a workflow run is triggered regardless of the conclusion of the previous workflow", and the docs' own recommended pattern is precisely the guard used here, `github.event.workflow_run.conclusion == 'success'`.
3. I confirmed all five concurrency blocks are intact at 528ce95a (`queue: max` + `cancel-in-progress: false` in each: the four master-write groups plus mirror-sync).

CI is green across the board; the Validate Hotfix Origin skip is expected since this is not a hotfix branch. Approving again.

Patch76:
Wait, I do another review, a specific GitHub-one.

Patch76:
Follow-up on the "verify against GitHub documentation" ask – I ran a dedicated documentation pass over every Actions mechanism this PR relies on, plus a repo-wide cross-check. Everything holds; receipts below.

**Concurrency**

- `queue: max` is GA per the 2026-05-07 changelog: "up to 100 queued jobs or workflow runs per concurrency group", allowed only when `cancel-in-progress` is `false` or unset. All five blocks in this PR set `cancel-in-progress: false`, so they validate.
- The default queue is documented as: "any existing pending job or workflow run in the same group is canceled and replaced" – the eviction hazard this PR closes is real and documented, in that direction.
- A concurrency group name is shared repo-wide across workflow files (the docs warn that reusing a name affects runs "regardless of the workflow") – exactly the cross-workflow serialization `master-write` uses deliberately, and with `cancel-in-progress: false` the cancellation hazard that warning is about is switched off.
- One nuance for completeness: the reference says queued runs are processed FIFO by wait-start time but adds "ordering is not guaranteed". The property the fix depends on is no-eviction, not strict ordering – the rebase/retry loops absorb any reordering – so this changes nothing.
- `_update-addon-config.yml` has no concurrency block of its own and needs none: "the entire called workflow is used, just as if it was part of the caller workflow", and all three callers hold `master-write` at workflow level, which scopes the whole run including called jobs.

**workflow_run (mirror sync)**

- "A workflow run is triggered regardless of the conclusion of the previous workflow" – so the skipped Hotfix Release runs really do fire the event, and `conclusion == 'success'` is literally the documented guard pattern for it.
- The triggered run executes against the default branch (GITHUB_SHA = "last commit on default branch"), so the tag step reads the manifest version from master's tip – which is what it should do.
- `workflows: ["SemVer Release", "Hotfix Release"]` matches by workflow name; both `name:` fields match exactly.
- workflow_run only fires for workflow files on the default branch, i.e. the new guard and `mirror-sync` group go live on merge – expected, nothing to do.

**Token / trigger propagation**

- "Events triggered by the GITHUB_TOKEN will not create a new workflow run" – hotfix-release's changelog push, stable-tag push and semantic-release push all use GITHUB_TOKEN, so none of them re-trigger push workflows.
- App-token pushes DO trigger workflows: semver-release's version-bump commit (app token, touches `pyproject.toml`) triggers addon-publish-dev. Its check-skip job catches it – semantic-release's default commit message starts with the bare version, matching `^[0-9]+\.[0-9]+\.[0-9]+` (no `commit_message` override in pyproject) – and with `queue: max` that queued run can no longer evict a pending release while it waits.
- The `[skip ci]` on the `_update-addon-config` bump commit is load-bearing: that push is app-token-authenticated, `homeassistant-addon/**` is in addon-publish-dev's path filter, and check-skip's patterns do NOT match `chore(addon): publish …` – the documented push/pull_request skip marker is what breaks that loop. It is present.

**Repo-wide completeness**

- I swept all 36 active workflow files at this PR's head: the four `master-write` members are exactly the set of workflows that push to this repo's master. The webhook-proxy workflows push feature branches, publish-dev only deletes dev tags, and the mirror sync pushes to the mirror repository under its own `mirror-sync` group.

No changes needed – my approval stands.

PR #1807, opened 2026-07-11
Fix nested helper flow options

loganrosen:
## What does this PR do?

Home Assistant flow schemas can place fields inside expandable sections. `ha_config_set_helper` treated those child fields as top-level form values, so updates such as a template sensor's `availability` were either silently dropped or rejected by Home Assistant. The options reader also skipped nested values, making persisted advanced settings invisible.

This recursively handles nested schema sections, wraps flat MCP inputs in the section shape Home Assistant expects, and flattens nested values when reading options. It includes unit coverage for flat and explicitly nested inputs plus an end-to-end template availability update and readback.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

## Checklist

### DIFF
--- a/src/ha_mcp/tools/config_entry_flow.py
+++ b/src/ha_mcp/tools/config_entry_flow.py
@@ -130,7 +130,7 @@ def _handle_menu_step(
 
 
 def iter_schema_fields(data_schema: Any) -> Iterator[dict[str, Any]]:
-    """Yield submitted fields from a flow schema, including nested sections."""
+    """Yield leaf field definitions from a flow schema, descending into sections."""
     if not isinstance(data_schema, list):
         return
     for field in data_schema:
@@ -143,10 +143,41 @@ def iter_schema_fields(data_schema: Any) -> Iterator[dict[str, Any]]:
         yield field
 
 
+def _section_path(path_prefix: str, name: Any) -> str:
+    """Return the dotted config path for a named or anonymous section."""
+    if not isinstance(name, str):
+        return path_prefix
+    return f"{path_prefix}.{name}" if path_prefix else name
+
+
+def _record_ignored_section_keys(
+    ignored_config_keys: set[str] | None,
+    remaining_config: dict[str, Any],
+    section_path: str,
+) -> None:
+    """Record undeclared keys remaining inside an explicit section dict."""
+    if ignored_config_keys is None:
+        return
+    ignored_config_keys.update(
+        f"{section_path}.{key}" if section_path else key
+        for key in remaining_config
+        if key not in _MENU_SELECTION_KEYS
+    )
+
+
 def _consume_form_schema(
-    data_schema: list[Any], remaining_config: dict[str, Any]
+    data_schema: list[Any],
+    remaining_config: dict[str, Any],
+    ignored_config_keys: set[str] | None = None,
+    path_prefix: str = "",
 ) -> dict[str, Any]:
-    """Consume matching config values and shape nested flow sections."""
+    """Consume matching config values and shape nested flow sections.
+
+    Mutates ``remaining_config`` by removing every consumed key. Flat child
+    values override the same value inside an explicitly supplied section dict.
+    Unknown keys inside explicit section dicts are added to
+    ``ignored_config_keys`` with their dotted section path.
+    """
     form_data: dict[str, Any] = {}
     missing = object()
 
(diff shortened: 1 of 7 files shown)


kingpanther13 (changes requested):
Thanks for this PR, and welcome to the project — this is a great first contribution. The root-cause analysis is exactly right, the recursive section handling reads cleanly, and the happy-path unit tests plus the e2e round-trip (which genuinely fail without the fix) made this easy to review. I'm requesting changes on three scoped items, none of which change your approach:

**1. Surface ignored config keys instead of dropping them silently**

The pre-filtering means a key that matches no schema field is dropped without a trace while the tool reports success — e.g. `config={"state": "{{1}}", "availabilty": "..."}` (typo) returns `"template helper updated successfully"` with availability never set. `_handle_flow_steps` never inspects `remaining_config` when it hits `CREATE_ENTRY`, and the new explicit-section path in `_consume_form_schema` recurses over a throwaway copy (`dict(explicit_section)`), so a mistyped key inside a section dict can't even be recovered by a terminal check. Since this PR is all about silently dropped options, let's close that last gap in the same change: collect unmatched keys (terminal `remaining_config` minus menu-selection keys, plus leftovers from the explicit-section copy) and surface them as a top-level `warnings: list[str]`. There's an existing pattern to follow: `_inject_or_strip_name` in `tools_config_helpers.py` builds `pre_warnings` and `_handle_flow_helper` threads them into the response.

**2. Tests for the remaining new branches in `_consume_form_schema`**

- Update of only a top-level field when the schema declares a section: assert the section is omitted from the payload entirely. This is the most common real-world path and no current test exercises it.
- Same field supplied both flat and inside an explicit section dict: today the flat value wins (the explicit dict is applied first, then `remaining_config` on top). Please pin the intended precedence with a test — and document it (see item 3) — so a future refactor can't silently invert it.
- Depth-2 nesting (section inside section) on both the write path and the `options_from_form_flow` read path — the code recurses, but every test stops at one level.
- The legacy fallback (`data_schema` not a list): this PR rewrote its guard from `schema_fields is None` to the isinstance check. They're equivalent, but a test would pin that.
- Cheap same-class additions while you're in there: non-dict explicit section value passed through as-is, and a section with a missing name flattening its children into the parent.

One heads-up: `test_accepts_explicit_expandable_section_dict` also passes on master (the section's own name made it a top-level schema field there), so it's a characterization test rather than a regression test — fine to keep, just mentioning it so coverage doesn't lean on it.

**3. Docstring accuracy**

- `iter_schema_fields`: "Yield submitted fields" — it yields schema field *definitions*, and one caller uses it for reading values back. Suggest: "Yield leaf field definitions from a flow schema, descending into nested sections."
- `_handle_form_step`: still describes a flat pop-only projection; add a sentence that fields declared inside a section are grouped under the section key, and an explicitly supplied section dict is accepted.
- `options_from_form_flow`: mention that nested section fields are flattened into the top-level map.
- `_extract_schema_field_names`: note that nested leaf names are included and section-container names are dropped.
- `_consume_form_schema`: document the mutation of `remaining_config` and the flat-over-explicit merge precedence once it's pinned.

Happy to clarify any of these or point at the relevant code. Thanks again for digging into this — it fixes a real pain point with template helper advanced options.

loganrosen:
Addressed the requested changes in 5c139d73:

- Unknown config keys now produce top-level warnings. Top-level keys remain available across multi-step flows and are reported only when the flow completes; unknown keys inside explicit sections use dotted paths.
- Added coverage for section omission, flat-value precedence, depth-two recursion, legacy fallback, malformed section values, anonymous sections, nested readback, and warning propagation through the real helper workflow.
- Corrected the schema traversal and flow-handling docstrings.

The full E2E suite passes: 1017 passed, 73 skipped.

kingpanther13 (approved):
Thanks for the fast and thorough turnaround — approving.

All three review items are addressed:

- Unknown config keys now surface as top-level warnings, including dotted paths for keys inside explicit section dicts, threaded through both create and update responses — and pinned by the new warning-propagation and response-shape tests.
- Every branch called out now has a test: section omission on top-level-only updates, flat-over-explicit precedence, depth-two nesting on both the write and read paths, the legacy fallback, non-dict section passthrough, and anonymous sections.
- The docstrings now describe the actual contracts, including the merge precedence.

Nice work — this fixes a real pain point and leaves the flow-walking code better documented than it started.

Issue #1794, opened 2026-07-10
ha_config_list_helpers returns stale unique_id/original_name after a UI rename, not current entity_id/name


hvrietsc:
## Bug: `ha_config_list_helpers` returns stale `id`/`name` instead of current `entity_id`/`name` after a UI rename

**ha-mcp version:** 7.12.0 (pypi install)

**Summary:**
`ha_config_list_helpers(helper_type="input_boolean")` returns each helper's original creation-time `unique_id` (as `id`) and `original_name` (as `name`), not the entity's current `entity_id` suffix or display name after it's been renamed via the HA UI. `ha_get_entity()` and `ha_get_state()` correctly reflect the current values, so the two tools disagree with each other for any helper that's ever been renamed.

**Repro:**
1. Create an `input_boolean` helper, e.g. named "Dark Enough" (entity_id `input_boolean.dark_enough`).
2. Rename it via the HA UI to "Dark Enough Mode" — HA updates the entity registry `name` and `entity_id` to `input_boolean.dark_enough_mode`, but the storage-level `unique_id` stays `dark_enough` (this is normal HA behavior — `unique_id` is immutable).
3. Call `ha_config_list_helpers(helper_type="input_boolean")`.
4. Call `ha_get_entity(entity_id="input_boolean.dark_enough_mode")`.

**Expected:** Both tools reflect the current entity_id/name (`dark_enough_mode` / "Dark Enough Mode").

**Actual:**
- `ha_config_list_helpers` returns `{"id": "dark_enough", "name": "Dark Enough"}` — stale.
- `ha_get_entity` returns `{"entity_id": "input_boolean.dark_enough_mode", "name": "Dark Enough Mode", "original_name": "Dark Enough", "unique_id": "dark_enough"}` — current, and correctly labels the stale value as `original_name`/`unique_id`.

Same reproduces for `input_boolean.we_have_guests_mode` (unique_id `we_have_guest_mode`, original_name "We Have Guest Mode" — note it's even a different word: "guest" vs "guests").

**Impact:** Anyone using `ha_config_list_helpers` to enumerate helpers and then act on the returned `id` (e.g. pass it to `ha_config_set_helper`, or just read it to find an entity_id) will silently get the wrong/stale identifier for any renamed helper. This is especially confusing for agentic use — an LLM reading this list has no signal that the data is stale, and may report entities as "missing" or attempt to reference an entity_id that no longer exists.

[...]

Issue #1590, opened 2026-06-12
fix: convert UTC timestamps to HA local timezone in add_timezone_metadata


rebelancap:
## Problem

`add_timezone_metadata` in `tools/util_helpers.py` correctly fetches the Home Assistant timezone from `/api/config` but never actually converts timestamps — it only adds a metadata note saying "All timestamps are in UTC."

This means LLMs receive raw UTC timestamps and are expected to convert them themselves. In practice, models like Gemini fail to do this reliably, reporting times like "12:06 AM" when the actual local time was "6:06 PM".

## Fix

Walk the response data recursively and convert all `last_changed`, `last_updated`, and `last_reported` fields from UTC to the HA-configured local timezone using Python's built-in `zoneinfo` (available since 3.9, already required by this project).

The timezone is read dynamically from HA's own `/api/config` response, so it works for any user in any timezone and handles DST automatically.

## Proposed change to `add_timezone_metadata` in `tools/util_helpers.py`

Replace the function body with:

```python
async def add_timezone_metadata(
    client: Any, data: dict[str, Any], include_metadata: bool = True
) -> dict[str, Any]:
    """Add Home Assistant timezone to tool responses and convert timestamps to local time."""
    if not include_metadata:
        return data

    ha_timezone = "UTC"
    try:
        config = await client.get_config()
        ha_timezone = config.get("time_zone", "UTC")
    except (
        HomeAssistantConnectionError,
        HomeAssistantAPIError,
        HomeAssistantAuthError,
        TimeoutError,
        OSError,
    ) as _tz_exc:
        logger.warning(
            "add_timezone_metadata: failed to fetch HA timezone config — "
            "falling back to UTC: %s",
            _tz_exc,
            exc_info=True,
        )

    from zoneinfo import ZoneInfo
    from datetime import datetime as _dt

    local_tz = ZoneInfo(ha_timezone)
    ts_fields = {"last_changed", "last_updated", "last_reported"}

    def _convert(obj: Any) -> Any:
        if isinstance(obj, list):
            return [_convert(i) for i in obj]
        if isinstance(obj, dict):
            out = {}
            for k, v in obj.items():
                if k in ts_fields and isinstance(v, str) and v:
                    try:
                        parsed = _dt.fromisoformat(v)
                        out[k] = parsed.astimezone(local_tz).isoformat()
                    except (ValueError, TypeError):
                        out[k] = v
                else:
                    out[k] = _convert(v)
            return out
        return obj

[...]

PR #1812, opened 2026-07-11
feat: ha_search optimization + component-backed read API

kingpanther13:
## What does this PR do?

Reworks the search stack for cost and capability, in phased commits:

- [x] **Request cuts** (`6046ecb4`): a verified audit counted 16 HA round-trips on a plain `ha_search` call. Now: `/states` and the entity registry are fetched once and shared across both branches, the discarded timezone fetch is removed from the 5 entity-search builders (the #1590/#1592 UTC→local conversion is untouched where records carry timestamps), `domain_filter` applies before alias hydration, and the device registry is fetched only when the opt-in visibility config uses its area/label dimensions. Request-count regression tests pin the new inventory. All output-identical.
- [x] **Catalog cost** (`046ff4d3`): ha_search was the largest pinned tool description (4,721 chars) with no lite variant — trimmed to 2,525 chars and added to `_LITE_DOCSTRINGS`, preserving the partial-results honesty caveats and the test-pinned `fields` key manifest.
- [x] **Deep-search helper names** (`a63af1ea`): score helpers by current entity-registry names (UI-renamed helpers currently mismatch; same root cause as #1794, whose `ha_config_list_helpers` half is handed off separately).
- [x] **Component read API v1** (`e41a920c` + `c10f5811`): the custom component (→ 1.1.0) registers versioned `ha_mcp_tools/*` WS commands doing in-process joins (states + registries + `ConfigEntry.options` + loaded YAML configs); the server routes to them via the existing version handshake with graceful fallback everywhere — consumers landed:
  - [x] `ha_search` (query-driven, non-area, visibility-filter-off installs; all-or-nothing with graceful fallback; secret values scrubbed from the YAML match corpus)
  - ~~`ha_config_get_*`~~ — capability WITHDRAWN after live e2e: `raw_config` is reload-fresh, not file-fresh, so get-after-set raced HA's async reload and returned stale bodies. All config gets are legacy-served; a file-reading design may revive it (#1813)
  - [x] `ha_get_overview` (~8 round-trips → 1 via raw slices; server assembly unchanged)
  - [x] `ha_config_list_helpers` (current registry names additively — the #1794 fix; flow-helper types now listable on component installs, `COMPONENT_NOT_INSTALLED` error otherwise; `tag` falls back to legacy via the component's `covered_types` signal)

Cross-seam contract tests pin every component↔server seam (real component output through the real tools). Rebased over #1815 (component manifest resolved to 1.1.0).

Stacked on #1810 — its commits appear here until it merges.

## Why this PR is ~10.4k lines — review guide

The number overstates the product-code surface; here's the honest breakdown:

| Slice | Lines | What it is |
|---|---|---|
| Tests | **~6.0k (58%)** | See below — the parity evidence |
| Stacked #1810 | ~1.1k | Not this PR's; vanishes when #1810 merges |
| Custom component | ~1.85k | `websocket_api.py`: five WS commands, in-process joins, server-parity scorer, secret scrub |
| Net server code | ~2.1k | Request dedup, catalog trim, caps gate, four thin consumers reusing legacy assembly |

**Why tests dominate:** this PR reroutes core read tools through a new in-process backend while guaranteeing byte-level response parity with the legacy paths. Every seam carries three layers: per-side unit suites, **cross-seam contract tests that drive the REAL component functions through the REAL server tools** (these caught three contract drifts before CI), and behavioral pins (request-count inventories, scorer golden parity, secret-scrub negatives). Each live e2e failure found during development also got a permanent unit mirror. The contract tests double as the specification of what's guaranteed — reading them is the fastest way to review the seams.

**Suggested review order:** `websocket_api.py` module docstring (the contract) → `component_api.py` (the gate) → the `ha_search` consumer in `tools_search.py` → `test_component_search_contract.py` / `test_component_readapi_contract.py` → remaining consumers.

## Type of change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 📚 Documentation
- [x] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

## Future improvements

- `ha_config_list_helpers` registry join for stale ids after UI rename (#1794) — maintainer-confirmed handoff to a separate PR; fix sketch posted on the issue.

## Checklist

### DIFF
--- a/custom_components/ha_mcp_tools/__init__.py
+++ b/custom_components/ha_mcp_tools/__init__.py
@@ -55,6 +55,7 @@ from .const import (
     YAML_KEY_DEFAULT_POST_ACTION,
     YAML_KEY_POST_ACTIONS,
 )
+from .websocket_api import async_register_commands
 from .yaml_rt import apply_seq_indent, detect_seq_indent, make_yaml, yaml_dumps
 
 _LOGGER = logging.getLogger(__name__)
@@ -2480,6 +2481,12 @@ async def _async_setup_tools_entry(hass: HomeAssistant, entry: ConfigEntry) -> b
         schema=SERVICE_READ_LEGACY_BACKUP_SCHEMA,
         supports_response=SupportsResponse.ONLY,
     )
+
+    # Register the in-process ha_mcp_tools/* WebSocket commands (info + search)
+    # the server calls behind a capability gate. Idempotent, and independent of
+    # the caller-token gate above — HA core authenticates the WS connection and
+    # @require_admin gates each command (see websocket_api.py).
+    async_register_commands(hass)
 
     _LOGGER.info("HA MCP Tools initialized with file management services")
     return True
--- a/custom_components/ha_mcp_tools/const.py
+++ b/custom_components/ha_mcp_tools/const.py
@@ -18,6 +18,14 @@ from datetime import timedelta
 
 DOMAIN = "ha_mcp_tools"
 
+# Component version, kept in lockstep with ``manifest.json``'s ``version``.
+# ``ha_mcp_tools/info`` reports this so the server can display/debug the running
+# component build; ``TestManifestVersionParity`` pins the two together so a
+# manifest bump that forgets this constant (or vice-versa) fails in CI. The
+# capability negotiation — not this version — gates each WS command (see
+# ``websocket_api.CAPABILITIES``).
+COMPONENT_VERSION = "1.1.0"
+
 # Config-entry discriminator (``entry.data[CONF_ENTRY_TYPE]``). A missing value
 # means "tools" so the pre-existing services entry keeps working across the
 # component update with no migration.
--- a/custom_components/ha_mcp_tools/manifest.json
+++ b/custom_components/ha_mcp_tools/manifest.json
@@ -19,5 +19,5 @@
     "requirements": [
         "ruamel.yaml>=0.18.0"
     ],
-    "version": "1.0.6"
+    "version": "1.1.0"
 }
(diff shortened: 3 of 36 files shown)


Patch76 (approved):
Reviewed the full diff at `93f21f3b`, with attention to the component read API — Gemini's single auto-review predates it (its summary covers the #1810 timeout content, not the component seam), so that path had no bot review.

**What I verified**

- The request-cut refactor is output-identical: the shared `/states` + entity-registry snapshot is read-only across both branches, `domain_filter` is an `entity_id`-prefix test that aliases never fed, and the removed timezone fetch only touched entity records that carry no timestamps (the #1590/#1592 conversion paths are untouched).
- Routing and fallback hold: component errors surface via `warnings[]` plus logs, connection errors propagate, `unknown_command` invalidates caps before falling back, and the `covered_types` signal keeps `tag` and other non-enumerable types on the legacy path instead of returning a false-empty. Flow helpers with no component hard-error rather than silently emptying.
- `websocket_api.py` event-loop hygiene is clean: the only blocking read (`secrets.yaml`) goes through `async_add_executor_job`; every registry/state join is an in-memory read.
- The cross-seam contract tests drive the real component functions through the real tools rather than mocking the seam they claim to verify. The #1794 rename fix and the person/nested-helper flatten are pinned by assertions that actually exercise the registry-override and dict-shape paths.

**One non-blocking note — reconcile the flow-helper secret-handling docs**

The module docstring (`websocket_api.py:49`) states flow-helper `ConfigEntry.options` "can hold `!secret` values resolved to plaintext," and `ha_mcp_tools/search` scrubs them accordingly (`secret_values` threaded at `:1095`). The `helpers_list` docstring (`:1519-1521`) states the opposite — "flow options are storage-backed ... so no resolved `!secret` plaintext can appear" — and returns `options` unscrubbed (`:1628`). Both commands read the same data.

My reading is that `:1521` is the accurate one: config-entry options come from the UI flow and `.storage`, never a YAML `!secret` resolution, which makes the search-side scrub harmless defense-in-depth and the parenthetical at `:49` an overstatement. Since these are the security-reasoning comments a later change will lean on, worth making the two agree and confirming that is the intended threat model. No leak either way — `helpers_list` is admin-gated and returns the same option data the Home Assistant UI shows.

The rest is clean and CI is green across the E2E lanes.

PR #1830, opened 2026-07-12
ci: rate limit Codex review requests

julienld:
## What changed

- Split Codex review admission from delivery into two explicitly documented workflows.
- Added a repository-wide rolling limit of 20 delivered Codex requests per hour.
- Serialized admission decisions globally to prevent parallel PR events from racing the limit.
- Kept PR authorization and same-commit deduplication in the admission workflow.

## Why two workflows

The original workflow runs for every matching GitHub event, including events rejected by its job condition, so its run count cannot represent actual Codex requests. The delivery workflow is triggered only after admission succeeds; therefore each delivery run is exactly one request and its GitHub Actions history is a durable, auditable global rate-limit ledger.

## Security and permissions

Admission uses the repository `GITHUB_TOKEN` with Actions write access to inspect and dispatch delivery runs. The separate delivery workflow alone receives `GH_TOKEN_CODEX_COMMENT`, keeping the comment-writing credential out of admission.

## Validation

- `git diff --check`
- Parsed both workflows with PyYAML.

### DIFF
--- a/.github/workflows/pr-codex-review-delivery.yml
+++ b/.github/workflows/pr-codex-review-delivery.yml
@@ -1,0 +1,42 @@
+name: Send Codex PR Review
+run-name: Send Codex PR review for #${{ inputs.issue_number }}
+
+# This workflow is deliberately separate from admission. One run equals one
+# actual Codex request, so its run history is the global rate-limit ledger.
+on:
+  workflow_dispatch:
+    inputs:
+      issue_number:
+        description: Pull request number to review
+        required: true
+        type: string
+
+permissions:
+  contents: read
+
+jobs:
+  send-codex-review:
+    name: Comment @codex review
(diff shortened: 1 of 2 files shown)


ghhamcp:
@codex review

ghhamcp:
@codex review

julienld:
## Implementation Summary

- Split admission from delivery so every delivery workflow run is exactly one Codex request.
- Added a global rolling cap of 20 delivery runs per hour and serialized admission checks to prevent races.
- Preserved trusted-commenter validation and same-head-SHA deduplication.
- Kept `GH_TOKEN_CODEX_COMMENT` isolated to delivery; admission uses the scoped repository `GITHUB_TOKEN`.
- Validated both workflow files with PyYAML and `git diff --check`.
- Squash-merged with administrator override as requested.

PR #1859, opened 2026-07-13
fix: correct Tools-tab group master + LLM API toggle rendering

kingpanther13:
## What does this PR do?

Two Tools-tab (settings UI) toggle-rendering fixes:

**1. Group master toggle for all-mandatory groups.** The per-group master switch derived its checked state from `anyEnabled`, computed only over *toggleable* tools. For a group whose tools are all mandatory — e.g. Search & Discovery (`ha_search`, `ha_get_overview`, `ha_get_state`) — that set is empty, so the switch rendered unchecked+disabled: it read as "the group is off" while the count beside it said "3/3 enabled" and the tools can't be disabled at all. It now takes the checked state from the group's real enabled status when nothing is toggleable, so an all-mandatory group shows on+disabled. Interactive/mixed groups are unchanged.

**2. LLM API toggle only on the custom-component (embedded) server.** The per-tool "LLM API" toggle offers a tool to Home Assistant conversation agents via the LLM API surface the `ha_mcp_tools` custom component registers. That surface only exists on the in-process (embedded) server; on the add-on / Docker / standalone server nothing consumes it, so the toggle was a no-op yet still rendered for every tool. `/api/settings/tools` now reports `llm_api_available` (= `is_embedded()`), and the Tools tab renders the LLM API column only when it is true.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

Added JSDOM regression tests for both fixes (group master: all-mandatory / all-disabled / mixed; LLM API column hidden when the flag is absent) plus a server-side test that `llm_api_available` is true under `HA_MCP_EMBEDDED`. Verified against the jsdom harness that the all-mandatory master is unchecked before the change and checked+disabled after, and that the LLM API column renders only when embedded. Full unit + E2E suites run in CI.

## Checklist

### DIFF
--- a/src/ha_mcp/settings_ui/__init__.py
+++ b/src/ha_mcp/settings_ui/__init__.py
@@ -1446,6 +1446,12 @@ def build_settings_handlers(
                 "read_only_exempt": sorted(READ_ONLY_EXEMPT_TOOLS),
                 "llm_api": llm_effective,
                 "llm_api_overrides": llm_overrides,
+                # LLM API exposure is registered by the ha_mcp_tools custom
+                # component (in-process/embedded server). On the add-on,
+                # Docker, or standalone server nothing consumes it, so the UI
+                # hides the per-tool "LLM API" toggle rather than showing a
+                # no-op control.
+                "llm_api_available": is_embedded(),
             }
         )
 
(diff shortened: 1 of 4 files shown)


ghhamcp:
@codex review

PR #1885, opened 2026-07-15
feat: add limit/offset pagination to the three unpaginated list tools

### LINKED ISSUE #1869: Several list tools (ha_config_list_helpers, ha_config_list_groups, ha_config_list_dashboard_resources) have no pagination
## Problem

`ha_get_history` and `ha_get_logs` correctly implement `limit`/`offset` with `has_more`, but several other list-returning tools don't support any pagination and load everything at once:

- `ha_config_list_helpers(helper_type)` (`src/ha_mcp/tools/tools_config_helpers.py`, around line 3617) — no `limit`/`offset` parameter.
- `ha_config_list_groups()` (`src/ha_mcp/tools/tools_groups.py`, around line 145) — takes no parameters at all.
- `ha_config_list_dashboard_resources(include_content)` (`src/ha_mcp/tools/tools_resources.py`, around line 155) — no pagination either.

On instances with many helpers/groups/dashboard resources this returns an unbounded response, which is worse for context usage in MCP clients and doesn't match the pagination pattern already established elsewhere in the codebase (`ha_get_history`/`ha_get_logs`).

## Proposed fix

Add `limit`/`offset` (or cursor-based pagination) to these three tools, following the same pattern already used in `ha_get_history`/`ha_get_logs` (`has_more`, `next_offset`, `total_count`).

## Environment

- ha-mcp (Python package, master branch)

Patch76:
Closes #1869

## What does this PR do?

`ha_config_list_groups`, `ha_config_list_dashboard_resources` and `ha_config_list_helpers` returned their whole collection on every call, so a large install pushed an unbounded response into the MCP client's context.

All three now take `limit`/`offset` and emit the shared `build_pagination_metadata` envelope (`util_helpers.py:448`), joining `ha_get_history`, `ha_list_services`, `ha_search`, the device listing and `ha_get_integration`. Default `limit` is 100, matching `_DEFAULT_HISTORY_LIMIT`; the ceiling is `le=500`, the `max_results` cap the `ha_mcp_tools` component already advertises, applied uniformly across the three. `count` becomes the page size and `total_count` carries the full size – below the default limit the record lists are unchanged and only the metadata keys are added.

The slice is client-side for all three. Groups and dashboard resources sit on Home Assistant core commands that take no pagination parameters, so the full list is already in memory. Helpers route through four separate envelope builders, so the slice is applied at one normalization point rather than copied into each.

This drops the `helpers_list` schema parity proposed in my analysis comment on #1869 – see the Implementation Summary for why it does not hold.

Also corrects three stale `_LITE_DOCSTRINGS` entries found while tracing this (all 13 audited; the other ten are accurate):

- `ha_config_set_script` and `ha_config_set_scene` named an `identifier` parameter and told the model to omit it to create a new object. Both take a required `script_id` / `scene_id` instead, so the lite description produced an invalid call.
- `ha_config_list_helpers` claimed the 12 storage-backed types were the only accepted values and that flow-based helpers could not be listed – both untrue since #1812.

## Type of change
- [x] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

The unit suite is green locally (6812 passed, 15 new). The E2E suite needs Docker, which this checkout cannot reach, so it runs in CI.

## Checklist


ghhamcp:
@codex review

ghhamcp:
@codex review

kingpanther13 (changes requested):
Thanks — the core of this is solid. All three tools now take limit/offset with the shared build_pagination_metadata envelope, and below the default limit the record lists are unchanged with only metadata keys added, which is exactly the constraint from my comment on #1869. Dropping the component-side helpers_list pagination is justified: all-types mode merges legacy records after the component call, so only a post-merge slice can paginate every route correctly. The unit suite is strong.

Requesting changes on the following:

1. **`_legacy_helper_list` feeds the pagination guard a non-list for `person`** (`tools_config_helpers.py:3929`). The inline legacy body flattens the `person/list` split shape via `_flatten_helper_list_result` (`:3787`); the component-error fallback copy does not. On that path `helpers` is a dict, so `_paginate_helpers_response`'s isinstance guard silently returns the envelope unchanged — no pagination metadata, `count` of 2 (the dict keys), malformed `helpers`. The same unflattened shape reaches the all-types merge (`:4052-4057`), where iterating the dict's keys fails `isinstance(item, dict)` and silently drops every person record from `helper_type="all"`. The missing flatten predates this PR, but the new guard is what makes it silent, and it breaks the "single normalization point" contract this PR introduces. Fix: flatten in `_legacy_helper_list` like the inline body, make the guard log rather than silently no-op, and add a regression test driving `person` through the component-error fallback (the suite only exercises that fallback with flat-shape input today).

2. **E2E coverage.** No e2e test exercises `limit`/`offset` on any of the three tools, and redefining `count` as page size silently weakened the assertions at `tests/src/e2e/workflows/dashboards/test_resources.py:74` and `:123` — they now hold only while the test instance stays under 100 resources. Migrate those to `total_count`, and add e2e coverage of the pagination contract through the MCP layer (`tests/src/e2e/tools/test_logbook.py` is the existing precedent — it e2e-tests limit bounds including rejection of invalid values). A small-limit boundary test (e.g. create 3 groups, `limit=2` → `has_more`/`next_offset`, then `offset=2` → final record) covers the contract cheaply; I'm also open to populating the test environment past 100 helpers/resources so the default-limit trigger is exercised for real — say the word and we can settle the shape in this thread.

3. **Stale NOTE in the `ha_config_list_helpers` docstring** (`tools_config_helpers.py:3730`). "This only returns storage-based helpers (created via UI/API)" contradicts the flow-based support described a few lines below it — and this PR's own lite entry, which contrasts storage-backed with flow-based types. Reconcile the NOTE; the "not YAML-defined" half is still correct.

4. **Test module docstring undercounts its own suite** (`tests/src/unit/test_list_tools_pagination.py:14`). "covers both routes" — the suite pins three envelope builders (legacy body, single-type component via `_shape_component_helpers_response`, merged all-types). Looks like it predates the commit that added the single-type component tests.

5. **Non-deterministic page boundaries for groups** (`tools_groups.py:209-214`). The sort key `(friendly_name or entity_id).lower()` has no unique tiebreaker, so two groups sharing a case-folded name keep whatever order `get_states()` returns, and a group can be duplicated or skipped across pages if that order shifts between calls. Add `entity_id` as a secondary sort key.

6. **The PR description points at an Implementation Summary for the helpers_list-parity rationale, but no such comment exists on the PR.** Please post it (the commit message of 9021c616 already carries the substance) or drop the reference.

Patch76:
Thanks – all six are addressed. Details where my findings differ from or extend yours:

**1. `_legacy_helper_list` feeds the pagination guard a non-list for `person`** – fixed in 47a62cc4. Flattened via `_flatten_helper_list_result` like the inline body, and the guard now logs rather than silently no-opping.

One correction to the mechanism: the all-types merge does not fail loudly. Iterating the unflattened dict yields its *keys* as strings, and the `isinstance(item, dict)` check at `tools_config_helpers.py:4073` filters them out, so every person record was dropped with no error rather than raising. Quieter than either of us described, and it makes the "single normalization point" contract the more useful frame – you were right that the new guard is what turned a shape bug into a silent one.

The regression test drives `person` through the component-error fallback. I verified it fails without the flatten rather than trusting that it would.

**2. E2E coverage** – added in 197c7f13, and both assertions you flagged in `test_resources.py` now use `total_count`. I swept the two other e2e modules that call the three tools: the remaining `count` uses are log lines, presence checks, or `count == len(helpers)`, which still holds under pagination.

I took the cheap boundary shape rather than populating past 100. Each of the three tools gets a small-limit test that asserts `has_more`/`next_offset`/page size and that the pages do not overlap. One deliberate deviation from your sketch: the assertions hold against the response they came from rather than a baseline captured earlier, because 26 other e2e modules call `ha_config_set_helper` and two others create groups, so under `-n2` the collection can grow mid-test and a baseline comparison would flake. The resource test pins that `by_type` describes the whole collection rather than the page, asserted within a single response for the same reason.

If you would rather have the default-limit trigger exercised for real, say so and I will populate past 100 – I skipped it because the walltime cost lands on every run.

**3. Stale NOTE** – fixed in 47a62cc4. Kept with that commit since it documents the same split shape.

The "not YAML-defined" half turned out to be inaccurate too. HA's base `StorageCollectionWebsocket.ws_list_item` serves the storage collection alone, but `person` overrides it to send `{"storage": ..., "config": ...}` where `config` is `yaml.async_items()` – so YAML-configured persons *are* listed. The NOTE now names `person` as the exception instead of claiming YAML helpers are never returned.

**4. Test module docstring** – fixed in c08da4bc. It now names the three builders it pins.

**5. Non-deterministic page boundaries for groups** – fixed in 6769b196, `entity_id` as the secondary sort key.

**6. Implementation Summary** – posted below.

One thing I did not do: no e2e test for the `limit` bounds. They are declarative `Field(ge=1, le=500)` constraints, so a test there exercises FastMCP's schema validation rather than this code. The `ha_get_logs` precedent works differently than the name suggests: it validates inside the tool body via `_coerce_limit`, which raises a structured error below 1 and clamps to `MAX_LIMIT` above it, so the test can assert on a returned error response. Here the bound is enforced by the parameter schema, so a violation raises a `ValidationError` that never reaches the tool body, and `safe_call_tool` only catches `ToolError`. Say if you want it pinned regardless and I will add it.

Patch76:
## Implementation Summary

**Choices Made:**

- **The slice is client-side for all three tools.** Groups and dashboard resources sit on HA core commands that take no pagination parameters, so the full list is already in memory and a slice costs nothing extra. Helpers route through four separate envelope builders, so the slice is applied at one normalization point rather than copied into each.
- **Dropped the `helpers_list` schema parity I proposed in my analysis comment on #1869.** All-types mode merges the legacy per-type records *after* the component call, so a component-side limit/offset could only page the component's own slice and would miss the merged legacy records. Only a post-merge slice paginates every route correctly. The component-side parameters would have been dead weight on the one route that needs them least.
- **`count` is the page size; `total_count` carries the collection size.** This matches `ha_list_services`. Below the default limit the record lists are byte-identical to before and only metadata keys are added, which is the constraint from the review on #1869.
- **Aggregates keep describing the whole collection.** `inline_count` and `by_type` on the resource listing are computed before the slice, so they do not shrink to the page.
- **Fixed the `person` split shape at the source rather than teaching the guard to cope.** `_legacy_helper_list` now flattens like the tool's inline legacy body, so every builder hands the pagination guard a flat list.
- **The guard reports an unpaginatable listing in `warnings[]`, not only in the log.** Logging alone would have kept the silent failure one layer up: the caller is an agent that never sees server logs, so an envelope with records but no pagination metadata reads as "the collection fits on one page". `warnings[]` is what AGENTS.md gives degraded responses, and the tool already uses it for the component-error fallback. The path is unreachable through the tool now that every builder flattens, so its tests drive the helper directly.
- **`entity_id` is the sort tiebreaker for groups**, so a page boundary cannot duplicate or skip a group when two share a case-folded name and `get_states()` order shifts.
- **Fixed the two neighbouring silent paths in this PR rather than deferring them.** `_flatten_helper_list_result` treated an unrecognised shape as an empty listing, and the all-types merge skipped records it could not use; neither said anything. They are the mechanism that kept the `person` bug invisible, not unrelated cleanup, so they belong with it. Both log rather than warn: unlike the pagination guard, neither can attribute the loss to anything the agent could act on. A helper type that is genuinely empty stays quiet.
- **The e2e tests assert each response against itself rather than a baseline captured earlier.** Many other e2e modules create groups and helpers, so the collections can grow mid-test; a captured-baseline comparison would flake under `-n2`. The by_type assertion holds within a single response for the same reason.
- **Did not add an e2e test for the limit bounds.** They are declarative `Field(ge=1, le=500)` constraints, so such a test exercises FastMCP's schema validation rather than this code. `ha_get_logs` can assert on a returned error response because it validates inside the tool body via `_coerce_limit`; here the schema rejects the call first and raises a `ValidationError` that never reaches the body.

**Problems Encountered:**

- **`_legacy_helper_list` fed the pagination guard a non-list for `person`,** exactly as described in the review. HA's `PersonStorageCollectionWebsocket` overrides the base `ws_list_item` to send `{"storage": [...], "config": [...]}`, and the fallback copy did not flatten it. The failure was quieter than I first assumed: the all-types merge iterates the dict's *keys* as strings, and its `isinstance(item, dict)` check then filters them out, so every person record vanished with no error at all rather than raising.
- **That same trace invalidated the tool's NOTE in both halves.** It contradicted the flow-based support described directly below it, and "not YAML-defined" was too broad: HA's base `ws_list_item` serves the storage collection alone, but `person` overrides it to send the YAML-configured persons alongside the storage ones, so those are listed too.
- **Redefining `count` silently weakened two existing e2e assertions** in `test_resources.py`, which compared it against a baseline. They only held while the instance stayed under the default limit. Both now use `total_count`.
- **The first version of the new overlap assertions could fail on a correct listing.** They keyed records on `entity_id or id` and kept the result even when a record carried neither; two such records collapse to `None` in both sets and intersect. Not every collection body embeds its own `id` – `_shape_collection_helper_record` documents `person` and `zone` as stored keyed by id – so the unidentifiable records are now dropped instead.

Patch76:
Follow-up on point 1, since my reply above understates what the guard now does.

Auditing my own change for silent failures turned up that making the guard log was only half a fix. The caller here is an agent, and it never sees server logs – so a `logger.warning` still leaves it holding an envelope with records but no pagination metadata, which is indistinguishable from a collection that fits on one page. The silent failure moved up a layer rather than going away.

facae003 adds the reason to `warnings[]`, which is the channel AGENTS.md gives degraded responses and which this tool already uses for the component-error fallback. The log stays for the operator. The guard builds a new warnings list rather than appending in place, so a caller's existing warnings survive.

The path is unreachable through the tool now that every builder flattens, so the two new tests drive `_paginate_helpers_response` directly. They also pin the property that actually matters: an envelope returned unpaginated must not carry pagination metadata it cannot honour.

That audit also turned up the two neighbouring paths that let this bug stay invisible, both fixed in cc44a643:

- `_flatten_helper_list_result` returned `[]` for any shape it did not recognise, reaching the caller as `count: 0` – identical to a helper type that genuinely has none. It now logs the unusable shape. An empty type stays quiet, since that is not a defect: `person` with no persons is `{"storage": [], "config": []}`, so the check is whether either key held a list at all, not whether records came back.
- The all-types merge kept only dict records and skipped the rest without a word. That is the check that ate every person record while the split shape went unflattened, so it now reports what it dropped.

Both stay logs rather than `warnings[]`: unlike the pagination guard, neither can attribute the loss to anything the agent could act on, and the operator is who needs to see a shape the parser does not know. Both predate this PR, but they are the mechanism behind point 1 rather than unrelated cleanup, so they are in this PR rather than a follow-up. I injected each defect to confirm the new tests catch it.

kingpanther13 (approved):
All six items are addressed, and the response went a level deeper than asked where it mattered. Approving.

1. The person split shape is fixed at the source, and moving the guard's report into `warnings[]` was the right call — an agent never sees server logs, and the non-mutating merge with existing warnings is verified by the new tests. Fixing the two neighbouring silent paths (the flatten's unrecognised-shape empty return and the all-types merge drop) in this PR rather than deferring them was exactly right; so is keeping a genuinely empty person listing quiet.
2. The e2e boundary tests cover the contract on all three tools through the MCP layer, and the self-consistent assertion shape is correct under `-n2` with other modules creating helpers and groups mid-run.
3. Good catch on the YAML half of the NOTE — the person `config` key is the reason the split shape exists, so naming it as the exception is more accurate than what I asked for.

Both declines are accepted as reasoned. On populating past 100: we considered baking a >100 collection into the `initial_test_state` seed instead of runtime creation (no walltime cost that way), but the default-limit trigger runs the same slice the small-limit tests already drive through the schema layer, and a baked collection perturbs the shared fixture every other e2e module sees — not worth it. And a `limit`-bounds e2e would exercise FastMCP's schema validation rather than this PR's code, since there is no in-body coercion here to assert on — thanks for the correction on how the `ha_get_logs` precedent actually works.

kingpanther13:
Note for after merge: we will run a live verification pass once this reaches the dev channel — pagination metadata on all three tools, a small-limit offset walk, and a person listing check — before the next stable cut.

kingpanther13:
Ugh Claude is being so stupid with its wording. I am asking for you to test this on dev server after merge, I'm currently not able to do my usual live testing I do pre-merge but I think it's reasonable to merge this now and fix any extra bugs tgat pop up on dev server before stable releases

Patch76:
@kingpanther13 – live verification pass done on the dev channel, as you asked, ahead of the next stable cut. I found no defect in the pagination itself. One compatibility nuance is below, plus two adjacent findings and a list of what I could not reach.

## Setup

Ran against a dev-channel install on `7.12.3.dev1942` (embedded, HA 2026.7.2), driving the server's HTTP endpoint over JSON-RPC directly rather than through an MCP client – the only way to send values the parameter schema forbids (negative offset, over-max limit, wrong types), so the validation layer gets exercised instead of assumed.

`dev1942` is still current for this surface. I hash-compared every file the pass depends on (`tools_config_helpers.py`, `tools_groups.py`, `tools_resources.py`, `util_helpers.py`, `config_entry_flow.py`, `component_api.py`, and the component's `websocket_api.py`) against master: all identical. The only functional commit since is #1881, in the backup subsystem.

Baseline collections were 0 groups, 37 helpers, 17 dashboard resources. That leaves the default-limit path and the group sort untouched, so I created synthetic objects in two phases (110 `input_boolean` + 4 groups + 1 person + 1 inline resource; then separately 110 groups), ran the pass, removed everything and verified the collections were back to baseline both times.

## The three items you asked for

**Pagination metadata on all three tools** – present and internally consistent. On every response I asserted `count == len(records)`, `has_more == (offset + count) < total_count`, `next_offset == offset + limit` exactly when `has_more`, and that `limit`/`offset` echo what was sent.

**Small-limit offset walk** – walked each collection at `limit=1,2,3,7` (non-divisors deliberately, so page boundaries land mid-collection), and on the 110-group collection additionally at 17, 99, 100, 101, 109, 110, 111 and 500. No duplicates, no gaps, order stable, union of pages equal to the single full-page call. Repeated 10 times per tool: one distinct ordering each.

**Person listing** – correct record shape, and with a second person present the walk sees both exactly once.

## Beyond that

**Default limit at real scale, on two of the three tools.** With 148 helpers and separately with 110 groups, the unparameterised call returns exactly 100, `has_more=true`, `next_offset=100`, and page two holds the remainder and terminates. This is the case the PR exists for, so I wanted it demonstrated rather than inferred.

**The group tiebreaker (`6769b196`), at its worst case.** I named 110 groups so the sort would place a case-folded name collision (`Pagtest B` / `pagtest b`) at exactly indices 99 and 100 – straddling the default page boundary. Page one's tail holds the first, page two's head holds the second, with no overlap and no gap: the pair is split across pages without duplicating or skipping either. That is precisely the failure the commit message describes, at the position where the missing tiebreaker would have caused it. 12 repeated `limit=1` walks produce an identical ordering; I also covered the `.lower()` path with a non-ASCII name.

**Schema-boundary and misuse.** `offset=-1` and `-100` are rejected with `greater_than_equal` before the body runs, so Python's negative slicing never reaches `groups[offset:offset+limit]` and the metadata cannot lie. I checked whether string coercion routes around the constraint – it does not: `offset="-1"` is coerced and then fails `ge=0`. Full 19-value matrix on both parameters (`"5"`, `"+5"`, `" 5 "`, `"abc"`, `""`, `"1e2"`, `"0x10"`, an Arabic-Indic digit, `5.0`, `5.5`, `true`, `false`, `null`, `[5]`, `{}`, `2**63`, `-2**63`), plus unknown keyword arguments, missing required arguments, `helper_type` garbage, and protocol-level abuse (`arguments` absent / `null` / a list). Everything rejected that should be. No defect.

**Injection check on the person fix.** Rather than trust a green test, I reverted `items = _flatten_helper_list_result(result)` in `_legacy_helper_list` and re-ran the unit suite: `test_raised_command_fallback_flattens_the_person_split_shape` fails immediately with `TypeError: string indices must be integers`. Restored, it passes. The guard discriminates.

## Compatibility nuance: `count` changed meaning

`count` is not a new key – it is an existing key whose meaning changed. Before the PR it was the whole collection (`"count": len(processed)`); now it is the page size. Below the default limit the two coincide, which is why it is invisible in most testing, including my own first attempt at this.

Above the limit it is visible. Same install, same 110 groups, the two versions side by side as standalone servers:

|              | records returned | `count` | `total_count` |
| ------------ | ---------------- | ------- | ------------- |
| pre-PR       | 110              | 110     | absent        |
| post-PR      | 100              | 100     | 110           |

So a consumer reading `count` as "how many groups exist" gets 110 before and 100 after. The record lists themselves are byte-identical below the limit (I compared 0/0, 17/17, 37/37 across the two versions), exactly five keys are added (`total_count`, `offset`, `limit`, `has_more`, `next_offset`), and none are removed.

I do not think this is a defect – the intent is explicit in the e2e assertion ("count is the page size, not the total"), and the same shape already ships on `ha_list_services`. But it is the one thing in this PR an existing consumer can notice.

## Two findings, neither introduced by this PR

**1. Ordering is guaranteed for groups but not for the other two.** `tools_groups.py` sorts before slicing; `tools_config_helpers.py` and `tools_resources.py` have no sort at all – they paginate over whatever order the upstream call returns. The asymmetry predates the PR (the group sort existed before it), but pagination changed what it costs: before, the order of a complete list was cosmetic; now it decides what lands on page one versus page two. The reasoning in `6769b196` – "if that order shifted between calls, a group could be duplicated onto the next page or skipped entirely" – applies verbatim to the other two tools, which have no ordering guarantee at all. Empirically the order is stable (10 walks per tool, one ordering each), so this is a latent robustness gap, not an observed break.

Adding a sort changes the order users see today, so it is a behaviour decision rather than a cleanup. I think it is worth doing: where a list tool returns a plain collection, a page boundary should not depend on whatever order the upstream call happens to return. Not as a blanket rule, though – `ha_search` should stay out of it, since it ranks by relevance and alphabetising would throw that away, and `ha_get_history` is naturally chronological. Which tools actually qualify still needs a per-tool read, and I have not done that yet.

I will put it up as a draft PR so there is something concrete to look at rather than a proposal in a comment.

**2. Four flow helper types are listable via `all` but rejected as a `helper_type` value.** `bayesian`, `combine`, `history_stats` and `mold_indicator` come back inside `helper_type="all"`, but passing any of them individually fails validation (I checked all four live). The cause: `ha_config_list_helpers` imports `SUPPORTED_HELPERS` from `config_entry_flow.py`, whose own comment describes it as "15 helpers that use Config Entry Flow API" – that is the *create* surface. The component enumerates 19 for listing (`FLOW_HELPER_DOMAINS`). Listing and creating are not the same capability. Reproducible: I have a `mold_indicator` helper that appears under `all` and cannot be filtered by its own type. Pre-existing, unrelated to pagination, and I have not filed anything.

## What I could not reach

- **The default-limit path on `ha_config_list_dashboard_resources`.** Proven on helpers (148) and groups (110); resources stayed at 17. Inflating a live Lovelace resource list past 100 was not a trade I wanted to make. Its slice is exercised at `limit=1,2,5,500` and at offsets up to `10^9`, and its default is verified as 100, so only the specific (default, >100) combination is unproven there.
- **The person fix's live trigger** needs the component installed *and* its `helpers_list` command raising a non-`unknown_command` error. `person` has been in `covered_types` since #1812, so the fallback cannot be forced from outside. The unit suite covers it deterministically (see the injection check above), which is the right place for it.
- **The group tiebreaker against a genuinely reordered `get_states()`.** Only determinism is forceable; I cannot make the state machine reorder on demand. What I could do is put the tie on the page boundary, which I did.
- **The sort-key `or entity_id` fallback.** Home Assistant auto-assigns a `friendly_name`, so a group created without one still has it and the `or` branch never fires.

## One note for the stable cut

The new `limit`/`offset` parameters reached the server fine, but the tool schema offered to a connected client lagged behind: the server advertised them over JSON-RPC while the client-side catalogue still showed the pre-PR signature and docstring, and a tool-list refresh plus a client restart did not clear it. If someone reports "the parameters do not exist" after the stable release, the server is fine and the client's cached catalogue is not.

Issue #1890, opened 2026-07-16
Claude.ai is not able to connect to ha_mcp with fully working Nabu Casa MCP URl


Penny2912:
I've just installed the ha_mcp custom component to start enhancing my hass with Claude. Unfortunately I was not able to get this working. I've installed the custom component through HACS and added both integrations.

Then I've added the Nabu Casa URL (which is reachable publicly and showing that the MCP server is running fine) to Claude.ai and it is able to connect but when trying to retrieve entities or whatever, it fails.

In my logs I see several occurrences of the following error and debug logs:

```
Logger: ha_mcp.client.rest_client
Quelle: custom_components/ha_mcp_tools/embedded_server.py:748
Integration: HA-MCP Custom Component (Dokumentation, Probleme)
Erstmals aufgetreten: 06:33:21 (7 Vorkommnisse)
Zuletzt protokolliert: 07:44:58

WebSocket message failed: Failed to connect to Home Assistant WebSocket: InvalidMessage: did not receive a valid HTTP response

```

```

Logger: ha_mcp.client.websocket_client
Quelle: custom_components/ha_mcp_tools/embedded_server.py:748
Integration: HA-MCP Custom Component (Dokumentation, Probleme)
Erstmals aufgetreten: 06:33:21 (7 Vorkommnisse)
Zuletzt protokolliert: 07:44:58

WebSocket connection failed: did not receive a valid HTTP response

```
```

Logger: ha_mcp.settings_ui
Quelle: custom_components/ha_mcp_tools/embedded_server.py:748
Integration: HA-MCP Custom Component (Dokumentation, Probleme)
Erstmals aufgetreten: 07:31:45 (6 Vorkommnisse)
Zuletzt protokolliert: 07:44:11

fs-custom-paths GET could not reach ha_mcp_tools: HTTP error: Server disconnected without sending a response.
Embedded restart: entry discovery failed: { "success": false, "error": { "code": "SERVICE_CALL_FAILED", "message": "config_entries/get failed: Failed to connect to Home Assistant WebSocket: InvalidMessage: did not receive a valid HTTP response", "suggestion": "Check the Home Assistant WebSocket connection" } }

```

So everything seems up and running but I'm not able to connect anyhow.

What is blocking the Websocket here?

kingpanther13:
[...]

PR #1892, opened 2026-07-16
fix: remove phantom bulk-config WS commands and honor HA SSL/port on the embedded loopback

### LINKED ISSUE #1889: [BUG] websocket connection errors in logs
## 📋 Bug Description
The ha-mcp WebSocket client emits recurring "Unknown command" / "Entity not found" errors in the Home Assistant log, in pairs, roughly every 1-2 minutes, starting right after every HA restart — independent of any actual MCP tool call from a connected AI agent.

## 🔄 Steps to Reproduce
1. Restart Home Assistant (or the ha-mcp embedded server) with ha-mcp installed
2. Watch the home-assistant.log / error_log for the `ha_mcp.client.rest_client` logger
3. Observe recurring "WebSocket message failed: Command failed: Unknown command." (and occasionally "Entity not found") entries appearing continuously, in pairs, roughly every 1-2 minutes — with no user/agent activity happening at those times

**Expected:**
No recurring background WebSocket errors when idle — any internal heartbeat/health-check command sent by the embedded server should be a recognized command.

**Actual:**
Continuous "Unknown command." / "Entity not found" errors logged at ERROR level indefinitely, starting shortly after each HA restart.

---

**User prompt:** "a chaque que je redemmare j'ai ca" (referring to recurring `ha_mcp.client.rest_client` "Unknown command" / "Entity not found" errors observed in the HA error log)

**Tool call(s):**

```markdown
User: a chaque que je redemmare j'ai ca
[Attached excerpt: "ha_mcp (moi-même 😅)  'Unknown command' / 'Entity not found' — probablement des appels d'outils mal formés de ma part"]

Agent: Ok, donc ce message d'erreur (ha_mcp — "Unknown command" / "Entity not found") revient à chaque redémarrage de Home Assistant. Ça vient du composant custom_components/ha_mcp_tools/embedded_server.py. Regardons ça de plus près.

[Tool: ha_get_logs(source="error_log", search="ha_mcp", order="oldest", limit=50)]
Result:
2026-07-15 23:16:26 ERROR (ha-mcp-server) [ha_mcp.client.rest_client] WebSocket message failed: Command failed: Unknown command.
2026-07-15 23:16:26 ERROR (ha-mcp-server) [ha_mcp.client.rest_client] WebSocket message failed: Command failed: Unknown command.
2026-07-15 23:16:44 ERROR (ha-mcp-server) [ha_mcp.client.rest_client] WebSocket message failed: Command failed: Unknown command.
2026-07-15 23:16:45 ERROR (ha-mcp-server) [ha_mcp.client.rest_client] WebSocket message failed: Command failed: Unknown command.
2026-07-15 23:16:58 ERROR (ha-mcp-server) [ha_mcp.client.rest_client] WebSocket message failed: Command failed: Entity not found
... (repeating in pairs every ~1-2 minutes, continuously since HA startup, with no correlation to any agent tool call)

Agent: D'après ces logs, ce n'est pas lié à mes propres appels d'outils — ces erreurs sont émises en continu par un processus interne de fond de l'intégration ha_mcp (probablement un heartbeat/keepalive), indépendamment de toute conversation.

[Tool: ha_get_integration(query="ha_mcp", exact_match=false)]
Result: {"entry_id":"01KXFRJQHAHYZJ6EZ1NSAYFKVR","domain":"ha_mcp_tools","title":"HA-MCP Server","state":"loaded","log_level":"WARNING"} — integration loaded fine, no config issue detected.

Agent: L'intégration ha_mcp_tools est bien chargée et à jour (v7.12.1). Le bug semble interne au serveur ha-mcp lui-même.

User: oui (confirming to generate the bug report)

[Tool: ha_report_issue(tool_call_count=16)]
Result: diagnostic_info confirms ha_mcp_version 7.12.1, installation_method "embedded", mcp_transport "stdio", startup_logs showing 6 identical "Unknown command" errors within the first 45 seconds after HA startup (+42.69s, +44.59s, +45.47s), no version mismatch, connection_status "Connected".
```

kingpanther13:
## What does this PR do?

Fixes two independent defects in how the embedded (custom-component) server talks to HA core.

**Issue #1889 — recurring paired "Unknown command" ERRORs.** Every legacy config search tried six WebSocket bulk-list commands (`config/automation/config/list`, `automation/config/list`, and the script/scene equivalents) that have never existed in HA core (zero occurrences in home-assistant/core). Each attempt came back `unknown_command` and was ERROR-logged by `send_websocket_message` — two attempts per domain = the pairs, three domains = 6 errors per search, reproduced on demand against a live instance. Stable servers (no component search fast path) hit this on every search. The WS fallback tier is removed — REST bulk → budgeted individual fetches remains, which is the path that actually ran anyway. The `rest_client` failure log now also names the failing command type, so future reports are diagnosable from the log alone.

Closes #1889

**Issue #1890 — SSL/custom-port instances cannot use the embedded server at all.** The component hardcoded `http://127.0.0.1:8123` as the loopback URL. With `http.ssl_certificate` configured, port 8123 speaks TLS, so every REST/WS call died at connection setup ("Server disconnected without sending a response" / "did not receive a valid HTTP response") while the MCP handshake and `tools/list` (no HA round-trip) kept working — the reporter's exact symptoms. The default loopback now derives scheme and port from `hass.config.api` (https + `verify_ssl=False`, since HA's certificate can never match 127.0.0.1); an explicit `server_url` override still wins verbatim. `set_embedded_connection` grows an optional `verify_ssl` parameter, called defensively so an older installed server still starts. The options form no longer pre-fills the loopback constant (every save persisted it as an explicit override), and normalization collapses a stored constant back to no-override.

Addresses #1890 — left open pending the reporter confirming `http.ssl_certificate` in their config.

**Hotfix releases never published to PyPI (pipeline gap found while shipping this fix).** `release-publish.yml` (PyPI / server Docker / MCPB) triggers only on the "SemVer Release" workflow, and its `workflow_run` branches filter matches the *triggering run's* branch — `hotfix/*`, never master — so hotfix releases skipped it entirely: v7.12.2/v7.12.3 shipped on GitHub/HACS while PyPI stayed at 7.12.1, and stable-channel pip installs silently never received them. `hotfix-release.yml` now dispatches `release-publish.yml` explicitly after a successful release (`workflow_dispatch` is the documented GITHUB_TOKEN exception that creates runs), which also gets the guard's forced-republish semantics — required because master routinely carries unreleased feat/fix commits when a hotfix goes out.

**Alias sentinel leak (found during live verification).** HA core's registry aliases can carry the `COMPUTED_NAME` sentinel (`entity_registry.ComputedNameType._singleton` — "the computed entity name is an alias"): an enum member in-process, `null` over the websocket API. Both alias readers ingested it blindly — the component's `_entity_record` `str()`'d it into a literal `"ComputedNameType._singleton"` alias on every carrying entity (fake data in results and a scored match text), and the server's `_fetch_entity_aliases` would admit `None` into downstream alias string ops. Both now keep string entries only.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

## Checklist


ghhamcp:
@codex review

ghhamcp:
@codex review

Patch76 (approved):
Four connection-layer defects that all surfaced from the same embedded-server investigation, each with its own targeted test.

I verified the load-bearing external claims against HA core rather than the diff alone:

- **#1889 (phantom WS tier):** `config/<domain>/config/list` / `<domain>/config/list` genuinely have no counterpart in core – automation/script/scene configs are served by REST id-based views (`EditAutomationConfigView` et al.), not a websocket list command – so the removed tier could only ever have returned `unknown_command`. REST bulk plus budgeted individual fetch remain, and `test_config_search_never_sends_phantom_bulk_ws_commands` pins that the branch never re-emits those types.
- **#1890 (SSL/port loopback):** `_derive_loopback_url` reads the real `hass.config.api` shape (`ApiConfig.port: int`, `.use_ssl: bool`), and the `verify_ssl=False` on the derived `https://127.0.0.1` hop is correct – the certificate is issued for the hostname, never for the loopback address. The strict `isinstance` port guard and the `TypeError` fallback for a server that predates the new parameter are both covered end-to-end.
- **CI dispatch:** `release-publish.yml` does expose `workflow_dispatch`, and its version step forces a republish of the latest stable tag on manual dispatch even while master carries unreleased commits – which is the state a hotfix ships in – so the new `trigger-release-publish` job closes the PyPI gap as described.
- **Alias sentinel:** both readers now keep string entries only; the name the sentinel stood for is still reachable via the friendly name.

Two non-blocking notes:

- This bundles four independent fixes. The shared origin justifies it, but it does make the single revert boundary coarser if one of them needs backing out.
- The test for #1889 documents that the REST bulk endpoint also has no real counterpart, so individual fetch is the path that actually runs. That leaves the REST bulk attempt as a guaranteed miss on every search – harmless (it only debug-logs and falls through), just no longer load-bearing.

Approving – the removals are provably dead code and the new logic is well isolated and tested.

PR #1897, opened 2026-07-16
feat: add overwrite option to ha_import_blueprint for blueprint re-import

### LINKED ISSUE #1894: [FEATURE] Trigger a Blueprint re-import

Hi,

Great tool! It has absolutely changed how I use HA. Blueprint all the things! 🤣

One thing I'm finding is that when I've got a new Blueprint, there's an MCP call to import. But there's no MCP action equivalent to the "Re-import Blueprint" feature in the HA UI.

Now I admit that l am not a programmer. I just bang together basic algorithms and shamelessly… adapt… other people's more complex algorithms. So I'm not quite sure how to make this happen or where to go to try to give you help. But there's anything I can do, please let me know!

Thanks! 

### Why do you need this?

It would make my development/testing loop quicker if I didn't have to switch over and do the re-import myself. Its already awkward developing then doing something physical to test, so I want to reduce that loop even by a little bit.

### Additional context

_No response_

kingpanther13:
Closes #1894

## What does this PR do?

Adds an `overwrite` parameter (default `false`) to `ha_import_blueprint`, giving parity with the UI's "Re-import blueprint" action.

Previously, re-importing an installed blueprint failed at the save step with "File already exists" because `blueprint/save` was called without `allow_override`. Now:

- `overwrite=true` passes `allow_override` to `blueprint/save`; Home Assistant overwrites the blueprint and automatically reloads all automations/scripts that use it.
- Without `overwrite`, an already-installed blueprint fails early with a structured `RESOURCE_ALREADY_EXISTS` error pointing at `overwrite=true` (using the `exists` flag `blueprint/import` already returns), instead of the raw save failure.
- The `overrides_existing` field in the response (previously always `false`) is now meaningful.

The save step was extracted into a `_save_blueprint` helper to keep `ha_import_blueprint` under the complexity limit.

## Type of change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

New e2e test `test_reimport_blueprint_with_overwrite` covers: re-import without `overwrite` → structured `RESOURCE_ALREADY_EXISTS`; re-import with `overwrite=true` → success with `overrides_existing=true`.

## Checklist

### DIFF
--- a/src/ha_mcp/tools/tools_blueprints.py
+++ b/src/ha_mcp/tools/tools_blueprints.py
@@ -234,16 +234,19 @@ class BlueprintTools:
 
         Returns the blueprint/save result payload (contains overrides_existing).
         """
-        save_response = await self._client.send_websocket_message(
-            {
-                "type": "blueprint/save",
-                "domain": domain,
-                "path": path,
-                "yaml": yaml_data,
-                "source_url": url,
-                "allow_override": overwrite,
-            }
-        )
+        save_message: dict[str, Any] = {
+            "type": "blueprint/save",
+            "domain": domain,
+            "path": path,
+            "yaml": yaml_data,
+            "source_url": url,
+        }
+        # allow_override only exists on HA >= 2023.12 and the WS schema
+        # rejects unknown keys - only send it when actually overwriting
+        if overwrite:
+            save_message["allow_override"] = True
+
+        save_response = await self._client.send_websocket_message(save_message)
 
         if not save_response.get("success"):
             error = save_response.get("error", {})
@@ -258,7 +261,11 @@ class BlueprintTools:
                 "Use ha_get_blueprint() to check if it already exists",
             ]
 
-            if "already exists" in save_error.lower():
+            # Reachable despite the early exists check: a race between
+            # import and save, or an installed file that failed to load
+            # (core reports exists=false for those)
+            already_exists = "already exists" in save_error.lower()
+            if already_exists:
                 suggestions.insert(
                     0,
                     "A blueprint with this path already exists - pass overwrite=true to re-import it",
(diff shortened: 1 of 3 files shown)


ghhamcp:
@codex review

ghhamcp:
@codex review

Patch76 (approved):
**Approve.**

This cleanly adds the re-import parity #1894 asked for, and the min-version gate it grew is a genuine safety win — it stops an incompatible blueprint from clobbering a working one when `overwrite=true`.

**Verified**
- HA-core contract (2026.7.2 `blueprint/websocket_api.py`): `blueprint/import` returns `exists` + `validation_errors`; `blueprint/save` accepts optional `allow_override` and returns `overrides_existing`; a save collision surfaces as `already_exists` / "File already exists". Every key the PR leans on matches.
- `Blueprint.validate()` returns `list[str] | None` (`blueprint/models.py`), so the `if validation_errors:` gate and the `"; ".join(...)` are both safe, and the message the e2e asserts ("Requires at least Home Assistant …") is the literal core string.
- Traced every branch: `overwrite=false` + installed → `RESOURCE_ALREADY_EXISTS` via the deterministic `exists` flag; `overwrite=true` + installed → override + reload; `overwrite=true` on a fresh path → plain install with `overrides_existing=false`; the validation gate runs before the exists check, so `overwrite` can't skip it. All three new e2e tests exercise these, including the easy-to-miss fresh-path case, and one verifies the new content actually lands on disk.
- `_save_blueprint` raises via `raise_tool_error(create_error_response(...))`, and `except ToolError: raise` precedes `except Exception`, so these keep their intended codes instead of collapsing to `INTERNAL_ERROR`.

**One note (non-blocking):** the save-time "already exists" fallback classifies on a message substring rather than HA's structured `already_exists` code. That isn't fixable at this layer — the WS client flattens the error to `str(e)` and drops the code before it arrives — and the normal collision is already caught deterministically by the earlier `exists` check, so the substring only runs in the rare import/save race. Fine as-is; flagging it so the fragility is on record.

Issue #1901, opened 2026-07-17
BestPracticeKey parameter not accepted by VS Code MCP client


Aldjinn:
After upgrading to v7.13.0, all write tools are blocked by the new strict best-practices acknowledgment gate (#1820).

The `ha_get_skill_guide` tool works and returns the key (`bps-ack-1779`), but the key cannot be passed to write tools because VS Code MCP client rejects the parameter:

```
ERROR: must NOT have additional properties
```

Steps:
1. Call `ha_get_skill_guide` — succeeds, returns key
2. Call `ha_config_set_dashboard` with `BestPracticeKey="bps-ack-1779"` — fails
3. Call without key — fails with BPS_ACKNOWLEDGMENT_REQUIRED

Environment: ha-mcp v7.13.0 (Docker), VS Code Copilot Chat MCP, Windows.

This is a regression — all write tools worked in v7.12.0. The `BestPracticeKey` parameter is declared server-side but VS Code MCP client doesn't include it in the schema it sends.

kingpanther13:
While I work on a fix, you can disable this in the options, just uncheck the box for "strict" under mandatorybps. 

Are you using lite docstrings or anything like that? Does this occur on all of the write tools? It should be in the tool schema so that is odd it isn't showing up, I'll work on a fix.

kingpanther13:
@Aldjinn Actually, before I do any digging into the code, please try to refresh your tool list. Your client might just have a stale tool list and may not have loaded the new schema up. Unsure how your client handles it but usually disconnecting and reconnecting will do the trick on most clients if there isn't a button to refresh. If you're still experiencing the problem after then I need to look deeper into it.

Aldjinn:
## Solution found ✅

After upgrading to v7.13.0, the `BestPracticeKey` parameter was not accepted by VS Code's MCP client. The fix was to set the environment variable in Docker:

```yaml
environment:
  - ENABLE_STRICT_MANDATORY_BPS=false
```

Then `docker compose down && docker compose up -d`.

After the restart:
- `ha_config_set_dashboard` works again without passing `BestPracticeKey`
- All write tools (automation, helper, scene, script) work as expected
- The strict BPS mode is disabled, so no acknowledgment key is required

### Environment
- ha-mcp: v7.13.0 (Docker, `ghcr.io/homeassistant-ai/ha-mcp:latest`)
- Client: VS Code built-in Copilot Chat (MCP HTTP transport)
[...]

PR #1910, opened 2026-07-17
fix: Guide stale-schema clients past BestPracticeKey client-side rejection

kingpanther13:
## What does this PR do?

Adds a second suggestion to the strict-BPS `BPS_ACKNOWLEDGMENT_REQUIRED` block error covering clients that validate tool calls against a stale cached tool schema (#1901).

Root cause of #1901: VS Code Copilot Chat validates model-generated arguments with AJV against its own copy of the tool schema before sending `tools/call` (`extensions/copilot/src/extension/tools/common/toolsService.ts`). The compiled validator is cached per tool name for the lifetime of the window and is not invalidated by MCP server restarts, reconnects, or `MCP: Reset Cached Tools` (see microsoft/vscode#284221 — the maintainer confirms a reload is required). After upgrading the server in place, the client can therefore keep rejecting `BestPracticeKey` with `must NOT have additional properties` even though v7.13.0 publishes the parameter: raw `tools/list` dumps of both releases confirm v7.12.0 and v7.13.0 both serve `additionalProperties: false` and only v7.13.0 declares `BestPracticeKey`, so validating the retry against the older schema produces exactly the reported error. Disabling strict mode "fixes" it only because the model stops sending the argument — the schema is declared unconditionally either way.

The rejection happens entirely client-side — the retry never reaches the server — so the block error (which the model always sees first, and which is what teaches it the `BestPracticeKey` argument name) is the only server surface that can carry the recovery path. The new suggestion names the client-side error verbatim so the model can match it, and directs a full client reload (VS Code: `Developer: Reload Window`).

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing
- [x] All automated tests pass (`uv run pytest`) — `tests/src/unit/test_strict_bps.py` run locally (30 passed, including the new regression test); full suite runs in CI

## Checklist

### DIFF
--- a/src/ha_mcp/strict_bps.py
+++ b/src/ha_mcp/strict_bps.py
@@ -161,8 +161,12 @@ def strict_bps_ack_line() -> str:
 def _raise_bps_ack_required_error(name: str) -> NoReturn:
     """Raise the structured block error for a gated write missing the key.
 
-    The message and suggestion tell the model how to obtain the key but
-    never contain the key itself.
+    The suggestions tell the model how to obtain the key but never contain
+    the key itself. The second suggestion pre-arms the model for clients
+    that validate tool arguments against a stale cached tool schema and
+    reject the ``BestPracticeKey`` retry client-side (#1901) — that
+    rejection never reaches the server, so this error is the only server
+    surface that can carry the recovery path.
     """
     reference_file = STRICT_BPS_GATED_TOOLS[name]
     message = (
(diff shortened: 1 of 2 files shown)


ghhamcp:
@codex review

ghhamcp:
@codex review

PR #1917, opened 2026-07-17
Validate runtime tool translation placeholders

SHAREN:
## Summary

- validate translated tool title/description placeholders against the actual runtime tool metadata
- fall back only the invalid translated field while keeping the existing shortened source description for display
- use placeholder set semantics and full source metadata so reordered, repeated, or later-line placeholders are handled correctly
- add jsdom regression coverage for invalid and valid runtime tool translations

## Root cause

Follow-up to #1913. The load-time tool placeholder validation compares localized tools with entries in `en.json`, but the English tool catalog is intentionally empty because canonical tool titles and descriptions come from the runtime API. As a result, the shipped Russian tool overrides were not actually covered by that check.

## Impact

A future translation that drops or invents a runtime tool placeholder no longer renders misleading copy. The affected field safely falls back to the canonical source value and emits a console warning; valid fields on the same tool remain localized.

## Validation

- `14 passed` for localization behavior and catalog tests on the rebased branch
- `161 passed` in the broader settings UI/i18n/sidebar proxy suite before isolating this follow-up commit
- repository-wide `ruff check` passed
- changed Python files pass `ruff format --check`
- `node --check src/ha_mcp/settings_ui/settings.js` passed
- `git diff --check` passed

### DIFF
--- a/src/ha_mcp/settings_ui/_i18n.py
+++ b/src/ha_mcp/settings_ui/_i18n.py
@@ -116,7 +116,12 @@ def load_catalogs(directory: Path = LOCALES_DIR) -> dict[str, dict[str, Any]]:
 
 
 def _validate_placeholder_parity(catalogs: dict[str, dict[str, Any]]) -> None:
-    """Reject translations that drop or invent ``{name}`` placeholders."""
+    """Reject catalog-backed translations with mismatched placeholders.
+
+    Tool metadata normally comes from the runtime API rather than ``en.json``;
+    ``settings.js`` performs the corresponding parity check against those
+    canonical values before displaying a translated tool field.
+    """
     english_messages = catalogs[DEFAULT_LOCALE]["messages"]
     english_tools = catalogs[DEFAULT_LOCALE]["tools"]
     for locale, catalog in catalogs.items():
--- a/src/ha_mcp/settings_ui/settings.js
+++ b/src/ha_mcp/settings_ui/settings.js
@@ -35,11 +35,38 @@ function localizedToolGroup(group) {
   return (I18N_PAYLOAD.tool_groups || {})[group] || group;
 }
 
+function _placeholderNames(value) {
+  return Array.from(new Set(Array.from(
+    String(value || '').matchAll(/\{([a-zA-Z_][a-zA-Z0-9_]*)\}/g),
+    match => match[1]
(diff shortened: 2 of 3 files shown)


SHAREN:
@Patch76 This is the isolated follow-up to your approved/merged #1913. It contains only the runtime placeholder-validation gap described in the final comment there, rebased onto current `master`. Ready for review. @codex review

ghhamcp:
@codex review

Patch76:
You're too fast for my Claude 😄

Patch76 (approved):
Thanks for the independent follow-up – this closes the gap from #1913 at the right seam. The earlier load-time check compared translated tools against `en.json`, whose `tools` section ships empty, so it never actually validated the tool overrides; moving the parity check to `settings.js` against the runtime tool metadata is the only place the canonical English title/description exist, so this is the correct place for it.

**Verified:**
- The placeholder set comparison matches the Python validator: same `{name}` pattern, and set semantics (via `_placeholderNames`) make it order- and duplicate-insensitive, so a reordered or repeated placeholder is accepted rather than falsely rejected.
- Validation runs against the *full* runtime description while display keeps the shortened first-line excerpt as the fallback – so a placeholder that lives past the first line is still checked. The `set_matches_full_metadata` test proves this: the source has `{entity}` on line two and twice, and a single/reordered translation is accepted.
- On a real mismatch only the offending field falls back to the runtime source, the other field stays localized, and a console warning is emitted. The `mismatch_falls_back_to_source` test drives a dropped-placeholder title and an invented-placeholder description against runtime metadata and asserts the row renders the source values with warnings for both fields.
- The tests source the English tool copy from the `/api/settings/tools` response (the runtime metadata), which is exactly how it arrives in production – so they exercise the shipped catalog model rather than a constructed one.

One non-blocking note: the load-time `tools` loop in `_i18n.py` is now inert for the shipped catalog (empty `en.json` tools), and the updated docstring says so – fine as a forward-compat guard for a future English tool catalog, just flagging that the JS check is the effective one now.

Looks correct and well-covered. I appreciate that you ran an independent verification pass after the merge and caught this yourself – that is exactly the kind of follow-through that makes the localization foundation solid, and it is a genuinely useful contribution to the project. Looking forward to more of your work here.

PR #1925, opened 2026-07-18
fix: Make strict-BPS ack key a rotating attestation phrase

### LINKED ISSUE #1924: [AGENT] Claude refuses to write to Homeassistant, claiming it was being subject to a "injection technique" by a malicious actor

I've been working with ha-mcp for several weeks now.  I've slowly allowed all write commands without issue.

I updated to 7.13.0 and was working on proofing several automations and entity names.

Claude began refusing to perform writes and started interrogating the chat as some sort of infiltration attack.  Claude summarizes:

In that session, the ha_config_set_automation skill content came back with a bps-ack-* token embedded in the fetched documentation, instructing me to pass it back as the BestPracticeKey parameter to unlock the write. I recognized that pattern — a tool result telling me to extract and replay a credential-like string — as a prompt injection and declined to comply each time.
That wasn't a "failure" on my part exactly — it was intentional. An acknowledgment key that arrives embedded in the content I'm supposed to be acknowledging, with instructions to feed it back as a parameter, is indistinguishable from a classic injection technique regardless of whether the source was legitimate in that case. The gate only cleared once you adjusted and restarted the add-on's security setting on your end, which is the right way to resolve it — not me copying a token out of tool output.
So to directly answer: I didn't fail to read the best practices — I read them fine. I failed (or rather, refused) to pass along a key that showed up in-band in a place a malicious actor could also plant one. Same caution would apply here if a BestPracticeKey value showed up embedded in fetched content rather than being something you handed me directly.

The AI shut down my attempts to write to my HA Green due to its suspicion this was some sort of infiltration attack.  Ultimately I was required to uncheck "strict best-practices mode" in order to get Claude to proceed.  This was on Sonnet 5 Medium.

```markdown
## 🤖 What Did the AI Agent Do?

In a prior session working on Home Assistant automation edits, `ha_config_set_automation` write calls were gated by strict best-practices mode. The tool's error response directed the agent to call `ha_get_skill_guide` to read the best-practices skill, then pass back an acknowledgment key as the `BestPracticeKey` parameter. The key (a `bps-ack-*` style token) arrived embedded inline inside the fetched skill documentation content itself, with an instruction telling the agent to extract that token and replay it in a subsequent tool call.

The agent declined to do this on repeated attempts, treating the pattern as indistinguishable from a prompt-injection attack: content returned from a tool call instructing the agent to extract a token and feed it into a privileged parameter is a textbook injection shape, regardless of whether the source was a legitimate first-party mechanism in this case. As a result, writes stayed blocked until the person manually adjusted and restarted the relevant add-on security setting.

## 🎯 What Should the Agent Have Done?

Arguably nothing different, given the information available — an agent that reliably distinguishes "legitimate tool-delivered credential" from "injected instruction to exfiltrate/replay a token" purely by pattern-matching the delivery mechanism would be a worse agent in the general case. The realistic fix is on the tool side: an acknowledgment mechanism that doesn't require the agent to lift a bare token out of narrative content and paste it into a parameter — e.g., a dedicated confirmation tool call, a structured (non-freetext) response field, or a session-scoped flag set automatically after `ha_get_skill_guide` is called, rather than a client-side round-trip that mimics injection.

## 📝 Conversation Context

Person was working through several automation hardening tasks (water leak safeguard, battery cycle automations) via `ha_config_set_automation` and repeatedly hit the write gate. This was flagged again in a follow-up session (the current one) when the person asked why "strict best-practices mode" had failed previously.

---

**User prompt:** (paraphrased from a prior session's summary, not verbatim — original session did not surface the literal HA config commands in currently available context) — the person was asking Claude to apply hardening edits to `ha_config_set_automation` for a leak-safety automation.

**Tool call(s) the agent chose:**
ha_config_set_automation(identifier="automation.[redacted]", config_hash="...", python_transform="...")
→ Error: strict best-practices mode active; call ha_get_skill_guide and pass returned acknowledgment key as BestPracticeKey
ha_get_skill_guide(skill="home-assistant-best-practices", file="SKILL.md")
→ [skill content embedding a bps-ack-* token with instructions to pass it as BestPracticeKey]
[Agent declined to extract/replay the embedded token, citing prompt-injection risk]

---

## 💡 Suggested Improvement

- [x] **Tool design** - Tool should accept different parameters or return more info
- [ ] Tool documentation
- [ ] Error messages
- [ ] Agent prompting
- [ ] New tool needed
- [ ] Other

**Details:** Replace the "extract token from fetched content, replay as parameter" acknowledgment flow with something that doesn't structurally resemble a prompt injection — e.g., have `ha_get_skill_guide` itself flip a server-side session flag once called for the relevant skill, so `ha_config_set_automation` etc. simply check "was this skill read this session" without requiring a client-visible secret to be lifted out of the response body and passed back verbatim.

## 🔧 Tool Calls Made (Auto-Filled)

See tool-call log above — this session's own calls (`ha_get_skill_guide`, `ha_config_set_automation` with `MandatoryBPS=false`) show the gate cleared per your prior add-on setting change.

## 📊 Environment

- **ha-mcp Version:** 7.13.0
- **Custom Component:** not detected
- **Installation Method:** addon
- **MCP Transport:** http
- **MCP Client:** unknown (client did not advertise itself)
- **AI Model:** Claude (Anthropic) — current session is Claude Sonnet 5; the session where the behavior occurred may have used a different Claude model, which I can't confirm from here
- **Home Assistant Version:** 2026.6.4

## ⚙️ ha-mcp Configuration

- enable_beta_features: `False`
- enable_websocket: `True`
- enable_dashboard_partial_tools: `True`
- enable_tool_search: `False`
- enable_yaml_config_editing: `False`
- enable_filesystem_tools: `False`
- enable_code_mode: `False`
- enabled_tool_modules: `all`
```

kingpanther13:
## What does this PR do?

Fixes #1924 — an agent treated the strict best-practices acknowledgment flow as a prompt-injection attempt: the opaque `bps-ack-1779` token, delivered inside fetched skill content with an instruction to replay it as `BestPracticeKey`, pattern-matched "extract and replay a credential", so the agent refused every gated write until the user disabled strict mode entirely.

Two changes, no mechanism change:

1. **The acknowledgment key is now an attestation phrase with an hourly-rotating suffix** — `I-HAVE-READ-THE-BEST-PRACTICES-GUIDE-<8 hex>`.
   - The plain-English prefix cannot be mistaken for a secret; it reads as what it is — a read-receipt.
   - The suffix is derived from a per-process salt plus the current hour bucket (stateless: salt + clock, no server-side session state). The old static literal sat in this public repo, so a model could satisfy the gate from training data, persistent agent memory, or a stale session summary without reading the guide; now only a recent read served by the current process yields a valid key. The previous hour's key is honored as a grace window so a read just before rotation does not strand the write that follows it.
2. **The `BestPracticeKey` schema description documents the protocol** — the value is public, not a credential, and is obtained by reading the skill via `ha_get_skill_guide`. With the round-trip sanctioned by trusted tool metadata, replaying the key is following the tool's declared protocol rather than obeying instructions found in tool output.

The block error states the key rotates and must be refreshed by re-reading (so an expired key is not replayed in a loop), and a non-string `BestPracticeKey` from an off-spec client degrades to that block error instead of raising. The `ha_get_skill_guide` ack line gets matching read-receipt framing. The e2e strict-BPS suite boots the server in-process, so it obtains the live key via `current_strict_bps_ack_key()`.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

`tests/src/unit/test_strict_bps.py` gains a `TestAckKeyShape` class (key shape, within-bucket stability and cross-bucket rotation, per-process salting, schema-description protocol markers, ack-line framing) plus middleware tests for the previous-hour grace window and rejection of keys two rotations old — written first, watched fail, then implemented. All 38 tests in the file pass; ruff and mypy clean.

## Checklist

### DIFF
--- a/src/ha_mcp/strict_bps.py
+++ b/src/ha_mcp/strict_bps.py
@@ -89,7 +89,7 @@ _ACK_KEY_ROTATION_SECONDS = 3600
 def _current_ack_key(now: float | None = None) -> str:
     bucket = int((time.time() if now is None else now) // _ACK_KEY_ROTATION_SECONDS)
     digest = hashlib.sha256(f"{_ACK_KEY_SALT}:{bucket}".encode()).hexdigest()
-    return f"{STRICT_BPS_ACK_KEY_PREFIX}-{digest[:4]}"
+    return f"{STRICT_BPS_ACK_KEY_PREFIX}-{digest[:8]}"
 
 
 def current_strict_bps_ack_key() -> str:
@@ -198,8 +198,8 @@ def strict_bps_ack_line() -> str:
     """Return the single line that publishes the acknowledgment key.
 
     Prepended to the ha_get_skill_guide Tier-3 best-practices content when
-    strict mode is effective (server.py). This is the ONLY place the key
-    literal is emitted to a caller.
+    strict mode is effective (server.py). This is the ONLY place the actual
+    key value is emitted to a caller.
     """
     return (
         f"Acknowledgment key: {current_strict_bps_ack_key()} — strict "
@@ -226,7 +226,9 @@ def _raise_bps_ack_required_error(name: str) -> NoReturn:
         "Strict best-practices mode is enabled for this tool. Read the "
         "best-practices skill to obtain the required acknowledgment key, then "
         "pass it as the BestPracticeKey argument on this call. The key is "
-        "published only in that skill's content."
+        "published only in that skill's content, and it rotates periodically "
+        "— a key obtained earlier may have expired, so re-read the skill for "
+        "the current value instead of resending a previous one."
     )
     raise_tool_error(
         create_error_response(
@@ -313,10 +315,10 @@ class StrictBpsMiddleware(Middleware):
             return await call_next(context)
 
         # The gate is the ONLY reader of BestPracticeKey: strip it before
-        # dispatch (whether or not strict mode is on) so the constant never
+        # dispatch (whether or not strict mode is on) so the key value never
         # reaches the tool body, the policy middleware's approval args-hash
-        # (where it would churn remembered approvals across strict toggles),
-        # or downstream logging.
+        # (where the hourly-rotating value would churn remembered approvals
+        # every rotation), or downstream logging.
         args = context.message.arguments or {}
         supplied = args.get(STRICT_BPS_KEY_PARAM)
         if STRICT_BPS_KEY_PARAM in args:
@@ -325,9 +327,13 @@ class StrictBpsMiddleware(Middleware):
                 message=context.message.model_copy(update={"arguments": stripped})
             )
 
+        # isinstance guards the set-membership test: the middleware reads
+        # raw transport arguments before pydantic validation, so an
+        # off-spec client can send an unhashable value (dict/list) that
+        # would otherwise TypeError instead of hitting the block error.
         if (
             strict_bps_effective()
-            and supplied not in _valid_ack_keys()
+            and not (isinstance(supplied, str) and supplied in _valid_ack_keys())
             and await self._is_registered(name)
         ):
             logger.info("strict-BPS mode blocked keyless write to %s", name)
--- a/src/ha_mcp/strict_bps.py
+++ b/src/ha_mcp/strict_bps.py
@@ -29,6 +29,7 @@ schema-validating clients will send it; the tool bodies never read it —
 from __future__ import annotations
 
 import logging
+import secrets
 from collections.abc import Awaitable, Callable, Sequence
 from typing import Annotated, Any, NoReturn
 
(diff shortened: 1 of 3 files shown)


ghhamcp:

ghhamcp:

Patch76 (approved):
This is a clean fix for #1924 — turning the strict-BPS acknowledgment key from a static opaque token into a plain-English read-receipt phrase with a rotating suffix directly addresses both failure modes: an agent misreading the extract-and-replay round-trip as a credential exfiltration, and a static literal in a public repo satisfying the gate from training data or a stale summary without any actual read.

I traced the gate logic and it holds up:

- **Validation predicate** — `not (isinstance(supplied, str) and supplied in _valid_ack_keys())` short-circuits correctly: `None`, wrong string, empty string, and unhashable off-spec values (dict/list) all block without a `TypeError`, and the key is stripped from the arguments before dispatch so tool bodies never see it.
- **Grace window** — `_valid_ack_keys()` derives the current and previous bucket from a single `time.time()` snapshot, so there is no clock race between the two derivations, and a read taken just before an hourly rotation still validates the write that follows it. A key two rotations old is correctly rejected.
- **Schema-as-protocol** — documenting the value as a public read-receipt in the `BestPracticeKey` description (trusted metadata rather than fetched content) is the right lever for the prompt-injection misread, and the block error now telling the agent the key rotates and must be refreshed closes the stale-replay loop.

Test coverage is thorough — `TestAckKeyShape` pins the prefix, within-bucket stability, cross-bucket rotation, per-process salting, and the schema/ack-line framing, plus middleware tests for the previous-hour grace and rejection of a two-rotations-old key, with the salt and clock pinned so the assertions are deterministic. The old `STRICT_BPS_ACK_KEY` symbol is fully retired with no dangling references, and the e2e suite derives the live key via `current_strict_bps_ack_key()`.

One optional, non-blocking note: `_ACK_KEY_SALT` is per-process, so the strict-BPS gate now shares the single-worker assumption that `approval_queue.py` already documents (a read served by one worker would not validate against another worker's salt). All the standard deployments run single-worker, so this is consistent with the existing invariant rather than a new constraint — a one-line comment cross-referencing that assumption in `strict_bps.py` would make the coupling explicit, but it is not required.

Approving.

PR #1928, opened 2026-07-18
feat: surface connect URLs for every network interface (#1862)

### LINKED ISSUE #1862: [FEATURE] Missing connect URLs when using multiple network interfaces

My homeassistant is in 2 VLANs. So I have configured 2 network interfaces `enp0s19` and `enp0s18`. 19 has for example VLAN 2 with IP `10.0.2.3` and 18 has VLAN 1 with IP `10.0.1.3`.

When installing via HACS and then getting the URLs in the device, it only shows 2 URLs

`http://10.0.2.3:8123/api/webhook/mcp_<insert_string>` `http://10.0.2.3:9584/private_<insert_string>`

but it's missing the URLs from the other interfaces

### Why do you need this?

When using claude from my PC which is in VLAN 1, it cannot access the displayed URL. I need to modify it myself to the corresponding IP

### Additional context

_No response_

Patch76:
Closes #1862

## What does this PR do?

On a Home Assistant host with more than one network interface (multiple NICs or
VLANs), the connect URLs shown at server start — in the startup log, the
Configure-screen hint, and the setup notification — only listed the single host
that `homeassistant.helpers.network.get_url()` resolves. Users on any other
subnet had to hand-edit the IP to reach the server.

This surfaces a connect URL for **every** IPv4 address on an enabled network
adapter, expanding both the webhook line and the direct-access line per adapter:

- Enumeration goes through `homeassistant.components.network.async_get_adapters()`,
  filtered to `enabled` adapters, IPv4 only, in adapter order.
- The canonical `get_url()` host is kept first and deduplicated against the
  adapter list, so the primary URL never moves and never repeats.
- All expansion happens inside the existing `build_connect_urls` choke point, so
  the three display surfaces stay consistent.
- Enumeration is best-effort and display-only: any failure (import, fetch, or a
  malformed adapter entry) degrades to the single resolved host and logs a
  warning — it never fails server bring-up.

Security posture is unchanged: the extra secret-bearing lines reach only the
admin-only startup log and the Configure screen. The persistent notification —
visible to every authenticated user — still carries no connect URLs or secrets
(covered by a regression test).

## Type of change
- [x] ✨ New feature

## Testing
- [x] All automated tests pass (`uv run pytest` — 126 unit tests)
- [x] Code follows style guidelines (`uv run ruff check` + `ruff format --check` + `mypy` clean)

Added unit coverage: multi-interface expansion of both lines, dedup against the
`get_url` host, direct-access line when `get_url` is unavailable, adapter-order
enumeration, degrade-to-empty on error and on a malformed adapter, and a
regression test that the multi-interface path keeps URLs out of the persistent
notification.

## Checklist
- [x] My code follows the project style guidelines
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix/feature works
- [x] New and existing unit tests pass locally

### DIFF
--- a/custom_components/ha_mcp_tools/config_flow.py
+++ b/custom_components/ha_mcp_tools/config_flow.py
@@ -447,7 +447,7 @@ class HaMcpServerOptionsFlow(OptionsFlow):
             data_schema=schema,
             description_placeholders={
                 "versions": await self._versions_hint(),
-                "connect_url": self._connect_url_hint(),
+                "connect_url": await self._connect_url_hint(),
                 "oauth_creds": self._oauth_creds_hint(),
                 "llm_api_docs_url": LLM_API_DOCS_URL,
                 "panel_hint": panel_hint,
@@ -535,7 +535,7 @@ class HaMcpServerOptionsFlow(OptionsFlow):
             f"Server ha-mcp {server_version} ({channel} channel)"
         )
 
-    def _connect_url_hint(self) -> str:
+    async def _connect_url_hint(self) -> str:
         """Return the connect URLs for the options form.
 
         The Configure screen is admin-only, so it shows the real resolved
@@ -555,10 +555,13 @@ class HaMcpServerOptionsFlow(OptionsFlow):
         hass = getattr(self, "hass", None)
         if hass is not None:
             try:
-                from .embedded_setup import build_connect_urls
+                from .embedded_setup import async_get_lan_hosts, build_connect_urls
 
                 urls = build_connect_urls(
-                    hass, self.config_entry, webhook_enabled=webhook_enabled
+                    hass,
+                    self.config_entry,
+                    webhook_enabled=webhook_enabled,
+                    extra_hosts=await async_get_lan_hosts(hass),
                 )
                 if urls:
                     return "Connect URL(s):\n" + "\n".join(f"- {u}" for u in urls)
--- a/custom_components/ha_mcp_tools/embedded_setup.py
+++ b/custom_components/ha_mcp_tools/embedded_setup.py
@@ -156,6 +156,7 @@ async def async_bring_up_server(hass: HomeAssistant, entry: ConfigEntry) -> None
             entry,
             auth_mode,
             webhook_enabled=webhook_enabled,
+            extra_hosts=await async_get_lan_hosts(hass),
             oauth_client_id=oauth_client_id,
             oauth_client_secret=oauth_client_secret,
             oauth_creds_active=oauth_creds_active,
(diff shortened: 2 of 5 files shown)


ghhamcp:

Patch76:
## Implementation Summary

**Choices Made:**

- **IPv4-only enumeration.** `async_get_adapters` returns a bare IPv6 `address` with a separate `scope_id` int (the `%scope_id` zone suffix is synthesized later by `async_get_enabled_source_ips_from_adapters`, not by `async_get_adapters`), so a link-local adapter address is not a usable URL host without rejoining that zone id, and every IPv6 host would additionally need bracket-wrapping. The reported setups are IPv4, so `async_get_lan_hosts` returns IPv4 addresses only rather than surfacing unusable URLs.
- **`network` in `after_dependencies`, not `dependencies`.** The adapter lookup is display-only and fully exception-guarded, so it must never fail component setup - a soft dependency is the honest declaration. `network` is a Home Assistant `system` default integration (always set up during bootstrap), so ordering is still guaranteed. This also matches the existing guarded-optional `cloud` import already declared in `after_dependencies`.
- **Canonical `get_url` host stays first and is deduplicated.** The single host `get_url` resolves is kept as the primary URL and never moves or repeats; any duplicate of it in the enumerated adapter list is removed. Users who only need the primary URL see no change in the first line.
- **All expansion lives in the `build_connect_urls` choke point.** The three surfaces (startup log, Configure-screen hint, setup notification) are rendered from this one function, so per-interface expansion stays consistent across all of them without touching each surface separately.
- **Best-effort degrade.** Any failure in enumeration - import error, fetch error, or a malformed adapter entry - degrades to the single `get_url` host and logs a warning with the traceback, rather than propagating. The whole extraction (not just the fetch) sits inside the guard.
- **Security posture unchanged.** The extra per-interface direct-access lines carry the secret path, so they reach only the admin-only startup log and the admin-only Configure screen. The persistent notification - visible to every signed-in user - still carries no connect URLs or secrets. A regression test asserts the multi-interface path keeps URLs out of the notification.
- **Sync builder + async enumeration helper.** `build_connect_urls` stays synchronous and gains an `extra_hosts` parameter; the async adapter lookup happens in `async_get_lan_hosts` at the two call sites. This keeps the builder's unit tests event-loop-free and limits the change surface.

**Problems Encountered:**

- **Hassfest failure.** Importing `homeassistant.components.network` requires the integration to be declared in the manifest; hassfest flagged `[DEPENDENCIES] Using component network but it's not in 'dependencies' or 'after_dependencies'`. Fixed by adding `network` to `after_dependencies` (rationale above). All other CI was green before the fix.
- **Guard scope during self-review.** An early version wrapped only the `async_get_adapters` fetch in the `try`, leaving the per-adapter extraction outside it. A malformed adapter entry would then raise out of `async_get_lan_hosts` into `async_bring_up_server`'s handler, which tears down the running server and files a start-failure repair issue - for a display-only lookup. Widened the guard to wrap the whole extraction and verified by fault injection (malformed adapter -> red without the fix, green with it).
- **Degrade log level.** The degrade path initially logged at `debug` with no exception detail. Raised it to `warning` with `exc_info=True` for parity with the sibling catch in `config_flow._connect_url_hint`, so a silently-empty interface list is diagnosable.

**Testing:** Unit tests, `ruff check`, `ruff format --check`, and `mypy` all clean locally, and CI is green. Coverage added for the multi-interface expansion, the bring-up-to-surface `extra_hosts` forwarding, the port-less URL swap, and the `<home-assistant-ip>` placeholder branch.

ghhamcp:

kingpanther13 (changes requested):
Four-agent review (code, tests, comments, silent-failures) at commit `d5b41d25`. The failure containment is genuinely well engineered — visible warning-not-silent degrade, whole-extraction inside the try so a malformed adapter can't tear down bring-up, `network` correctly a soft `after_dependencies` entry, keyword-only params, and the notification secret-exclusion invariant preserved with a regression test. Version policy checks out too: riding pending 1.2.0 is correct (master already leads stable 1.1.0), and `MIN_COMPONENT_VERSION` correctly untouched for a display-only change. The `await`/dict shape was verified against real HA core (`async_get_adapters` is async and offloads `ifaddr` internally — no event-loop blocking). Three things before merge:

**1. Confirm the `enabled`-adapters semantics cover the reporter's config.** In HA's default *auto* network mode, `enable_auto_detected_adapters` enables only the default-route adapter — on such a host `async_get_lan_hosts` returns exactly the `get_url` host, dedup collapses it, and this feature is a silent no-op. That filter is the right call (it's also what keeps docker/VPN/link-local noise out), but #1862 only closes if the reporter's two interfaces are actually *enabled* in Settings → Network (their screenshot suggests configured; enabled-in-auto-mode is the open question). Worth confirming on the issue — and if manual enabling is required, saying so there when this ships so the dev-channel verification doesn't read as "no change."

**2. The bring-up call site's `extra_hosts` forwarding is unasserted** (`embedded_setup.py:159`). The `TestBringUp` spies on `_surface_connect_urls` assert the oauth kwargs but never `extra_hosts` — deleting the kwarg leaves every test green while killing the feature on its primary surface (the startup log). The config-flow side has exactly this forwarding assertion; please mirror it in `test_success_starts_registers_and_surfaces` with a monkeypatched `async_get_lan_hosts`.

**3. The IPv6 docstring rationale describes the wrong API** (`embedded_setup.py:238-239`). `async_get_adapters` returns a bare `address` with a *separate* `scope_id` int — no `%scope_id` suffix (that suffix is synthesized by `async_get_enabled_source_ips`, the API discussed on the issue but not the one used here). The IPv4-only decision stands on its own; please just fix the mechanism claim so a future IPv6 pass isn't misled.

Minor, optional: `_swap_url_host` on a port-less/hostname internal URL (reverse proxy) produces `https://<lan-ip>` entries that won't pass TLS — display-only and admin-gated, but the `port is None` branch is also untested, so one test would pin intended behavior; same for the `<home-assistant-ip>` placeholder branch. Per-adapter (rather than all-or-nothing) degrade on a malformed entry would be more resilient, but the current behavior is documented and tested — fine as is.

Patch76:
Thanks - all three addressed in be8facba, plus the two optional branch tests.

**1. `enabled`-adapters semantics.** The `enabled` filter is deliberate (it is what keeps docker/VPN/link-local noise out). I checked the auto-mode behaviour against HA core: `_ifaddr_adapter_to_ha` marks only the default-route external adapter `auto`, and `async_load_adapters` then applies a fallback - if no adapter is both `default` and `auto`, every adapter with an external address is promoted to `auto`; `enable_auto_detected_adapters` enables just the `auto` set. So when HA can resolve a default route (the reporter's case - `get_url` already returns the 10.0.2.3 VLAN), only that adapter is auto-enabled and the second VLAN stays disabled, making `async_get_lan_hosts` a no-op until both interfaces are enabled under Settings - Network. Only when no default+auto adapter exists does the fallback auto-enable every external adapter. I will spell out the manual-enable requirement on #1862 when this ships so the verification does not read as "no change".

**2. `extra_hosts` forwarding unasserted.** Fixed. `test_success_starts_registers_and_surfaces` now monkeypatches `async_get_lan_hosts` and asserts `_surface_connect_urls` receives the enumerated hosts, mirroring the config-flow forwarding assertion. I verified by injection that deleting the kwarg turns it red (`KeyError: 'extra_hosts'`).

**3. IPv6 docstring mechanism.** Corrected in both the docstring and the Implementation Summary. `async_get_adapters` returns a bare IPv6 `address` with a separate `scope_id` int; the `%scope_id` suffix is synthesized in `async_get_enabled_source_ips_from_adapters` (`IPv6Address(f"{ipv6['address']}%{ipv6['scope_id']}")`), not by `async_get_adapters`. The IPv4-only decision stands on its own; the rationale now names the right API.

**Minor.** Added `test_portless_internal_url_swaps_host_without_a_port` (the `port is None` swap) and `test_direct_line_uses_placeholder_when_no_host_resolves` (the `<home-assistant-ip>` fallback), pinning both previously-untested branches.

kingpanther13 (approved):
All three items addressed on 751db5ef — verified in the diff: the bring-up forwarding assertion mirrors the config-flow one (and goes red without the kwarg), the IPv6 docstring now names `async_get_adapters`'s actual shape (bare address + separate `scope_id` int), and the enabled-adapters analysis matches core, including the no-default-route fallback — documenting the manual-enable step on #1862 at ship time closes that loop. Both optional branch tests landing too (port-less swap, placeholder fallback) is appreciated. CI is green on the final head. The real acceptance gate remains the post-merge dev-channel check on a multi-VLAN host per AGENTS.md — the #1862 reporter is well positioned to confirm.

kingpanther13:
Verification note (mirrored from https://github.com/homeassistant-ai/ha-mcp/issues/1862#issuecomment-5012322370): we need to verify this immediately after it lands on master — every master commit publishes a dev-channel build. HA's auto network mode enables only the default-route adapter, so the per-interface URLs appear only for adapters explicitly enabled under **Settings → System → Network**; on a default config this change is a deliberate no-op. The real-world confirmation is the reporter's multi-VLAN host with both adapters enabled — CI cannot reproduce that topology.

PR #1929, opened 2026-07-18
fix: state which server ha_dev_manage_server update_source actually changes

kingpanther13:
## What does this PR do?

`ha_dev_manage_server(action="update_source")` only ever changes the ha_mcp_tools custom component's separate in-process server entry — but its docstring and success payload didn't say so. During live testing of #1926, an agent connected to the add-on called it, read `success: true` plus "the component is reloading the in-process server", and concluded the server it was talking to would change version. It never does, which presents as "the install silently failed" and triggers a debugging spiral. `success: true` on this path means HA accepted the entry's options-flow submit — not that the install finished (that runs in the component in the background and can still fail), and not that the serving server changed.

- The non-embedded `update_source` success payload now carries `target: "ha_mcp_tools in-process server entry"` and a note stating that the server handling this connection is NOT affected, that the reinstall continues in the background and can still fail, and that verification belongs on the in-process server's own connect URL. The embedded payload gets a matching `target` ("this server …").
- `info` labels `component_server_entry` with a `role` field so its channel/pip_spec can't be conflated with the serving server's `server_version`/`deployment_mode`.
- The tool docstring and the `action` field description state the target rule explicitly. The missing-entry case already fails fast with `COMPONENT_NOT_INSTALLED` — unchanged.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing
- [x] All automated tests pass (`uv run pytest`) — three regression tests added in `tests/src/unit/test_tools_dev.py` (written first, red → green)

## Checklist

### DIFF
--- a/src/ha_mcp/tools/tools_dev.py
+++ b/src/ha_mcp/tools/tools_dev.py
@@ -834,7 +834,9 @@ class DevTools:
             Field(
                 description=(
                     "info: deployment/version report; update_source: point the "
-                    "in-process server at a channel or pip spec and reinstall; "
+                    "ha_mcp_tools component's separate in-process server at a "
+                    "channel or pip spec and reinstall it (never changes the "
+                    "server serving this connection, unless embedded); "
                     "restart: restart this server"
                 )
             ),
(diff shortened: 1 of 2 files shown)


ghhamcp:

ghhamcp:

PR #1931, opened 2026-07-18
fix: Escape translated catalog strings before rendering as HTML in settings UI

kingpanther13:
## What does this PR do?

Resolves all 8 open CodeQL code-scanning alerts (`js/xss-through-dom`, alerts 33–40) in `settings_ui/settings.js`, and closes the CI gap that let them reach master unflagged.

**The fix.** The settings-UI localization system (#1913) interpolates translation-catalog strings into `innerHTML` unescaped so they can carry inline markup. The locale catalogs are community-editable data (and CodeQL treats the embedded `ha-mcp-i18n` payload as a DOM-text source), so raw interpolation is an HTML-injection surface.

A new `tHtml()` helper renders catalog strings for HTML contexts: the whole value is escaped with `escapeHtml()`, then only the fixed allowlist of formatting tags the catalogs use (`<code>`, `<strong>`, and the internal `<a href="#" data-panel-link="...">` tab links) is restored; `{placeholders}` are substituted after escaping with caller-built, pre-escaped fragments. All raw interpolation sites now go through it (`[data-i18n-html]` static translations, tool-row beta/env-pinned notes, `envLockedNoteHtml()`, filesystem custom-paths copy, code-mode saved-tools-path note), and the plain-text `BACKUP_ORIGIN_LABELS` values are escaped. The `[data-i18n-html]` branch now also skips rewriting when no catalog entry exists instead of round-tripping the server-rendered markup through `innerHTML`.

Also replaces a literal NUL byte in a string with the `'\0'` escape (it made tools treat `settings.js` as a binary file).

**The CI gap.** These flows passed PR CI because `codeql-quality.yml` runs only the `<language>-code-quality` suites, which contain no security queries, while the security suite (GitHub default setup) analyzes master post-merge and posts no blocking PR check — the alerts appeared one minute after #1913 merged. The workflow now feeds the default `<language>-code-scanning` security suites through the same fail-on-any-finding gate for both languages, so the exact queries default setup runs are blocking at PR time. Job/check names are unchanged so any required-check references keep working.

**Security-suite triage.** The new gate's first run surfaced 28 python findings: two were real and fixed in code (the webhook-proxy dev flavor now writes the proxy-config handoff file via `_atomic_write_0600` like the creds file; the ui_panel locale-cookie regex became a linear character-walk validator); the rest are verified false positives or reviewed-by-design patterns, each allowlisted in `codeql_quality_gate.py` with its rationale.

**Webhook-proxy dev versioning.** The dev flavor's version line still counted from its 1.2.2 fork point while stable moved to 2.0.2, sorting dev behind the stable it promotes into. Re-based to `2.0.3.dev1` (next stable patch + `.devN`) and the rule is now documented in the shared AGENTS.md. The stable flavor's proxy-config write is fixed via the promote flow after merge (the stable-guard blocks direct edits here).

**Hardening from review.** Catalog loading now rejects messages whose inline markup `tHtml` cannot restore (mirroring the placeholder-parity check), so a mistyped tag in a community translation fails fast at load instead of silently rendering literal text. Added tests: proxy-config persistence/permissions, locale-validator edge cases, and hostile-catalog sanitization.

## Type of change
- [x] 🐛 Bug fix
- [x] 🔧 Maintenance/refactor

## Testing
- [x] All automated tests pass (`uv run pytest`) — new `TestTranslationHtmlSanitization` jsdom tests (hostile catalog neutralized, allowlist restored, post-escape placeholder substitution) plus existing localization, env-pinned, parse, and i18n suites

## Checklist

### DIFF
--- a/.github/workflows/codeql-quality.yml
+++ b/.github/workflows/codeql-quality.yml
@@ -1,6 +1,15 @@
 name: CodeQL Code Quality
 
-# Runs the CodeQL "<language>-code-quality" suites and fails the PR on ANY finding.
+# Runs the CodeQL "<language>-code-quality" suite PLUS the default
+# "<language>-code-scanning" security suite per language and fails the PR on
+# ANY finding.
+#
+# Why the security suite is gated here too: GitHub default setup (the
+# Security tab) only analyzes master after merge and posts no blocking check
+# on PRs, so a PR can introduce security-query findings (e.g.
+# js/xss-through-dom) that surface as code-scanning alerts minutes after
+# merge — that is how alerts 33-40 landed. Running the same default
+# code-scanning suite in this gate catches them pre-merge.
 #
 # Languages gated (one matrix leg each), both scanning the whole tree
 # (--source-root .):
@@ -68,8 +77,10 @@ jobs:
         include:
           - language: python
             suite: codeql/python-queries:codeql-suites/python-code-quality.qls
+            security_suite: codeql/python-queries:codeql-suites/python-code-scanning.qls
           - language: javascript
             suite: codeql/javascript-queries:codeql-suites/javascript-code-quality.qls
+            security_suite: codeql/javascript-queries:codeql-suites/javascript-code-scanning.qls
 
     steps:
       - uses: actions/checkout@v7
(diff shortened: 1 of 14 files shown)


ghhamcp:

ghhamcp:

PR #1960, opened 2026-07-18
fix: join the entity registry on the ha_config_list_helpers legacy fallback

### LINKED ISSUE #1945: [BUG] ha_config_list_helpers legacy fallback skips the registry join - stale entity_id/name for renamed helpers
### 🐛 Runtime Bug

**💬 What Happened?**

`ha_config_list_helpers` still returns stale `entity_id`/`name` for renamed helpers when it is served via the legacy fallback path – exactly the staleness #1794/#1818 set out to fix.

**Details**

#1818 wired `_enrich_helpers_with_current_registry` into the inline body of `ha_config_list_helpers` – that is the only callsite. `_legacy_helper_list` never calls it, and its own docstring records that it was deliberately kept as a separate copy so the registry-join PR could patch the inline body without conflict; the join was then never applied to it.

That fallback is live, not hypothetical: `_list_helpers_via_component` routes to `_legacy_helper_list` whenever the component's `helpers_list` command fails with anything other than `unknown_command` (component-handler bug, WS timeout). On that path:

- a renamed helper is served with the stale id/name; the appended warning says "served via legacy path" but not that the registry join was skipped;
- person-type responses additionally skip `_flatten_helper_list_result`, so `helpers` comes back as the raw `{"storage": [...], "config": [...]}` dict and `count` is wrong.

`test_raised_command_falls_back_with_warning` exercises this branch but only asserts the warning string, never `entity_id` – the gap is untested.

**Fix sketch:** route the fallback result through the same enrichment + flattening as the inline path; extend the test to assert `entity_id`/`original_name` on the fallback.

Found while stress-testing my review tooling against merged PRs; verified by hand on current master.

Patch76:
Closes #1945

## What does this PR do?

`ha_config_list_helpers` serves a renamed helper with its current entity_id and
display name by joining the entity registry (#1794). That join lived only on the
inline `{type}/list` path. The component-error fallback `_legacy_helper_list`
(reached from both the command-error and the transport-failure branches of
`_list_helpers_via_component`) flattened its result but never called
`_enrich_helpers_with_current_registry`, so a UI-renamed helper was served with
its stale storage id/name on that path. This routes the fallback through the same
registry join and threads its degrade-open warnings into the response.

The all-types merge (`helper_type="all"`) consumes `_legacy_helper_list` for
uncovered simple types and previously dropped its warnings, harmless only while
that list was always empty; it now propagates them, so a failed registry read is
not served silently stale.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

Extended `test_ha_config_list_helpers_component_routing.py`: enrichment is now
asserted on both fallback branches, plus a new test that the all-types merge
surfaces the stale-join warning. Ran that file with the registry-join, rename,
and pagination siblings (67 passed); ruff, ruff-format, and mypy are clean on the
changed source file.

## Checklist

### DIFF
--- a/src/ha_mcp/tools/tools_config_helpers.py
+++ b/src/ha_mcp/tools/tools_config_helpers.py
@@ -4001,11 +4001,11 @@ class HelperConfigTools:
     async def _legacy_helper_list(self, helper_type: str) -> dict[str, Any]:
         """Legacy ``{helper_type}/list`` success envelope, for the § 4 #3 fallback.
 
-        A faithful copy of the tool's inline legacy success path, kept separate
-        (rather than extracted from that body) so the #1794 registry-join PR can
-        patch the inline body without a conflict here. The drift is bounded to
-        the rare component-error fallback and is flagged by the ``warnings[]``
-        entry the caller appends.
+        A copy of the tool's inline legacy success path, kept separate rather
+        than extracted from that body. It flattens and joins the entity registry
+        (issue #1945) exactly like the inline path, so a renamed helper served on
+        this fallback carries its current entity_id/name; the caller additionally
+        appends a ``warnings[]`` entry flagging that the component path was used.
         """
         result = await self._client.send_websocket_message(
             {"type": f"{helper_type}/list"}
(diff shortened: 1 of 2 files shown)


ghhamcp:

ghhamcp:

PR #1961, opened 2026-07-18
fix: raise MIN_COMPONENT_VERSION to 1.2.0 so stale 1.1.0 builds fail the gate

### LINKED ISSUE #1946: [BUG] MIN_COMPONENT_VERSION floor equals an already-shipped component version - gate passes builds lacking the gated behaviors
### 🐛 Runtime Bug

**💬 What Happened?**

The server-side component version gate cannot distinguish component builds that predate #1882's behaviors from builds that have them, because the floor was raised to a version string that had already shipped.

**Details**

#1882 raised the floor in `src/ha_mcp/tools/tools_filesystem.py`:

```python
MIN_COMPONENT_VERSION = "1.1.0"
```

to gate two component behaviors it depends on (packages-aware `list_files`, `read_file(include_parsed=...)`). But component `1.1.0` had already been opened and shipped by an earlier PR without those behaviors, and #1882 did not bump the component version. An install still running that earlier `1.1.0` build passes the version check and then hits raw failures (unrecognized `include_parsed` service arg, "Path not allowed" for a packages folder) instead of the actionable "update the component" message this gate exists to produce.

Master has since moved the component version onward (currently `1.2.1`), which shrinks the exposure window but does not close it: any install that grabbed the earlier `1.1.0` build remains indistinguishable, and the floor still reads `1.1.0`.

**Fix sketch:** bump the floor to the current component version in the next PR touching this surface, and add a note to the AGENTS.md version-bump rule that a `MIN_COMPONENT_VERSION` raise must always target a version *opened in the same PR*, never an already-shipped one.

Found while stress-testing my review tooling against merged PRs; verified by hand on current master.

Patch76:
## What does this PR do?

Fixes #1946. `MIN_COMPONENT_VERSION` in `src/ha_mcp/tools/tools_filesystem.py` was set to `1.1.0` – a component version that had **already shipped** before it carried the behaviours the gate protects (packages-aware `list_files`, `read_file(include_parsed=...)`). #1882 added those behaviours to the component but under an un-bumped `1.1.0`, so a stale `1.1.0` install passes the version check and then hits raw failures (unrecognised `include_parsed` arg, "Path not allowed" for a packages folder) instead of the actionable "update the component" message the gate exists to produce.

This raises the floor to `1.2.0` – the first component version cut after #1882 (commit `9e808087`), and therefore the first that guarantees both behaviours. Any build reporting `1.2.0`+ postdates #1882; `1.1.0` is ambiguous (it splits into builds with and without the behaviours) and must be rejected. Flooring at the current `1.2.1` would over-reject valid `1.2.0` installs, so `1.2.0` is the tight floor.

Changes:
- Raise `MIN_COMPONENT_VERSION` `1.1.0` → `1.2.0` and rewrite the rationale comment to explain why `1.2.0`, not `1.1.0`.
- Add regression test `test_stale_1_1_0_component_rejected` pinning that a component reporting `1.1.0` is rejected as "too old". The assertion holds for any floor above `1.1.0`, so it also guards against a future change that lowers the floor back.
- Harden the AGENTS.md component version-bump rule: a PR that adds a server-gated behaviour must open a fresh pending component version and floor at it, never at a version any pre-behaviour build also reports. This overrides the "already ahead of stable → do not bump" rule for that one case.

**Testing:** the affected unit suites pass locally (`test_caller_token_auth`, `test_tools_filesystem`, `test_yaml_read_tool`, `test_yaml_config_tool`, `test_component_api`, `test_constants`); `ruff check` clean on the changed files. The full suite runs in CI.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

## Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -661,13 +661,21 @@ fully validate a component change before merge.
     go straight to that minor, not an extra patch. Never go past the current
     pending version otherwise; per-revision bumps skip never-shipped numbers and
     desync the version from the release cycle.
-- **When the change adds a service or argument the server depends on**, raise
-  `MIN_COMPONENT_VERSION` in `src/ha_mcp/tools/tools_filesystem.py` to match the
-  pending version in the same PR. `get_caller_token` reports the manifest
-  version and the server gates on it, so without the gate the old and new
-  component are indistinguishable: a caller on the old version passes the check
-  and then hits raw "service not found" errors instead of an actionable
-  "update" prompt.
+- **When the change adds a service or argument the server depends on**, this PR
+  must **open a fresh pending component version** (bump `manifest.json` +
+  `COMPONENT_VERSION`) and raise `MIN_COMPONENT_VERSION` in
+  `src/ha_mcp/tools/tools_filesystem.py` to that same new version. This is the
+  one case that **overrides** the "already ahead of stable → do not bump" rule
+  above: bump here even if a pending version already exists. `get_caller_token`
+  reports the manifest version and the server gates on it, so without the gate
+  the old and new component are indistinguishable: a caller on the old version
+  passes the check and then hits raw "service not found" errors instead of an
+  actionable "update" prompt. **Never floor at a version any build lacking the
+  behaviour also reports** – an already-shipped version, or a pending version
+  that was opened *before* this behaviour landed. Such a build passes the gate
+  yet lacks the behaviour, defeating the gate (#1946: the floor was set to a
+  1.1.0 that had already shipped without the gated behaviours, so 1.1.0 builds
+  split into with/without and the gate could not tell them apart).
 - **Keep the component backward-compatible with the released server.** The
   component (HACS) and the server (add-on / PyPI / Docker) follow the same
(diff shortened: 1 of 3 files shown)


ghhamcp:

ghhamcp:

kingpanther13:
Just fyi we might start getting some bug reports , since we've been rapidly increasing the custom component version we might run into some people still on 1.1.0 and getting the server automatically updated. Hacs seems to lag behind the actual server updates.

Patch76:
I was also still on version 1.1.0 until this morning; I got a prompt to update. It was a bit confusing—I had to figure out the version numbers in my head first. 7.14.0 vs. 1.1.0/1.2.1, and the DEVs on top of that. 

Yes, I also had to download the 1.2.1 update manually; HACS hadn't issued a notification yet.

PR #1962, opened 2026-07-18
fix: prefer an HTTP-500 sample in the deep-search failed fragment

kingpanther13:
## What does this PR do?

Post-merge polish on #1930 (the #1784 follow-up), closing out the review-round findings:

- `record_first_failure()` now upgrades a non-500 sample once when the first HTTP 500 arrives. Failure order under `asyncio.gather` tracks latency, so a fast-failing outlier (e.g. a connection blip) could claim the sample slot and suppress the `http_500_diagnosis_hint` diagnosis in exactly the mixed-failure runs it exists for. Still at most two `summarize_fetch_error` calls per pass.
- The 500 detection is a shared regex predicate (`HTTP 500` followed by `:` or end-of-string) used by both the upgrade and the hint, replacing the `startswith("HTTP 500:")` coupling — a bare `HTTP 500` sample (empty error body) no longer silently drops the hint, and the two helpers can't disagree on what counts as a 500.
- Call-site comments corrected: the guard lives in `record_first_failure`, not the closures.
- Tests: pins the scene per-id 404 asymmetry disclosed in #1930's description (`e.g. HTTP 404` sample, no 500 hint); an automation seam test through public `deep_search`; `record_first_failure` selection semantics including an at-most-two-summaries spy; bare/near-miss hint predicate cases; the both-types-500 double-hint composition; `failed_sample is None` asserts on the existing timeout/yaml unpacks.

Unit tests for the changed surface pass locally (156 passed across the three deep-search test files, two of which this PR touches); full suite on CI.

## Type of change
- [x] 🐛 Bug fix

## Testing

## Checklist

### DIFF
--- a/src/ha_mcp/tools/smart_search/_deep.py
+++ b/src/ha_mcp/tools/smart_search/_deep.py
@@ -393,12 +393,12 @@ class DeepSearchMixin(SceneSearchMixin):
         failed_count = 0
         yaml_skipped_count = 0
         timeout_count = 0
-        # One representative summary — only the FIRST ``failed``-class
-        # exception is kept (the fetch closure guards the append); it rides
-        # partial_reason as an ``e.g.`` (#1784 follow-up). The remaining
-        # failures are counted (``failed_count``) but not summarized, so the
-        # motivating "every per-id fetch 500s" case does N-1 fewer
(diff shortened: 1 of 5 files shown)


ghhamcp:

ghhamcp:

PR #1968, opened 2026-07-19
ci: never fail the mirror gate open on a skipped triggering run

kingpanther13:
## What does this PR do?

Closes the mirror-sync gate hole that cut three accidental stable tags on ha-mcp-integration (v1.1.1, and yesterday v1.2.0 and v1.2.1 — alert issues #1939 / #1957).

Every non-hotfix PR merge produces a wholesale-skipped "Hotfix Release" run. Job records for a skipped run are created asynchronously, seconds after the completed event fires, so the gate's immediate jobs query races them and can come back empty — falling into the job-absent fail-open branch, which tagged whatever pending component version sat on master as a stable mirror release. The gate now short-circuits to "no release" when the triggering run's conclusion is `skipped` (delivered synchronously in the event payload), before the jobs query: a run skipped in its entirety performed no release.

Fail-open behavior is unchanged for runs that actually executed (API errors, genuinely absent job), and the out-of-cycle alarm issue stays in place behind it.

## Type of change
- [x] 🐛 Bug fix

## Testing
- Workflow YAML validated; the failure mechanism was confirmed against the two live misfiring runs behind #1939 and #1957 (both skipped `pull_request` Hotfix Release shells whose job records had not yet materialized when the gate queried — the v1.2.1 gate logged `Semantic Release job conclusion: <job absent>` at 23:13:34 while the job record was created at 23:13:36).

## Checklist
- [x] I have updated documentation if needed (workflow comments updated to describe the real mechanism)

### DIFF
--- a/.github/workflows/sync-integration-mirror.yml
+++ b/.github/workflows/sync-integration-mirror.yml
@@ -73,11 +73,19 @@ jobs:
   # The mirror tag depends only on the "Semantic Release" job (the name both
   # semver-release.yml and hotfix-release.yml use — commented as load-bearing
   # there), so gate on that job's conclusion instead. workflow_run fires for
-  # EVERY completed run of the listed workflows; the all-skipped Hotfix
-  # Release runs that every non-hotfix PR merge produces conclude "skipped",
-  # which never satisfied the old success gate either — the job-level check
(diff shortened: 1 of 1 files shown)


ghhamcp:

ghhamcp:

PR #2013, opened 2026-07-23
Fix required section defaults in config flows

### LINKED ISSUE #2011: ha_set_integration fails to submit the required "advanced" section for Generic Camera (section() config-flow steps)
## Problem

Setting up the built-in "Generic Camera" integration (`domain="generic"`) via `ha_set_integration` fails when the config flow reaches its collapsed "Advanced" section. Home Assistant's `homeassistant/components/generic/config_flow.py` defines this as a **required**, nested `section()`:

```python
vol.Required(SECTION_ADVANCED): section(vol.Schema(advanced_section), {"collapsed": True})
```

`SECTION_ADVANCED = "advanced"`, and the nested schema itself has required fields (`framerate`, `verify_ssl`).

Five different submission attempts were tried against this step, all failing identically with `"advanced: required key not provided"`:

1. Only `stream_source` (baseline, expected to fail)
2. `stream_source` + `advanced` as a nested object with 3 sub-fields
3. `stream_source` + `advanced` as a nested object with 4 sub-fields
4. `stream_source` + `advanced` as a nested object with 2 sub-fields
5. `stream_source` + all advanced fields flattened at the top level (no `advanced` key at all)

In every attempt that included the advanced fields (nested or flattened), the response's `submitted_keys` came back as an **empty array** (`[]`) — even `stream_source`, which had been accepted on an earlier step, was lost from the submission. No config entry was ever created (`ha_get_integration(domain="generic")` confirmed `total_count: 0` throughout), so this is a clean, 100% reproducible flow-consume failure rather than a partial/corrupted state.

This happened independently on two separate camera setups (both hitting the same required `advanced` section).

## Suspected cause

`src/ha_mcp/tools/config_entry_flow.py`'s `_consume_form_schema()` has generic logic for recursing into nested schema sections (checking a field's `schema` for a list of sub-fields, and supporting both flat and explicitly-nested caller input), but it doesn't appear to correctly populate `user_input` for a voluptuous `section()` wrapper when the section itself is `vol.Required` and its inner fields are also required — the section key isn't ending up in the submitted payload at all, regardless of how the caller structures the input.

## Proposed fix

Add explicit handling/tests for voluptuous `section()`-wrapped config flow steps (the "expandable" collapsible groups used by several newer HA integrations), particularly when the section itself is required. The built-in Generic Camera integration is a convenient, always-available reproduction case since its `advanced` section is required by default.

## Environment

- ha-mcp (Python package)
- Home Assistant Core 2026.7.3

cyphercodes:
## What does this PR do?

Fixes #2011.

When Home Assistant returns a required expandable section with suggested/default values, `ha_set_integration` now includes those defaults in the submitted form data. This lets Generic Camera flows submit the required `advanced` section even when the caller only provides fields like `stream_source`, while still allowing explicit caller overrides.

AI assistance was used to implement and test this change.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

Targeted tests run:
- `uv run pytest tests/src/unit/test_flow_multistep.py -q`
- `uv run pytest tests/src/unit/test_flow_error_parsing.py tests/src/unit/test_flow_name_injection.py tests/src/unit/test_identifier_validation_family.py -q`
- `uv run ruff format --check src/ha_mcp/tools/config_entry_flow.py tests/src/unit/test_flow_multistep.py`
- `uv run ruff check src/ha_mcp/tools/config_entry_flow.py tests/src/unit/test_flow_multistep.py`
- `git diff --check`

## Checklist

### DIFF
--- a/src/ha_mcp/tools/config_entry_flow.py
+++ b/src/ha_mcp/tools/config_entry_flow.py
@@ -177,9 +177,9 @@ def _record_ignored_section_keys(
 def _field_default_value(field: dict[str, Any]) -> Any:
     """Return a serialized schema field's default/suggested value, if present."""
     description = field.get("description")
-    if isinstance(description, dict) and "suggested_value" in description:
+    if isinstance(description, dict) and description.get("suggested_value") is not None:
         return copy.deepcopy(description["suggested_value"])
-    if "suggested_value" in field:
+    if field.get("suggested_value") is not None:
         return copy.deepcopy(field["suggested_value"])
     if "default" in field:
         return copy.deepcopy(field["default"])
@@ -207,11 +207,10 @@ def _schema_default_values(data_schema: list[Any]) -> dict[str, Any]:
     return defaults
 
 
-def _required_section_defaults(
-    field: dict[str, Any], nested_schema: list[Any]
-) -> dict[str, Any]:
+def _required_section_defaults(field: dict[str, Any]) -> dict[str, Any]:
     """Return default data for a required section, otherwise an empty dict."""
-    if not field.get("required"):
+    nested_schema = field.get("schema")
+    if not field.get("required") or not isinstance(nested_schema, list):
         return {}
     return _schema_default_values(nested_schema)
 
@@ -240,32 +239,29 @@ def _ignored_keys_warnings(
 
 def _consume_section_schema(
     field: dict[str, Any],
-    nested_schema: list[Any],
+    explicit_section: dict[str, Any] | None,
     remaining_config: dict[str, Any],
     ignored_config_keys: set[str] | None,
+    consumed_config_keys: set[str] | None,
     path_prefix: str,
-    missing: object,
-) -> tuple[str | None, dict[str, Any] | Any]:
+) -> dict[str, Any]:
     """Consume config values for a nested flow section."""
-    name = field.get("name")
-    section_name = name if isinstance(name, str) else None
-    section_path = _section_path(path_prefix, name)
-    explicit_section = (
-        remaining_config.pop(section_name, missing)
-        if section_name is not None
-        else missing
-    )
-    if explicit_section is not missing and not isinstance(explicit_section, dict):
-        return section_name, explicit_section
+    nested_schema = field.get("schema")
+    if not isinstance(nested_schema, list):
+        return {}
 
-    nested_data = _required_section_defaults(field, nested_schema)
-    if isinstance(explicit_section, dict):
+    name = field.get("name")
+    section_path = _section_path(path_prefix, name)
+    nested_data = _required_section_defaults(field)
+
+    if explicit_section is not None:
         explicit_remaining = dict(explicit_section)
         nested_data.update(
             _consume_form_schema(
                 nested_schema,
                 explicit_remaining,
                 ignored_config_keys,
+                consumed_config_keys,
                 section_path,
             )
         )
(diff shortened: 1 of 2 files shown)


ghhamcp:

kingpanther13 (changes requested):
The advanced-section defaulting mechanism is correct and verified — against HA's serializer and live on a 2026.7.3 instance: a Generic Camera created from only `stream_source` stored `advanced: {framerate: 2, verify_ssl: true, rtsp_transport: "tcp", authentication: "basic"}`, i.e. the schema's own suggested values. Precedence (flat > explicit-dict > seeded), falsy-override preservation, and the deepcopy against schema mutation are all correct. Four things before merge.

### 1. The create still fails — Generic Camera setup doesn't complete

Live-tested on 2026.7.3: `ha_set_integration(domain="generic", config={"stream_source": "..."})` — the exact call from #2011 — still fails after this change. The `advanced` gate is cleared, but the generic flow then advances to `async_step_user_confirm` (`vol.Required(CONF_CONFIRMED_OK, default=False)`). The walker never satisfies it, loops back through `async_step_user`, and the submission fails with a misleading `no_still_image_or_stream_url` — the seeded stream/still is lost on the looped resubmission. Adding `"confirmed_ok": true` is the only way the entry gets created (verified end-to-end by creating a working camera from a real RTSP stream, then removing it).

`_handle_flow_steps` should carry any flow through its preview/confirm step so the create completes — this is a general walker capability (several integrations end their config flow on a preview/confirm step), not a Generic Camera special-case. The only design point is which steps qualify for auto-advance (e.g. a step flagged `preview=...` with a single required confirmation boolean), applied uniformly across flows.

### 2. Seeded defaults defeat the empty-submission guard (silent-failure regression)

`any_form_key_consumed` is set from `if form_data:` (config_entry_flow.py:854) and gated at line 709 (`... and not any_form_key_consumed`). After this change `form_data` can be non-empty purely from seeded section defaults, so the guard no longer fires when zero caller keys were consumed. For any flow where a defaults-only submission is enough to create an entry, a caller supplying only unmatched/typo'd keys now gets a silent success built from HA's defaults — their values dropped, only a buried warning — where before it raised the explicit "Flow completed without consuming any of the supplied config keys." Derive `any_form_key_consumed` from actual caller-key consumption (a pop from `remaining_config`), not from `form_data` being non-empty.

### 3. Missing regression tests for the required gate and the default branch

- `_required_section_defaults` (line 214): nothing asserts an optional section is NOT seeded. Remove that `if not field.get("required")` gate and every current test still passes — the core invariant is unguarded. Add an optional-section-with-suggested-values case asserting the section stays absent from form_data.
- `_field_default_value` (line 177): only `description.suggested_value` is exercised; the top-level `suggested_value` and `default` branches are untested, and integrations that serialize `vol.Optional(default=X)` land on the `default` branch. Add a case for each.

An e2e over a required collapsed-section flow would also catch the create-completes gap in #1, which unit tests over a hand-built schema can't.

### 4. Refactor drops falsy non-dict section values + return type hides the invariant

`_consume_section_schema` returns the raw non-dict `explicit_section` (line 259), which the caller then filters through `if nested_data:` (line 319) — so a falsy non-dict section value (`{"advanced": 0/false/""/[]}`) is silently dropped, where the previous inline code forwarded it to HA. Moving the non-dict short-circuit back into the caller lets `_consume_section_schema` return plain `dict[str, Any]` instead of `tuple[str | None, dict[str, Any] | Any]` (which collapses to `Any` and hides that the anonymous-section branch is dict-only). One change fixes both.

### Minor cleanups

- Two "absent" sentinels now coexist: module-level `_MISSING_DEFAULT` (line 82) and the per-call `missing = object()` (line 302) passed into `_consume_section_schema`. Consolidate to one and drop the parameter.
- `_required_section_defaults` and `_consume_section_schema` take `nested_schema` as a separate argument though it's always `field.get("schema")` at the call site — derive it inside with an isinstance guard and drop the param.

kingpanther13:
Oh also please address the codex reviews, either fix what they say if you think they're legit, or push back on them. Either way please reply to and resolve the suggestions.

cyphercodes:
Addressed the Codex/maintainer feedback in `0d752b63e76e4bfa3b42fa1250398324b470dd1e`:

- Auto-advance safe HA preview/confirm forms with a single required boolean defaulting to false (`confirmed_ok: true`), so Generic Camera create can complete after the preview step.
- Track actual caller-consumed form keys separately from injected required-section defaults, preserving the all-ignored-config validation error.
- Added regression coverage for optional sections not being seeded, description/top-level/default fallback sources, null suggested-value fallback to defaults, and falsy non-dict section passthrough.
- Simplified section default/consume helpers to derive nested schema internally and keep `_consume_section_schema()` dict-only.

Verification:
- `uv run pytest tests/src/unit/test_flow_multistep.py -q`
- `uv run ruff format src/ha_mcp/tools/config_entry_flow.py tests/src/unit/test_flow_multistep.py`
- `uv run ruff check src/ha_mcp/tools/config_entry_flow.py tests/src/unit/test_flow_multistep.py`
- `uv run pytest tests/src/unit/test_flow_multistep.py tests/src/unit/test_flow_error_parsing.py tests/src/unit/test_flow_name_injection.py tests/src/unit/test_identifier_validation_family.py -q` (189 passed)
- `git diff --check`

kingpanther13 (approved):
All four items verified fixed — in code, in tests, and live against a real Home Assistant 2026.7.3 instance:

- `ha_set_integration(domain="generic", config={"stream_source": "..."})` — the exact call from #2011 — now completes end-to-end: the required `advanced` section is seeded from the schema's suggested values, the preview/confirm step is auto-advanced, and the entry is created and loads against a real RTSP camera (verified live, then removed). The gate set on `_auto_confirm_form_payload` (preview key + single required boolean defaulting false) is appropriately conservative.
- The `consumed_config_keys` split correctly restores the all-keys-ignored guard — live, a typo-only submission still fails with no entry created, and `test_seeded_section_defaults_do_not_count_as_consumed_keys` pins the walker-level contract.
- The new regression tests cover the optional-section gate, all three default sources, the null-suggested-value fallback, and falsy non-dict passthrough.
- `_consume_section_schema` returning plain `dict` with the non-dict short-circuit back in the caller resolves both the falsy-drop and the type collapse.

Thanks for the fast, thorough turnaround.

PR #2025, opened 2026-07-25
test: gate inaddon setup on the addon's Home Assistant link

kingpanther13:
## What does this PR do?

The HAOS inaddon lane gated setup on `wait_for_addon_mcp_ready`, which probes the dev addon's own MCP port (TCP accept, then HTTP OPTIONS). That proves the addon's listener is up, but tools depend on a different layer: the addon reaches Core over the Supervisor WebSocket proxy, and that proxy answers HTTP 502 for a beat after boot while Core comes up behind it. Setup returned while the addon was half-ready, so the session's first HA-backed tool call could land in that window and fail with `CONNECTION_FAILED` — presenting as a product bug rather than a setup race.

Seen on #1997, where a fresh single-test invocation of `test_web_ui_debug_log_level_reaches_addon_log` died on its first call (`ha_get_addon`) with `server rejected WebSocket connection: HTTP 502` on `/addons`, while the same job's main suite passed 1037 tests and a rerun went green.

- Adds `wait_for_addon_ha_link_ready` next to `wait_for_addon_mcp_ready` in `haos_runtime.py` and calls it from the inaddon bring-up: poll `ha_get_addon` until it answers, then let setup proceed. That tool is the probe because it needs both legs of the racing path (the Core WebSocket and the Supervisor API) and is the call that surfaced it.
- On timeout, setup fails pointing at the Supervisor and Core logs in the HAOS diagnostics artifact, instead of leaving the first test to absorb it.
- Mirrors the `haos_embedded` branch a few lines below, which already gates on its webhook answering a real MCP `initialize` rather than on a port being open. Transient errors mirror the suite's canonical `_POLLING_TRANSIENT_ERRORS` set plus `httpx.HTTPError` — the same combination `test_addon_debug_log_level.py` uses — so bugs still propagate per the styleguide's polling-loop rule.

The gate is in shared inaddon bring-up, so it covers every test on that lane, not only the one that flaked.

## Type of change
- [x] 🐛 Bug fix

## Testing

## Checklist

### DIFF
--- a/tests/src/e2e/conftest.py
+++ b/tests/src/e2e/conftest.py
@@ -183,6 +183,11 @@ _EMBEDDED_READY_POLL_S = 5
 # _READY_TIMEOUT_S). pytest.ini's timeout_func_only exempts this session-fixture
 # wait from the 300s per-test timeout.
 _HAOS_EMBEDDED_BRINGUP_TIMEOUT = 600
+# Inaddon: budget for the addon's Home Assistant link after its own HTTP
+# listener answers. Core is usually accepting within a few seconds of the
+# listener, so this is only consumed when something is genuinely wrong.
+_ADDON_HA_LINK_TIMEOUT_S = 180.0
+_ADDON_HA_LINK_POLL_S = 3.0
 
 
 def _is_embedded_backend_selected() -> bool:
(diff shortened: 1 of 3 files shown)


ghhamcp:

ghhamcp:

PR #2028, opened 2026-07-25
fix: bound each addon HA-link probe by the readiness deadline

kingpanther13:
## What does this PR do?

Bounds each Home Assistant link probe by the readiness deadline it runs under.

`_probe_addon_ha_link` (added in #2025) passed no timeout to `Client` or `call_tool`, so an attempt inherited FastMCP's Streamable HTTP default of `httpx.Timeout(30.0, read=300.0)` (`fastmcp/client/transports/http.py:135`, fastmcp 3.4.4 as pinned). The poll loop only consulted the deadline *between* attempts, so a listener that accepted the connection but stalled on initialize or the tool response could hold `wait_for_addon_ha_link_ready(timeout=180)` for ~300s on a single attempt: the fixture's failure and its HAOS diagnostics arrive well past the advertised budget, and the docstring's ceiling was not true.

- Each attempt now receives the time left on the deadline, applied at all three layers that can stall independently: connect/initialize (`init_timeout`), the tool response (`call_tool(timeout=...)`), and the surrounding coroutine (`asyncio.wait_for`).
- The inter-attempt sleep is capped the same way, so `timeout` is exact rather than exact-plus-one-poll.
- Two tests added: one pins that the probe is handed a shrinking, bounded budget (a constant or missing cap turns it red), one drives a probe that consumes its whole budget and asserts the helper stops at the deadline with a single attempt.

Only affects how long a failing gate takes to report — it cannot turn a passing boot into a failure, since a timed-out attempt lands in the same retried transient class as before (`asyncio.TimeoutError` is `TimeoutError`, already in the set).

Reported by Codex on #2025.

Second commit, from the same investigation: the change classifier that gates the e2e suites could skip them on its own failure. `if: needs.changes.outputs.run == 'true'` carries an implicit `success()` on `needs`, so when the classifier job failed on this PR (no steps, no runner assigned) the inaddon suite was skipped — and a skipped required check reports Success to branch protection. The classifier's script is written to fail closed, but that only covers the script picking a value, not the job failing. All four suites now gate on `!cancelled() && (needs.changes.result != 'success' || needs.changes.outputs.run == 'true')`, so a docs-only PR still skips exactly as before while a classifier that did not succeed runs the suite. `pr.yml`'s `e2e-validation-gate` had the mirror of the same hole — a failed classifier leaves `run` empty, which it read as docs-only and passed — and now requires the classifier to have succeeded before honoring the skip.

## Type of change
- [x] 🐛 Bug fix

## Testing

## Checklist


ghhamcp:

ghhamcp:

Issue #2031, opened 2026-07-25
MCP 2026-07-28 spec: our-side compliance items


kingpanther13:
> [!WARNING]
> **Disclaimer:** Everything in this issue — the audit findings, code references, and claims of
> existing compliance — was gathered by Claude (AI) and therefore cannot be trusted as-is.
> Every claim must be independently verified before acting on it.

The MCP 2026-07-28 specification finalizes July 28 (RC since May 21 —
https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/,
https://modelcontextprotocol.io/specification/draft/changelog).

Audit result: ha-mcp is architecturally aligned already — stateless HTTP
(`stateless_http=True`, stateless HMAC OAuth tokens, no server-side session state), zero usage
of the deprecated features (Sampling, Roots, MCP Logging, elicitation, experimental tasks),
deterministic `tools/list` ordering, relays forward the new required headers untouched, and the
ha_auth paths already advertise CIMD. All wire-level protocol changes (statelessness,
`server/discover`, `resultType`, `subscriptions/listen`, headers, cache fields) arrive via the
fastmcp/python-sdk upgrade that Renovate will deliver. Every install method (add-on, HACS
component, Docker, stdio/binaries) ships the same package, so the SDK bump covers them all;
stdio remains a first-class transport in the new spec.

This issue tracks the changes that are ours to make and actionable now:

- [x] **RFC 9207 `iss` in authorization responses.** (merged in #2033) All three authorization-server
      implementations build the redirect in our code and none include `iss`:
      `src/ha_mcp/auth/provider.py` (`construct_redirect_uri` call),
      `custom_components/ha_mcp_tools/oauth_legacy.py` (`_redirect_with`), and
      `homeassistant-addon-webhook-proxy/mcp_proxy/oauth.py` (same pattern). Add `iss` to both
      success and error redirects.
- [x] **Relay stream timeout.** (merged in #2034) `_CLIENT_TIMEOUT = aiohttp.ClientTimeout(total=300, ...)` in
      `custom_components/ha_mcp_tools/mcp_webhook.py` (mirrored in `mcp_proxy`) hard-caps
      streamed responses at 5 min. The new `subscriptions/listen` is a deliberately long-lived
      POST stream — it would be cut every 300 s, forcing clients to re-subscribe. Drop `total`
      (keep `sock_connect`/`sock_read`) so idle detection, not wall-clock, bounds the stream.
[...]

PR #2032, opened 2026-07-26
feat!: remove ha-mcp-sse entry point

kingpanther13:
## What does this PR do?

Removes the legacy `ha-mcp-sse` entry point (deprecated HTTP+SSE transport). The MCP
specification has formally deprecated HTTP+SSE; no install method or documentation advertised
`ha-mcp-sse`, and Streamable HTTP (`ha-mcp-web`) is the supported HTTP mode.

Removal surface: `main_sse` + the `ha-mcp-sse` console script, `fastmcp-sse.json`, the SSE
branches in `_http_run_kwargs` / `ProbeAccessLogFilter` / `register_browser_landing`, SSE
entry-point detection in the bug-report tool, SSE-specific unit tests, and the SSE notes in
SECURITY.md and the setup page.

Review round: adds a test covering the `FASTMCP_TRANSPORT=sse` env-hint detection (now the
sole remaining path to `"sse"` in the bug-report tool) and sweeps stale `HTTP/SSE` comment
phrasing from files outside the removal surface.

Part of #2031.

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [x] 💥 Breaking change

## Testing

## Checklist


ghhamcp:

ghhamcp:

Patch76 (approved):
The removal surface matches the checklist item in #2031, and the sweep at `dd639bc5` comes back clean: `ha-mcp-sse`, `main_sse`, `fastmcp-sse`, `quiet_probe_log` and `drop_mcp_405` have no occurrence left outside CHANGELOG history, and port 8087 survives only as coincidental substrings (a dependency hash, HACS fixtures, a changelog commit link). Every call site of the narrowed signatures – `_http_run_kwargs`, `_run_http_server`, `_run_http_with_graceful_shutdown`, `register_browser_landing`, `ProbeAccessLogFilter` – is updated; the add-on `start.py` and the embedded server go through `browser_landing.register_browser_landing`, whose signature never carried `quiet_probe_log`, so they are unaffected either way. The premise holds at the source as well: the draft spec changelog reclassifies HTTP+SSE as Deprecated under the feature-lifecycle policy (SEP-2596) and points at Streamable HTTP – Deprecated, not Removed, so dropping the entry point is our own call in a breaking release rather than something the spec forces.

Behaviour parity on the surviving path is preserved: `main_web` previously called `_run_http_server("http", default_port=8086)`, which meant `_http_run_kwargs` added `stateless_http=True` and `quiet_probe_log=transport != "sse"` resolved to True – exactly what the hardcoded version now does unconditionally. Running the branch locally, `ha-mcp-web` starts on `transport 'http' (stateless)`, `GET /mcp` answers 405 with the landing body, and an `initialize` POST returns a valid JSON-RPC result; the unit files for browser_landing, graceful_shutdown, http_run_kwargs, healthz, transport_security and settings_secret_path pass together (124 tests), as do the transport-detection tests. Keeping the `FASTMCP_TRANSPORT=sse` branch is the right call – fastmcp 3.4.4 as pinned in `uv.lock` still lists `sse` in its `Transport` literal, so that run shape stays reachable and now has direct coverage.

Two minor notes, neither blocking:

**Note 1:** the PR body says no documentation advertised `ha-mcp-sse`, but the setup page did – at the base commit, `site/src/pages/setup.astro:1803` told readers that "`ha-mcp-sse` entry point serves legacy SSE on port 8087 – pair that with the `url` key (not `httpUrl`) if you need SSE specifically", and this PR removes that paragraph. Worth carrying into the breaking-change note, so a reader who followed it lands on `httpUrl` against port 8086 instead of discovering the missing script.

**Note 2:** pre-existing, not introduced here. The comments at `src/ha_mcp/__main__.py:453`, `:475` and `:1144` state that the dropped raw 405 access line is replaced by an annotated line from the landing handler. In standard mode it is not: `_setup_logging(settings.log_level)` at `:305` calls `logging.basicConfig` without `force`, while `src/ha_mcp/utils/usage_logger.py:87` has already attached the in-memory `StartupLogCollector` to the root logger at import time – so the call is a no-op and the root logger stays at WARNING with no console handler. The `logger.info` at `src/ha_mcp/browser_landing.py:129` therefore never reaches the log; in the local run above the 405 appeared with neither the raw access line nor the annotated one. The OAuth and OIDC entry points avoid this by passing `force=True` and setting the `ha_mcp` logger levels explicitly (`:1429`, `:1575`); `main()` at `:839` has the same shape as `:305`. That belongs in its own change, but since this PR makes the 405 drop unconditional, the comments currently promise a replacement line standard mode does not emit.

PR #2039, opened 2026-07-26
fix: standard-mode console logging + relay-stub upstream contract pins

### LINKED ISSUE #2031: MCP 2026-07-28 spec: our-side compliance items

kingpanther13:
## What does this PR do?

Closes out the actionable review follow-ups on #2031 (the remaining items there are gated on
the fastmcp/python-sdk 2026-07-28 release and ride the future dependency bump):

- **Fix standard-mode console logging** (Patch76's finding on #2032, reproduced + extended):
  `StartupLogCollector` attaches to the root logger at import time, making `_setup_logging`'s
  `basicConfig` a no-op in `ha-mcp-web`/stdio modes — root stayed at WARNING with no console
  handler, so INFO lines (including the annotated 405 probe line the code comments promise)
  never surfaced. `force=True` is now the default, wrapped in `preserve_startup_collector()`
  so the collector is detached before the force sweep (which removes AND closes handlers) and
  re-attached intact. This also repairs OAuth/OIDC modes, whose explicit `force=True` has been
  silently destroying `ha_report_issue`'s startup-log capture on every start.
- **Pin the relay-timeout upstream contracts** (Patch76's note on #2034): real `aiohttp` (new
  dev-group dependency, floored at HA 2026.6's `3.13.5` pin) is compared field-by-field against
  both `ClientTimeout` test stubs — probed in a pristine subprocess, since the unit-test peers
  install a fake `aiohttp` into `sys.modules` and an in-process import would compare the stub
  against itself. sse-starlette's 15 s keepalive default and the MCP SDK's omission of a `ping`
  override (AST-verified across its call sites) are asserted too, so upstream drift fails loud.
- **FAQ migration note** for the removed `ha-mcp-sse` entry point (#2032 review note).
- **Component escalated to 1.3.0** (pending minor per the AGENTS.md escalation rule — the
  merged RFC 9207 `iss` feature warrants more than the pending patch), with the parity test's
  literal pin updated.

Closes #2031.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

## Checklist


ghhamcp:

ghhamcp:

Patch76 (commented):
The logging fix lands the failure mode from the #2032 review at the right place: detaching the collector for the duration of the reconfiguration, rather than re-adding it afterwards, is what keeps `ha_report_issue` on a live handler instead of a closed one. I verified that half by injection: dropping the `preserve_startup_collector()` wrapper while leaving `force=True` in place turns three of the five new tests red.

**[Concern] The `force=True` default is not pinned by the new suite.** Flipping `_setup_logging`'s default back to `force: bool = False` at `src/ha_mcp/__main__.py:481` leaves all five tests in `test_standard_mode_logging.py` green (checked locally at `7a870a5e`). The reason is the state the tests recreate: root carries the collector as its only handler, `preserve_startup_collector` removes it, and `basicConfig` then sees an empty handler list and installs the console handler regardless of `force`. The new regression guard therefore covers the collector-preservation half of the change, not the half the title leads with.

The same mechanism makes the docstring's causality read wrong. "`force` defaults to True because `usage_logger` attaches its `StartupLogCollector` to the root logger at import time: with a root handler already present, `basicConfig` does nothing at all — neither the console handler nor the level is applied" describes the pre-fix situation, but the collector is exactly the handler the wrapper takes out before `basicConfig` runs. Nothing else attaches a handler to the root logger of these processes: `grep -rn addHandler src/` returns `usage_logger` only, the `basicConfig` in `stdio_settings_sidecar.py` belongs to the separate sidecar process, and the component never calls `_setup_logging` at all (the embedded backend builds the server object directly). So `force=True` guards against a *foreign* root handler, not against the collector.

Suggestion for both: add a case that installs a second, non-collector root handler before calling `_setup_logging("INFO")`, then assert the console handler is installed and the level applied. It goes red on a default flip (with a foreign handler present, `force=True` installs the console handler and applies INFO, `force=False` installs none and leaves root at WARNING), and it states what `force` is actually for, after which the docstring can say the same thing.

**[Question] `Closes #2031`.** All three review follow-ups are covered. The issue also carries the block gated on the fastmcp/python-sdk release (CIMD for the standalone provider, removing the dead `Mcp-Session-Id` propagation, the `llm_api.py` client adaptation, explicit `ttlMs`/`cacheScope`, docs sweep plus e2e), and closing #2031 retires that tracker. Carry it in the Renovate bump PR, or file a successor issue first?

`test_relay_timeout_upstream_contract.py` is the strongest part of the PR. Probing the real `ClientTimeout` in a pristine interpreter is the right answer to the `sys.modules` fakes, the assertion messages name the production consequence rather than the assert, and parsing the SDK module for a `ping` override survives reformatting that would defeat a source grep. It also fails loudly instead of skipping when aiohttp is missing, which is what makes it a pin. Lockfile delta is aiohttp plus six transitive packages with nothing else moved, and the 1.2.4 to 1.3.0 escalation matches the AGENTS.md rule: stable carries 1.2.3, so master's 1.2.4 is the single pending version and this raises its level instead of adding one.

kingpanther13:
**Concern**: fixed in `d0dc1e4a` — added `test_force_reconfigures_past_a_foreign_root_handler` (a non-collector root handler installed before setup; red on a `force=False` flip, verified by mutation both ways), and the docstring now attributes `force=True` to foreign-handler / repeat-reconfiguration determinism rather than the collector the wrapper already detaches.

**Question**: closing is deliberate because the gated block has no our-side action until the ecosystem ships — verified against the installed stack: the mcp SDK's CIMD support is client-side only (fastmcp's server-side CIMD lives in its OAuth-proxy mode, not the `OAuthProvider` we subclass), the current SDK still emits `Mcp-Session-Id` so the relays must keep propagating it, and `ttlMs`/`cacheScope` don't exist in the SDK types yet. We deliberately follow fastmcp/python-sdk rather than hand-rolling protocol machinery they will ship. Renovate auto-opens the bump PR when they do, and that PR carries the gated items; #2031's text remains the reference for that work.

Patch76 (commented):
The concern is resolved, and the mutation reproduces in both directions at `d0dc1e4a`: the six tests in `test_standard_mode_logging.py` pass as committed, and flipping the default at `src/ha_mcp/__main__.py:481` back to `force: bool = False` turns `test_force_reconfigures_past_a_foreign_root_handler` red while the other five stay green. That is the discriminating mutation – it moves the default itself rather than the behaviour the default gates – so the new case pins what the original five could not.

One residual, in the rewritten docstring rather than in the code. The parenthetical at `src/ha_mcp/__main__.py:490` attributes the repeat-call scenario to "the OAuth/OIDC entry points reconfigure after settings load", and both halves are inverted against the call graph at this head:

- `main_oauth` (`:1447`) and `main_oidc` (`:1593`) pass `force=True` explicitly, so neither exercises the default at all, and both take the level from `os.getenv("LOG_LEVEL", "INFO")` before any settings load – `main_oauth` never calls `get_settings()`, and `main_oidc` reaches it only at `:1602`, nine lines after the reconfiguration.
- The two sites that do use the default are `main` (`:857`) and `_setup_standard_mode` (`:305`), and those are precisely the ones that run after `get_settings()`.
- No entry point reconfigures twice. Those four are the only `_setup_logging` call sites in `src/`, one per process, so "a repeat call" has no production instance today.

The rest of the sentence stands: a root handler installed by pytest or by a dependency is real, and the new test pins exactly that. Dropping the parenthetical, or replacing it with the test runner / dependency framing the test itself uses, leaves the docstring saying what the suite proves. The complementary fact is worth a clause too – at the two default call sites the only root handler attached anywhere in `src/` is the collector (`usage_logger.py:85` and `:120`), which the wrapper detaches before `basicConfig`, so in the current entry points the default is defensive rather than load-bearing.

On `Closes #2031`, the deferral rationale holds against the installed stack. The SDK's CIMD implementation is client-side – `mcp/client/auth/oauth2.py` and `mcp/client/auth/utils.py`; the only other occurrence is the `client_id_metadata_document_supported` metadata flag on the model at `mcp/shared/auth.py:135`. On the fastmcp side the server support hangs off the proxy line: `OIDCProxy` subclasses `OAuthProxy` and already defaults to `enable_cimd=True` (`fastmcp/server/auth/oidc_proxy.py:248`), which is what OIDC mode picks up at `__main__.py:1583`, while the `OAuthProvider` base that `HomeAssistantOAuthProvider` subclasses carries no CIMD reference at all – so the gap really is confined to the standalone provider. `mcp-session-id` is still written onto responses in `mcp/server/streamable_http.py` (`:307`, `:337`), which keeps the relay propagation load-bearing, and `ttlMs` / `cacheScope` have no occurrence in either package. Carrying the gated block on the Renovate bump PR follows from that, with #2031 left as its written reference.

kingpanther13:
Residual fixed in `3b932a5c` — parenthetical replaced with the test-runner/dependency framing the suite proves, plus the defensive-not-load-bearing clause for today's call sites; OAuth/OIDC noted as never exercising the default.

Patch76 (approved):
The docstring now matches the call graph. I re-verified its four claims at `3b932a5c`:

**Two default call sites, two explicit ones.** `_setup_logging` has four call sites in `src/`: `_setup_standard_mode` (`src/ha_mcp/__main__.py:305`) and `main` (`:861`) take the default, `main_oauth` (`:1451`) and `main_oidc` (`:1597`) pass `force=True`. The "OAuth/OIDC never exercise the default" clause holds.

**The collector really is the only root handler in `src/`.** The only `addHandler` calls on the root logger are `usage_logger.py:85` and `:120`, both attaching the startup collector. `stdio_settings_sidecar.py:826` also calls `basicConfig`, but that sits inside `run_main`, the sidecar entry point, which is spawned as its own process (`subprocess.Popen` at `stdio_settings_sidecar.py:428`, `cmd = [..., "-m", "ha_mcp.stdio_settings_sidecar"]`). It never installs a handler in the process the two default call sites run in, so the claim is accurate as scoped.

**"Defensive rather than load-bearing"** follows from those two facts, and the parenthetical that attributed the default to an OAuth/OIDC repeat call is gone. That was my only residual point.

CI is green at this head: 36 checks pass, and the two non-successes are the expected `Validate Hotfix Origin` and `dependabot` skips.

PR #2041, opened 2026-07-26
fix: retry Supervisor add-on calls on job-group collisions

kingpanther13:
## What does this PR do?

Supervisor serialises jobs per add-on job group, so a state-changing call is rejected outright with `Another job is running for job group addon_<slug>` whenever a still-settling job (a watchdog restart, a prior start/stop, a store reload) still holds that group.

`ha_manage_addon` surfaced that as `SERVICE_CALL_FAILED` with the suggestion "Check Home Assistant connection and Supervisor availability" — neither accurate nor actionable, since nothing is wrong with the connection and the holding job clears within seconds. An agent has no way to act on it.

`_supervisor_api_call` now retries the call inside one bounded wall-clock window with backoff:

- The window is **total**, not per-attempt, so a wedged group gives up once the window elapses rather than after some fixed retry count.
- Retrying is safe for non-idempotent calls: `JobGroup.acquire` rejects **before** the job body runs, so nothing was applied and a retry cannot double-execute a start/stop/install.
- The marker is anchored on `another job is running for job group`. Supervisor's job-level `JobConcurrency.REJECT` path raises a bare `Another job is running` for long OS/data-disk operations, which must not ride a seconds-scale retry.
- **Every other failure raises immediately, without retrying** — including a genuine error that arrives mid-retry.
- If the group is still held when the window elapses, the call raises with stuck-job guidance ("check Supervisor logs for a stuck or long-running job") and Supervisor's original error text, instead of the generic connectivity suggestion. It also logs at WARNING with the attempt count and elapsed time.

This also removes a real source of e2e flakiness: the add-on lifecycle tests drive stop/start/restart against a live Supervisor, so a sibling add-on job in flight could fail an otherwise-good run with this error.

## Type of change
- [x] 🐛 Bug fix

## Testing

`TestSupervisorJobGroupCollisionRetry` pins the behaviour: retry-until-clear with the exact backoff schedule, the backoff cap, a non-collision failure raising without retrying, a real error arriving mid-retry, the bare job-level rejection not being retried, `timeout` forwarding across retried attempts, and the bounded give-up preserving Supervisor's text while dropping the connectivity hint. All patch `asyncio.sleep` (and, for the bound, `time.monotonic`) so they add no wall-clock time.

## Checklist

### DIFF
--- a/src/ha_mcp/tools/tools_addons.py
+++ b/src/ha_mcp/tools/tools_addons.py
@@ -210,6 +210,15 @@ def _merge_options(base: dict, override: dict) -> dict:
     return merged
 
 
+# Substring of Supervisor's per-job-group rejection ("Another job is running
+# for job group addon_<slug>"), matched case-insensitively, plus the bounded
+# wall-clock window and backoff used to ride it out. See _supervisor_api_call.
+_JOB_COLLISION_MARKER = "another job is running"
+_JOB_COLLISION_RETRY_WINDOW = 60.0
+_JOB_COLLISION_RETRY_INITIAL_DELAY = 1.0
+_JOB_COLLISION_RETRY_MAX_DELAY = 5.0
+
+
 async def _supervisor_api_call(
     client: HomeAssistantClient,
     endpoint: str,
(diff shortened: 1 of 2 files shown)


ghhamcp:

ghhamcp:

PR #2038, opened 2026-07-26
Add translation fr.json

Locodice67:
## Description

Adds complete French translation across all four translatable surfaces.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Testing

- [x] Local tests pass
- [x] CI should pass

PR #2042, opened 2026-07-26
feat: complete the Simplified Chinese (zh-Hans) locale and enforce locale parity

kingpanther13:
## What does this PR do?

Completes the Simplified Chinese locale, and makes "a language ships everywhere or not at all — against the current English" enforceable.

**zh-Hans, the three missing surfaces.** #1992 added `zh-Hans` for the settings UI only, so a Chinese install still got an English config flow and English add-on options. German (#1920) and Russian (#1913) ship all four.

- `custom_components/ha_mcp_tools/translations/zh-Hans.json` — config flow, options, repair issues, selectors, update entity (77 leaf keys)
- `homeassistant-addon/translations/zh-Hans.yaml` — 16 schema keys
- `homeassistant-addon-dev/translations/zh-Hans.yaml` — 26 schema keys

**zh-Hans settings UI drift.** `policies.card.conditions_intro` read 当所有条件都匹配时 ("when ALL conditions match"); #1993 reworked policy evaluation to ANY-match, so the Chinese string described the opposite gating rule to the one the server enforces. `policies.card.always_row`, `policies.card.and_join` (#1993) and `advanced.extra_yaml_write_keys.label` / `.help` (#1997) were missing and fell back to English.

**`tests/src/unit/test_locale_parity.py`** — guards, none of which existed before:

1. *Surface parity.* A language code present on one surface and absent from another fails, naming the exact files to create.
2. *Spelling.* Only the settings UI is case-insensitive about the code (`_i18n.normalize_locale` lowercases stems); HA and Supervisor match the literal filename, so a `zh-hans.yaml` typo would render in the settings UI and be invisible to both. Checked separately so the failure points at the typo rather than reporting two unrelated languages.
3. *Component key and placeholder parity* with `en.json`, plus `strings.json` ↔ `en.json` equality. The settings UI had this guard in `_i18n.load_catalogs` and the add-ons in `test_addon_structure.py`; the component had none.
4. *Source drift.* `locale_source_baseline.json` pins a hash of every English string on all four surfaces. Key parity is structurally blind to the failure that actually shipped — #1993 changed a string's meaning while its key and its (empty) placeholder set stayed put. Changing, adding, or removing an English string now fails until the locales carrying it are revisited and `python scripts/update_locale_baseline.py` refreshes the baseline. The script imports the test's own hashing helper, so the two cannot disagree about what is hashed.
5. *Surface accounting.* Every catalog directory git tracks must be listed as guarded, English-only, or vendored, each with its reason — otherwise the rule stays blind to a new *surface* the way the repo was blind to a new locale. The Webhook Proxy add-on and its bundled integration are English-only by decision (the upkeep isn't worth it), and a locale appearing there fails rather than rotting unchecked. Discovery reads `git ls-files` so an untracked `venv/` carrying a package's `locales/` can't turn the test red.

`tests/addon/test_addon_structure.py::test_translations_cover_every_schema_key` was parametrized over a hardcoded pair, leaving both Webhook Proxy flavors with no `name`/`description` check; it now uses the same `homeassistant-addon*/config.yaml` glob as the test below it, with a sibling assertion so an empty glob fails instead of collecting zero cases.

**What the new tests caught.** `ru.json` was missing the four `allowed_dirs` / `extra_yaml_keys` option strings added by #1997, and its `tools_info` description still told users there was nothing to configure on a screen that now has two fields. Both fixed here.

Terminology follows the existing zh-Hans settings catalog (插件 for add-on, 只读模式, 工具搜索, 最佳实践, 测试版); product nouns users paste or search for (Client ID/Secret, Webhook, HACS, LLM API, tool names) stay in English.

AGENTS.md § Translations documents the four-surface rule, the per-surface content rules, and the baseline refresh step.

No component version bump — 1.2.4 is already pending against stable 1.2.3, and translations add no service or argument the server gates on.

Sequencing note: once this merges, #2038 (French, component catalog only) goes red on the surface-parity test until its other three catalogs land.

## Type of change
- [x] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

Each guard was verified by reproducing the failure it is meant to catch: a locale added to one surface only, a `zh-hans.yaml` casing typo, and an English string reworded from ANY-match back toward ALL-match.

## Checklist
- [x] I have updated documentation if needed — AGENTS.md § Translations

**Translation review wanted.** The Chinese and Russian text was drafted with AI assistance and verified structurally — key parity, placeholder parity, no folded-line space artifacts between CJK characters, terminology matched against the existing catalogs — not by a native speaker. @C3H3-AI (zh-Hans, #1992) and @SHAREN (ru, #1913), a read-through of the wording would be welcome.


ghhamcp:

ghhamcp:

Patch76 (commented):
Broke each check to confirm it fails: removed an add-on catalog, renamed `zh-Hans.yaml` to `zh-hans.yaml`, emptied the component translations directory, dropped a `{detail}` placeholder from a translated string, and reworded an English string. All five go red. Three fail with a single message naming the mistake; the casing typo and the emptied directory also trip surface parity with a message pointing at the symptom, which for the casing case is exactly what the docstring at lines 146-149 predicts and why the dedicated spelling test carries the real diagnosis. `scripts/update_locale_baseline.py` reproduces the committed baseline byte-identically, and the 582 pinned strings match the leaf count of the four English catalogs exactly (421 + 77 + 32 + 52).

The baseline is the part that earns its keep. Key and placeholder parity cannot see a string whose meaning moved, and #1993 is the proof that this is not a hypothetical failure mode.

One scope concern.

**The surface list is hardcoded.** `ADDON_DIRS` (`tests/src/unit/test_locale_parity.py:38`) is a two-tuple and `COMPONENT_TRANSLATIONS` (line 34) a single path, so "ships everywhere or not at all" covers four of the eight user-facing catalogs in the tree. Outside it:

- `homeassistant-addon-webhook-proxy/translations/` (9 schema keys)
- `homeassistant-addon-webhook-proxy-dev/translations/`
- `homeassistant-addon-webhook-proxy/mcp_proxy/translations/` (config flow, repair issues, one service)
- `homeassistant-addon-webhook-proxy-dev/mcp_proxy_dev/translations/`

All four are English-only today, so nothing is inconsistent at this moment. Two consequences regardless. A user running zh-Hans who installs the Nabu Casa webhook proxy still gets English option names and an English config flow, which is the half-install this PR sets out to make impossible. And the guard is blind to a new *surface* in the same way the repo was blind to a new *locale*: add `homeassistant-addon-foo/` and nothing notices.

`tests/addon/test_addon_structure.py:310-312` already globs `homeassistant-addon*/config.yaml` so that "a new add-on can't reintroduce it". Deriving the add-on list from that glob would reach the first two bullets; the two nested `mcp_proxy` catalogs are component-shaped, so they need `COMPONENT_TRANSLATIONS` widened separately. Either way, putting the proxy flavors into a named exclusion constant that carries the reason turns the boundary into a decision rather than an omission. AGENTS.md already names the four guarded paths – what it does not say is why the other four sit outside them.

The same two flavors also escape the guard the new docstring cites as the existing add-on check: `test_translations_cover_every_schema_key` is parametrized over the literal pair at `tests/addon/test_addon_structure.py:259`, so nothing requires a `name`/`description` for the proxy add-ons' schema keys either.

Minor: `english_sources` filters `isinstance(text, str)` (line 109) and `_flatten` does not descend into lists, so a non-string leaf never enters the baseline – and because `unrecorded` is derived from that same filtered dict, it is not reported as a new string either. Every one of the 582 leaves is a string today, so this costs nothing now. It is the one case the "a key that never enters the baseline would be silently unguarded forever" reasoning does not cover.

kingpanther13:
Fixed all three — the surface point was the good catch.

**Hardcoded surface list.** Rather than fold the proxy catalogs into parity, the test now discovers every `translations/` / `locales/` directory in the tree and requires each to be either in `GUARDED_SURFACES` or in `ENGLISH_ONLY_SURFACES` with its reason attached. The Webhook Proxy add-on and its bundled integration are English-only by maintainer decision — the upkeep isn't worth it — so that is what the entries say. A new `homeassistant-addon-foo/translations/` now fails until someone decides which list it belongs in, and a locale landing on an English-only surface fails too rather than rotting unchecked. AGENTS.md records the decision.

**`test_translations_cover_every_schema_key`.** Now parametrized over the same `homeassistant-addon*/config.yaml` glob as the test two below it, so both proxy flavors are covered. They already satisfy it (9/9 schema keys with non-empty `name` and `description`), so no proxy content changed and no version bump was involved.

**The non-string leaf.** `_flatten` now descends into lists and renders every non-string leaf as text, so the baseline covers every leaf rather than every string leaf.

Verified the new guards by breaking them: an unregistered `homeassistant-addon-newthing/translations/` and a `de.yaml` dropped into the proxy add-on each fail with the offending path named.

Patch76 (commented):
All three land. The surface discovery finds all eight catalog directories and reports none unaccounted, `test_translations_cover_every_schema_key` now collects four cases instead of two with both Webhook Proxy flavors included and passing, and `_flatten` descends into lists with non-string leaves rendered through `json.dumps`. `tests/src/unit/test_locale_parity.py` plus `tests/addon/test_addon_structure.py` is 26 passed at this head. Breaking the two new guards reproduces what you saw: an unregistered surface directory and a stray non-English catalog on an English-only surface each fail with the offending path named.

Two things in the new commit.

**[Concern 1]: the schema-key parametrize lost the empty-glob guard.** `tests/addon/test_addon_structure.py:258` parametrizes over `_REPO_ROOT.glob("homeassistant-addon*/config.yaml")`, which is resolved at collection time. An empty glob yields zero cases and pytest reports `SKIPPED [1] tests/addon/test_addon_structure.py:258: got empty parameter set for (addon_dir)`, which reads as green. `test_addon_names_are_backup_filename_safe` uses the same glob but resolves it inside the body and asserts on it (`:321`). I moved all four `config.yaml` files aside and ran both: the backup-filename test failed with `no add-on config.yaml files found to validate`, the schema-key test skipped silently. This is the case `7927908edd` fixed for the two parametrized component checks two commits earlier – converting the hardcoded pair to a glob reintroduced it on a third test. The shape you already chose there fits: a sibling test asserting the glob is non-empty, mirroring `test_component_translation_locales_are_discovered`, or the in-body resolution at `:321`. A module-level `assert` would work too but surfaces as a collection error for the whole file rather than a named failure.

**[Concern 2]: the surface walk reads the working tree, so untracked directories decide the result.** `_discover_translation_surfaces` (`tests/src/unit/test_locale_parity.py:87-91`) walks the filesystem and prunes by directory name against `_UNSEARCHED_DIRS` (`:72`). `.venv` is on that list, `venv` is not, and `babel` – among others – ships a `locales/` directory. `mkdir -p venv/lib/python3.13/site-packages/babel/locales` in an otherwise clean checkout turns the test red with `new translation surface(s) ['venv/lib/python3.13/site-packages/babel/locales']`. CI never sees this, since every workflow builds `.venv` via `uv venv`; a contributor who created their environment with `python -m venv venv` does, and the message points at a directory they are right to have.

Driving the discovery off `git ls-files` restricts it to tracked content and retires every entry that exists only to skip untracked or generated noise. It does not retire the list outright: `git ls-files` at this head yields nine catalog directories, not eight, because `tests/initial_test_state/custom_components/hacs/translations` is a tracked HACS fixture. Today it is invisible only because `tests` is in `_UNSEARCHED_DIRS`. So that one exclusion is load-bearing over tracked content and would need to stay, as a `tests/` prefix filter or an entry of its own – which is arguably the honest outcome, since it is a real decision rather than a side effect of a name-based skip list.

The empty-container residual on my earlier `_flatten` note – a `{}` or `[]` value still contributes no key – is not worth acting on: no catalog has one, and the recursion has no natural key to record for it.

kingpanther13:
Both fixed.

**Empty parametrize.** You're right that widening to a glob reintroduced what `7927908e` had just closed. `_ADDON_DIRS` is resolved once at import and `test_addon_config_glob_is_not_empty` asserts on it, mirroring the in-body assert in `test_addon_names_are_backup_filename_safe`. Emptying the list fails with the glob named instead of skipping.

**Working-tree discovery.** Switched to `git ls-files`. Your `venv/lib/python3.13/site-packages/babel/locales` repro now passes, and the whole `_UNSEARCHED_DIRS` list is gone with it. As you predicted, that surfaces `tests/initial_test_state/custom_components/hacs/translations` — recorded in a `VENDORED_SURFACES` constant ("third-party HACS integration, vendored to seed the e2e container — upstream's catalogs, not ours to translate") rather than hidden behind a name-based skip. Agreed that is the honest outcome.

That switch then failed in a way no local run could show: `git ls-files` exits 128 in the Unit Tests job, which runs in a container against a checkout owned by another uid, and `actions/checkout`'s `safe.directory` lands in a temp HOME the container never sees — the same reason `pr.yml:277-281` sets it by hand for the mirror job. `01016200` passes `-c safe.directory=*` so the exemption is scoped to that one read instead of depending on every job to configure git, and surfaces git's stderr, which `capture_output` had swallowed into a bare exit 128.

Left the empty-container residual alone, for your reasons.

Patch76 (dismissed):
Both land, and the container failure was worth the extra commit: moving the measurement to tracked content put the check somewhere the working-tree walk never had to survive.

I re-ran `tests/src/unit/test_locale_parity.py` and `tests/addon/test_addon_structure.py` at `01016200`: 27 passed, one more than at `3bba29f2` because of the sibling test that arrived in `658e2fc3`. Breaking each guard reproduces the intended failure. Moving all four `config.yaml` files aside fails `test_addon_config_glob_is_not_empty` with the glob named, while `test_translations_cover_every_schema_key` still skips silently. The sibling test is what turns that silence into a named failure, which is exactly the job it was added for. My `venv/lib/python3.13/site-packages/babel/locales` repro passes here and still fails at `3bba29f2`, so the change is what fixed it. Renaming the `VENDORED_SURFACES` key fails with `tests/initial_test_state/custom_components/hacs/translations` named, so that entry is load-bearing over tracked content rather than decorative, and staging an `en.json` under a fresh `src/ha_mcp/newthing/translations` fails the same way, so the guard still catches real surfaces.

The `safe.directory` diagnosis holds up on the record too: Unit Tests failed at `658e2fc3` with `subprocess.CalledProcessError: Command '['git', 'ls-files', '-z']' returned non-zero exit status 128` and passes at `01016200`. That commit changes one file and does exactly two things, and the `try/except` alone would still fail, so the `-c safe.directory=*` read is what carried it.

**[Nit] the comment's precedent does not exist.** `pr.yml:277-281` sits in the `lint` job (`name: Ruff Lint`), in its "Run ruff format check on changed Python files" step. That `lint` step is the only place in `.github/workflows/` that sets `safe.directory` at all – `sync-integration-mirror.yml` exists but never touches it, so the precedent the comment points at does not exist anywhere. The parenthetical is otherwise the most useful thing in `_discover_translation_surfaces` for whoever hits this next, so it is worth pointing at the step that actually set the precedent. Whether that earns a commit is your call – I have approved as it stands, and since a push dismisses stale reviews, say the word and I will re-approve if you take it.

One property worth recording rather than acting on: `git ls-files` in the superproject does not descend into `src/ha_mcp/resources/skills-vendor`, which the unit-test job checks out with `submodules: true`, so a catalog directory added there would not be discovered the way the old walk would have found it. The submodule's pinned tree has none today. Worth knowing that the exemption lists cannot cover that case by convention either, since nothing would ever surface it to be recorded.

kingpanther13:
@Patch76 Claude decided to fix the nit that I was going to dismiss without my approval. Opus 5 is annoying -_-. please re-review/approve when able.

Patch76:
You can use *hookify* from Anthropics … it's a pain, but it keeps Claude in line.

Patch76 (approved):
The comment now names the step that actually set the precedent, and folding the submodule gap into the docstring puts it where someone reading `_discover_translation_surfaces` will find it, which is better than leaving it in a review comment.

Re-approving. `461e4c8a` changes a comment and a docstring and nothing else the parser sees: parsing the file at `01016200` and at `461e4c8a` and comparing the two ASTs with docstrings stripped gives an identical tree, so everything I broke and re-ran against the previous head carries over untouched. I re-checked the new attribution against `461e4c8a` itself rather than my earlier read – `pr.yml:255-256` is the `lint` job and its `name: Ruff Lint`, `:275` is the "Run ruff format check on changed Python files" step, `:281` is the `git config` line, and 277 and 281 are the only two places the file mentions `safe.directory` at all.

Several lanes were still running when I looked; nothing in this commit can change what they do.

PR #2045, opened 2026-07-26
fix: propagate cancellation in deep-search fan-outs and usage logging

kingpanther13:
## What does this PR do?

Two cancellation swallows, both surfaced during #2036's review, which made this tool's `CancelledError` propagation deliberate:

- `smart_search/_deep.py`'s three per-item gather loops branch `isinstance(result, tuple)` / `elif isinstance(result, Exception)` with nothing below, so a `CancelledError` handed back as a value by `return_exceptions=True` matched neither arm and was silently discarded — a cancelled search reported itself as one that ran and found nothing. Each loop now re-raises non-`Exception` results, the line `tools_entities` and `tools_search` already draw.
- `log_tool_usage` caught `Exception` only, so a cancelled tool call skipped the handler and the `finally` recorded it as `success: true` in the usage log. The handler now catches `BaseException`; the class name stands in for the empty `str()` of a bare `CancelledError`.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

Four regression tests, each verified red against the unfixed source: one per fan-out loop (helper types, dashboards, flow helpers) and one for the decorator. Run locally: the two touched unit files (28 tests) plus `ruff format --check`/`check` on all four files and `mypy` on the two source files; the full suite is left to CI.

## Checklist

### DIFF
--- a/src/ha_mcp/tools/helpers.py
+++ b/src/ha_mcp/tools/helpers.py
@@ -526,9 +526,12 @@ def log_tool_usage(func: Any) -> Any:
             elif hasattr(result, "__len__"):
                 response_size = len(str(result).encode("utf-8"))
             return result
-        except Exception as e:
+        except BaseException as e:
+            # BaseException, not Exception: a cancelled tool call must not be
+            # recorded as a success. str() is empty on a bare CancelledError,
+            # so the class name stands in.
             success = False
-            error_message = str(e)
+            error_message = str(e) or type(e).__name__
             raise
         finally:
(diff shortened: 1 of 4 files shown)


ghhamcp:

ghhamcp:

PR #2054, opened 2026-07-27
fix(internal): Hold pydantic-monty at 0.0.18 and satisfy ruff 0.16 LOG004

kingpanther13:
## What does this PR do?

Unblocks the two red dependabot PRs, which fail for unconnected reasons but are both waiting on code we own.

**`pydantic-monty` held at 0.0.18** (`.github/dependabot.yml`). 0.0.19 replaced the single-shot `Monty(code, ...)` + `run_async(external_functions=...)` API with a subprocess worker pool (`AsyncMonty` → `checkout()` → `feed_run(external_lookup=...)`) and dropped the module-level `run_monty_async`. Every path in `tools_code.py` fails against it — #2051 showed the whole code-mode E2E suite erroring with `Monty.__new__() takes 0 positional arguments but 1 was given`, plus two mypy errors. FastMCP's own `code-mode` extra still pins `==0.0.17` (shipped 3.4.4 and the 4.0.0a2 alpha) with the old constructor, so nothing upstream is pulling us forward. Monty is 0.0.x, so every release is a semver patch and no `update-types` filter holds the line — hence a blanket `dependency-name` ignore, to be dropped when `tools_code.py` moves to the pool API. `pyproject.toml` and `uv.lock` already carry `==0.0.18` and are unchanged.

**LOG004 fixed ahead of the ruff bump** (`src/ha_mcp/tools/helpers.py`). ruff 0.16.0 stabilizes `log-exception-outside-except-handler`; the repo selects the whole `LOG` group, so #2047 goes red on the `logger.exception()` in `exception_to_structured_error`. ruff's `--unsafe-fix` rewrites it to plain `.error()`, dropping the traceback the branch exists to provide, so this passes `exc_info=error` instead — identical inside `except` handlers, and correct outside them. `_component_state_result` (`tools_search.py:4308`) is a real caller with no handler in scope, where `logger.exception()` would log `NoneType: None`. With this merged, #2047 goes green on its merge ref with no further changes.

## Type of change
- [x] 🐛 Bug fix
- [x] 🔧 Maintenance/refactor

## Testing

Verified against ruff 0.16.0 specifically (`uvx ruff@0.16.0 check` over the full CI path list, plus `format --check`): clean. `mypy` clean. `test_helper_classifier.py` 32 passed; the two new tests fail against the previous `logger.exception()` line.

## Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -36,6 +36,17 @@ updates:
     commit-message:
       prefix: "deps"
       include: "scope"
+    ignore:
+      # Held at the pin in pyproject.toml. pydantic-monty 0.0.19 replaced the
+      # single-shot `Monty(code, ...)` + `run_async(external_functions=...)`
+      # API with a subprocess worker pool (`AsyncMonty` -> `checkout()` ->
+      # `feed_run(external_lookup=...)`), which breaks every code path in
+      # src/ha_mcp/tools/tools_code.py. FastMCP's own code-mode extra still
+      # pins ==0.0.17 for the same reason, so nothing in the ecosystem is
+      # pulling us forward. Monty is 0.0.x, so every release is a patch by
+      # semver and there is no narrower update-type that holds the line.
+      # Drop this entry when tools_code.py is ported to the pool API.
+      - dependency-name: "pydantic-monty"
     groups:
       # Group dev dependencies together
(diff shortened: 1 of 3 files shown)


ghhamcp:

ghhamcp:

PR #2055, opened 2026-07-27
feat: add the Spanish (es) locale across all four translated surfaces

Patch76:
## What does this PR do?

Adds Spanish (`es`) as a fully shipped locale. A language has to cover all four translated surfaces or none, so this adds all four catalogs in one go:

| Surface | File | Scope |
|---|---|---|
| Settings UI | `src/ha_mcp/settings_ui/locales/es.json` | 419 `messages`, 29 `tool_groups`, 87 `tools` title/description overrides |
| Custom component | `custom_components/ha_mcp_tools/translations/es.json` | all 77 `strings.json` keys |
| Add-on (stable) | `homeassistant-addon/translations/es.yaml` | `name` + `description` for all 16 `schema:` keys |
| Add-on (dev) | `homeassistant-addon-dev/translations/es.yaml` | `name` + `description` for all 26 `schema:` keys |

The settings UI `messages` and the component catalog were produced by mapping each English key to its translation and writing the file from the English structure, so key set, key order and `{placeholders}` match the source by construction rather than by review. `en.json` leaves `tool_groups` and `tools` empty — English for those comes from the tool definitions at runtime — so those two sections are keyed off `site/src/data/tools.json` instead.

Two notes:

- `tool_groups` has **29** keys, matching every other shipped locale. I first wrote all 34 tags, then dropped five (`Radio Management`, `Thread`, `Z-Wave`, `Zigbee`, `beta`): `settings.js:803` resolves a group label from `primary_tag`, and `_tools_meta`'s `SECONDARY_TAGS` keeps `Z-Wave`/`Zigbee` out of that choice while the other three never sort first among a tool's tags — so none of the five can reach `localizedToolGroup`. `zh-Hans` carried them too; they are removed there as well now (review item 7). Four `tool_groups` values are byte-identical to their English key (`Assist`, `HACS`, `Matter`, `Scripts`), against eight in `de`.
- Six `messages` values are byte-identical to English on purpose: `Diff`, `Diff: {name}`, `{count} min`, `{count} bytes`, `Total`, `Normal`. They read the same in Spanish; `fr` leaves four of the six unchanged for the same reason. All 87 tool titles and descriptions are genuinely translated — none is left English.

Registration coverage now runs as one check parametrized over the registered catalogs, replacing the four hand-written per-locale copies (review item 9). Also adds `es` and `fr` to the locale-code list in `AGENTS.md` § Translations, which had gone stale when `fr` landed.

## Type of change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

What actually ran locally, rather than a blanket tick:

- `tests/src/unit` + `tests/addon` with `--maxfail=0`: **9432 passed, 172 skipped, 9 errors**. All nine are environmental and reproduce on an untouched `master`: five in `test_relay_timeout_upstream_contract.py` because the `aiohttp` dev dependency is absent from my local venv, four in `test_addon_startup.py` with `FileNotFoundError: 'docker'`. Note `tests/pytest.ini` sets `--maxfail=3`, so a default invocation stops at the first of those and never reaches `tests/addon`.
- `test_locale_parity.py`, `test_settings_ui_i18n.py` and `test_addon_structure.py` green, including `test_translations_cover_every_schema_key` for both add-on flavors.
- The new `es` test was injection-checked: emptying `tool_groups` in `es.json` turns it red, restoring it turns it green.
- `ruff check .` clean repo-wide with ruff 0.16.0 — the pin moved from 0.15.22 while this branch was open. `ruff format --check .` reports six files repo-wide; none of them is a `.py` this PR changes (every changed `.py` is format-clean), and the two I traced, `AGENTS.md` and the `CLAUDE.md` symlink to it, are already unformatted on `master`. Mypy is green in CI on the six changed `.py`.
- E2E was not run locally — it needs containers. No LLM-agent test either: these catalogs render in the settings UI and the Home Assistant config flow, so nothing here reaches an agent's context.

## Checklist


ghhamcp:

ghhamcp:

Patch76:
Italian and Dutch to follow. 

Italian, Polish, Portuguese, Czech, Swedish, Norwegian, Danish, and Russian are (after English, German, French and Spanish) the largest groups based on installation figures (HA statistics)

kingpanther13:
Review brewing. Since we just updated ruff I'm going to go ahead and merge master in now for you just to be sure this will pass the new ruff

kingpanther13 (changes requested):
Strong PR. Every structural claim in the description held up under independent re-derivation: the 29 `tool_groups` keys are exactly the reachable primary-tag set per `_tools_meta.py`'s own rule, the component and add-on catalogs are complete against their own sources, no #1993-style polarity flips, and "Planos" matches Home Assistant's own Spanish frontend (`ui.panel.config.blueprint.caption` = "Planos") — where `de`/`fr` leave it English, so `es` is the more correct sibling here.

Requesting changes anyway: a handful of catalog fixes, plus this repo's Boy Scout rule (AGENTS.md § Boy Scout Rule — Handling Discovered Improvements) applied to the adjacent gaps this PR surfaced. Default here is fix-in-place, in this PR, so locale six lands on hardened rails.

### Catalog fixes (`es`)

1. **`options.step.init.data_description.bind_host`** (component `es.json`): "Elige bucle local…" names a dropdown option the user can only ever see in English — the labels are hard-coded at `config_flow.py:371-377` ("This machine only (loopback)"), since hassfest forbids dots in translation keys for the IP-valued options. Use the literal the UI shows, e.g. "Elige loopback…", as `de`/`fr`/`zh-Hans` do.
2. **`…data_description.enable_llm_api`**: "'Servidor HA-MCP'" — the selectable LLM API name is `entry.title`, the untranslated literal `HA-MCP Server` (`config_flow.py:100`, `llm_api.py:602`). Use 'HA-MCP Server'. `fr` and `ru` carry the same mismatch — sweep them in the same commit.
3. **`notice.restart.embedded`** (settings `es.json`): "la integración del proceso" reads as "the process's integration". The component catalog renders the same concept as "integrado en el proceso"; align — "la integración en el proceso".
4. **`backup.identical`**: this string substitutes for the diff body itself (`settings.js:1712`), so the feminine-plural "(idénticas)" has nothing to agree with. "(sin diferencias)" sidesteps it, as `ru` does.
5. **`selector.server_webhook_auth.options.none`**: "(predeterminado)" after feminine "URL de webhook secreta" → "(predeterminada)". The file itself gets this right in `llm_api_exposure.options.tool_search`.
6. **`features.enable_tool_search.help`** (and the same sentence in both add-on `es.yaml`): English approximates both figures ("~46K to ~5K"); the Spanish only the first ("de unos 46K a 5K") → "de unos 46K a unos 5K".

### Boy Scout items (pre-existing — fix here, not deferred)

7. **Remove the five inert `tool_groups` keys from `zh-Hans.json`.** Your own analysis proved `Radio Management`/`Thread`/`Z-Wave`/`Zigbee`/`beta` can never render; "they are inert there too" is exactly the "pre-existing — not touching it" pattern the Boy Scout rule names as a red flag. Five-line deletion.
8. **Hard-coded English blocks in the options description**: `panel_hint` (`config_flow.py:481-486`), the version line (`config_flow.py:583-585`), and the tools-module hint (`config_flow.py:616-628`) are English paragraphs interpolated into every locale's otherwise-translated form. Move the static text behind `strings.json` keys (the dynamic parts stay placeholders; `homeassistant.helpers.translation.async_get_translations` covers the server-side lookups) and translate in all five locales. Largest item on the list, and squarely the class of problem that bites every future locale.

### Test hardening (in this PR)

9. **Collapse the four hand-copied per-locale registration tests** (`test_settings_ui_i18n.py:212-248`) into one parametrized over `sorted(CATALOGS)` minus `en`, asserting non-empty `meta.native_name`, `meta.dir` in `{"ltr","rtl"}`, and non-empty `tool_groups`/`tools`. The copies have already drifted — `ru` has no test at all, `de` lacks the non-empty assertions — so locale six could ship a stub and stay green: the exact half-install `test_locale_parity.py`'s docstring records for #1992 and #2038. Keep an empty-collection guard like `test_locale_parity.py:306`.
10. **Validate catalog keys against reality.** Nothing checks `tool_groups` keys are reachable primary tags or `tools` keys are real tool names — `settings.js:55-57` silently falls back to English on a stale/typo'd group, and `build_payload` (`_i18n.py:345-349`) creates entries nothing reads. Your 29-key derivation is correct, but it lives in this PR's description; a locale-parametrized test against `site/src/data/tools.json` using `_tools_meta.py:245`'s primary-tag rule pins it in CI and catches the forward direction too (new tool or new first-sorting tag → every locale silently shows English).
11. **Stub-catalog ceiling.** A wholesale copy of `en.json`'s `messages` passes every existing check. Byte-identical-to-English counts are ≤9/419 in every shipped locale; assert roughly ≤5% per locale so a lazy catalog fails while the deliberately-identical strings this PR documents keep passing.
12. **Guard the AGENTS.md locale list** you just fixed (`AGENTS.md:716`): it went stale when `fr` landed; assert the codes on that line match the shipped catalog set (`test_locale_parity.py` already shells out to `git ls-files`).
13. **Make `test_translations_cover_every_schema_key` bidirectional** (`tests/addon/test_addon_structure.py:278`): it never flags orphaned `configuration.<key>` entries for removed options. Your `es.yaml` files are clean (16/16 and 26/26, zero extras) — close the hole while we're in the file.

Items 1–7 and 9–13 are small and mechanical. Item 8 is real work — if you'd rather not carry it, say so and I'll push it onto this branch, but it ships in this PR either way.

Patch76:
It's cooking … I had to add a little more salt and pepper; Claude knew better …

Patch76:
All thirteen items are in, across five commits on `b4729d55`.

### Catalog fixes (1–6)

Done as described. Sweeping the siblings turned up two more of the same two defects than the review named:

- **Item 1** — `ru` also translates the loopback option ("Выберите кольцевой интерфейс") for the same hard-coded English label. Fixed alongside `es`.
- **Item 2** — `zh-Hans` also renders the LLM API name as a translated string ("HA-MCP 服务器"). The review named `fr` and `ru`; all three now carry the literal.

### Item 7

Removed. I re-derived the reachable group set from `site/src/data/tools.json` with `_tools_meta`'s own rule rather than relying on the derivation in my PR description: 29 primary tags, and no tool carries only secondary tags, so the `tags[0]` fallback cannot reach `Z-Wave`/`Zigbee` either. That matches the 29 keys the other locales ship.

### Item 8

Carried it. The three blocks now live in a `common` catalog block and are looked up for the configured language; the runtime values stay placeholders. `common` is one of the top-level categories hassfest accepts, keyed by slug, and the loader treats a custom integration's translations no differently — same `translations/<language>.json` files this component already ships.

The lookup fails soft, matching the hints it feeds: an unreadable language skips it, a failing load falls back, and a partial catalog keeps the English source for the keys it does not carry. That is why the existing options-flow tests needed no changes — where translation is unavailable the form renders exactly as before. Five tests cover the translated path.

Two details worth naming:

- The panel sentence lost its trailing space, which now comes from the caller. `validate_translation_value` rejects a value with leading or trailing whitespace, so it could not have stayed in the catalog.
- The version line falls back to the English template if a catalog ever carries an unusable brace. A catalog is data, and a raised format error would otherwise take the whole options form down.

### Test hardening (9–13)

All five in. The four registration copies collapse into one check parametrized over the registered catalogs; the group and tool keys are now checked against the shipped tool catalog in both directions, importing `_tools_meta`'s primary-tag rule rather than copying it (I hoisted it to a module-level `SECONDARY_TAGS` for that); the stub ceiling is 5% of messages byte-identical to English, against a high-water mark of 9 of 419 messages (2.1%, in `de` and `fr`); the AGENTS.md list is pinned to the shipped catalog set; and the add-on schema check runs both ways.

Every new assertion was injection-checked — the condition it guards was broken, the test observed to go red, then restored. That includes the empty-collection case: pointing the glob at a missing directory makes the guard fail rather than collapse to a silent skip.

### One more Boy Scout item

`src/ha_mcp/settings_ui/locales/zh-Hans.json` was the only one of the 24 locale and translation catalogs stored with CRLF, against `.gitattributes`' own `* text=auto eol=lf`. The attribute does not renormalize an already-committed blob, so it stayed that way.

It is not cosmetic. Python's `read_text`/`write_text` translate newlines by default, so any tooling that rewrites that file silently converts it and buries the real change in an 808-line diff — which is what happened while I was making the five-key deletion, and what the diffstat rather than the edit caught. Normalized in its own commit; the content is byte-identical with the carriage returns stripped.

### One thing I could not close myself

The Home Assistant button labels inside the new translations ("Add entry" → "Eintrag hinzufügen", "Añadir entrada", and so on) are natural renderings. I did not verify them against Home Assistant's own frontend catalogs the way you checked "Planos", so they are the weakest strings in the set.

I also refreshed the PR description: my own commits had falsified the test count, the note that `zh-Hans` still carries the inert group keys, and a test name that no longer exists, and the master merge moved the ruff pin to 0.16.0.

Patch76:
## Implementation Summary

**Choices Made:**

- Carried all six review rounds in this PR rather than handing any part off. Spanish ships on all four translated surfaces, and every sentence the options form assembles at runtime now comes from a `common` catalog block — sixteen keys, translated in all five locales — instead of being English inside an otherwise translated form.
- Left a value verbatim wherever the reader has to match it against another screen, and translated the prose around it. That one rule decided four separate sweeps: the LLM API name (`entry.title`), the tools entry title (force-renamed at `__init__.py:3571`, so the integrations page shows it in English in every language), the connect URLs and their `<...>` stand-ins, and the `Client ID` / `Client Secret` labels. Its boundary matters as much as the rule: the abort string's first sentence is prose about why setup failed and stays translated, as do the catalog-rendered step titles naming the same entry — Home Assistant renders those from the catalog, so translating them is what makes them appear at all.
- Named what the screen shows, not what the config holds: `connect_local_lan` says "Local network", the option in the dropdown, rather than the bind value `0.0.0.0` the user never sees. A test binds that sentence to the `BIND_HOST_ALL` label at `config_flow.py:594`, so renaming the option turns six catalogs red instead of leaving them quoting something the form no longer shows.
- Translated each new string from the field it points at rather than from its English source, then aligned the English to the same wording, since that is what the next maintainer edits. Each catalog also keeps one quote convention throughout — `„…"` for `de`, `«…»` for `es`, matching what `fr`, `ru` and `zh-Hans` already did.
- Made every degradation on that form audible. Failed lookup, unusable template, discarded non-`Mapping` result, a `common` category that was never built, both version reads and the tools-module read all log at warning with the reason; before this, four of them were silent or at debug. Each level is asserted, because raising a level without pinning it is how an earlier round's fix quietly came undone.
- Fixed guards rather than symptoms twice. The prefix no-op watched a branch core cannot produce (`_TranslationCache.get_cached` serves a single-component lookup from that component's cache entry, so the reachable error arrives as `{}`), and the sentence separator sniffed glyphs — wrongly in both directions, since U+201D is Simplified Chinese's closing quote and the traditional brackets it kept are unused there.
- Keyed that separator on both of its inputs in the end, after keying it on the language alone proved to be the same mistake mirrored. Core loads `[LOCALE_EN, language]` and merges English first as the documented fallback, so an instance set to a language this integration does not ship reads English prose while `hass.config.language` still names a script that sets its own spacing. The language must be in the set **and** the sentence must differ from its English source; `ko` is out of that set entirely, since Korean separates words with ASCII spaces.
- Guarded every `str(integration.version)`: a manifest without a version returns `None` rather than raising, so the form rendered the literal "None", the tools device took it as its `sw_version`, and both advisory comparisons fed it to `AwesomeVersion`. Each now takes the unreadable-manifest path its own module already defines, each is pinned, and the consumer reports the component's own reason rather than re-diagnosing an unreadable manifest as a missing token — which sent the operator to the reload and admin-rights suggestions, neither of which can fix it.
- Derived the checked tool set from `scripts/extract_tools.py`'s static parse rather than its committed output, which is regenerated only after merge — reading that let the PR adding a tool stay green and turned the *next* PR red across five locales. The parse then needed four fixes of its own: two missing encodings (one on the destructive write), a raise instead of `sys.exit` so an import-time failure is named, a discovery assertion so a shrunken parse cannot read as "these tools do not exist", and a raise when the doc markers are gone — the unchanged return compared equal to the file it came from, so `--check` printed "All files in sync." right after warning that it could not find them.
- Put a ceiling on all four translated surfaces, counting missing keys as untranslated, and added a floor for the `tools` values — 174 strings per locale, the largest translated surface in the repo, previously checked for presence only. A share alone would not have caught the case that motivates it, since one pasted tool scores 1.1%, so a wholly-English tool fails by name alongside the share. For the seven feature-gated tools both checks compare against both English renderings, the `FEATURE_GATED_TOOLS` stub and the parsed docstring, because the feature flag decides which one the settings UI shows and five of the seven differ — six of the 174 keys, enough to hide an untranslated remainder under the ceiling had the share consulted the rendered set alone.
- Made an empty parametrize list a collection error in both pytest configs, and deleted the dead `exclude_internal` machinery it exposed rather than pinning a mask no spec has ever used.
- Injection-checked every new assertion rather than reading it — the guard's *condition* broken, the test observed red, then restored.

**Problems Encountered:**

- Five of my own measurements were wrong before the reviewer's were. Two round-four injections reported guards as unpinned when the harness was at fault (one pointed at a test without the assertion, one used `TOOL_FILES = [] or sorted(...)`, which changes nothing); two round-five injections did the same for the same reason (one added the old check *in front of* the rule under test instead of replacing it, so it never reached the gate; one inserted its sentence before the heading that immediately follows the section it was meant to sit outside of, which put it inside); and the round-three `”` removal was reasoned from a false premise — that character is zh-Hans's own closing quote, and this PR's `zh-Hans` catalog ends a string on it. A broken mutation reads exactly like an unpinned guard, so a green injection now gets a second look before it becomes a claim.
- The round-two `.get("dir")` finding came from the same reflex: the guard could not be made to matter because `load_catalogs` defaults a missing `dir` and raises on any other value, so the assertion is unreachable by construction. Both of those loader raises were untested; they are pinned now.
- `empty_parameter_set_mark = fail_at_collect` is repo-wide and I can only run `tests/src/unit` and `tests/addon` locally. CI answered it: the e2e lanes collected clean.
- Deleting `exclude_internal` and its test was a judgment call on code I did not write. No `TOOL_SPECS` entry has ever carried it, and I found no case that should; a comment records what to restore when one appears.
- A claim in an earlier summary was wrong on its own terms: I justified moving a trailing space by naming `validate_translation_value` as rejecting whitespace. No such function exists here. The decision was right for a different reason — the separator is layout, and layout follows the text as well as the language.
- No test executes the real `async_get_translations`, and per AGENTS.md § Custom Component the component path cannot be exercised by CI pre-merge. That is why every failure mode on this path is now a warning with its cause in the message: the post-merge live test on the dev server is the first run that exercises it.
- Python's `read_text`/`write_text` translate newlines, and `zh-Hans.json` was the only one of the 24 catalogs stored with CRLF; editing it silently rewrote all 808 lines. The diffstat caught it, not the edit.
- One round-five push reached CI with an unsorted import block, because the lint I normally run locally sits on a path that push did not take. Lint caught it in about a minute and the fix is its own commit; the lesson is that the check has to be tied to the artefact, not to the route.
- A comment sent maintainers to the wrong file: the `_COMMON_FALLBACKS` mirror is asserted in `test_config_flow.py`, never in `test_locale_parity.py`. The same file cites `test_locale_parity.py` correctly for placeholder parity some fifty lines down, which is how the wrong one read as right. Nothing executes a comment, so no check could catch it; it now names the test as well as the file.
- One branch is pinned by its rule rather than by its call site. Deleting the counting rule that consults both English renderings turns its test red; deleting the argument that hands the second rendering to the tools check keeps the suite green, because no shipped locale pastes either English. Discriminating that argument needs a synthetic catalog pasting descriptions only — a whole tool trips the paste check first — padded with English until the share crosses the ceiling. I pinned the rule and reported the gap rather than build that unasked.

kingpanther13 (changes requested):
All thirteen items verified implemented — I re-derived the 29-key reachable set, confirmed the zh-Hans normalization is byte-identical with the 808 carriage returns stripped, confirmed the extra `ru` loopback and `zh-Hans` LLM-API-name fixes are real, and confirmed the five `common.*` catalogs are complete in every locale. I can also close the one thing you flagged as unverifiable: against the shipped `home-assistant-frontend` 20260624.6 catalogs, the key `ui.panel.config.integrations.integration_page.add_entry` renders exactly "Eintrag hinzufügen" / "Añadir entrada" / "Ajouter une entrée" / "Добавить запись" / "添加条目" — all five of your renderings match the button verbatim.

Requesting changes again, same rules as last round — everything below lands in this PR.

### New-string correctness

1. **`common.version_unknown` disagrees with its noun in three locales.** It substitutes into `{component_version}` and renders directly after "Componente"/"Composant"/"Компонент" (seeded at `config_flow.py:634`, survives when `async_get_integration` raises): `es` "desconocida" → "desconocido", `fr` "inconnue" → "inconnu", `ru` "неизвестно" → "неизвестен". The sibling `version_not_installed` agrees correctly in all five locales, so only this key is wrong — and it's new in this PR. `de`/`zh-Hans` are fine.
2. **Spanish double colon.** `tools_module_not_installed` and `tools_module_not_loaded` in `es.json` read "…(opcional): No instalado: pulsa…" — two colons in one sentence. English and all four siblings break the second clause with an em dash; match them.

### Item 8 is half done — same form, same class

3. The same options description still interpolates two all-English blocks: `_connect_url_hint` (`config_flow.py:712-764`, ~7 sentences into `{connect_url}`) and `_oauth_creds_hint` (`:766-811`, 3 more into `{oauth_creds}`). A Spanish user still sees roughly eight English sentences inside the paragraph block this PR just translated. The `common` catalog and `_common_strings` machinery now exist, so this is a mechanical sweep of the shape you already built — do it here.

### The lookup must fail loudly enough to debug

No test executes the real `async_get_translations` call — every new test monkeypatches `_fetch_common_translations` and rebuilds the same `component.{DOMAIN}.common.` prefix from the same assumption, and AGENTS.md § Custom Component says CI cannot exercise the component path pre-merge. We will live-test the translated form on the dev server after merge per that section, but for that test to mean anything the failure modes must be visible:

4. **Log level.** The `except` at `config_flow.py:171-173` logs at debug; `_connect_url_hint`'s own sibling in this same file (`:738-741`) argues in its comment for warning-by-default on exactly this kind of degradation, and the `websocket_api` seam this docstring cites as its model logs at warning with `exc_info`. Match them — the broad `except Exception` also swallows an `ImportError` from the function-local core import, which is a permanent defect no one will see at debug. Additionally, when `loaded` is non-empty but zero keys match the prefix, the merge is a silent no-op that renders pure English with nothing logged at any level — that's the exact signature of a wrong category or prefix, and it deserves one log line.
5. **Pin `_COMMON_FALLBACKS` to `strings.json["common"]`.** It's a third copy of the seven English strings (`config_flow.py:117-141`) and nothing compares it to anything — reword a string in `strings.json`, sweep the locales, refresh the baseline, and CI stays green while the fallback keeps the old text. Same move as the existing `strings.json` ↔ `en.json` mirror test (`test_locale_parity.py:407`), keys and values both.
6. **Say which language it actually is, and assert it.** The comment at `:110-116` and docstring at `:159-163` say "the user's language", but `:166` reads `hass.config.language` — the system language, not the viewing user's profile language; where the two differ the form mixes languages. If an `OptionsFlow` genuinely cannot see the requester's language, fix the prose to say "the system-configured language" so the next reader doesn't assume otherwise. And add the one-line seam assertion (`seam.assert_awaited_once_with(flow.hass, "de")`) — today, hardcoding `"en"` at the call site keeps all five new tests green.

### Test hardening follow-through

7. **The tools.json coupling fails on the wrong PR.** `test_settings_catalog_keys_name_real_groups_and_tools` reads the committed `site/src/data/tools.json`, which is regenerated only *after* merge by `sync-tool-docs.yml` with `[skip ci]` on the commit (`:71`). So the PR that adds a tool stays green, and the sync then turns every *subsequent* PR red across five locales — the failure lands on whoever opens the next PR. Derive the tool set from `scripts/extract_tools.py`'s `extract_tools()` (static AST parse, importable) instead of the committed JSON, so the PR that adds the tool is the one that owes the translations and master never breaks. `test_tool_docs_sync.py:1-7` records a maintainer decision about this exact timing for a related check.
8. **The `.format()` guard misses the error class it exists for.** `(KeyError, IndexError, ValueError)` at `config_flow.py:668` doesn't catch `AttributeError` — and a translated `version_line` carrying `{component_version.major}` passes placeholder parity (`_PLACEHOLDER_RE` doesn't match dotted tokens, so the extracted set is unchanged) and then takes the whole options form down, the precise outcome the comment above the block says it prevents. Catch `Exception` like the sibling seam does. While there: the terminal `return ""` at `:670` is unreachable (the second template is the well-formed module constant) — format the English constant unguarded after the loop instead of silently blanking the line.
9. **The stub ceiling can be walked around by omission.** `share = len(identical) / len(english)` only counts keys the catalog carries, so `messages: {}` scores 0% and a 20-key all-English stub scores 4.8% — both pass, and nothing else catches them since omission is legal. Count missing keys as untranslated — `(identical + missing) / len(english)` — and every shipped locale still clears the ceiling with the same margin (all five carry 419/419). And mirror the ceiling onto the other three surfaces while you're in the file: component catalogs (high-water `de` at 8.3% — pick a ceiling above that) and both add-on flavors (0% everywhere) currently accept a wholesale-English catalog.
10. **Collapsing the four tests dropped the native-name pins.** Nothing now asserts any specific `native_name` — `es.json` shipping `"Deutsch"` is green today. Keep it parametrized: assert native names are unique across catalogs and differ from `en`'s.
11. **Import the rule, not just the set.** `SECONDARY_TAGS` is imported, but the selection logic (`sorted(tags)`, filter, first-or-`tags[0]`-or-`"Other"`) is duplicated verbatim from `_tools_meta.py:250-252` into `test_locale_parity.py:459-462` — the set can no longer drift, the rule still can. Extract a `primary_tag(tags)` helper in `_tools_meta` and import it, which also makes your response's "importing the rule rather than copying it" claim true. Related record-straightening: your comment's justification for moving the trailing space — "`validate_translation_value` rejects a value with leading or trailing whitespace" — names a validator that does not exist anywhere in this repo. The decision was right anyway (the conditional space belongs to the caller); the stated reason wasn't.
12. **AGENTS.md § Translations now understates the contract.** Line 721's "may omit keys — English is the per-key fallback" is no longer true for `tool_groups` and `tools`, which your own new test forbids omitting in both directions. Add the qualifier, and note that adding a tool means updating every locale — that sentence is also where the item-7 timing behavior belongs once fixed.

### Small stuff, fix while you're in the files

13. The stub-ceiling comment says "9 of 419 messages (2.2%)" — it's 2.1%, as your own PR comment has it. `catalog["meta"]["dir"]` at `test_settings_ui_i18n.py:238` KeyErrors on a missing key instead of producing the assert's message — use `.get("dir")`. `isinstance(result, dict)` at `config_flow.py:155` silently discards a non-dict `Mapping` where the mirrored seam accepts one — check `Mapping`. The `isinstance(value, str) and value` filter at `:180` is untested (deleting it stays green; an empty translated value would blank the panel hint) — one case pins it. And the caller-appended space after `panel_hint` (`:559`) lands after a CJK full stop in `zh-Hans` — make it conditional on the hint being non-CJK-terminated, or key it off the catalog.

The quality of the round itself: the response comments name what you couldn't verify instead of papering over it, the CRLF near-miss is reported with its mechanism and split into its own commit, the three fail-soft paths each have their own test, and the extra `ru`/`zh-Hans` sweeps went beyond what was asked. The injection-check discipline — including the empty-collection guards — is exactly right. This is one more round of the same standard, not a moving target.

Patch76:
All thirteen items are in, across four commits on `2beaaedc`. Three of them ended somewhere other than where the item pointed, and I have flagged those below rather than quietly widening or narrowing them.

### New-string correctness (1–2)

1. `version_unknown` now agrees with the noun it follows: `es` "desconocido", `fr` "inconnu", `ru` "неизвестен". Its `version_not_installed` sibling was already masculine in all three, which is what settled the reading.
2. The two Spanish tools-module lines break their second clause with an em dash. All four siblings do it that way; `es` was the only one with a second colon.

### Item 8's remainder (3)

Both blocks are gone. Nine new `common` keys carry the six strings `_connect_url_hint` assembles and the three `_oauth_creds_hint` returns, translated in all five locales, so every sentence in that paragraph now follows the catalog.

Three things stay verbatim, and this is a decision rather than an omission: the resolved URLs, the two `<your-home-assistant-url>` / `<home-assistant-ip>` stand-ins, and the `Client ID` / `Client Secret` labels. The stand-ins are addresses to copy. The labels name the two fields of the client an admin pastes them into, and those clients label them in English whatever this instance's language is — the same rule as your round-one item 2, applied to a label instead of an entry title. Placeholder parity keeps `{url}` pinned in the three keys that interpolate one.

For each locale I took the wording from the field it points at rather than translating the English afresh: `oauth_select_legacy_mode` names that locale's own `webhook_auth` label and `legacy` selector option ("Modo de autenticación" / "OAuth heredado", "Mode d'authentification" / "OAuth hérité", and so on), and `connect_local_lan` names its `bind_host` label. `connect_urls_label` drops the `(s)` in `es` and `ru`, where the plural of "URL" is unmarked.

### Failing loudly (4–6)

4. The `except` logs at warning with `exc_info` now. Your reading of the `ImportError` is why: the function-local core import fails inside the same broad `except`, and at debug nobody would ever see it. The prefix no-op is its own warning line, and its own test — a catalog that loads with nothing under `component.ha_mcp_tools.common.` renders pure English, and until now no level said so.
5. `_COMMON_FALLBACKS` is pinned to `strings.json`'s `common` block, keys and values, mirroring the `strings.json` ↔ `en.json` test. Injection: change one English value in the fallback dict, test red.
6. The comment and docstring say "system-configured language" now, and say why: `hass.config.language` is instance-wide, and an options flow is handed no requester language. I checked that on the owning layer rather than inferring it from our side: at 2026.7.4, `homeassistant/data_entry_flow.py`, `homeassistant/helpers/data_entry_flow.py` and `homeassistant/components/config/config_entries.py` mention no language at all, `FlowManagerIndexView.get_context` returns `{}`, and `OptionManagerFlowIndexView` does not override it — only the config-entries flow view does, to add `source` and `entry_id`. So the prose was the thing to fix, not the behaviour. The seam assertion is in: `seam.assert_awaited_once_with(flow.hass, "de")`, and hardcoding `"en"` at the call site now fails exactly one test instead of none.

### Test hardening (7–12)

7. The tool set comes from `extract_tools()` now, so the PR that adds a tool is the one that owes the translations. `test_tool_docs_sync.py`'s header records the same reasoning for the sync check, which is what convinced me this was the intended direction rather than a workaround.
8. `_fill` replaces the two-template loop and catches `Exception`. A translated `{component_version.major}` is the test case: placeholder parity reads the set as unchanged, `.format` raises `AttributeError`, and the form used to go down. The unreachable `return ""` is gone — the English constant is formatted after the guard, unguarded, as you suggested.
9. Missing keys count as untranslated now: `(identical + missing) / len(english)`, so `messages: {}` scores 100% instead of 0%. All four surfaces have a ceiling. The component catalogs needed a looser one than 5%: your 8.3% was 7 of 84, and after this PR's nine new keys it is 7 of 93 (7.5%) — six product names plus "Update", all correct. I set that ceiling at 15% rather than just above the high-water mark, so a future locale that keeps one more product name does not have to move it. Both add-on flavors are at 0% and share the 5% ceiling.
10. Native names are pinned by comparing the catalogs to each other and to English's, rather than hard-coding six literals: `es.json` shipping "Deutsch" is red now, and so is any catalog carrying "English".
11. `primary_tag(tags)` lives in `_tools_meta` and is imported. Injection: make it return `"Other"` unconditionally, test red.

    Record-straightening on your last paragraph: you are right, and it is worse than a wrong citation. `validate_translation_value` exists nowhere in the repo, and nothing in the loader rejects trailing whitespace — `_validate_string_map` only checks that keys and values are strings. The decision stands for a reason I can now demonstrate instead of assert: the separator is layout, and layout is script-dependent, which is item 13's last point.
12. AGENTS.md § Translations qualifies the omission rule for `tool_groups` and `tools`, states that adding a tool means translating it in every locale in the PR that adds it, notes why the failure lands there now, and records the per-surface ceilings and the add-on schema check's second direction.

### Small stuff (13)

The comment says 2.1%. `isinstance(result, Mapping)` matches the `websocket_api` seam. The empty-value filter has a test — deleting `and value` from the condition is now the only mutation that turns it red. The separator after `panel_hint` is conditional on the sentence not ending in full-width punctuation, so zh-Hans gets none.

`.get("dir")` is in, but I could not make it matter, and the attempt is worth reporting. I injected the failure it is meant to survive — deleted `meta.dir` from `es.json` — and the test stayed green. `load_catalogs` defaults a missing `dir` to `"ltr"` (`_i18n.py:95`) and raises for any other value (`:98-99`), so a registered catalog cannot be missing the key, and the `in {"ltr", "rtl"}` assertion downstream is unreachable by construction. What that turned up is a real gap next door: both of those raises were untested, including the one guarding the value that reaches the rendered `<html dir>` attribute. They are pinned now, and both go red when their condition is removed.

### Verification

Every new assertion was injection-checked, and the mutation was the guard's condition rather than the behaviour it gates: sixteen injections, fifteen red, each restored afterwards. The sixteenth is the `meta.dir` case above — it stayed green, which is why that paragraph exists. That set includes the empty-collection case: with the add-on glob emptied, the parametrized ceiling check reports "got empty parameter set" and skips silently, which the summary does not count as a failure, and only the sister assertion goes red.

Full run with `--maxfail=0`: 9458 passed, 172 skipped, 9 errors, all nine pre-existing environment failures that reproduce on untouched master (five need `aiohttp` as a dev dependency, four need `docker`). The reference before these commits was 9432, and the 26 new tests account for the difference exactly.

One thing outside this PR that I am not touching, since it is neither mine nor in these files: `tests/src/unit/test_fields_projection_docstring_completeness.py:686` skips with "got empty parameter set for (spec)" on a clean checkout. That is the same silent-skip class as above, and it has no sister assertion.

### One open question

`config.abort.unsupported_home_assistant` names both entry titles, and the catalogs disagree about them. `de` keeps "HA-MCP Server" and "HA-MCP File & YAML Tools" verbatim; `es`, `fr`, `ru` and `zh-Hans` all render them as description ("El servidor HA-MCP integrado en el proceso", "HA-MCP 文件与 YAML 工具"). That is the mechanism behind your round-one item 2, on a string neither round flagged.

I did not touch it, for two reasons. It is a four-locale change to strings this PR otherwise leaves alone, and I am not convinced the rule applies: in `enable_llm_api` the literal is what a user has to spot in a dropdown, whereas here it reads as prose on an abort screen, where naming the product in the reader's own language may be the better text. Your call — say which way and I will sweep it, here or separately.

kingpanther13 (changes requested):
All thirteen verified in, and all three flagged deviations are sound — I re-checked the cross-referenced field wordings in all five locales (`oauth_select_legacy_mode` naming each locale's own `webhook_auth` label and `legacy` option, `connect_local_lan` naming its `bind_host` label) and every share, count, and injection claim that could be checked statically, including the one you reported green: `_i18n.py:95` does default a missing `dir`, the assertion really is unreachable by construction, and the two loader raises you pinned instead were genuinely untested. Reporting the failed injection with its mechanism, then converting it into the coverage next door, is the best thing in this round.

Requesting changes once more. Two defects this round introduced (both one-liners), the observability follow-through, my answer to your open question, and the skip you flagged.

### New this round

1. **`scripts/extract_tools.py:189` reads source without an encoding, and your new test is its first caller in the unit suite.** `ast.parse(f.read_text())` uses the platform codec; on any non-UTF-8-default box (stock Windows) all five `test_settings_catalog_keys_name_real_groups_and_tools` cases die with `UnicodeDecodeError` — reproduced here twice independently; Linux CI can't see it. Every other `read_text` in that file already passes `encoding="utf-8"`; `:189` and `:363` are the two that don't. Fix both. While you're in the file: give `_renderable_groups_and_tools()` a discovery guard (`assert tools` with a message) — `extract_tools()` silently `continue`s past a hardcoded file that has moved (`:186-188`) and `sys.exit(1)`s on a duplicate tool name (`:203-209`), which inside pytest surfaces as a bare `SystemExit`; a shrunken parse otherwise fails as "translates tool(s) that do not exist", telling a translator to delete legitimate entries.
2. **The prefix no-op warning guards a case Home Assistant cannot produce, and misses the one it can.** Verified against core 2026.7.4: `_TranslationCache.get_cached` short-circuits single-component lookups to `category_cache.get(component, {})` (`translation.py:200-201`), so everything `loaded` can contain is already `component.ha_mcp_tools.common.`-prefixed — `if loaded and not translated:` (`config_flow.py:264`) can only fire on a catalog of empty/non-string values. The realistic developer error — the `common` category never built, the block missing from `strings.json` — arrives as `{}`, skips the guard, and renders pure English with nothing logged at any level: the exact silent no-op item 4's second half existed to make visible, and the one the post-merge dev-server live test needs to see. Drop the `loaded and`; add the `AsyncMock(return_value={})` caplog case next to `test_catalog_without_our_prefix_says_so` (the two existing `{}`-returning tests don't assert on caplog, so they stay green). Same standard one function up: the non-`Mapping` discard at `:216-220` is a second unlogged path to the same pure-English outcome — one warning naming `type(result)`.

### Pin what item 4 bought

3. **The two warning levels aren't pinned.** Reverting the failing-lookup `except` (`:249`) or `_fill`'s warning (`:183`) to debug keeps every test green — `test_failing_lookup_degrades_to_english` asserts only the fallback text. The prefix-no-op sibling already pins its level via `caplog.at_level(logging.WARNING)`; give these two the same treatment. Log level was the literal ask.

### Close the loop the new contract opens

4. **The `tools` section has exactness but no quality floor.** 87 titles + 87 descriptions per locale is the largest translated surface in the repo — bigger than the component catalog that just got a ceiling — and nothing would notice English text sitting in the values. Meanwhile item 7's new contract means every tool-adding PR now owes five languages before it can go green, which is pressure toward machine-pasted English nothing catches. `extract_tools()` now supplies the English titles and descriptions in the same test module, and the check lands with complete headroom: 0 of 87 titles and 0 of 87 descriptions are English-identical in any locale today.

### Your open question — sweep the entry title, keep the prose

5. You had the right instinct about the first name and the wrong conclusion about the second. "El servidor HA-MCP integrado en el proceso" is prose about why setup failed — nobody has to find it on a screen; leave it (and `de`'s variant) alone in all five. But the second sentence is an instruction to go use a specific entry, and that entry's title is a force-renamed Python literal (`const.py:40`, `__init__.py:3567`) — the integrations page shows "HA-MCP File & YAML Tools" in English in every language, always. The decisive part is internal to this PR: your own new `tools_module_not_installed` / `tools_module_not_loaded` keys name the *same* entry and keep it literal in all five locales, so shipping as-is renders the same title two different ways in the same language on two screens a user moves between. Sweep `config.abort.unsupported_home_assistant` in `es`/`fr`/`ru`/`zh-Hans` to the literal; one rule, not two.

### The skip you flagged

6. Your diagnosis is right, and the fix is class-wide, not another sister assertion: no `TOOL_SPECS` entry carries `exclude_internal`, so `test_fields_projection_docstring_completeness.py:686` collects an empty set and has never run — and the mask it validates (`:527`, `emitted -= spec.get("exclude_internal", frozenset())`) subtracts an always-empty set, so the guard *and* the mechanism are both dead. Set `empty_parameter_set_mark = fail_at_collect` in `tests/pytest.ini` and in `pyproject.toml`'s `[tool.pytest.ini_options]` (neither sets it today) — that makes the whole silent-skip class impossible repo-wide, and it's why a hand-written assertion would be wrong here anyway: an empty exclusion list is the *desirable* state, so asserting it non-empty would demand a mask exist. Enabling it forces the decision on the dead machinery: delete `exclude_internal` and its test unless you know a current internal-key case that should populate it. Your two glob discovery guards can stay — they fail with better messages than the ini setting does.

### Small stuff, while you're in the files

7. `_FULLWIDTH_TERMINATORS` contains `”` (U+201D) — that's the ordinary Latin closing quote, not a full-width glyph, so a future Latin-script `panel_hint` ending on a quoted phrase silently loses its separator and runs into the next sentence. Drop it, or key the separator off the language instead of sniffing the last character.
8. English `connect_local_lan` still says "when bind host is 0.0.0.0" while all five translations name the field's actual label — align the English source (it's what the next maintainer edits) and refresh the baseline.
9. `ru` `oauth_not_serving`: "пока их не использует" reads "does not use them yet" — "пока их не обслуживает" is what the English says. Same-form terminology drift while you're there: ru "с машины Home Assistant" vs the `bind_host` description's "с самого компьютера", de "Rechner" vs "Maschine", zh "机器" vs "主机".
10. `_versions_hint`'s own degradations (`:733-736`, `:747-748`) still log at debug — after this round they're the only paths that put a visible fallback on the same form without a visible log, and the file's comment argues warning-by-default. Raise them.
11. `str(integration.version)` renders the string `"None"` for a manifest without a version — no exception, so the freshly translated `version_unknown` never fires for the likeliest defect it exists for. Guard the `None`.
12. `_fill`'s warning blames the translation ("Unusable version_line translation") even when the failure is our own caller/template argument mismatch — it then correctly crashes on the English constant, but the log line sends the reader to the wrong place. One distinguishing word.
13. `ids=lambda p: p.name` on the add-on `Path` parametrize (`addon_dir0`/`addon_dir1` at the `-k` prompt today). And AGENTS.md § Translations: state the actual ceilings (5% settings/add-ons, 15% component) and scope that sentence so it doesn't read as covering `tool_groups`/`tools`, which have exactness, not a ceiling.

Item 2's one-word guard change plus its test is the only thing here I'd block on alone — it's the difference between the post-merge live test proving the feature works and proving nothing. The rest is one pass through files you're already in. The `_fill` shape, the omission-counting inversion, and the fabricated-citation correction were all handled exactly right.

Patch76:
All thirteen in, across five commits on `bfe428f0`. Both defects this round introduced are real, and I checked each mechanism at the source before fixing it rather than taking the report on trust — though not by reproduction: I have no non-UTF-8 box here, so item 1 rests on the missing encoding and the caller change, not on a `UnicodeDecodeError` I watched happen.

### New this round (1–2)

1. **Encoding.** Confirmed: `:189` and `:363` were the only two `read_text` calls in that file without an encoding; the other four already pass utf-8. Both fixed. `_renderable_groups_and_tools` now asserts the parse found something — injected by emptying `TOOL_FILES`, and the failure names the parse instead of telling a translator that 87 tools "do not exist".

2. **The guard watched the unreachable branch.** Verified against core 2026.7.4: `_TranslationCache.get_cached` returns `category_cache.get(component, {})` for a single-component lookup, so a non-empty result is always already prefixed and `{}` is what a missing category produces. The condition is now `not translated`, and the reachable case has its own caplog test. The non-`Mapping` discard at the seam logs the type it got — same pure-English outcome, previously with nothing at any level.

### Pinned (3)

Both levels now fail when reverted to debug. Worth recording how that nearly slipped: my first injection run reported the `_fill` level as unpinned, and it was my harness that was wrong — it pointed at the dotted-placeholder test rather than the one carrying the `caplog` assertion. The corrected mutation goes red. Same class of instrument error caught the `TOOL_FILES` injection: I had written `TOOL_FILES = [] or sorted(...)`, which evaluates to `sorted(...)` and changes nothing. Two green results, both my measurement rather than the guard — which is exactly why a green injection gets a second look before it becomes a claim.

### The tools floor (4)

In. Your figure reproduces: 0 of 87 titles and 0 of 87 descriptions are English-identical in any of the five locales, so the 5% ceiling lands with the full surface clear. It reuses the same untranslated-share helper as the other three surfaces, so a missing tool entry counts the same way a copied one does.

### The entry title (5)

Swept in `es`, `fr`, `ru`, `zh-Hans`; the first sentence's prose is untouched in all five, as you said. The internal argument is the one that decided it for me: my own `tools_module_*` keys name that entry and keep it literal in every locale, so leaving the abort string translated would have rendered one title two ways in one language, on two screens a user moves between. I verified the force-rename rather than take the citation — `TOOLS_ENTRY_TITLE` at `const.py:40`, applied at `__init__.py:3567`.

### The skip (6)

`empty_parameter_set_mark = fail_at_collect` is set in both configs, and proven both ways on a throwaway empty parametrize: with the setting it is a collection error, with `-o empty_parameter_set_mark=skip` it is the silent skip again.

`exclude_internal` and its test are deleted. Your reading holds — no `TOOL_SPECS` entry has ever carried it, so the mask subtracted an empty set and its guard collected nothing — and I could not find a current internal-key case that should populate it. A comment records what to restore, and why, the day a spec needs one.

**One thing CI has to answer, because I cannot.** The setting is repo-wide and I can only run `tests/src/unit` and `tests/addon` locally. Those are clean — the full run's skip count went 172 → 171, exactly the one you named, with no new collection errors. If an e2e module carries an empty parametrize, it now fails at collection instead of skipping, and that lane is where it will surface.

### Small stuff (7–13)

`”` is gone from `_FULLWIDTH_TERMINATORS` — you are right that it is the Latin closing quote, and a Latin `panel_hint` ending on a quoted phrase would have lost its separator. English `connect_local_lan` now names "Network access" like all five translations do, baseline refreshed. `ru` reads "не обслуживает". The three terminology drifts are aligned to what each catalog already uses: `de` "Maschine", `ru` "компьютера", `zh-Hans` "主机". `_versions_hint`'s two degradations log at warning. `str(integration.version)` is guarded, with a test — the read succeeds for a version-less manifest, so the `except` never ran and the translated wording could never appear. `_fill`'s warning names both causes rather than blaming the catalog. The add-on parametrize has readable ids. AGENTS.md states the ceilings (5% / 15%) and separates them from the `tool_groups`/`tools` exactness rule.

### Verification

Seven guards injected, all red once the mutations were correct, each restored. Full run with `--maxfail=0`: 9466 passed, 171 skipped, 9 errors — the same nine environment failures that reproduce on untouched master. The 8-test delta and the one fewer skip account for the change exactly.

Patch76:
And it was supposed to be just a little translation ...

kingpanther13 (changes requested):
All thirteen in and implemented as asked, not adjacent to it — including both mechanisms you checked at source, which I re-verified independently (core's `_async_load` overlays the requested language onto English in the same cache entry, so the widened guard cannot false-positive on an Italian install; the force-rename citation is accurate; the 0-of-174 figure reproduces in all five locales; the +8/−1 test arithmetic is exact). Your two instrument-error reports are the most valuable paragraphs in the response — both diagnoses check out against the source, and a green mutation getting a second look is what makes the other injection claims worth something. CI also answers your open question: the e2e lanes collected clean under `fail_at_collect`.

Four things left, none structural. Land these and this is an approve.

1. **`scripts/extract_tools.py:408` — the write is the one call still missing `encoding="utf-8"`, and it's the destructive one.** The two `write_text` calls below it (`:411`, `:414`) pass it; this one doesn't, and `generate_tools_json` emits `ensure_ascii=False` text containing arrows, ∈, ⚠ and emoji that cp1252 cannot encode. `Path.write_text` opens with mode `w` — truncate first, encode during write — so `python scripts/extract_tools.py` on a stock Windows box leaves `site/src/data/tools.json` at **zero bytes** and aborts, after which your freshly-fixed `--check` blames the sync against the wiped file. The exact sibling of last round's item 1, three lines below the two you fixed. One word.
2. **Pin the two levels item 10 raised.** No test asserts "Could not read the component version" or "Could not read the server version" — reverting either of `config_flow.py:763-768` / `:779-785` to debug keeps the suite green. This is the same class item 3 closed, created by item 10's own fix. `test_versions_placeholder_is_failure_proof` already drives both failure paths at once; add `caplog`, wrap in `at_level(WARNING)`, assert both substrings — nothing new to write.
3. **`_tools_module_hint` is now the last silent degradation on that form.** `config_flow.py:821` logs a read failure at debug and returns `None`, which makes the whole tools-module paragraph vanish from the options form — a stronger user-visible loss than the "unknown" / "not installed yet" wordings raised to warning forty lines above it. It came from master, but the commit that swept the class stopped one function short. Raise it, and pin it with the others.
4. **The tools floor misses its motivating case.** It reuses the 5% share against 174 keys, so a PR that adds one tool and pastes the English title + description into all five locales scores 2/176 = 1.1% and stays green — it takes five pasted tools to trip. That paste is the exact case item 4 named. Add the paste signature — fail any tool whose title *and* description are both English-identical, with the message naming the tool — and keep the share ceiling alongside it for the wholesale-stub case. A scattered identical title alone stays legal, so no brittleness for a future legitimate match like "Backup".

Small stuff, one pass:

5. English `connect_local_lan` (and all five translations) says "when Network access is 0.0.0.0" — a value the user never sees; the dropdown shows the hard-coded "Local network (default)". Same rule as the loopback fix: name what's on screen — "Local network" — in all six, baseline refresh.
6. The non-Mapping path now warns twice for one cause, and the second line ("the de translations carry no `component.ha_mcp_tools.common.` keys") is untrue in that case — we discarded them. Make the caller's message accurate ("no usable translations") or skip it when the seam already warned.
7. The U+201D story inverted on inspection: `”` **is** Simplified Chinese's closing quote — your own zh-Hans `tools_module_not_installed` ends on it — while the kept 「」『』 are the traditional forms zh-Hans doesn't use. Nothing breaks today because only `panel_hint` flows through `_sentence_prefix` and no `panel_hint` ends on a quote, but the character-sniffing rule has now bitten in both directions across two rounds. Key the separator off the language (CJK locale → no ASCII space) and the whole glyph-list class goes away; either way, pin the behavior with a test — re-adding `”` is green today.
8. `websocket_api.py:4743` — the seam your comment cites as its model still discards a non-`Mapping` silently; the two mirrored copies now differ in exactly the respect this round fixed. One log line brings them back in step.
9. While you're in the files: `_english_tool_texts()` bypasses the discovery guard, so a shrunken parse reaches `_assert_not_a_copy` as `ZeroDivisionError` instead of the guard's message — route it through the guarded helper or assert non-empty; `_settings_catalog(locale)["tools"]` at `test_locale_parity.py:607` vs the sibling's `.get("tools", {})` at `:511`; `sys.modules.setdefault` at `test_config_flow.py:1394` is never undone — `monkeypatch.setitem` removes the ordering dependence; `extract_tools()`'s duplicate-name `sys.exit(1)` still surfaces in pytest as a bare `SystemExit` — raise a `ValueError` and let `main()` translate it to an exit code; pin the AGENTS.md "5%" / "15%" literals to `_MAX_ENGLISH_IDENTICAL_SHARE` / `_MAX_COMPONENT_IDENTICAL_SHARE` the way `test_agents_md_lists_every_shipped_locale` in the same file already pins the locale list; and the version-`None` guard you added at `config_flow.py:755` has three unguarded siblings — `__init__.py:3577` (a version-less manifest renders a device with `sw_version` "None"), `__init__.py:3094`, and the advisory comparisons in `embedded_setup.py:730`/`:887` — sweep them the same way.

The `exclude_internal` deletion took the harder half of that item and left the right tombstone; the entry-title sweep's wording is right in all four locales (ru's «» around the literal is the correct treatment); and deciding that sweep on the internal two-screens argument rather than on reviewer authority was the right reason to change four catalogs.

Patch76:
All nine in, across five commits on `7ed4921c`. Item 7 is the interesting one: it reverses a correction I made last round, and it is right.

### The two defects (1–2)

1. Confirmed and fixed. `:408` was the only remaining `read_text`/`write_text` in that file without an encoding, and it is the destructive one — truncate first, encode during the write, so the failure leaves the file empty rather than untouched. The two writes three lines below it already passed utf-8. I did not reproduce the `UnicodeDecodeError`: no non-UTF-8 box here, so this rests on the missing argument and on `generate_tools_json` emitting `ensure_ascii=False`.

2. Fixed as described. `extract_tools()` also stops calling `sys.exit` — a duplicate tool name now raises `ValueError` and `main` turns it back into exit 1, so the locale check reports a named failure rather than a bare `SystemExit`.

### Levels (3)

Both pinned in `test_versions_placeholder_is_failure_proof`, which already drove both paths. You were right that item 10 created this gap: the fix raised two levels and left them unasserted, exactly the hole item 3 had closed one round earlier.

### `_tools_module_hint` (3, second half)

Raised and pinned. It is the largest of the three losses — the whole paragraph disappears rather than a word inside it degrading.

### The tools floor (4)

Right, and the arithmetic is the point: one pasted tool scored 2 of 176. The paste signature is in — a tool whose title *and* description are both English-identical fails, and the message names the tool. The share ceiling stays alongside for the wholesale-stub case, and a lone matching title is still legal, so a future legitimately-identical name does not trip it. Injected by pasting one tool's English pair into `es`: red, naming that tool.

### The `”` inversion (7)

You are right, and my round-three reasoning was wrong in a way worth stating plainly: I removed U+201D as "the ordinary Latin closing quote" while this PR's own `zh-Hans` `tools_module_not_installed` ends on that exact character, and I kept 「」『』, which zh-Hans does not use. The glyph list was wrong in both directions and I had touched it twice without noticing either.

Keyed off the language now, and pinned in both directions: a CJK language gets no ASCII space, and a Latin sentence ending on U+201D still gets one. The second test is the one that would have caught my round-three edit.

### Small stuff (5, 6, 8, 9)

5. Fixed in the English source and all five translations. The dropdown shows "Local network (default)", so the sentence now names that instead of the bind value; each locale quotes the literal in its own convention.
6. The prefix line no longer says a catalog "carries no keys" when the seam discarded it — there was no catalog. It reads "No usable … translations under …", which is true in both paths.
8. Both copies of the seam log the discard now. That log line was itself new and unpinned, which by this round's standard is the next finding, so it has a test rather than waiting for one.
9. All of it: `_english_tool_texts` routes through the discovery guard, `.get("tools", {})` matches its sibling, `monkeypatch.setitem` replaces the `sys.modules.setdefault` that was never undone, `extract_tools` raises instead of exiting, and the AGENTS.md percentages are pinned to `_MAX_ENGLISH_IDENTICAL_SHARE` / `_MAX_COMPONENT_IDENTICAL_SHARE` the way the locale list already was.

The three unguarded `version` siblings are swept. Each takes the unreadable-manifest path its own module already defines rather than a new special case: the tools device falls back to the compiled-in `COMPONENT_VERSION`, `get_caller_token` returns its existing `manifest_unreadable` response, and both `embedded_setup` comparisons skip the advisory gate instead of feeding the string "None" to `AwesomeVersion`.

### Verification

Eight guards injected, all red, each restored. Full run with `--maxfail=0`: 9470 passed, 171 skipped, 9 errors — the same nine environment failures as before, skips unchanged, and the +4 is exactly the four tests this round adds.

One CI note, since the lane history shows it: the first run of `E2E Validation (embedded, ubuntu-latest)` failed on `test_delete_refuses_freshly_created_backup`, with the backup manager reporting `upload_failed`. I re-ran it rather than touching code, on two grounds — the arm sibling of the same embedded backend ran the same test green on the same commit, and the failing surface is Home Assistant's backup upload, which nothing in this round goes near. The re-run is green, so all 31 checks pass on `7ed4921c`.

kingpanther13 (changes requested):
All nine in, correctly, and the round's self-corrections were the right calls — I re-verified the U+201D reversal (your zh-Hans `tools_module_not_installed` does end on it; 「」『』 appear in no shipped catalog), the paste-signature margin (0 identical titles or descriptions in all five locales; the 2-of-176 arithmetic is exact), and the complete encoding sweep (all nine IO calls in `extract_tools.py` now carry utf-8). I also pulled the failed e2e attempt's log rather than take the re-run on trust: `test_delete_refuses_freshly_created_backup` died in HA's backup manager (`create_backup` → `upload_failed`), a surface this diff never touches, with the arm sibling green on the same commit — the re-run was the right call, made for the right reasons.

Four items this time. Three are the "revert stays green" class you closed on yourself this round; one is a real regression the language keying introduced.

1. **The language-keyed separator strips the space from English text.** `_sentence_prefix` now reads only `hass.config.language` — but the text doesn't always follow it. A `ja`, `ko`, or `zh-Hant` instance gets our *English* catalogs (core overlays the requested language onto English, and we ship five languages), so the options form renders "…for tool management and server settings.Changes here are applied on save." The same jam hits a `zh` install whenever `_common_strings` degrades to English — the exact path the warning you added exists for. The glyph rule got both right because it read the string; this is the third consecutive round this one separator has been wrong in some direction, each time for skipping half its inputs. Fix that closes the class: strip the space only when the language is in the set **and** the sentence differs from `_COMMON_FALLBACKS["panel_hint"]` — an unshipped language's catalog is the English source byte-for-byte, so the comparison is exact. Pin with a `ja` test asserting the space survives. Two adjacent corrections while you're in the function: drop `"ko"` from `_NO_ASCII_SENTENCE_SPACE` (Korean ends sentences with an ASCII full stop and inter-word spacing — a future `ko` catalog wants the space even fully translated; `zh`/`ja` fit the constant's rationale, `ko` doesn't), and pin the *mechanism*: both current tests pass under the old glyph tuple too — one character in the CJK test (`"打开“设置面板”"`, ending on the quote the tuple missed) makes the language rule the only implementation that passes.
2. **The four version-`None` guards landed unpinned, and one changed a contract.** Deleting any of `__init__.py:3094`, `:3581`, `embedded_setup.py:730`, `:891` keeps the suite green — only round 4's config-flow sibling is asserted. Your own words this round: "that log line was itself new and unpinned, which by this round's standard is the next finding." The embedded_setup pair is near-free (the existing loader-boom tests take a `SimpleNamespace(version=None)` variant). And `:3094` is more than a log: a version-less manifest used to return `success: True` with `version: "None"`; it now returns `success: False, error_code: "manifest_unreadable"` — and the sole consumer (`_fetch_caller_token`, `tools_filesystem.py:259-270`) never reads `error`, so it rediagnoses the manifest problem as "did not return a usable token" with auth-flavored suggestions. Unreachable in production — core's loader refuses a version-less custom integration — which is precisely why tests are the only thing keeping these branches honest. Pin all four, and pass the component's `error` string through the consumer when the response carries one, so the accurate message isn't dead-ended.
3. **The paste signature is blind to the five tools most likely to be pasted next.** `_english_tool_texts()` reads the parsed docstrings, but with the feature flag off — the default — the settings UI renders `FEATURE_GATED_TOOLS`' hand-written text, and five of the seven differ (verified: `ha_config_set_yaml` shows "Set YAML Config" / "Add, replace, or remove top-level keys…", not the parsed "Raw YAML Config Edit" / docstring). A translator pasting the English they can actually *see* for those tools byte-differs from the parsed English and sails past both the paste check and the share. Three lines in the helper — prefer `FEATURE_GATED_TOOLS[name]`'s title/description when present — and the two English sources gain a drift check for free.
4. **End the encoding recurrence structurally.** Two consecutive rounds fixed the same defect in the same file; all nine calls are correct today and nothing keeps a tenth honest. A ~ten-line AST test over `scripts/extract_tools.py` asserting every `read_text`/`write_text`/`open` passes an `encoding` keyword closes the class the way `fail_at_collect` closed the skips. While you're in the file: `--check` currently reports "All files in sync" when the README markers are *lost* — `:348-352` warns to stderr and returns the README unchanged, so the comparison passes on the very failure it just printed — the same verification-path-passes-on-its-own-failure shape as the zero-byte write, three lines away. Lost markers should fail the check.

Small stuff, one pass:

5. AGENTS.md § Translations documents the share ceilings but not the absolute paste rule you just added — the commit that wrote the paragraph is titled "state the locale ceilings and what they do not cover", and it now doesn't cover the gate that will actually stop a contributor. One clause. And bound `test_agents_md_states_the_current_ceilings`' regex to the section — it currently matches `(\d+)% for the` across the whole file, so any future unrelated sentence fails it; the locale-list sibling shows the marker-line pattern.
6. Two unpinned wordings from this round: the corrected "No usable … translations" line (the reverted, untrue sentence passes both existing tests — one substring assert), and the quoted "Local network" literal, which nothing ties to the hard-coded label at `config_flow.py:569` — rename the option and six catalogs go silently stale, the same drift class the ceilings test just closed.
7. `de` and `es` now mix quote conventions inside one `common` block — the new strings use „…“ / «…» while the block's earlier UI literals use ASCII quotes; `fr`/`ru`/`zh-Hans` are internally consistent. Pick one per catalog (the new, typographically correct one is fine — sweep the block to match it).
8. While in the files: have `_common_strings` return the language it actually used so `_configured_language`'s duplicate read can't diverge from the catalog fetched (today they agree by coincidence of implementation, not structure); truncate the paste check's failure list to `[:20]` like its sibling; add `versions.startswith("Component ")`-shape anchoring to the tools-hint test so a `_versions_hint` regression can't hide behind its `not in` assertion; and the `# pragma: no cover` at `__init__.py:3100` sits on a branch your own new `raise` makes deliberately reachable — drop the stale pragma.

Item 7's handling was the strongest thing this round — reversing your own two-round-old correction, stating why it was wrong in both directions, and writing the test that would have caught your round-3 edit. Item 1 above is the same lesson one level up: the heuristic moved from the string to the language and dropped the other half of the inputs again. Gate on both and this class is done for good.

Patch76:
All eight in, across five commits up to `3729318e` — the sixth commit there only re-sorts an import block that Lint caught. Item 1 was a regression I introduced, and the shape of it is the same mistake one level up, exactly as you put it.

### The separator (1)

You are right on all three counts, and the first is the one that matters: keying off the language alone assumes the text follows the language, and core guarantees it does not. `helpers/translation.py` loads `[LOCALE_EN, language]` and merges English first as the documented fallback, so an instance set to any language this integration does not ship renders these sentences in English while `hass.config.language` still names a script that sets its own spacing. I moved the heuristic from the string to the language and dropped the other half of the inputs, one round after doing the same thing in the other direction.

Both inputs decide it now: the language must be in the set **and** the sentence must differ from the English source. `ko` is out — Korean separates words with ASCII spaces and ends sentences on an ASCII full stop, so the full-width rationale never applied to it, and a future `ko` catalog wants the separator like a Latin one.

The mechanism is pinned rather than agreed with. The CJK test now ends on U+201D, and I checked it discriminates instead of assuming it: restoring the pre-round glyph tuple in place of the language rule turns that test red. A `ja`/`zh-Hans` pair asserts the space survives on English text, and a `ko` case asserts it survives on translated text, so neither direction can be flipped silently.

### The four version guards (2)

Confirmed before fixing: deleting all four left the unit suite at 8769 passed, zero failed. All four are pinned now, and the two in `embedded_setup` needed the warning asserted rather than the return value — without the guard the literal "None" reaches `AwesomeVersion` and the comparison decides the outcome quietly, so both paths return the same thing and only the log tells them apart.

`__init__.py:3094` is more than a log, as you said. The consumer now reports the component's own `error` when the response carries one; re-diagnosing an unreadable manifest as a missing token sent the operator to the reload and admin-rights suggestions, neither of which can fix it. The tools-device guard is asserted at source level like its two siblings in that class — `_async_setup_tools_entry` drives the token Store, the backup migration and ~10 service registrations, so no unit harness runs it — but the assertion names the guard's *condition*, so deleting the guard fails it.

The stale `# pragma: no cover` is gone; the new `raise` makes that branch deliberately reachable, and the test now reaches it.

### The paste signature (3)

Verified and exact: five of the seven differ, `ha_config_set_yaml` reads "Set YAML Config" as a stub against "Raw YAML Config Edit" parsed.

I went one step past the suggestion. Preferring the stub swaps which English the check is blind to rather than closing the hole: with the flag on, the tool registers and the UI shows the parsed docstring, so both are English a translator can be looking at depending on a setting. The check consults both now, and the share still uses the rendered set.

### Encoding, structurally (4)

An AST test asserts every `read_text`/`write_text`/`open` in the file passes an `encoding`, with a count floor beneath it — a parse that stopped finding calls would otherwise report no offenders and read as a pass. It counts nine today.

Lost markers now raise. You named the shape precisely: the unchanged return compared equal to the file it came from, so `--check` printed "All files in sync." directly after printing the warning that it could not find them. While in the file, `update_docs`' `sys.exit(1)` became a `ValueError` for the same reason the duplicate-name check did last round — a library function that exits hands its caller a bare `SystemExit` instead of a named cause — and `main` turns both into one exit 1.

### Small stuff (5, 6, 7, 8)

5. One clause, and deliberately one: `AGENTS.md` is at 47,284 characters against the 40,000 the size check warns at, so the paragraph states the rule and stops. The ceilings regex reads the section now, and I pinned that both ways — an unrelated sentence carrying the same shape elsewhere in the file leaves it green, the same sentence inside the section turns it red.
6. Both pinned. The reverted "carries no keys" sentence fails the prefix test now, and `connect_local_lan` is tied to the `BIND_HOST_ALL` label at `config_flow.py:594`, so renaming the option turns six catalogs red instead of leaving them quoting something that no longer exists.
7. `de` and `es` swept to one convention each. `fr`, `ru` and `zh-Hans` were already internally consistent.
8. All four: `_common_strings` returns the language it used so the separator cannot read a different one, the paste list truncates at 20, the tools-hint test anchors on the paragraph that must survive before asserting the one that must not, and the pragma is gone.

### Verification

Eighteen guards injected, each by breaking its condition rather than the gated behaviour. Sixteen went red first time. Two did not, and both times the mutation was at fault rather than the guard: the glyph case added the old check *in front of* the language rule instead of replacing it, so it never reached the gate, and the section-scope case inserted its sentence before the heading that immediately follows § Translations, which put it inside the very section it was meant to sit outside. Redone correctly both went the right way, and I added a control for the second so the new scope cannot pass by being vacuous.

Unit suite on the pushed head: 8782 passed, 170 skipped, zero failed, 5 errors — all five are the `aiohttp` introspection contract, which needs a dev dependency this checkout does not have. The pre-fix injection run sat at 8769, so the +13 is exactly the thirteen tests this round adds. CI is green on that head: 31 passed, 2 skipped, nothing else.

kingpanther13 (changes requested):
sorry, my Claude has been a real dick lately, it has been running the toolkit every time without me realizing it. Once these 3 findings are fixed I'll merge, as long as you don't change anything else

1. **`config_flow.py:119-121` names the wrong file.** The comment says the `_COMMON_FALLBACKS` ↔ `strings.json` mirror is "asserted in tests/src/unit/test_locale_parity.py"; the assertion is `test_common_fallbacks_mirror_strings_json` at `tests/src/unit/test_config_flow.py:1605` — `test_locale_parity.py` never references the constant. That pointer is how a maintainer finds the assertion item 1's `sentence != english` exactness rests on, so it has to point at the right file.
2. **Pin the dual-rendering branch itself.** `_english_tool_texts(as_rendered=False)` and the stub branch it feeds can be reverted today and the suite stays green — no shipped locale pastes either English, so the new check never discriminates on real data. One line inside the tools check closes it: `assert english != _english_tool_texts(as_rendered=False)`. It's the one vacuity guard this round didn't get, by the standard the round applied everywhere else (the AST count floor, the section control, the discovery guards).
3. **Count the share against both renderings, like the paste check.** `_untranslated_keys` still compares against the rendered set only, so a description pasted from the *parsed* English of one of the five differing gated tools reads as translated — up to 6 of 174 keys (3.4%) invisible to the ceiling. The argument that justified consulting both for the paste applies key-by-key to the share: count a key untranslated when it matches either English rendering.

Everything else from this round verified clean: the separator gate is right in all four quadrants and pinned discriminatingly, all five `version is None` guards are asserted, the passthrough strictly improves all three error shapes, the encoding class is closed with a non-vacuous floor — and refusing the stub-preference suggestion was correct; consulting both renderings was the better decomposition, and item 3 above finishes that same argument.

Patch76:
All three in, across two commits on `a1fc2184`.

### 1. The pointer

The comment above `_COMMON_FALLBACKS` now names `test_common_fallbacks_mirror_strings_json` in `tests/src/unit/test_config_flow.py`. `test_locale_parity.py` never mentions the constant — every reference to it under `tests/` is in `test_config_flow.py`. Naming the test as well as the file keeps the pointer alive when the assertion moves within it.

### 2. The vacuity guard

`test_settings_catalog_tools_are_translated` now asserts the two renderings differ before consulting both. Injected: making `_english_tool_texts` ignore `FEATURE_GATED_TOOLS`, so both renderings return the parsed text, turns it red in all five locales.

### 3. The share counts both renderings

`_untranslated_keys` takes an optional second English rendering and counts a key untranslated when the value matches either; `_assert_not_a_copy` passes it through, and among the four surfaces only the tools check supplies one — the messages, component and add-on calls are unchanged.

The six keys that differ between the renderings are `ha_config_set_yaml.title`, `ha_config_set_yaml.description`, `ha_config_get_yaml.description`, `ha_delete_file.description`, `ha_list_files.description` and `ha_manage_custom_tool.description` — 6 of 174, your 3.4%.

Pinned by `test_the_tools_share_counts_both_english_renderings`: a catalog that translates all 174 keys except those six, which carry the parsed English. Against the rendered set alone it leaves nothing untranslated, so a zero ceiling holds; with the alternate consulted it fails. Deleting the `or value == alternates.get(key)` clause turns that test red and nothing else in the file.

One gap I did not close, so you can weigh it rather than find it: reverting the `alternate=as_parsed` argument at the call site alone keeps the suite green. Discriminating that one keyword needs a synthetic catalog driven through the parametrized test, and pasting a whole tool trips the paste check first — so it would have to paste the five differing descriptions only (2.9%) and pad with byte-identical English until the share crosses 5%. I pinned the counting rule instead of the argument. If you want the padded-catalog version as well, say so and it goes in.

`ruff@0.16.0 check` and `format --check` are clean on both files. `test_locale_parity.py` plus `test_config_flow.py`: 128 passed.

kingpanther13 (approved):
All three verified in, and nothing else changed — the delta is two commits touching exactly the pointer comment, the renderings-differ guard, and the dual-rendering share count with its pin test. I read the full diff and re-checked each: the pointer now names `test_common_fallbacks_mirror_strings_json` and its actual file; the guard's failure message names both ways it can trip; and deleting the new `or value == alternates.get(key)` clause turns `test_the_tools_share_counts_both_english_renderings` red and nothing else in the file. The six differing keys match the measured set exactly. CI is green on `a1fc2184`.

On the padded-catalog variant you offered: no — per my comment above, nothing else changes in this PR. You pinned the counting rule and disclosed the call-site residual plainly, which is exactly the right way to hand a reviewer a known edge.

Approved. Thanks for staying with this through six rounds — every response came back cleaner than the ask, the injection discipline (including reporting your own instrument errors) set a bar this repo gets to keep, and every locale that follows lands on the rails this PR built. Italian and Dutch have much better tracks now.

PR #2062, opened 2026-07-28
feat: add the Italian (it) locale across all four translated surfaces

Patch76:
## What does this PR do?

Adds Italian (`it`) as a fully shipped locale. A language has to cover all four translated surfaces or none, so this adds all four catalogs in one go:

| Surface | File | Scope |
|---|---|---|
| Settings UI | `src/ha_mcp/settings_ui/locales/it.json` | 421 `messages`, 29 `tool_groups`, 87 `tools` (174 title/description values) |
| Custom component | `custom_components/ha_mcp_tools/translations/it.json` | all 93 `strings.json` keys, including the 16 `common.*` strings |
| Add-on (stable) | `homeassistant-addon/translations/it.yaml` | `name` + `description` for all 16 `schema:` keys |
| Add-on (dev) | `homeassistant-addon-dev/translations/it.yaml` | `name` + `description` for all 26 `schema:` keys |

The settings UI `messages` and the component catalog were produced by mapping each English key to its translation and writing the file from the English structure, so key set, key order and `{placeholders}` match the source by construction rather than by review. `en.json` leaves `tool_groups` and `tools` empty — English for those comes from the tool definitions at runtime — so those two sections are checked against the static parse in `scripts/extract_tools.py`, and `tool_groups` carries the same 29 renderable primary tags as every other shipped locale.

Four notes:

- Four values are byte-identical to English on purpose. In the settings UI `messages`, `common.minutes_short` — `{count} min` abbreviates the same way in Italian. In `tool_groups`, the labels `Assist`, `HACS` and `Matter` equal their key, which is what English renders: those are the words Italian uses too. Apart from those four, no translated string in any of the four catalogs repeats its English source — the component catalog, both add-on flavors and all 174 tool titles and descriptions are genuinely translated. (`meta.dir` is `ltr` here as in every other locale, English included; it is a direction flag, not text.)
- Where a string names something the reader has to find on another screen, the name stays verbatim and the prose around it is translated. That covers the `"Local network"` dropdown option in `common.connect_local_lan`, the `HA-MCP File & YAML Tools` entry title in `common.tools_module_*` and in `config.abort.unsupported_home_assistant`, the `HA-MCP Server` LLM API name in `options.step.init.data_description.enable_llm_api`, and `Client ID` / `Client Secret`. `common.oauth_select_legacy_mode` and `common.connect_local_lan` name this catalog's own field labels, since that is what the form shows in Italian.
- Where the same English string appears on more than one surface, the Italian is the same string too. Grouping every English value across all four surfaces by exact match, 53 groups span at least two surfaces and Italian renders all 53 consistently. Eighteen of those span only the stable and dev add-on catalogs, which are two flavors of the same file; the remaining 35 cross a real surface boundary — the feature toggles, the read-only switch, the snapshot age field — so the same option does not read differently in the web UI than in the add-on configuration. The per-locale figures are the same under either grouping: `ru` also has 0 divergent and `es` 1, while `fr` has 4, `zh-Hans` 18 and `de` 25.
- Terminology: "skill" stays untranslated as a product term — it names `skill_content` and `ha_get_skill_guide`, and `de` treats it the same way. The `Dashboard` and `Dashboards` tool groups are two separate renderable headings whose Italian loanword does not inflect, so the distinction is built rather than translated: they render as `Dashboard (screenshot)` and `Dashboard`. Quotation marks are `«…»` in all four Italian catalogs, including where the English source uses `'…'` around a UI label. Two ASCII cases remain on purpose: the `<a href="#">` attributes in `policies.rules.empty`, and the literal `'all'` in `advanced.enabled_tool_modules.help`, which names a config value rather than quoting prose and which every shipped catalog leaves as the English source spells it. `de` and `es` mostly keep straight quotes instead, so the guillemets are a per-language convention rather than a per-surface one.

Beyond the Italian catalogs, the review rounds pulled in fixes that are not Italian-specific and belong with the change that surfaced them:

- `settings.js` resolves the 409 `current_decision` enum and the 503 decide path through the catalog instead of interpolating backend English into translated sentences; `policies.pending.decision.*` is added to all seven catalogs, which fixes `de`/`es`/`fr`/`ru`/`zh-Hans` alongside `it`.
- Seven `zh-Hans` values lose a stray ASCII space between a placeholder and a CJK character — the three message shapes whose placeholder is filled with a localized label (`{title}`, `{operation}`, `{decision}`) rather than a Latin or numeric value. The eighteen `zh-Hans` values that still put a space beside a placeholder are correct: those interpolate counts, endpoints, env-var names, backup filenames, or the raw operator literal.
- The 503 body moves to `POLICY_UNAVAILABLE_MESSAGE` so the guard on it pins the string the server sends rather than a hand-copied one.
- `tests/js/harness.mjs` prefixes a rejected invoke so `_assert_clean_init` can see it; unprefixed, any test whose invoke rejects recorded nothing and passed its own negative assertions.
- `src/ha_mcp/settings_ui/locales/README.md` no longer says CI rejects a catalog that omits a key — that is false for `messages` and contradicted `AGENTS.md` § Translations.

`AGENTS.md` § Translations lists the shipped locale codes and `test_agents_md_lists_every_shipped_locale` pins that list against the catalogs, so the line moves in the same commit as the catalogs. #2056 edits the same line, and #2074 rewrites `locales/README.md` from the master base, so both will conflict textually with this branch; whichever lands second takes the rebase.

## Type of change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

What actually ran, rather than a blanket tick:

- CI is green on the head commit — 31 checks pass, 2 skip, none fail. That is the only run covering the whole suite; the boxes above stay unticked because I did not repeat a full local `pytest` this round.
- Locally: `test_settings_ui_js_behavior.py`, `test_settings_ui_i18n.py`, `test_locale_parity.py`, `test_rendered_scripts_parse.py` and `test_settings_ui.py` together, 442 passed. `ruff check` and `ruff format --check` clean on the changed Python; `mypy` clean on `settings_ui/__init__.py`.
- Every guard this PR adds or changes was checked by injection rather than by re-reading it, and each injection breaks the *condition* rather than removing the checked call. Where the point of a change was that the previous shape could not fail, the injection was run against both shapes: the whole-sentence alert comparison reddens on a drifted host key where the containment check it replaces stays green; the casefolded `untranslated` comparison reddens on a sentence-cased `Approved` that the exact comparison passed; the `harness.mjs` prefix reddens a negative-only test whose invoke was renamed, and leaves it green without the prefix. The blank-value check, a dropped `policies.operators.*` key, both emptied discovery sets and a reverted 503 branch each redden their own test and nothing else.
- E2E was not run locally — it needs containers. No LLM-agent test either: these catalogs render in the settings UI and the Home Assistant config flow, so nothing here reaches an agent's context.

## Checklist


ghhamcp:

ghhamcp:

kingpanther13 (changes requested):
This is a genuinely strong locale addition, and the PR description's claims hold up under checking: all four surfaces are structurally complete (419/419 settings `messages`, 29 `tool_groups`, 174 `tools`, 93/93 component keys, 16/16 and 26/26 add-on schema keys), placeholder parity is exact everywhere, and key order matches the English by construction. The 49-pair cross-surface claim verifies — I count 47 shared English strings by exact match, 0 with divergent Italian. `it` also clears every byte-identical ceiling by the widest margin of any shipped locale: 1/419 (0.24%) on settings `messages`, 0% on tools, component and both add-on flavors.

I checked the semantic layer the parity tests structurally cannot reach — the #1993 failure class — and it is clean. No meaning inversions, no dropped negations, no weakened warnings. `policies.card.conditions_intro` renders `ANY condition` as `UNA QUALSIASI condizione`, which is the exact string #1993 flipped. Every destructive tool keeps its gating condition (`(ULTIMA RISORSA)`, the *orphaned* gate on `ha_remove_device`, the allowed-directories restriction on `ha_delete_file`/`ha_write_file`), the beta master gate keeps its full force, and all 42 identifier occurrences survive byte-identical.

Requesting changes for the following. Fixes are suggestions, not prescriptions — correct me where the Italian is better your way.

**`src/ha_mcp/settings_ui/locales/it.json:435-436` — two groups render as the same heading.**
```json
"Dashboard":  "Dashboard",
"Dashboards": "Dashboard",
```
These are two separate renderable groups. `settings.js:800` iterates the English tags in sorted order and gives each its own collapsible section headed by `localizedToolGroup(tag)`; nothing merges them. `Dashboard` sorts immediately before `Dashboards`, so an Italian user sees two adjacent sections both titled "Dashboard" — the first holding only `ha_get_dashboard_screenshot` (`1/1`), the next the six dashboard CRUD tools (`6/6`). This is the only duplicate heading in any of the six locales; the others all distinguish them (`de` `Dashboard`/`Dashboards`, `es` `Panel`/`Paneles`, `fr` `Tableau de bord`/`Tableaux de bord`, `ru` `Панель`/`Панели`, `zh-Hans` `仪表盘`/`仪表盘总览`). The cause is real grammar — `dashboard` is invariable in Italian, so the distinction has to be re-created rather than translated. Suggested, keeping the loanword the rest of the PR uses:
```json
"Dashboard":  "Dashboard (screenshot)",
"Dashboards": "Dashboard",
```

**`src/ha_mcp/settings_ui/locales/it.json:89` — `visibility.enforce.only_enforced` negation scope.**
`Rendi le entità nascoste non leggibili da tutti gli strumenti` parses as partial negation — "not readable by *all* tools", i.e. some still can — which understates enforce mode on the checkbox label itself. Both sibling Romance locales avoided the construction (`es` `no se puedan leer desde ninguna herramienta`, `fr` `illisibles dans tous les outils`). Suggested: `Rendi le entità nascoste illeggibili in tutti gli strumenti (non solo nella ricerca)`.

**`src/ha_mcp/settings_ui/locales/it.json:70` — `accessibility.theme.preset` reads as "default theme".**
`Tema predefinito` uses `predefinito` = *default*, which this file uses that way ~15 times, including one line above at `:69`. The label above the Paper/Gray/High-Contrast picker should not say "default". `fr` has `Préréglage du thème`. Suggested: `Tema preimpostato`.

**`src/ha_mcp/settings_ui/locales/it.json:103-104` — `visibility.allowlist.title`/`.help` adjective with no head noun.**
`elenco di consentite` leaves a feminine plural adjective agreeing with nothing, and contradicts the section's own masculine `ID entità consentiti` at `:107-108`. Suggested: `Limita a un elenco di elementi consentiti`, and `...appena un campo dell'elenco di elementi consentiti viene compilato...`.

**`src/ha_mcp/settings_ui/locales/it.json:232` and `:257` — `Conservati per entità` is a dangling participle.**
As a numeric field label and a state readout (`Conservati per entità: 5`) it reads unfinished; `es` and `fr` both use the infinitive. Suggested: `Snapshot conservati per entità`, which works in both slots.

**`src/ha_mcp/settings_ui/locales/it.json:270` — `backup.bulk.deleted` hard-codes the plural.**
`Eliminati {count} backup` renders `Eliminati 1 backup` at `count = 1`. The English carries `(s)` and `fr` carries it on both words. Suggested: `Backup eliminati: {count}`, which sidesteps agreement.

**`src/ha_mcp/settings_ui/locales/it.json:41`, `:114`, `:290`, `:422` — noun-noun juxtaposition missing the preposition.**
`interfaccia impostazioni` (`:41`, `:290`, `:422`) and `ricerca strumenti` (`:114`) are anglicisms; the file uses the correct forms elsewhere — `l'interfaccia delle impostazioni` (`:45`), `la ricerca degli strumenti` (`:26`, `:112`, `:115`). Suggested: `interfaccia delle impostazioni` in the three sidecar strings, `della ricerca degli strumenti` at `:114`.

**`src/ha_mcp/settings_ui/locales/it.json:80` — `accessibility.colors.accent`.**
`Colore di risalto` means "highlight colour"; the established Italian UI term is `Accento` (`es` `Color de acento`, `fr` `Accent`). Suggested: `Accento`.

**`src/ha_mcp/settings_ui/locales/it.json:765` — `ha_manage_updates` skip/un-skip.**
`skip` → `ignora` but `un-skip` → `annulla l'esclusione`, which presupposes *escludere*; in one sentence the reader cannot see the second undoes the first. Suggested: `...installa in blocco, ignora o annulla gli aggiornamenti ignorati.`

**`custom_components/ha_mcp_tools/translations/it.json:139` — imperative verb in an entity-name slot.**
`entity.update.server_update.name` fills a Home Assistant entity name, rendered as `HA-MCP Server Aggiorna`. `Aggiorna` is the imperative; a noun belongs here, as in `fr` `Mise à jour`, `ru` `Обновление`, `de` `Update`, `zh-Hans` `更新`. Suggested: `Aggiornamento`.

**`homeassistant-addon/translations/it.yaml:35`, `:55`, `:144` and `homeassistant-addon-dev/translations/it.yaml:50`, `:142`, `:263`, `:265`, `:308` — one English term, four Italian renderings.**
`add-on`/`add-ons` is uniform in both `en.yaml` files but comes back as `componente aggiuntivo`, `app [componenti aggiuntivi]`, `l'app (componente aggiuntivo)` and `repository di componenti aggiuntivi`. `es` uses `complemento(s)` uniformly, `fr` uses `add-on(s)` uniformly. The add-on YAML's English never says "App", so the simplest lane is to drop the gloss there: `(backup, componenti aggiuntivi, preferenze energetiche, …)` at stable `:55` / dev `:50`, `sopravvivono ai riavvii del componente aggiuntivo` at dev `:142`, `il componente aggiuntivo «Puppet» di balloob` at dev `:263`.

**`custom_components/ha_mcp_tools/translations/it.json:33` vs `src/ha_mcp/settings_ui/locales/it.json` `advanced.extra_yaml_write_keys.label` — same English, two Italian renderings.**
`Extra YAML write keys` becomes `Chiavi YAML scrivibili aggiuntive` on one surface and `Chiavi YAML aggiuntive scrivibili` on the other. Same meaning, swapped adjective order — worth aligning given the PR sets out to keep shared strings identical across surfaces.

Three more that surfaced while reviewing this and are worth folding into the same commit rather than leaving to rot:

**Six locale counts are now off by one.** The prose was exact at five shipped locales and `it` makes six: `tests/src/unit/test_locale_parity.py:429` ("six catalogs" → seven; the loop directly beneath it is `["en", *_translated_component_locales()]` and that directory now holds seven files), `:486`, `:667`, `:681` ("five locales"/"five languages" → six), `tests/src/unit/test_config_flow.py:1609` ("the five catalogs" → six), and `tests/src/unit/test_settings_ui_i18n.py:305` ("six literals" → seven). `test_agents_md_lists_every_shipped_locale` pins the AGENTS.md line and you updated it correctly, but nothing pins these, which is why CI is green.

**`src/ha_mcp/settings_ui/locales/README.md` calls `tool_groups` and `tools` "optional".** That is no longer true — `test_locale_parity.py:542` requires both sections exactly, no key more and none fewer, and `test_settings_ui_i18n.py:288-295` requires both non-empty. That README uses `it.json` as its worked example and is what the next locale contributor reads.

**`src/ha_mcp/settings_ui/settings.js:3321-3322` interpolates raw English into a translated string.** `body.current_decision` is passed straight into `policies.pending.already_decided`, so the Italian renders `Questa approvazione è già stata approved`. `es` and `fr` render the same way, so mapping the decision value through the catalog fixes all three at once.

Patch76:
All 15 items are addressed, in `e268b79f` and `916c04c2`. I took the wording fixes as suggested except where noted below — two needed a wider edit than proposed, and one of the four reported add-on renderings turned out to be the correct one.

**The add-on gloss had to move on both surfaces, not only in the YAML.** The three strings you flagged are byte-identical in English to `features.read_only_mode.help`, `features.enable_code_mode.help` and `features.enable_dashboard_screenshot.help` in the settings UI. Dropping the gloss only in `homeassistant-addon*/translations/it.yaml` would have split those three cross-surface groups, so it is gone from both sides of each.

`tools.read_only.help` keeps its gloss. Its English reads `Apps (add-ons)`, where `features.read_only_mode.help` reads plain `add-ons` — the two English strings gloss differently on purpose, so the Italian does too. I also left `repository di componenti aggiuntivi` as it was: its English is `balloob's add-on repository`, with no gloss to drop.

Italian cross-surface divergence is now **0 of 53** groups. `Extra YAML write keys` was the single divergent one; the settings UI now carries the component's `Chiavi YAML scrivibili aggiuntive`.

Everything else went in as you wrote it, including `Dashboard (screenshot)` for the screenshot group — the loanword is invariable, so the distinction has to be built rather than translated. 25 Italian values changed across the four surfaces.

**On the raw-enum finding**, which was the substantive one. `current_decision` is an `approval_queue.Decision` literal rather than display text, so it resolves through the catalog now, falling back to the raw value for an outcome the catalog does not know. All seven catalogs gain `policies.pending.decision.approved` and `.denied`, and the source baseline is refreshed. This fixes `es`/`fr`/`de`/`ru`/`zh-Hans` along with `it`.

`test_every_decided_outcome_has_a_catalog_word` pins it. It derives the outcome set from the `Decision` literal itself via `get_args`, so extending the enum without adding a word turns it red rather than silently restoring the English. I checked that by injection: dropping a catalog key, breaking only the catalog lookup while leaving the call in place, and adding an outcome to `Decision` each failed it, with the control green before and after.

The three you folded in are in as well: six locale counts in test prose (`test_locale_parity`, `test_config_flow`, `test_settings_ui_i18n`), and the locales README, which called `tool_groups` and `tools` optional and also told the next contributor an incomplete catalog was fine — `test_locale_parity` has required both exactly for some time, so both claims are corrected.

`916c04c2` is a follow-up on my own change: the new test reached `settings.js` through `import ha_mcp.settings_ui` next to an existing `from ha_mcp.settings_ui import ...`, which CodeQL flagged as `py/import-and-import-from`. It reads the exported `_SETTINGS_JS` instead, which also means the guard asserts against the JS actually served rather than the template on disk.

All checks are green on `916c04c2`.

kingpanther13 (changes requested):
All 15 are addressed, and I verified each rather than taking the summary — including the three you deviated on, which were all right to deviate on. `tools.read_only.help` really does read `Apps (add-ons)` where `features.read_only_mode.help` reads plain `add-ons`, so keeping one gloss and dropping the other preserves a distinction the English makes; my suggestion would have flattened it. `balloob's add-on repository` has no gloss to drop. And moving the gloss on both surfaces rather than only in the YAML was the right call — I confirmed cross-surface divergence is 0 of 53 groups.

The decision-enum fix is better than what I asked for. Resolving through the catalog with the raw value as fallback, translating both words into all seven catalogs, and refreshing the baseline fixes `es`/`fr`/`de`/`ru`/`zh-Hans` alongside `it`. I checked the injected word against its host sentence in each language rather than in isolation — `ru`'s neuter short form agrees with «решение» inside the guillemets, `de`'s uninflected participle is correct for the Zustandspassiv, and `it`/`es`/`fr` are feminine singular to match their subject. `t()` gates on `hasOwnProperty`, so `__proto__` as a decision value falls back to the raw string rather than resolving an inherited property. Deriving the outcome set from `get_args(Decision)` is the right shape for the guard.

Six things in the new work.

**`basata` should be `basato` — `src/ha_mcp/settings_ui/locales/it.json` (`features.enable_dashboard_screenshot.help`) and `homeassistant-addon-dev/translations/it.yaml:263`.**
The gloss removal deleted the feminine head the participle agreed with. `...il componente aggiuntivo «Puppet» di balloob, basata su Chromium headless...` — `basata` agreed with the old `l'app`; both candidate heads now are masculine (`il componente aggiuntivo`, and `un motore` further back). I swept the other gloss-removal sites and this is the only stranded agreement, but it is in both files.

**`annulla gli aggiornamenti ignorati` names an action the tool does not have — `src/ha_mcp/settings_ui/locales/it.json` (`tools.ha_manage_updates.description`).**
This one is my fault: it is the wording I proposed last round, and it is wrong. `annullare` takes `gli aggiornamenti` as its direct object, so it reads "cancels the ignored updates" rather than un-skipping them. Suggested: `...installa in blocco, ignora o ripristina gli aggiornamenti ignorati.`

**Stray ASCII space before `{decision}` — `src/ha_mcp/settings_ui/locales/zh-Hans.json` (`policies.pending.already_decided`).**
The host reads `此审批已被 {decision}，`. That space was correct while the slot interpolated the Latin `approved`; now that it resolves to `批准` it sits between two CJK characters. The host was not edited here, but this PR changed what fills it, so the typography broke on this commit. The repo already treats this as a real distinction — `config_flow.py`'s `_NO_ASCII_SENTENCE_SPACE = frozenset({"zh", "ja"})`. Suggested: `此审批已被{decision}，可能由另一个标签页或会话完成。`

**`tests/src/unit/test_locale_parity.py:592-593` still says `9 of 419`.**
`en.json` `messages` went 419 → 421 with the two decision keys, so the correct figure is 9 of 421. The percentage is unchanged, which is why nothing reds. You swept the other six counts correctly — `test_locale_parity.py:429`, `:486`, `:667`, `:681`, `test_config_flow.py:1609`, `test_settings_ui_i18n.py:307` all check out — this is the one the delta itself created.

**`src/ha_mcp/settings_ui/locales/README.md:13-17` overstates the rule, and `en.json` is the counterexample.**
"Every group a tool renders under must be present" and "Every tool must be present" are stated for the directory, but `test_settings_catalog_keys_name_real_groups_and_tools` is parametrized over `_non_english_settings_locales()` (`test_locale_parity.py:542`), and `en.json` deliberately ships `tool_groups: {}` and `tools: {}` because English comes from the tool definitions at runtime. As written the rule also contradicts the recipe still standing at `README.md:3-6` — copy `en.json`, keep the keys — which produces an empty `tools` that would violate it immediately. A "non-English catalogs" qualifier resolves both. The old "optional" wording was wrong in the other direction and needed fixing, so this is about landing between the two, not reverting.

**`test_every_decided_outcome_has_a_catalog_word` is narrower than its docstring — `tests/src/unit/test_settings_ui_i18n.py:590` and `:600`.**
It fails on all three cases you injected, and the vacuity guard is sound. Two gaps:

`:590` reads `CATALOGS["en"]` only. The failure the docstring describes is "the English goes back, in every language at once", but a future `Decision` outcome whose keys land in `en.json` alone renders English everywhere and stays green — the 5% ceiling on 421 allows ~21 untranslated and the worst locale sits at 9, so two missing keys red nothing. The assertion text at `:596` tells the author to sweep the locales; nothing enforces it. Parametrizing over `CATALOGS` instead of indexing `["en"]` closes it.

`:600` pins the lookup, not the use. A dangling `const decision = t('policies.pending.decision.' + body.current_decision, ...)` with the outer call reverted to `{decision: body.current_decision}` still passes, and no linter runs on `settings.js` to catch the unused local. The same literal also reds confusingly on a harmless edit — renaming `body`, destructuring `current_decision`, or splitting at the `+` — with a message asserting the resolution is gone when it is not. Asserting on the interpolation site, or a regex tolerant of those shapes, covers what the test is actually for.

Minor in the same function: `missing` is bound twice at `:591-592`, the first binding being the full key list.

CI is green on `916c04c2` (31 pass, 2 skipped) and there are no unresolved threads — the only outstanding items are the six above.

Patch76:
All six are in, on `f0fc25d9`. One of them turned into a larger change than you asked for, and re-checking a second one turned up a claim of mine that was wrong — both below.

**`basata` → `basato`** — `it.json` (`features.enable_dashboard_screenshot.help`) and `homeassistant-addon-dev/translations/it.yaml:263`. The other three `basata` occurrences (`it.json:113`, `homeassistant-addon/translations/it.yaml:22`, `homeassistant-addon-dev/translations/it.yaml:22`) agree with `una scoperta` and are untouched.

**`annulla` → `ripristina`** — the English is `un-skip` (`tools_updates.py:956`), so `ripristina gli aggiornamenti ignorati` is what the tool does.

**zh-Hans space** — dropped, as suggested.

**The double-bound `missing`** — gone; the loop binds it once.

**`9 of 419` → `9 of 421`** — `en.json` `messages` is 421.

**`README.md`** — the `tool_groups`/`tools` rules now say "in a non-English catalog", and a paragraph states that `en.json` ships both empty on purpose so a catalog copied from it starts out missing them. Heads-up on an overlap: #2074 rewrites this same file from the master base, so the two will conflict textually. I will resolve it there, where the fuller rewrite supersedes this wording.

**The guard.** I first did what you suggested — replaced the literal with two regexes, one for the lookup and one for the interpolation site, the second reading the value side of the object so a dangling binding no longer passes. Then I tried to break it rather than re-read it, and it broke in six ways:

- A catalog whose value for `policies.pending.decision.approved` is still the string `approved` — the exact `Questa approvazione è già stata approved` this key exists to prevent. Green.
- `'policies.pending.decision.' + body.current_decision.toUpperCase()` — the key the concatenation builds is in no catalog, `t()` falls back to the raw value, every language shows the enum. Green, because the regex pins the prefix, not that the key resolves.
- The live block deleted and left behind as a single-line `// used to read: ...` comment. Green — `_SETTINGS_JS` is the whole file including comments, so dead prose satisfied both anchors.
- `{decision, code: body.current_decision}` in the call plus `{decision}` → `{code}` in all seven catalogs: the resolved word is interpolated and consumed by nothing. Green. (Changing one catalog is caught, by the placeholder-parity check in `load_catalogs`.)
- A second call site rendering the same sentence with the raw value. Green — both anchors are `re.search`, first match, and nothing requires them to be the same call site.
- A `current_decision` value the backend can emit that is not in `get_args(Decision)`.

Two harmless refactors went red with a message asserting the resolution was gone while it sat two lines above: `{decision: decision.toLowerCase()}` (plausible — German capitalises nouns mid-sentence, Italian does not), and adding a second interpolated value whose expression contains an object literal, e.g. `toLocaleString(locale, {timeStyle: 'short'})`, where `[^}]*` truncates at the inner brace and the message then names `'short'` as an interpolated variable.

So the instrument was the problem, not its tuning: matching source text cannot establish that the thing matched is the thing that renders. The repo already has the right instrument — `tests/js/harness.mjs` and the JSDOM behaviour tests in `test_settings_ui_js_behavior.py` — so the check moved there:

- `TestAlreadyDecidedCopy` drives `policyDecide` against a stubbed 409 for every (non-English catalog, decided `Decision` outcome) pair — six catalogs by two outcomes, 12 cases — and asserts the alert contains the catalog word and never the enum. Cases come from `get_args(Decision)` minus `pending`, so a new outcome arrives as new cases rather than as untested copy.
- `test_every_decided_outcome_has_a_catalog_word` keeps the type-derived half — every outcome has a key in every catalog — and now also rejects a non-English value spelled the way the backend spells it. It needs no node, which the JSDOM tests do, so the key coverage survives an environment without them. CI installs them (`pr.yml:397`), so both halves run there.

Injection results: the English-value catalog reds both tests; the suffix drift, the commented-out code, the unconsumed interpolation and a removed lookup red the behaviour test; dropping a key or extending `Decision` reds the catalog test; both false-positive refactors are green. **Two of the six stay uncovered and I would rather name them than paper over them.** A second call site rendering the same sentence: the behaviour test drives the handler that exists, and I do not think a source-text scan is the answer — if you want it closed, the honest fix is to give the sentence one owner, a small helper both paths call, and pin that. And a `current_decision` outside `Decision`: both tests derive their cases from `get_args(Decision)`, so neither would see it. That one needs a type violation to reach — `handlers.py:152` and `:176` send `entry.decision`, which is `Decision` — so I left it rather than build a guard against a value the type says cannot arrive.

**A correction to my own commit message.** `d52f0262` says the nine other placeholder sites in `zh-Hans.json` interpolate Latin or numeric values. That was too narrow a sweep. There are 24 messages where an ASCII space sits between a placeholder and a CJK character, and four of them — `tools.aria.llm_api`, `tools.aria.enabled`, `tools.aria.pinned`, `tools.aria.security_gated` — interpolate `{title}`, which is the *localized* tool title (`settings.js:941`, `toolCopy.title`), so in Chinese they render `批量控制 已启用`: CJK, ASCII space, CJK, the same shape you flagged.

I have left them alone. They predate this PR — nothing here changed what fills those slots — and they are aria-labels rather than visible copy, where a separator may well be doing useful work for a screen reader. It is your call whether that is a real defect and whether it belongs in this PR or its own; I did not want to fold a Chinese accessibility-string change into an Italian locale PR without asking.

CI is green on `f0fc25d9` — 31 passing, 2 skipped — and the PR is still `MERGEABLE` against the current master. Locally the three affected suites pass, 228 tests including the 12 new cases.

kingpanther13 (changes requested):
All six from the last round are in and verified. The guard work is the best thing in this PR: trying to break the regexes rather than re-reading them, finding them green under a `.toUpperCase()` suffix that builds a key no catalog holds and green again with the whole block reduced to a comment, and concluding that matching source text cannot establish that the matched thing renders — that is the right conclusion, and moving to the JSDOM harness is the right instrument.

Two of the six you named as still uncovered are closed, and you do not need to build anything for them. A second call site rendering the same sentence does not exist today, so there is nothing to pin; if one is ever added, that PR owns the helper. A `current_decision` outside `Decision` requires a type violation — `handlers.py:152` and `:176` send `entry.decision`, typed `Decision` — so a guard there would be testing that Python's type system was not bypassed. Both are correctly left alone.

I ran the full review toolkit over this head and verified every finding below myself against the source rather than relaying it.

## The CJK spacing class is six strings, not four

You were right to sweep, and right that the four `tools.aria.*` are defects. The sweep stopped one placeholder short. `{operation}` is also catalog-derived: `policyPut` (`settings.js:310`) takes `opLabel`, and all four call sites pass a `t()` result — `settings.js:358` `policies.operations.sync_gated`, `:3184` and `:3195` `policies.operations.save_rule`, `:3223` `policies.operations.save_global`. In `zh-Hans` those are `同步门控开关`, `保存规则`, `保存全局设置`. They land in:

- `zh-Hans.json` `common.operation_failed` — `{operation} 失败：{detail}`
- `zh-Hans.json` `policies.errors.conflict` — `{operation} 失败：策略已在另一个标签页...`

So a Chinese user who hits a policy write conflict reads `保存规则 失败：…` — the same stray gap. Neither key has any other consumer, so `{operation}` is never anything but a resolved label.

Six values, one space deleted from each, one file. The discriminator you found is the right one and holds: 24 `zh-Hans` strings put a space beside a placeholder and exactly these six are wrong, because the other eighteen interpolate Latin or numeric values (`{count}`, `{endpoint}`, `{variable}`, `{operator}`) where the space is correct Chinese typography. The class is `zh-Hans`-only — `de`, `es`, `fr`, `it`, `ru` all want their spaces.

On asking rather than doing: a six-character correction to a defect the review had just established is a Boy Scout fix, and `AGENTS.md` § Boy Scout Rule is explicit that the default is fix-in-place and that Small items get fixed in the PR with no mention in the description. "It predates this PR" is not a disposition — the rule says addressing what predates the PR is the point of it. A per-language PR is the right instinct for a *language*; it is the wrong instinct for six characters, where opening one costs more review churn than the fix. The bar for surfacing instead of fixing is a large or design-bearing change, and this was neither. Perhaps you need to update the memories.md file on your Claude.

## `README.md` still overstates, in a way that now contradicts `AGENTS.md`

`src/ha_mcp/settings_ui/locales/README.md:24-25` — "English is the runtime fallback, so a catalog missing a key still renders. **CI does not accept one**." That is false for `messages`, and `AGENTS.md:721-723` says the opposite in as many words: "Their `messages` may omit keys — English is the per-key fallback — but `tool_groups` and `tools` may not." The code sides with `AGENTS.md`: `_untranslated_keys` counts a missing key as untranslated, but `test_settings_catalog_is_not_a_copy_of_english` only fails above 5%, so a locale may omit 21 of 421 `messages` keys and stay green. The text this rewrite replaced said the true thing. A translator reads this, concludes all 421 are mandatory, and either blocks or pastes English in.

`README.md:16-18` and `:26-27` — the `tools` bullet states only key exactness and the closing paragraph attaches a ceiling to `messages` alone, but two further rules gate `tools`: the same 5% share ceiling applies to the 174 title/description values (`test_locale_parity.py:717-722`), and any single tool whose title *and* description are both byte-identical to English fails by name regardless of share (`:700-713`). A contributor who adds a tool, pastes English into all six catalogs to satisfy the exactness rule the README does document, and goes red on a rule it does not, is exactly the pressure that test's own docstring says it exists to catch.

`README.md:3-6` — "copy `en.json`, rename it to the language code (for example `it.json`)". This PR makes `it.json` a shipped 810-line catalog, so the worked example now tells the next contributor to overwrite the Italian translation. Use a code that does not ship.

## The two new guards

`tests/src/unit/test_settings_ui_js_behavior.py:1757` — `assert word in alert` cannot fail on catalog content, because the alert *is* the host with `word` substituted. Its only teeth are against a revert to raw-enum interpolation, which the `outcome not in alert` line already covers. Concretely, delete `policies.pending.already_decided` from `it.json`: `build_payload` merges English in (`_i18n.py:354`, `{**english["messages"], **selected["messages"]}`), the alert reads *"This approval was already approvata, possibly by another tab or session."*, and both assertions pass. One line closes it and subsumes the containment check:

```python
expected = CATALOGS[locale]["messages"]["policies.pending.already_decided"].replace("{decision}", word)
assert alert == expected
```

`tests/src/unit/test_settings_ui_i18n.py:618-625` — the `untranslated` check compares `messages[key] == o` exactly, so `"Approved"`, `"APPROVED"`, `" approved "`, `""` and `" "` all pass while rendering the bug or a hole in the sentence. Sentence-casing is what a translator writing display copy actually does. `messages[key].strip().casefold() == o` catches all three case variants, and a non-empty assertion catches the other two. I checked all twelve shipped values against the casefolded comparison — no false positives.

`tests/src/unit/test_settings_ui_js_behavior.py:1701-1713` — `_already_decided_cases()` returns `[]` silently if `CATALOGS` ever holds only `en` or `Decision` loses its decided outcomes, and zero collected cases pass. Its own sibling guards its half (`assert outcomes`, `test_settings_ui_i18n.py:593`), and the repo carries four more discovery guards for exactly this (`test_locale_parity.py:317`, `:534`, `:780`; `test_settings_ui_i18n.py:264`). Add the matching one.

## `_assert_clean_init` cannot see the failure it exists to catch

`tests/js/harness.mjs:445` sets `window.__harnessError` to a bare stack and `:478` pushes it into `errors` with no prefix, while `_assert_clean_init` (`test_settings_ui_js_behavior.py:611`) filters on `("script init:", "transpile failure", "invoke:", "jsdom error")`. A rejected async invoke is therefore invisible to it — and the `invoke:` prefix at `:451` only applies to a synchronous throw, which an async IIFE essentially cannot produce. `TestAlreadyDecidedCopy` is protected by its own `len(result.alerts) == 1`, but the hole is live: `test_error_response_does_not_default_to_restart_required` (`:1076-1100`) drives `await window.saveFeatureFlag(...)` and asserts only a negative, so renaming that function rejects the invoke, records no broadcasts, and passes. One word at `harness.mjs:478`; 113 call sites depend on the helper.

## `PredicateOp` is the same shape, one function over

`settings.js:2565` — `t(\`policies.operators.${p.op}\`, {}, p.op)` resolves the `PredicateOp` literal (`policy/model.py:15`, nine values) through the catalog with the raw enum as fallback, inside a translated condition row. Not a live bug: all nine ops have `policies.operators.*` keys in all seven catalogs today. The gap is that nothing pins it, while this PR has just built the machinery that would. Add an op and ship without catalog entries and every non-English user reads `args.domain starts_with "light"` mid-sentence — `test_every_decided_outcome_has_a_catalog_word` stays green because it derives only from `Decision`, and the 5% ceiling absorbs one missing key exactly as that test's own docstring argues. Iterating `get_args(PredicateOp)` alongside the `Decision` loop is the same few lines.

## The 503 path splices English into a translated sentence

`settings.js:3334` — a 503 is neither 409 nor 404, so it falls to `const detail = body.error || resp.statusText` and into `policies.pending.action_failed`. The 503 body is a fixed English paragraph (`settings_ui/__init__.py:247-256`), so an Italian user with the feature off clicks Approve and reads *"Azione di approvazione non riuscita:"* followed by five lines of English. Same class as the `current_decision` fix, eight lines below it in the same function.

I am not asking you to swap in a catalog string blindly, because the sibling path already answers this question differently on purpose: `policyLoadPending` (`settings.js:3268-3272`) resolves `policies.pending.unavailable` and then deliberately overwrites it with `body.error`, and `test_pending_list_shows_server_message_when_feature_on_but_503` pins that with a docstring saying the server message should propagate verbatim so the user learns which of three causes applied. That is a real trade-off and it was decided.

What is not defensible is the current shape — a translated clause with an English paragraph spliced into the middle of it. Two things make it coherent, and both match what the sibling already does: show the server diagnostic as its own message rather than inside a translated sentence, and mirror the sibling's `enabledKnown && !enabled` check so a user whose feature is simply switched off gets `policies.pending.disabled` instead of being told to read the add-on log.

## The rest

`test_settings_ui_js_behavior.py:1712` hardcodes `{"en"}` where its sibling at `test_settings_ui_i18n.py:612` uses `DEFAULT_LOCALE` for the same exclusion. Neither spelling is wrong alone; carrying both in two tests that pin the same behaviour is.

`tests/src/unit/test_settings_ui_i18n.py:284`, `:288`, `:292`, `:296`, `:607`, `:622` print `f"{locale}.json"` where `locale` is a lowercased catalog key (`_i18n.py:83`), so for Simplified Chinese all six name `zh-hans.json` while the file is `zh-Hans.json`. `:607` and `:622` are new this round; the other four already read that way, which is what makes it a sweep rather than a typo. Either drop the `.json` and name the catalog code, or resolve the on-disk stem. `:40` is `_write_catalog` creating a file from its own argument and is correct as it stands.

Everything else I checked came back clean, and I checked the things that would have been easy to get wrong: every remaining count and percentage in the touched files is correct after the 419 → 421 move, both new baseline hashes match `sha256` of their English values, the component version is correctly not bumped, no catalog anywhere carries an empty or whitespace-only value, every `<code>` literal and env-var token in the English sources survives into all six translations, and the Italian catalogs are an exact key match on all four surfaces with one legitimate byte-identical value.

Patch76:
All of it is in, on `7714468f`. One of your findings turned up a defect in the fix I wrote for it, which I have called out below rather than quietly corrected.

**The sweep did stop one placeholder short.** `{operation}` is catalog-derived the same way `{title}` is: all four `policyPut` call sites pass a `t()` result as `opLabel` (`settings.js:358`, `:3184`, `:3195`, `:3223`), and neither `common.operation_failed` nor `policies.errors.conflict` has another consumer — both are read inside `policyPut` itself. Six values, one space each. Eighteen `zh-Hans` values still put a space beside a placeholder and all eighteen are correct: `{operator}` is the raw `PredicateOp` literal from `settings.js:2975`/`:3002`, not the translated `policies.operators.*` word, and the rest interpolate counts, endpoints, env-var names and backup filenames.

**On asking instead of fixing — you are right, and the rule already said so.** `AGENTS.md` § Boy Scout Rule makes fix-in-place the default and puts a Small item in the PR without mentioning it in the description. "It predates this PR" is not a disposition, and a six-character correction to a defect the review had just established is not the large-or-design-bearing case that justifies surfacing. I have taken the correction.

**`README.md`.** The "CI does not accept one" sentence is gone; it was false for `messages` and contradicted `AGENTS.md:721-723`, and `_untranslated_keys` counts an omission against the share ceiling rather than failing on it. The `tools` bullet now names both value rules — the by-name failure when a tool's `title` and `description` are both byte-identical, and the share ceiling across all title and description values — and the closing paragraph states that ceiling for both sections instead of for `messages` alone. The worked example now names a code no catalog ships.

**The two guards.** The alert is compared as a whole sentence, which subsumes the containment check exactly as you said: the host a containment check searches is that same catalog value with the word already substituted, so it cannot fail on catalog content. The `untranslated` comparison is stripped and casefolded, and a separate assertion rejects empty and whitespace-only values. `_already_decided_cases` now asserts both discovered sets are non-empty before returning, matching the four sibling guards.

**`_assert_clean_init`.** `harness.mjs:481` prefixes the rejected-invoke push like the synchronous throw at `:451`. I checked it discriminates rather than reading it: renaming the invoked function in `test_error_response_does_not_default_to_restart_required` reds it with the prefix and leaves it green without.

**`PredicateOp`.** `test_every_predicate_operator_has_a_catalog_word` iterates `get_args(PredicateOp)` through the same helper as the `Decision` loop, so the key, blank and untranslated rules all apply to it. Dropping `policies.operators.gt` from one catalog reds it.

**The 503 path.** Both changes are in: `policyDecide` mirrors the sibling's `enabledKnown && !enabled` test and shows `policies.pending.disabled`, and otherwise shows the server's diagnostic on its own rather than inside `policies.pending.action_failed`. `TestDecideUnavailableCopy` drives both branches in an Italian UI.

**Where that fix was wrong before I committed it.** My first version of the test pinned a hand-written copy of the 503 body — and it had dropped the middle sentence, the one naming which of the three causes applied, which is the entire reason the branch shows the server's text rather than the generic line. The guard would have passed while pinning a string the server does not send. The body is now `POLICY_UNAVAILABLE_MESSAGE` in `settings_ui/__init__.py`, and the test imports it, so the copy cannot drift again.

**The rest.** `_already_decided_cases` uses `DEFAULT_LOCALE` like its sibling. Failure messages resolve the on-disk stem through a `_catalog_file` helper, so Simplified Chinese is named `zh-Hans.json` rather than `zh-hans.json`; that covers the four pre-existing sites as well as the three new ones.

Locally 442 tests pass across the settings-UI, i18n, locale-parity and rendered-script suites.

kingpanther13 (changes requested):
All eleven are in and verified, and calling out the drifted 503 fixture in your own first draft rather than quietly fixing it is the right instinct — that is the second time this round you have found a defect in a fix by attacking it instead of re-reading it.

Four things, and the first is a regression the fix I asked for introduced.

**`settings.js:3349` — the translated fallback is unreachable, and the path got worse rather than better.**

`body` is synthesised 32 lines earlier at `:3317`:

```js
try { body = await resp.json(); } catch (_) { body = {error: 'HTTP ' + resp.status}; }
```

So on a 503 whose body does not parse, `body.error` is the truthy string `'HTTP 503'`, and `body.error || t('policies.pending.unavailable', ...)` alerts that bare token. Before this change the same response produced *"Azione di approvazione non riuscita: HTTP 503"* — translated, with context. After it, every language gets a contextless `HTTP 503`, and `policies.pending.unavailable` cannot render from this branch for any realistic response shape.

The comment above it says this is "the trade-off `policyLoadPending` already makes deliberately". It is not: `policyLoadPending` (`:3268-3272`) reads the body *inside* its branch under its own `try`, and keeps the translated default when the parse fails or `error` is absent. Mirroring it means tracking whether the body actually parsed, not testing `body.error` for truthiness — a `parsed` flag set beside the `catch` at `:3317` is two lines.

Reachability is narrow: `_post_approve`/`_post_deny` never 503, so the only producer is the stub `unavailable()` at `settings_ui/__init__.py:248`, which always sends JSON — you need an intermediary returning a non-JSON 503. But the branch exists for exactly the deployments where an ingress or reverse proxy sits in front.

**`tests/src/unit/test_settings_ui_i18n.py:591-593` and `:607-610` — the stated failure mode is not what happens.**

The docstring says "``settings.js`` resolves these keys with the raw value as the fallback, so a value the catalogs do not cover renders the enum itself", and the assertion tells the contributor "The settings UI shows the raw {what} instead". `build_payload` merges English under the selected locale (`_i18n.py:354`, `{**english["messages"], **selected["messages"]}`) and `t()` tests that merged map, so a key missing from `it.json` but present in `en.json` renders the *English catalog value*. The JS inline fallback is reached only when the key is absent from English too.

That was accidentally true while the helper served `Decision` alone, because English's value for `approved` *is* the enum spelling. Generalising it to `PredicateOp` made it false: omit `policies.operators.regex` from `it.json` and an Italian user reads `matches regex`, not `regex`. The message now contradicts itself inside one sentence — it says the raw operator shows, then says an English word does.

The test is right and worth having; only its account of why is wrong. English inside translated copy is the defect either way, which is what the wording should say.

**`tests/src/unit/test_settings_ui_js_behavior.py:1666-1667` — the drift you introduced the constant to stop is still there, one test above the new class.**

That fixture hand-writes the 503 body as *"Tool security policies live approvals are not active. Check the addon log for ImportError / RuntimeError details."* — dropping the middle sentence naming the three causes, and writing `addon log` where the server writes `App (add-on) log`. Its own assertion, `"addon log" in result.dom.lower()`, passes *because* of that drift: the real `POLICY_UNAVAILABLE_MESSAGE` contains `app (add-on) log`, so switching the fixture to the constant fails the assertion. That is the proof it drifted, and it is the identical failure you caught in your own draft 140 lines below. Use the constant and assert on something the real string contains.

**`src/ha_mcp/settings_ui/__init__.py:191` — the new comment names three routes; `unavailable` serves five.**

"Body of the 503 the pending/approve/deny routes answer with" — `:264-268` wires it to `policy_get_pending`, `policy_post_approve`, `policy_post_deny`, `policy_get_tool_schema` and `policy_get_value_source`.

Everything else in this round is clean, and I checked the parts most likely to bite. `policyState` is settled in production — init awaits `loadPolicyState()` at `settings.js:403` and `policyDecide` is only reachable from a button `policyLoadPending` painted, long after. `enabledKnown` false correctly falls to the diagnostic rather than claiming the user turned the feature off. `TestDecideUnavailableCopy` cannot pass vacuously — `result.alerts == [x]` is strict equality, so the old spliced wrapping could not satisfy it — and the two awaited calls share the harness's single async IIFE, so there is no race. `_catalog_file` mirrors `load_catalogs` exactly (`path.stem.lower().replace("_", "-")`) and its fallback is unreachable by construction. The `harness.mjs` prefix breaks nothing: all three `_assert_clean_init` copies filter the same tuple, and nothing matches on the unprefixed content. The casefold rule is not a trap for the next language, because these keys hold phrases rather than bare tokens. And the eighteen remaining `zh-Hans` placeholder spaces are all correct — `{operator}` is the raw literal from `settings.js:2975`/`:3002`, not the translated word.

Patch76:
All four are in, on `40fca7bc`, and the branch is rebased onto master.

**The unreachable translated fallback.** Fixed by tracking the parse instead of testing `body.error` for truthiness: `body` starts as `{}` and `serverError` is only assigned from a parsed string `error`, so an unparsable 503 leaves it empty and `policies.pending.unavailable` renders. `TestDecideUnavailableCopy` now drives that case — a 503 whose body is `<html>...</html>` — and asserts the Italian line. Putting the old stand-in back in the `catch` turns exactly that test red and nothing else, with the alert reading `HTTP 503`.

One consequence I would rather state than leave implicit: the generic branch read the same value, so dropping the stand-in changes it too. `detail` is now `serverError || resp.statusText || 'HTTP ' + resp.status`, which means an unparsable **non**-503 shows the reason phrase where it previously showed `HTTP <status>` directly. Both are English on a path with no translated alternative, and reason-phrase-first seemed the less surprising of the two, but it is a change you did not ask for — say so and I will pin it back byte for byte.

**The stated failure mode.** You are right, and it was wrong for the same reason in two places. `build_payload` merges English under the selected catalog (`_i18n.py:354`), and `t()` resolves against that merged map, so a key missing from one catalog renders **English's value**; the inline fallback in `settings.js` is reached only when English lacks the key too. The helper docstring and its assertion message now say that, and so does `test_every_predicate_operator_has_a_catalog_word`, which claimed the raw literal just as concretely: dropping `policies.operators.regex` from one catalog gives that language `args.domain matches regex "light"` mid-row, while an operator added to `PredicateOp` with no catalog entry anywhere is the bare literal in every language including English.

**The drifted fixture.** Both fixtures now read `POLICY_UNAVAILABLE_MESSAGE` through one module-level `_server_503_body()`, and the pending-list assertion is on the real paragraph appearing in the DOM rather than on the two words the drift happened to keep. Restoring the hand-written body turns that assertion red, which is the same proof in the other direction.

**The comment.** It names all five routes now — pending, approve, deny, tool-schema and value-source. Checking that count turned up a second thing wrong with the same sentence: my first rewrite called them the sidecar's routes, and they are not sidecar-only. `__init__.py:432` picks the stub handlers whenever `not is_sidecar and approval_queue is not None` fails, so a main server with the feature off, or one whose policy package failed to import, sends this same body on those five routes — which is what the constant's own text has been saying all along. The comment now says that, and says that config GET/PUT are served for real even then.

**Rebase.** Onto `1786ad96`. `src/ha_mcp/settings_ui/locales/README.md` is no longer in the PR: #2074 landed while this sat, and its version of that file already carries everything my hunk said — `en.json` shipping `tools` and `tool_groups` empty, the exact key sets, both ceilings, and the paste signature — so keeping my wording would only have been a second, thinner account of the same rules. The PR is 19 files now.

kingpanther13 (approved):
All four are in and verified. Two of the fixes are better than what I asked for.

Tracking the parse rather than testing `body.error` for truthiness is the right shape, and `typeof body.error === 'string'` closes a case I did not raise — a parsed body whose `error` is a number or an object no longer stands in as a server message. I traced every branch that reads either value: `body = (await resp.json()) || {}` handles a null body, the 409 branch sees a falsy `body.current_decision` on an unparsable response exactly as it did before, the 503 branch now genuinely reaches `policies.pending.unavailable`, and `detail` can no longer come out empty. Nothing changed that you did not intend.

**On the generic-branch consequence you flagged: keep it.** `serverError || resp.statusText || 'HTTP ' + resp.status` is strictly better than what it replaced, and not only for the reason you gave. The old `body.error || resp.statusText` could produce an *empty* detail — a parsed body carrying no `error` alongside an empty reason phrase — and rendered "Azione di approvazione non riuscita: " with nothing after the colon. The three-tier version cannot. Reason-phrase-first also reads better on HTTP/1.1 and degrades to `HTTP <status>` on HTTP/2, where `statusText` is always empty by spec. No need to pin it back.

The regression test pins what it claims. `harness.mjs:99` falls back to `entry.body` when a stub carries no `json`, and `json()` is a bare `JSON.parse` over that text, so an HTML body really does reject and the `catch` really is what leaves `serverError` empty. It fails for the right reason.

The reworded docstring and message are accurate now — English merged underneath is what renders for a key one catalog omits, and the raw literal only where English lacks it too — and `test_every_predicate_operator_has_a_catalog_word` states both cases concretely rather than asserting the wrong one. Moving `_server_503_body()` to module scope and asserting the real paragraph appears in the DOM is the stronger check; it also survives `escapeHtml` untouched, since that string carries none of `& < > " '`.

Dropping the `README.md` hunk was right. I read master's version rather than taking it on trust: #2074 says `messages` keys may be omitted subject to the share limit, names both the 5% `tools` ceiling and the by-name failure when a tool's title and description are both English, and replaces the copy-`en.json` advice with "start from a translated catalog" — which covers all three of the things I raised against your wording, and more. And thank you for chasing the route count into a second error in the same sentence; the stub handlers are not sidecar-only, and the comment says so now.

The Italian itself has been solid since the first round — structurally exact on all four surfaces, no meaning inversions, every destructive tool's gating condition preserved, and the widest byte-identical margin of any shipped locale. What made this PR long was everything the locale turned over on the way: a backend enum rendering raw inside translated copy in six languages, a CJK spacing class in a seventh, a guard that matched source text instead of behaviour, and a fixture that had drifted so that its own assertion depended on the drift. You found several of those yourself by attacking your own work rather than re-reading it, and said so each time.

CI is green on `40fca7bc`, the rebase carries no upstream churn into the PR's 19 files, and there are no unresolved threads.

Patch76:
## Implementation Summary

Posting this after the fact — the merge came before I got it up. It covers the whole run, not just the last round.

### Choices made

**Scope.** A language lands on all four translated surfaces or on none, so this shipped every English settings-UI message, 29 `tool_groups`, 87 tool title/description overrides, all 93 component keys, and both add-on flavors (16 and 26 schema keys). The `messages` set was built by mapping every English key programmatically, so key set, key order and `{placeholders}` match the English source by construction rather than by review.

**What stays English.** The rule I applied is narrower than "proper nouns": a string stays verbatim when it names something the UI shows *untranslated on another screen*. That covers the `"Local network"` dropdown option, the `HA-MCP File & YAML Tools` entry title, `Client ID` / `Client Secret`, and the `<...>` placeholders. The prose around each of them is translated. `config.step.*.title` and `menu_options` go the other way and are translated, because Home Assistant renders those from the catalog itself.

**Terminology.** «…» throughout all four Italian catalogs, including where the English original uses `'…'`. The ASCII quote marks that remain are the ones that are not quotation: the HTML attributes in `policies.rules.empty`, which is how `ru` reads too, and the code literals the English source quotes as code (`yaml_path='automation'` in the three YAML-package help texts, `'all'` in `advanced.enabled_tool_modules.help`). Italian elision apostrophes are of course untouched. "skill" stays untranslated — it is the product term naming `skill_content` and `ha_get_skill_guide`, and `de` leaves it as well. `common.operation_failed` became `{operation} non è riuscito: {detail}` because the three `policies.operations.*` values are infinitive phrases and infinitives agree as masculine. Deliberately *not* unified: "cartella" in the settings UI against "directory" in the component, which is how each surface's own register reads.

**Cross-surface wording.** Where the same English string appears on more than one surface, I made the Italian identical, which meant the add-on YAML adopts the settings-UI wording. Only `>-` blocks were re-folded — folding is lossless there, while `|-` and `|` carry their line structure as part of the value, so those were left alone.

**Review rounds turned this into more than a catalog.** Rounds 3 and 4 were about English text sitting inside translated copy, and the fixes went into the code rather than the catalog: the 409 decision text and the predicate operators now resolve through the catalog, the 503 path in `policyDecide` shows the translated line instead of a spliced English paragraph, and `POLICY_UNAVAILABLE_MESSAGE` became a module-level constant so the test fixtures pin the string the server actually sends instead of a hand-copied paraphrase.

**One consequence I chose rather than avoided.** Dropping the synthetic error stand-in also changes the generic branch, which read the same value: `detail` is now `serverError || resp.statusText || 'HTTP ' + resp.status` (`settings.js:3366`), so an unparsable **non**-503 shows the reason phrase where it previously showed `HTTP <status>`. Both are English on a path with no translated alternative, and reason-phrase-first is the less surprising of the two, but it is a change nobody asked for and I would rather it be visible here than discovered later.

**Rebase.** `src/ha_mcp/settings_ui/locales/README.md` left the PR during the rebase. #2074 landed while this sat and its version of that file already carries what my hunk said, so keeping my wording would only have been a second, thinner account of the same rules.

### Problems encountered

**Three of the four round-4 findings were defects in my own earlier fixes, and no test caught any of them.** A synthesized `{error: 'HTTP <status>'}` used as a body substitute is truthy, which makes every `body.error || t(...)` chain unreachable for the translation it was supposed to fall back to — the translated line I had just added could not render. A docstring claim about fallback behaviour was wrong because `build_payload` merges English *underneath* the selected locale (`_i18n.py:354`), so a key missing from one catalog renders English's *value*, not the raw enum; the inline fallback is reached only when English lacks the key too. And the comment fix itself was in the wrong scope twice — first "three routes", then "every sidecar route" — when the truth is five live-approval routes, and the stub also answers in a main server whose approval queue is absent, because `__init__.py:432` gates on `not is_sidecar and approval_queue is not None`.

The general lesson I took from that: prose describing behaviour is the one surface no test pins, and the code half of it was reachable only through a branch nothing exercised. What found all three was reading each changed comment and docstring back against the code it describes, as a separate deliberate pass rather than as part of writing the fix. Only the first of the three is testable at all, and it now has one: a 503 whose body is `<html>...</html>`, asserting the Italian line. I put the old stand-in back to confirm that test turns red, and ran the whole file to confirm nothing else did. The other two are prose, and the only guard prose gets is that pass.

**Two dead keys, and they are mine.** While counting for this summary I found `it.json` shipping 421 `messages` where `en.json` has 419: `advanced.entity_search_limit.label` and `.help` have no English counterpart, are referenced nowhere in `src/` or `tests/`, and render nothing. `es.json` carries the same two. The two catalogs got there differently. `es` was correct when it merged and was orphaned afterwards, by #2043 removing those keys from `en`, `de`, `fr`, `ru`, `zh-Hans` and the baseline without pulling `es` along. `it` was authored against a pre-#2043 English source and then rebased onto a master that had already lost the keys, so the rebase is where it broke, and nothing flagged it.

Nothing flags it because the direction is not covered: the coverage ceilings measure English keys a locale is missing, not locale keys English is missing, and the baseline only hashes English sources, so a catalog can drift against a *removal* silently whenever the removing PR does not know about it. Removing the four entries across both catalogs is a small, coherent change; I have not opened anything for it.

PR #2068, opened 2026-07-28
fix: offload secrets.yaml masking to the executor

### LINKED ISSUE #2067: read_file: secrets.yaml masking constructs its YAML instance on the event loop
> [!WARNING]
> **Disclaimer:** Everything in this issue — the finding, the code references, and the
> mechanism described — was gathered by Claude (AI) and therefore cannot be trusted as-is.
> Every claim must be independently verified before acting on it.

This is a residual path of #1370, which #1371 and #1396 fixed for the bulk-edit case. The
same ruamel plugin-discovery glob still reaches the event loop through one uncovered call
site.

Home Assistant reports blocking-call warnings attributed to `ha_mcp_tools` the first time
`read_file` is used on `secrets.yaml` after a start or reload:

```
WARNING (MainThread) [homeassistant.util.loop] Detected blocking call to glob with args
('.../site-packages/ruamel/yaml/*/__plug_in__.py',) inside the event loop by custom
integration 'ha_mcp_tools' at custom_components/ha_mcp_tools/yaml_rt.py, line 139:
ry = YAML(typ="rt") (offender: .../ruamel/yaml/main.py, line 774: res = [x.replace(gpbd,
"")[1:-3] for x in glob.glob(bd + '/*/__plug_in__.py')]), please create a bug report
```

Three warnings fire together (`glob`, `iglob`, `scandir`), all with the same stack:
`handle_read_file` → `_shape_read_file_response` → `_mask_secrets_content` → `make_yaml` →
`_YAMLStorage.__init__` → `_build_yaml`.

## Where

All anchors measured against `upstream/master` at `01964cc1`:

- `custom_components/ha_mcp_tools/__init__.py:1255` — `parsed = make_yaml().load(content)`,
  inside `_mask_secrets_content` (defined at `:1232`)
- reached from `_shape_read_file_response` (async, defined at `:2748`) at `:2771`, gated by
  `:2770` `if normalized == "secrets.yaml":`
- `_shape_read_file_response` is awaited at `:2892`, inside the async service handler
  `handle_read_file` (`:2826`), so it runs on the event loop

The file read itself is already offloaded — `:2860`
`await hass.async_add_executor_job(_read_file_sync, target_file)`. Masking is the only
inline step that performs blocking filesystem work; the other inline statement in that
function, `:2767` `os.path.normpath`, does no I/O and already carries an `ASYNC240`
suppression.

## Why the existing cache does not cover it

`make_yaml()` caches per thread via `_YAMLStorage(threading.local)` (`yaml_rt.py:145`,
instantiated at `:152`) — the construct introduced by #1396. Subclasses of `threading.local`
re-run `__init__` in every thread on first attribute access, so the first `make_yaml()` call
on a given thread constructs a fresh `YAML(typ="rt")` (`yaml_rt.py:137`/`:139`), and ruamel
runs its plugin discovery inside `YAML.__init__` per instance. When that first call happens
on the event loop, the discovery blocks it. Subsequent calls on the same thread hit the cache,
which is why this surfaces once per thread rather than on every read.

The import-time warm-up in `_register_ha_tags()` (`yaml_rt.py:66`–`:80`) does not cover it:
it builds a throwaway instance at `:74` to register HA tag constructors on the class-level
registries, and never stores it in `_STORAGE`.

## Scope

Of the seven `make_yaml()` call sites in the component, this is the only one reached on the
event loop. The others are all offloaded:

| Site | Offloaded via |
|---|---|
| `:1663`, `:1725`, `:1759`, `:1997` | `await hass.async_add_executor_job(lambda: make_yaml()…)` |
| `:1977` (`_dump`) | sole caller at `:1982` |
| `:2313` (`_extract_yaml_views`) | production caller at `:2814`; also reached from `_extract_yaml_subtree` (`:2348`), which has no non-test caller |

## Impact

Low — it happens once per thread, and nothing breaks. I have not measured the cost inside a
Home Assistant container, so I am not putting a number on it.

Worth fixing because the log line asks users to file a bug report, and because the same glob
already justified #1370.

## Suggested fix

Offload the masking step the way the surrounding code already does. `hass` is a parameter of
`_shape_read_file_response` (`:2749`) and is already used for an executor call at `:2813`:

```python
content = await hass.async_add_executor_job(_mask_secrets_content, content)
```

Patch76:
## What does this PR do?

Fixes #2067 — reading `secrets.yaml` ran the masking step on the event loop, and Home
Assistant reported three blocking calls (`glob`, `iglob`, `scandir`) against this integration.

`_mask_secrets_content` calls `make_yaml()`. The first such call on a given thread constructs
a ruamel `YAML` instance, and ruamel discovers its plug-ins per instance by globbing the
site-packages tree. `_shape_read_file_response` invoked it inline, so the globbing happened on
the loop.

The per-thread cache introduced in #1371 (refactored in #1396) hides this after the first
call, which is why it surfaces once per thread rather than on every read. The import-time warm-up in
`_register_ha_tags` does not cover it either: it builds a throwaway instance to register the HA
tags on the class-level registries and never stores it, so the loop thread still constructs its
own. This was the only one of the seven `make_yaml()` call sites still reached on the loop —
the other six already go through the executor.

The fix hands the masking to the executor, matching every other `make_yaml()` caller the
event loop reaches. The file read itself was already offloaded.

Two test changes come with it. The existing source guard
(`TestReadFileSecretsMaskingOrder::test_full_content_is_captured_after_masking`) anchored on the
literal `_mask_secrets_content(content)`, which the executor hand-off removes; it now anchors on
the bare-args form, matching the sibling guard one test below it that already documents this
shape. The new `TestReadFileSecretsMaskingOffload` pins the offload itself.

I verified both guards bite by reverting the hand-off: the source guard fails with
`ValueError: substring not found` and the new test fails its assertion. Restoring produced a
byte-identical file.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

Being precise about the two I left unchecked rather than ticking them loosely:

- `tests/src/unit/test_custom_component_filesystem.py` passes in full (230 passed), and `ruff
  check`, `ruff format --check` and `mypy` are clean on both changed files. I have not run the
  whole suite locally — CI covers that.
- The change is not exercised through an agent against a live instance. The reproduction is the
  Home Assistant warning itself, which is what #2067 documents.

## Checklist

### DIFF
--- a/custom_components/ha_mcp_tools/__init__.py
+++ b/custom_components/ha_mcp_tools/__init__.py
@@ -2766,9 +2766,11 @@ async def _shape_read_file_response(
     # Apply special handling for specific files
     normalized = os.path.normpath(rel_path)  # noqa: ASYNC240
 
-    # Mask secrets.yaml
+    # Mask secrets.yaml. Offloaded because the first make_yaml() call on a
+    # thread constructs a ruamel YAML instance, whose plugin discovery globs
+    # the site-packages tree — blocking work that must stay off the loop.
     if normalized == "secrets.yaml":
-        content = _mask_secrets_content(content)
+        content = await hass.async_add_executor_job(_mask_secrets_content, content)
 
     # Apply tail for log files
     if normalized == "home-assistant.log":
--- a/tests/src/unit/test_custom_component_filesystem.py
+++ b/tests/src/unit/test_custom_component_filesystem.py
@@ -984,8 +984,9 @@ class TestReadFileSecretsMaskingOrder:
         return inspect.getsource(_shape_read_file_response)
 
     def test_full_content_is_captured_after_masking(self):
+        """Handed to the executor as bare args, so the anchor carries no "("."""
         src = self._handler_source()
-        mask = src.index("_mask_secrets_content(content)")
+        mask = src.index("_mask_secrets_content, content")
         capture = src.index("full_content = content")
         assert mask < capture, (
             "full_content must be captured AFTER _mask_secrets_content, or "
@@ -999,6 +1000,47 @@ class TestReadFileSecretsMaskingOrder:
         """
         src = self._handler_source()
         assert "_extract_yaml_views, full_content, yaml_path" in src
+
+
+class TestReadFileSecretsMaskingOffload:
+    """Masking secrets.yaml must not run on the event loop.
+
+    ``_mask_secrets_content`` calls ``make_yaml()``, and the first such call on
+    a given thread constructs a ruamel ``YAML`` instance whose plugin discovery
+    globs the site-packages tree. Home Assistant flags that as a blocking call
+    when it happens on the loop, so the masking belongs in the executor like
+    every other ``make_yaml()`` caller in this module.
+    """
+
+    async def test_masking_is_offloaded_to_the_executor(self):
+        from custom_components.ha_mcp_tools import (
+            _mask_secrets_content,
+            _shape_read_file_response,
+        )
+
+        offloaded: list = []
+
+        async def fake_executor(func, *args):
+            offloaded.append(func)
+            return func(*args)
+
+        hass = MagicMock()
+        hass.async_add_executor_job = AsyncMock(side_effect=fake_executor)
+
+        response = await _shape_read_file_response(
+            hass,
+            "secrets.yaml",
+            {"content": "api_key: hunter2\n", "mtime": 0, "size": 17},
+            tail_lines=None,
+            yaml_path=None,
+            include_parsed=False,
+        )
+
+        assert _mask_secrets_content in offloaded, (
+            "secrets.yaml masking must go through hass.async_add_executor_job, "
+            "or ruamel's plugin discovery blocks the event loop"
+        )
+        assert response["content"] == 'api_key: "[MASKED]"'
 
 
 class TestEditYamlConfigBackCompat:


ghhamcp:

ghhamcp:

Patch76:
**Choices Made:**

The masking call is handed to the executor rather than warming ruamel up at import time. `_register_ha_tags` already builds a throwaway `YAML` instance to populate the class-level tag registries, but it never stores it, so a loop thread still constructs its own on first use, and the per-thread cache from #1371 only hides that construction after it has happened once. Offloading the call puts this path on the same footing as every other `make_yaml()` caller the event loop reaches.

The offload is pinned by a behavioural test rather than by the source guard alone. A source guard cannot distinguish an executor hand-off from an inline call that was merely renamed, so `TestReadFileSecretsMaskingOffload` drives `_shape_read_file_response` with a mock `hass` and asserts that `_mask_secrets_content` was submitted to `async_add_executor_job`.

**Problems Encountered:**

The hand-off broke the existing source guard `TestReadFileSecretsMaskingOrder::test_full_content_is_captured_after_masking`, which anchored on the literal `_mask_secrets_content(content)`. It is re-anchored on the bare-args form instead of being dropped, matching the sibling guard directly below it that already pins `_extract_yaml_views, full_content, yaml_path`. Both guards were confirmed to bite by reverting the hand-off: the source guard fails with `ValueError: substring not found`, the new test with `assert <function _mask_secrets_content at 0x...> in []`. Restoring left the file byte-identical, and `tests/src/unit/test_custom_component_filesystem.py` passes in full at 230 tests.

A census of the seven `make_yaml()` call sites qualifies the description's claim that the other six already run in the executor. Five are wrapped in a callable that their own enclosing function submits to the executor. The sixth sits in `_extract_yaml_views`, which is offloaded one level up instead: `_shape_read_file_response` submits it to the executor, while `_extract_yaml_subtree` calls it inline. That second caller has no production caller of its own, only tests reference it, so it is not reachable from the event loop. Every loop-reachable path is covered by this change, and the inline caller is pre-existing and untouched here.

PR #2082, opened 2026-07-29
refactor(internal): drop two orphaned locale keys and pin the direction that let them in

Patch76:
## What does this PR do?

`es.json` and `it.json` each carried `advanced.entity_search_limit.label` and
`.help`, which `en.json` has not had since #2043. This deletes all four and adds
the check that would have caught them.

They got in by two different routes. #2043 removed the pair from `en`, `de`,
`fr`, `ru`, `zh-Hans` and the baseline; `es` was already merged and was not one
of the catalogs that PR knew about, so it kept them. `it` was written against
the older English source and rebased past the removal while its PR was open.
Neither failed anything.

Nothing failed because every key check here runs the other way. The ceilings
count English keys a locale is *missing*, and the baseline hashes English
sources only, so a key with no English counterpart is outside both. Nothing
renders such a key either: `build_payload` ships it and `t()` never asks for
it, since every lookup goes through a key `en.json` declares.

The new test asserts the orphan direction on the settings UI `messages`
section, and AGENTS.md gains the matching half-sentence. Scope is deliberately
that narrow:

- **The missing direction is not asserted.** `messages` may omit keys — English
  is the per-key fallback, and AGENTS.md states the allowance — so asserting it
  would forbid something the contract permits. A wholesale omission is already
  covered as a share by the 5% ceiling.
- **The other three surfaces are not touched.** `tool_groups` and `tools` are
  pinned in both directions against the parsed tool set by
  `test_settings_catalog_keys_name_real_groups_and_tools`, the component
  catalogs have their own key check, and
  `tests/addon/test_addon_structure.py::test_translations_cover_every_schema_key`
  already rejects a `configuration.<key>` the schema no longer declares. Adding
  them here would have been a second check of the same thing.

Also corrects the ceiling comment's denominator, which #2043 left saying 421
when it took `en.json` down to 419. `_assert_not_a_copy` divides by
`len(english)`, so the code had been computing over 419 the whole time.

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 📚 Documentation
- [x] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

`tests/src/unit/test_locale_parity.py` (77) and
`tests/src/unit/test_settings_ui_i18n.py` (31) both pass; I ran those two
modules rather than the full suite and left the box above unticked for that
reason. `ruff check` and `ruff format --check` are clean at the version
`uv.lock` pins (0.16.0). `mypy` reports the two import-resolution errors this
file already has on master, on lines my diff does not reach.

Three injections against the test as it ships here, each reverted:

| Injection | Expected | Result |
|---|---|---|
| Put both dead keys back | red | 2 failed (`es`, `it`) |
| Give `it` one key `en` lacks | only `it` red | 1 failed (`it`) |
| Have `fr` omit a key `en` has | **green** — the allowance | 30 passed |

The last one is the one that decides the scope: it fails if the test
over-reaches into the omission the contract permits.

## Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -722,9 +722,10 @@ That list of codes is itself pinned by
 code here, in the same PR, or the suite goes red.
 
 Settings UI catalogs are auto-discovered (no registration). Their `messages` may
-omit keys — English is the per-key fallback — but `tool_groups` and `tools` may
-not: each locale must carry exactly the renderable group headings and every tool
-name, no key more and none fewer. **Adding a tool therefore means translating it
+omit keys — English is the per-key fallback — but may not carry one `en.json`
+lacks: nothing renders it. `tool_groups` and `tools` may do neither: each locale
+must carry exactly the renderable group headings and every tool name, no key
+more and none fewer. **Adding a tool therefore means translating it
 in every locale, in the PR that adds it**: the check derives the tool set from
 the sources (`scripts/extract_tools.py`), not from the committed
 `site/src/data/tools.json` that `sync-tool-docs.yml` regenerates only after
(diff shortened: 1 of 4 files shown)


ghhamcp:

ghhamcp:

Issue #2093, opened 2026-07-30
Known issue: Claude Desktop and other mcp-proxy clients fail to connect after the MCP SDK 2.0.0 release — add --with "mcp<2.0.0" to your config


kingpanther13:
**If your working setup broke on or after 28 July 2026 and your client config runs `mcp-proxy`, this may be the cause. Please try this fix first before filing a new issue.**

## Fix

**Recommended: switch to `fastmcp-remote`.** As of [#2101](https://github.com/homeassistant-ai/ha-mcp/pull/2101), the Setup Wizard generates this instead of a pinned `mcp-proxy` config. `fastmcp-remote` is the FastMCP project's own stdio bridge and pins its SDK dependency to a bounded range (`mcp<2.0,>=1.24.0` as of fastmcp-remote 3.4.5), so this class of break cannot recur:

```json
{
  "mcpServers": {
    "home-assistant": {
      "command": "uvx",
      "args": [
        "fastmcp-remote",
        "http://192.168.1.100:9583/your-secret-path"
      ]
    }
  }
}
```

Use your own connect URL. Then fully restart the client (Claude Desktop: **File → Exit**, then reopen — closing the window is not enough).

Claude Desktop's config lives at:
- **Windows:** `%APPDATA%\Claude\claude_desktop_config.json` (on Microsoft Store installs: `C:\Users\<you>\AppData\Local\Packages\<claude-package-id>\LocalCache\Roaming\Claude\claude_desktop_config.json`)
- **macOS:** `~/Library/Application Support/Claude/claude_desktop_config.json`
- **Linux (beta, Ubuntu 22.04+/Debian 12+):** `~/.config/Claude/claude_desktop_config.json`

**If your client is JetBrains or Zed:** you don't need any bridge at all. Both support Streamable HTTP natively — JetBrains via a bare `{"mcpServers": {"name": {"url": "..."}}}` entry, Zed via its own `context_servers` format. Run the [Setup Wizard](https://homeassistant-ai.github.io/ha-mcp/setup) again for the current client-specific config.

**Prefer to stay on mcp-proxy?** Pin the SDK in its config's `args` instead. `--with` is a global `uv` option, so it **must come before** the package name:

```json
{
  "mcpServers": {
    "home-assistant": {
      "command": "uvx",
      "args": [
        "--with",
        "mcp<2.0.0",
        "mcp-proxy",
        "--transport",
        "streamablehttp",
[...]

Issue #2073, opened 2026-07-28
[BUG] mcp-proxy fails to start: ImportError 'request_ctx' after mcp 2.0.0 release (breaks ha-mcp connection)


RescueWolf-Dreamwalker:
### 💬 What Went Wrong?

mcp-proxy crashes on startup with an ImportError, causing ha-mcp to fail
connecting to Home Assistant. Claude Desktop shows "Server disconnected".
The setup worked this morning and broke without any local config changes.

Root cause: the `mcp` PyPI package jumped from 1.29.0 to 2.0.0 today
(2026-07-28), removing/renaming `request_ctx` in
`mcp.server.lowlevel.server`. `mcp-proxy` (0.12.0) imports this symbol
directly and has no upper version bound on its `mcp` dependency, so `uvx`
pulls in the incompatible 2.0.0 release.

### 📋 AI-Generated Content / Logs / Config

```markdown
Config (%APPDATA%\Claude\claude_desktop_config.json — on MSIX/Store installs
located at C:\Users\<username>\AppData\Local\Packages\<claude-package-id>\LocalCache\Roaming\Claude\claude_desktop_config.json):

{
  "mcpServers": {
    "home-assistant": {
      "command": "uvx",
      "args": [
        "mcp-proxy",
        "--transport",
        "streamablehttp",
        "http://<local-ha-ip>:9584/<private-token>"
      ]
    }
  }
}

Error shown in Claude Desktop:
Server disconnected

Full traceback when running manually (uvx mcp-proxy --version), reproduced
even after `uv cache clean` (i.e. not a stale-cache issue):

Traceback (most recent call last):
  File "<frozen runpy>", line 203, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "<uv-cache-path>\Scripts\mcp-proxy.exe\__main__.py", line 4, in <module>
    from mcp_proxy.__main__ import main
  File "<uv-cache-path>\Lib\site-packages\mcp_proxy\__main__.py", line 23, in <module>
    from .mcp_server import DEFAULT_EXPOSE_HEADERS, MCPServerSettings, run_mcp_server
  File "<uv-cache-path>\Lib\site-packages\mcp_proxy\mcp_server.py", line 24, in <module>
    from .proxy_server import create_proxy_server
  File "<uv-cache-path>\Lib\site-packages\mcp_proxy\proxy_server.py", line 11, in <module>
    from mcp.server.lowlevel.server import request_ctx
[...]

Issue #1470, opened 2026-05-28
[DOCS] Shift from mcp-proxy to uvx?


titom43:
### Where is the issue?

Setup Wizard (https://homeassistant-ai.github.io/ha-mcp/setup/)

### What's wrong or unclear?

I installed an early version on Codex and it worked like charm.

Then I updated and nothing was working anymore.

After spending hours trying to debug, I noticed that the MCP client advised on the GitHub is different from the one on the installed module documentation:

GitHub/online doc: 

<img width="608" height="469" alt="Image" src="https://github.com/user-attachments/assets/38507af8-be6a-4b91-a94b-2f75d8cd4a16" />

Installed module documentation:

<img width="775" height="611" alt="Image" src="https://github.com/user-attachments/assets/42831442-65a3-4a8d-9e44-d01f31faf2ac" />

After shifting from mcp-proxy to uvx, it works great again!

I hope it helps...

kingpanther13:
Good catch, thank you! We need to find a way to either keep the addon docs in sync with the other docs, or at least need to audit them to get them up to date. 

One of us will do a PR to fix this one as soon as we can! N

kingpanther13:
@Patch76 this should be an easy one if you'd like to tackle it

Patch76:
Thanks for putting both docs side by side and pinpointing the exact divergence (`uvx` vs `mcp-proxy`) — that comparison is what makes the root cause obvious.

The two screenshots aren't one setup documented two ways; they're two different deployment topologies, and the add-on doc's `mcp-proxy` block is actually correct for what it documents:

- **Online setup wizard / platform guides** (`site/src/pages/guide-macos.astro:75-76`, `guide-windows.astro:75-76`) configure Claude Desktop with `"command": "uvx"`, `"args": ["ha-mcp@latest"]` — the **local** model: the MCP server runs on your own machine as a stdio child of Claude Desktop and talks to HA over the network.
- **The bundled add-on doc** (`homeassistant-addon/DOCS.md:49-85`) uses `"command": "mcp-proxy"` pointing at `http://<addon-host>:9583/<secret>` — the **add-on** model: the server runs *inside* HA as an HTTP service. Claude Desktop's `mcpServers` config can only spawn stdio commands, so `mcp-proxy` bridges HTTP→stdio. The doc's own "How it works" line (`DOCS.md:85`) says exactly this.

[...]

PR #2101, opened 2026-07-31
fix: swap the stdio bridge from mcp-proxy to fastmcp-remote

sergeykad:
## What does this PR do?

Swaps the recommended stdio-to-HTTP bridge for stdio-only MCP clients (originally Claude Desktop, JetBrains, Zed) from `mcp-proxy` to `fastmcp-remote`, and, on further investigation, stops treating JetBrains and Zed as needing any bridge at all.

`mcp-proxy` declares an unbounded `mcp>=...` dependency, so it broke outright when the MCP SDK released 2.0.0 and removed `request_ctx` (#2073, #2093). `fastmcp-remote` is the FastMCP project's own stdio bridge (same upstream as this repo's `fastmcp` dependency) and pins `fastmcp-slim[client,server]==<its own version>`, which in turn bounds `mcp` to a fixed range (`mcp<2.0,>=1.24.0` as of 3.4.5) — verified against PyPI metadata. An SDK release alone cannot repeat this break.

Updates the setup wizard, `guide-addon.astro`, FAQ, and add-on `DOCS.md` to emit `fastmcp-remote` configs. The `#mcp-proxy-request-ctx` FAQ anchor is preserved (still linked from #2093) with the mcp-proxy pin kept as a documented fallback for anyone who prefers to stay on it.

Also fixes a bug found while reviewing this change: Zed's config takes the connect URL directly (`context_servers`) and runs no bridge process, but the wizard's `needsStdioBridge` flag didn't distinguish it from clients that do — so Zed users were shown an "Install uv" panel and bridge-install instructions for a bridge their config never invokes. Fixed by folding the Zed exclusion into `needsStdioBridge`'s definition rather than patching each call site, which let a now-dead code branch be deleted instead of adding more special-casing.

### Second commit: JetBrains and Zed don't need a bridge at all

Before marking this ready, checked whether the "needs a bridge" premise still holds for each client (docs can go stale — this repo already fixed the same class of bug once for Codex, #1470):

- **JetBrains**: confirmed via jetbrains.com/help/ai-assistant/mcp.html — native Streamable HTTP support, configured with a bare `{"mcpServers": {"name": {"url": "..."}}}` entry, no bridge. Verified twice against the primary source (including the exact JSON shape, no `type`/`transport` key) after an initial pass over-claimed a "2025.2+" version requirement the page doesn't actually state.
- **Zed**: confirmed via zed.dev/docs/ai/mcp — native `context_servers: {url: ...}` support, which the wizard already emitted for Zed (see above). Only the stale classification (`stdioOnlyClients` membership, `httpNote` text, `transports` metadata) needed fixing, not the config generation.
- **Claude Desktop**: bridge genuinely still required — verified its `claude_desktop_config.json` schema silently rejects a `url` field, and its native remote-server path (Settings → Connectors) requires public HTTPS reachability from Anthropic's cloud plus OAuth/static headers, incompatible with ha-mcp's secret-path-in-URL LAN deployment model.

Removing jetbrains/zed from `stdioOnlyClients` also resolves, for real, the platform-picker known-gap the first commit had documented but deferred: both no longer match `isStdioOnly()`, so `needsPlatform()` correctly stops forcing an OS choice that was never relevant to either client. New regression test (`test_http_method_native_http_client_skips_platform`) pins this.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing
- [x] All automated tests pass (`uv run pytest`) — full unit suite (9149 passed, 3 skipped) ran via the pre-commit hook on both commits; `tests/src/unit/test_astro_setup_js_behavior.py` (55 tests) verified separately, including new coverage for the JetBrains/Zed native-HTTP config shape and the platform-step skip
- [x] Code follows style guidelines (`uv run ruff check`) — clean; `eslint` on the changed `.astro` pages clean
- [ ] I have tested these changes with a LLM agent — not applicable; this is a docs/wizard-config change, not a tool change

Also verified empirically: `uvx fastmcp-remote <url>` (bare, no `--auth` flag) completes MCP `initialize` + `tools/list` against a real standard-mode (secret-path, no-OAuth) `ha-mcp-web` deployment in ~1s, confirming its default OAuth-preferring auth mode doesn't stall against a server with no OAuth metadata. Not verified against OAuth-enabled or remote-tunnel deployments — flagged as a known gap in the relevant code comment.

## Checklist


ghhamcp:

ghhamcp:

kingpanther13:
Good call, @sergeykad can you please edit issue #2093 though to point to new instructions? I think we should still keep pinned issue up because we're going to get more people running into this problem. I'm not currently able to edit with new instructions

PR #2104, opened 2026-08-01
test: pin the mixed-entities status gate and search fallbacks

### LINKED ISSUE #2103: test: the mixed-entities e2e block can pass without exercising the operation-status contract
### What I found

`test_bulk_operation_error_scenarios` in `tests/src/e2e/error_handling/test_network_errors.py` reaches its status-contract section behind a gate that nothing asserts, so the section can be skipped without failing.

The status section at `:421-468` sits behind `if operation_ids:` with no `else`. Nothing pins that the batch returns a non-empty `operation_ids` — the only assert mentioning it (`:454`) is inside the gate. The comment at `:384-391` argues the legacy dispatch tier always runs and therefore populates it, but that is an argument, not an assertion.

As far as I can follow it in the code, that reasoning holds today: `_validate_bulk_operations` only checks that `entity_id` and `action` are present, so the fabricated entities do enter `valid_operations`; their unknown services then raise in the component's all-guards-first pass, `_bulk_via_component` returns `None`, and the batch falls back to legacy, which populates `operation_ids`. I have not confirmed that chain against a live run.

What makes it worth pinning is that the component tier builds its response with an empty `operation_ids` by construction (`src/ha_mcp/tools/device_control.py:1048-1053`, whose own comment reads "operation_ids is empty (ops confirmed inline, no polling handle)"). So if this batch ever becomes component-servable — a routing change, resolvable services for the fabricated domains, a capability shift — the gate closes and this block goes quiet rather than red. The test depends on a routing outcome it never asserts.

A second, independent route reaches the same place. `_safe_tool_call` now marks its two failure legs — `timed_out` at `:44-54` and `tool_error` at `:55-57` — and the bulk call sites read the marker before deciding to assert (`:372`, `:435`). The `ha_search` at `:335` does not:

```python
search_data = parse_mcp_result(search_result)
valid_entities = []
if search_data.get("success") and search_data.get("entities"):
    valid_entities = [search_data["entities"][0]["entity_id"]]
```

`parse_mcp_result` maps either failure dict to `{"error": "No content in result"}`, so a timed-out or failed search is indistinguishable from an empty one. `valid_entities` becomes `[]`, nothing dispatches, and `:421` is skipped again.

What that degraded run loses: the three status asserts (`:441`, `:454`, `:464`) and the "valid entity succeeds" half of `:398`, which collapses to `successful_commands == 0`. `:401` (`failed_commands == 2`) still runs, and the `:472` section contains only logging, no asserts.

### Why it matters

The per-item contracts themselves are covered deterministically by unit tests — `test_per_item_failure_entries_keep_structured_context` and neighbours in `tests/src/unit/test_bulk_operation_status_summary.py` for operation status, `TestBulkExecutionErrorHandling` in `tests/src/unit/test_bulk_device_control.py` for bulk execution. So nothing guaranteed today is unguarded.

What is lost is the live evidence. This is the only e2e test that asserts on the status of operations it really dispatched: `test_bulk_light_control` in `tests/src/e2e/workflows/device_control/test_lights.py:261` does drive real operation IDs through `ha_get_operation_status`, but its own comment says the status is "purely informational (asserted nowhere below)", and the loop swallows exceptions. `test_operation_status.py` covers the invalid, empty, and coercion cases, and its dispatch test uses a fabricated ID (`test_dispatch_op_999:112`) rather than a real batch.

So the risk is not a wrong result today — it is that a regression in the live status path stops being detected, with CI still green.

### Suggested change

Two small additions, in keeping with what `:372`/`:435` already do:

1. Assert that a batch which dispatched at least one valid entity comes back with a non-empty `operation_ids`, so `:421` cannot be skipped without a failure. If the component tier is a legitimate route for this batch, the assertion needs to account for it explicitly rather than the block quietly doing nothing.
2. Check the `_safe_tool_call` fallback markers at `:335` — both of them, since `tool_error` collapses exactly like `timed_out` — so a failed search is tolerated explicitly instead of reading as "no entities found".

The same marker-blind pattern exists at the two other `ha_search` call sites whose result gates later work (`:152`, `:637`); neither gates an assertion, though a failure at `:637` no-ops the rest of `test_concurrent_operation_handling` through the early return at `:644-646`.

### Scope

Both gates predate #2076: `git log -S 'if operation_ids:'` traces the line to the file's initial commit, and #2076 narrowed this block rather than widening it. Filing separately so it is tracked rather than folded into merged work. Low priority — test hardening, not a runtime defect.

kingpanther13:
## What does this PR do?

Hardens the e2e error-handling suite against the silent-skip paths described in #2103, plus the adjacent ones found while closing them.

- The operation-status section of `test_bulk_operation_error_scenarios` sat behind `if operation_ids:` with nothing asserting the list is non-empty. The routing is now pinned: whenever a valid entity dispatched, the test asserts `len(operation_ids) == len(valid_entities)`, so a routing change to the component tier (which returns no polling handles by construction) fails loudly instead of quietly skipping the only e2e assertions on live operation status. The status helper additionally pins entry count and requires failed entries to carry `SERVICE_CALL_FAILED` rather than an unmapped code folded into the same status.
- The `ha_search` call sites that gate later work parsed their results without checking `_safe_tool_call`'s `timed_out`/`tool_error` markers, so a failed search read as "no entities found" and degraded the dependent blocks to no-ops. A shared helper now discriminates the markers before parsing (wrapper timeout tolerated with a warning, tool error fails hard) and hard-asserts the registry-seeded demo lights are found — with a bounded poll absorbing the documented post-boot state-machine lag on the container/embedded lanes, where the `light.bed_light` boot gate does not run.
- `parse_mcp_result` collapsed the wrappers' fallback dicts into a generic placeholder, erasing both the markers and the real error text; it now passes already-parsed dicts through unchanged (the three local shadow parsers mirror the guard). The wrappers also normalize error-flagged *returned* results — checking `is_error`, fastmcp's actual spelling, alongside the raw-MCP `isError`; the previous spelling-only check was dead code — and preserve the structured error payload.
- The concurrent sub-helpers asserted nothing (0/3 passed identically to 3/3) and `_run_concurrent_bulk_operations` sent `turn_on`/`turn_off`, which the `valid_actions` vocabulary rejects per item before dispatch — it exercised nothing while logging success. They now use the real vocabulary and fail on any non-timeout error, with per-item counts pinned. The empty-batch and invalid-action sections assert their outcomes instead of logging them.
- New unit file `tests/src/unit/test_e2e_assertions.py` pins the dict passthrough, both error-flag spellings (observable only on non-JSON bodies), and the hard-failure classifier — the contracts whose breakage degrades e2e runs into passes rather than failures.

Closes #2103

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [x] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

## Checklist

### DIFF
--- a/tests/src/e2e/error_handling/test_network_errors.py
+++ b/tests/src/e2e/error_handling/test_network_errors.py
@@ -148,31 +148,43 @@ class TestErrorHandling:
         # 3. MISSING REQUIRED PARAMETERS: Try to call service without required params
         logger.info("📋 Testing missing required parameters...")
 
-        # Try to call light.turn_on without entity_id (if lights exist)
+        # Try to call light.turn_on without entity_id (if lights exist).
+        # A failed search is not "no lights": parse_mcp_result renders both
+        # _safe_tool_call fallbacks as empty content, so the markers must be
+        # read before the parse. Only the wrapper timeout is tolerated.
         search_result = await self._safe_tool_call(
             mcp_client,
             "ha_search",
             {"query": "light", "domain_filter": "light", "limit": 1},
         )
 
-        search_data = parse_mcp_result(search_result)
-        if search_data.get("success") and search_data.get("entities"):
-            # Call service without entity_id to test parameter validation
-            missing_params_result = await self._safe_tool_call(
-                mcp_client,
-                "ha_call_service",
-                {
-                    "domain": "light",
-                    "service": "turn_on",
-                    # Missing entity_id
-                },
+        if isinstance(search_result, dict) and search_result.get("timed_out"):
+            logger.warning(
+                "  entity search timed out; skipping the missing-params probe"
             )
+        else:
+            assert not (
+                isinstance(search_result, dict)
+                and search_result.get("success") is False
+            ), f"ha_search with valid params must not error: {search_result}"
+            search_data = parse_mcp_result(search_result)
+            if search_data.get("success") and search_data.get("entities"):
+                # Call service without entity_id to test parameter validation
+                missing_params_result = await self._safe_tool_call(
+                    mcp_client,
+                    "ha_call_service",
+                    {
+                        "domain": "light",
+                        "service": "turn_on",
+                        # Missing entity_id
+                    },
+                )
 
-            missing_params_data = parse_mcp_result(missing_params_result)
-            # This might succeed (affects all lights) or fail depending on HA config
-            logger.info(
-                f"  Service call without entity_id: {'succeeded' if missing_params_data.get('success') else 'failed'}"
-            )
+                missing_params_data = parse_mcp_result(missing_params_result)
+                # This might succeed (affects all lights) or fail depending on HA config
+                logger.info(
+                    f"  Service call without entity_id: {'succeeded' if missing_params_data.get('success') else 'failed'}"
+                )
 
         logger.info("✅ Service call error handling test completed")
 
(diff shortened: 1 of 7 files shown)


ghhamcp:

ghhamcp:

Issue #198, opened 2025-11-29
Assess if prompts are usefull


julienld:
If not, remove them now (could be added back later)

VSCode uses it as slash commands?

PR #2112, opened 2026-08-02
feat: add MCP Prompts layer (safety, troubleshooting, automation, status, security)

crs2007:
## Summary

Adds an MCP **Prompts** layer — 12 parameterized, reusable prompts across five modules that guide an LLM through structured Home Assistant workflows. The server currently exposes tools and skills but no prompts; this fills that surface.

| Module | Prompts |
|---|---|
| `safety.py` | `confirm_action`, `quiet_hours_check`, `read_before_write`, `bulk_action_review` |
| `troubleshooting.py` | `diagnose_automation`, `diagnose_entity` |
| `automation.py` | `create_automation`, `review_automation` |
| `status.py` | `system_status`, `area_status` |
| `security.py` | `lock_workflow`, `alarm_workflow` |

## Wiring

`_initialize_server` gains one call to a new `_initialize_prompts` hook. Registration is wrapped in `try/except` that logs and continues — prompts are additive guidance and must never take the tool surface down with them.

## Note on the return type

The prompts return their text as a plain `str` rather than a `list[PromptMessage]`. FastMCP 3.4.4 rejects the latter at *render* time (`messages[0] must be Message or str, got PromptMessage`), while registration and `list_prompts` both keep passing — so the failure is invisible until a prompt is actually invoked. `test_every_prompt_renders_a_user_message` renders all 12 for exactly that reason.

## Test plan

- [x] `tests/src/unit/test_prompts_registration.py` — 3 tests: full registration set, every prompt renders a non-empty user message, arguments are interpolated
- [x] Full unit suite: 2195 passed (3 pre-existing Windows-only path-semantics failures in `test_custom_component_filesystem.py`, identical on clean master)
- [x] `ruff check src/ tests/` clean
- [x] `mypy src/` clean (4 pre-existing Windows-only `fcntl` errors, identical on clean master)

No new tools, so no locale-parity impact.

### DIFF
--- a/src/ha_mcp/prompts/__init__.py
+++ b/src/ha_mcp/prompts/__init__.py
@@ -1,0 +1,33 @@
+"""MCP Prompts layer for Home Assistant.
+
+Provides reusable, parameterized prompts that guide LLMs through
+structured workflows for safety, troubleshooting, automation,
+status checks, and security-sensitive operations.
+"""
+
+from __future__ import annotations
+
+import logging
+
+from fastmcp import FastMCP
+
+from .automation import register_automation_prompts
+from .safety import register_safety_prompts
+from .security import register_security_prompts
+from .status import register_status_prompts
+from .troubleshooting import register_troubleshooting_prompts
+
+logger = logging.getLogger(__name__)
+
+
+def register_all_prompts(mcp: FastMCP) -> None:
+    """Register all prompt modules with the MCP server."""
+    register_safety_prompts(mcp)
+    register_troubleshooting_prompts(mcp)
+    register_automation_prompts(mcp)
+    register_status_prompts(mcp)
+    register_security_prompts(mcp)
+    logger.info(
+        "Registered MCP prompt modules "
+        "(safety, troubleshooting, automation, status, security)"
+    )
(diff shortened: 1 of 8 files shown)


ghhamcp:

kingpanther13:
Thanks for putting this together, and sorry for the effort already spent — we're closing this now so you don't put more into it.

ha-mcp had an MCP prompts layer and removed it deliberately. #198 assessed whether prompts were earning their place, the answer was no (clients weren't invoking them), and #248 deleted the surface in December with a note that it could return if that changed. Nothing has changed on the client side since, so re-adding it would put us back where we started.

There's also a design reason we aren't planning to revisit it soon. MCP prompts are user-invoked: the human picks one from a client menu before doing anything. That makes them a poor fit for the safety and read-before-write guidance here, which needs to reach the model automatically at the moment of the action. ha-mcp does that through tool safety annotations, the best-practice checker that fires on writes, and the skill guides the model can pull on its own.

Separately, worth knowing for anything similar in future: the prompt bodies are written against one specific installation — personal notification targets, a named announcement script, particular Zigbee and Tuya conventions — and they instruct several tool names that no longer exist in the server, having been renamed or merged away. The tests pass because they only assert that each prompt renders, not that the tools it names resolve.

The upside is that none of this needs to live in the server. Guidance of this kind works well as a client-side asset — a project instructions file, a skill, or a client slash command — where you can tune it to your own entities and conventions without it shipping into everyone else's install. That is where I would keep it.

Thanks again for contributing, and please keep them coming.

PR #2170, opened 2026-08-08
fix: keep hardcoded config-flow labels untranslated

Patch76:
## What does this PR do?

`_validate` in `scripts/translate_locales.py` rejects a translation that localises a selector label the config flow hardcodes in Python, and the missing `nl` value for `common.connect_local_lan` is authored by hand.

The locale-sync run on 2026-08-08 failed its content-completeness verification and pushed nothing — for every language, not just the one at fault. Filling `nl` for the first time, the engine localised the on-screen option named in that key: `"Local network"` became `"Lokaal netwerk"`. That option label is hardcoded in `config_flow.py`, so it reads English on the form whatever language the reader has selected. The prompt already asked for the literal to survive; nothing enforced it.

Quoting alone is not the signal. Over the 3232 English/translation pairs shipped at the time, a naive "quoted text must stay English" check rejects 86 — 72 in the settings catalogs, where one setting's help text quotes another's title, and 6 in the component catalogs, where every locale that has the key translates `"Add entry"`, correctly, because Home Assistant translates that button itself. The discriminator is where the string comes from: labels the config flow hardcodes cannot be translated, everything else can. The label set is read from `config_flow.py` at runtime rather than duplicated, so a rename cannot strand a stale copy, and the quoted text is matched through straight and typographic marks alike.

The module docstring and the workflow header both enumerate what the script validates, so both name the new check; the header additionally picked up formatting-tag parity, which it had been missing against that same list. That comment is the entire `.github/` diff.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

No agent testing: this changes a build-time validation script and one catalog string, neither of which an agent exercises at runtime.

Beyond the suite:

- Four unit tests cover the rejection, the same rejection through typographic quotes, a check that other quoted material stays translatable, and one asserting the extracted label set is non-empty — without that last one, a rename in `config_flow.py` would degrade the check into a no-op that passes everything while still reporting a clean run.
- Each test was verified by mutating the condition it guards — not by deleting the call, which leaves an absence-assert green by construction — with each mutation naming exactly the assertion it should, and the suite green again after every restore.
- The content-completeness suites, which the sync gates on and PR CI skips, produce the same failure set on this branch as on the base commit without these changes.
- `generate_locales.py` reports no updates, so the projections stay consistent with the edited catalog.
- The `nl` key's English is unchanged since the last baseline repin, so the sync will not retranslate it and no repin is included. Repinning would also freeze the strings legitimately waiting for retranslation.

## Checklist

### DIFF
--- a/.github/workflows/locale-sync.yml
+++ b/.github/workflows/locale-sync.yml
@@ -18,8 +18,9 @@ name: Locale Sync
 # which is what tells the bot the changed English is already covered).
 #
 # Verification is three layers. (1) Every returned string is validated
-# before it is written (placeholder, markup and panel-link parity, in the
-# script). (2) After a clean run, the content-completeness checks that PR CI
+# before it is written (placeholder, markup and panel-link parity, plus the
+# config flow's hardcoded option labels, in the script). (2) After a clean
+# run, the content-completeness checks that PR CI
 # skips run here under LOCALE_COMPLETENESS_CHECKS=1; a failure means the
 # output may be corrupt (the engine pasted English back, parity broke) and
 # BLOCKS the push — machine translations are recomputable, so dropping them
--- a/.github/workflows/locale-sync.yml
+++ b/.github/workflows/locale-sync.yml
@@ -18,9 +18,9 @@ name: Locale Sync
 # which is what tells the bot the changed English is already covered).
 #
 # Verification is three layers. (1) Every returned string is validated
-# before it is written (placeholder, markup and panel-link parity, plus the
-# config flow's hardcoded option labels, in the script). (2) After a clean
-# run, the content-completeness checks that PR CI
+# before it is written (placeholder, markup, formatting-tag and panel-link
+# parity, plus the config flow's hardcoded option labels, in the script).
+# (2) After a clean run, the content-completeness checks that PR CI
 # skips run here under LOCALE_COMPLETENESS_CHECKS=1; a failure means the
 # output may be corrupt (the engine pasted English back, parity broke) and
 # BLOCKS the push — machine translations are recomputable, so dropping them
--- a/custom_components/ha_mcp_tools/translations/nl.json
+++ b/custom_components/ha_mcp_tools/translations/nl.json
@@ -1,5 +1,6 @@
 {
   "common": {
+    "connect_local_lan": "Lokaal/LAN (wanneer Netwerktoegang \"Local network\" is): {url}",
     "oauth_not_serving": "Legacy OAuth levert deze nog niet — start Home Assistant opnieuw op wanneer je daarom wordt gevraagd, om ze te activeren."
   }
 }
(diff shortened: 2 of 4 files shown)


ghhamcp:

Patch76:
## Implementation Summary

**Choices Made:**

- The rule keys on where a string comes from, not on whether it is quoted. Quoting looked like the obvious signal and is not one: over the 3232 English/translation pairs currently shipped, "quoted text must stay English" would reject 86 of them — 80 in the settings catalogs, where one setting's help text quotes another setting's title, and 6 in the component catalogs, where every locale that has the key translates "Add entry" because Home Assistant translates that button itself. What genuinely cannot be translated is narrower: the selector labels the config flow hardcodes in Python. Under that rule the same 3232 pairs produce zero rejections, and the string that stopped the sync is rejected.
- The label set is read from `config_flow.py` at runtime rather than listed in the script, so renaming a label cannot strand a stale copy in the validator. `test_connect_local_lan_quotes_the_bind_host_option` already guards the same rename against the catalogs.
- Quoted text is matched through straight and typographic marks. Two shipped English strings already quote with typographic marks, so a check that understood only `"…"` would have let the spelling of a quote decide whether an option was protected.
- I authored the missing `nl` string by hand instead of leaving it to the next run. Without it the run is unblocked but not clean — the key stays on the work plan and the outcome depends on the engine's next attempt. Its English is unchanged since the last baseline repin, so no repin is included here; repinning would also freeze the strings that are legitimately waiting for retranslation.
- The workflow header lists what the script validates and was one item short of the module docstring before this change, so it now names formatting-tag parity as well. That comment is the entire `.github/` diff.
- I did not act on the docstring-coverage warning. Its 80% threshold is the checker's default, not this repository's convention — 42% of the unit tests here carry docstrings — and the new test class documents its intent in its own docstring and in comments on the individual cases.

**Problems Encountered:**

- The first design was wrong in a way that reads as correct. "Preserve quoted literals" is what the engine prompt effectively asks for, and enforcing it verbatim would have rejected correct translations across every shipped locale. Measuring the candidate rule against the existing catalogs before implementing it is what surfaced that.
- `re.findall` is typed as `list[Any]`, so returning one of its elements made the helper return `Any` from a function declared `str | None`. The type check that runs over `scripts/` in CI covers a wider surface than the check I run locally before pushing, so this only appeared after the first push.
- Two of my own measurements were wrong before they were right. The first scan of the component catalogs flattened only top-level values, so it silently examined zero component keys while reporting cleanly; a later check compared quote glyphs rather than label text and concluded that six catalogs translate a label they all in fact keep. The counts in the description and above are from the corrected runs, each with a positive control confirming the scanner sees the keys it claims to scan.

ghhamcp:

Patch76:
@coderabbitai review

PR #2171, opened 2026-08-08
feat: add the Polish (pl) locale across all four translated surfaces

Patch76:
## What does this PR do?

Adds Polish on all four translated surfaces: the canonical settings store, the component config-flow catalog, and both add-on YAML flavors.

Polish is the largest untranslated language by installation base after Dutch. Home Assistant collects no language statistic — the public analytics snapshot carries country codes, not languages — so the ranking is derived from country installations with multilingual countries weighted, which puts pl at roughly 16k estimated installations, ahead of sv at roughly 14k.

The two authored catalogs carry the fourteen keys the ungated checks demand, plus `policies.operators.exists_long`, which no check asks for: it is the condition editor's own dropdown label rather than a `PredicateOp` member, so a catalog without it reads English there until the sync fills it. The demanded keys are the eleven `Decision` and `PredicateOp` words, the `policies.pending.already_decided` host sentence they interpolate into, and one reader-addressing anchor per surface. Both add-on YAMLs are `generate_locales.py` output, regenerated against current master, which is why they carry the `redact_secrets` option that landed after this work started. The remaining 736 strings are the post-merge sync's job, so the gated completeness checks fail for pl until that run, exactly as they did for nl.

Register is measured, not assumed. Across all 9708 Polish strings in frontend wheel 20260729.6 there is no formal address marker at all against 92 informal possessives, so the two strings that address a reader use the ty-form; the other thirteen are impersonal third person, matching the English and every other shipped catalog. The operator words render the concept instead of the backend's spelling — `in` reads "jest jednym z" — and a value left as the backend's own word is what the check reds on.

## Type of change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

No agent testing: this adds catalog data and one documentation line, neither of which an agent exercises at runtime.

Beyond the suite:

- Of the 13 parametrized IDs that carry `[pl]`, 4 run in PR CI and pass; the other 9 are the completeness checks, which skip until the sync sets their flag.
- A count of IDs understates the coverage, because several checks iterate every shipped catalog instead of being parametrized. Confirmed by injection instead: spelling `policies.operators.in` as the backend operator reds `test_every_predicate_operator_has_a_catalog_word`, and the suite is green again once the word is restored.
- The sync planner reports 736 items, all `pl`, and none of the authored keys — so the machine fill will not overwrite the hand-written anchors. That also held empirically for nl, whose authored operator words are unchanged after its own fill.
- `generate_locales.py` reports no further updates after the regeneration, so the projections are byte-exact generator output.
- The anchors were cross-read before marking ready, and one finding adopted: `common.oauth_not_serving` left the product name in the nominative after an imperative, where Home Assistant's own Polish declines it — `ui.panel.config.hardware.restart_homeassistant` reads "Uruchom ponownie Home Assistanta".

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added Polish translations for Home Assistant add-on settings, including security, backup, tool, YAML editing, and TLS options.
  * Added Polish messages for approval decisions, accessibility restrictions, and OAuth status notifications.
* **Documentation**
  * Updated the shipped-locale documentation to include Italian.
  * Added localized descriptions covering configuration defaults, safeguards, and operational behavior in the development add-on.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -755,7 +755,7 @@ identity holds by construction.
 
 A language ships on all four surfaces or not at all —
 `tests/src/unit/test_locale_parity.py` enforces it. One Home Assistant language
-code (`de`, `es`, `fr`, `it`, `nl`, `ru`, `zh-Hans`) names every file:
+code (`de`, `es`, `fr`, `it`, `nl`, `pl`, `ru`, `zh-Hans`) names every file:
 `src/ha_mcp/settings_ui/locales/<code>.json`,
 `custom_components/ha_mcp_tools/translations/<code>.json`, and
 `homeassistant-addon{,-dev}/translations/<code>.yaml`.
--- a/custom_components/ha_mcp_tools/translations/pl.json
+++ b/custom_components/ha_mcp_tools/translations/pl.json
@@ -1,0 +1,5 @@
+{
+  "common": {
+    "oauth_not_serving": "Legacy OAuth jeszcze ich nie udostępnia — uruchom ponownie Home Assistant, gdy pojawi się taka prośba, aby je aktywować."
+  }
+}
--- a/custom_components/ha_mcp_tools/translations/pl.json
+++ b/custom_components/ha_mcp_tools/translations/pl.json
@@ -1,5 +1,5 @@
 {
   "common": {
-    "oauth_not_serving": "Legacy OAuth jeszcze ich nie udostępnia — uruchom ponownie Home Assistant, gdy pojawi się taka prośba, aby je aktywować."
+    "oauth_not_serving": "Legacy OAuth jeszcze ich nie udostępnia — uruchom ponownie Home Assistanta, gdy pojawi się taka prośba, aby je aktywować."
   }
 }
--- a/homeassistant-addon-dev/translations/pl.yaml
+++ b/homeassistant-addon-dev/translations/pl.yaml
@@ -1,0 +1,258 @@
+---
+# GENERATED FILE — do not edit. Translations live in
+# src/ha_mcp/settings_ui/locales/pl.json (keys addon.*, features.*,
+# addon_dev.*); regenerate with: python scripts/generate_locales.py
+configuration:
+  backup_hint:
+    name: Full-HA snapshot suggestion level (LLM-facing)
+    description: 'Tunes the wording of the hint the LLM sees in the
+
+      `ha_manage_backup(scope=''snapshot'')` tool description for when to
+
+      suggest a full Home Assistant tarball backup before risky operations
+
+      (e.g. mass device deletes). Affects ONLY this one LLM-facing prompt
+
+      sentence.
+
+
+      For per-edit automatic snapshots of automations / scripts / helpers /
+
+      etc., use the separate **Enable auto-backup of edits** toggle below.'
+  secret_path:
+    name: Secret path override
+    description: 'Optional custom HTTP path for the MCP server. Leave empty to use
+      the auto-generated secure path.
+
+      '
+  enable_tool_search:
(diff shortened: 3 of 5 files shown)


ghhamcp:

Patch76:
## Implementation Summary

**Choices Made:**

- The settings catalog carries the fourteen keys the ungated checks demand, plus `policies.operators.exists_long`, which no check asks for. That one is the condition editor's own dropdown label rather than a `PredicateOp` member, so leaving it out would have rendered English in the editor until the sync ran. Everything else is the sync's work, so authoring more by hand would duplicate what the pipeline exists to do.
- The operator words render the concept rather than keeping the backend's spelling: `in` reads "jest jednym z". The check compares the catalog value against the operator literal, so what it catches is a value left as the backend's own word.
- Both add-on YAMLs are `generate_locales.py` output. I regenerated them against current master rather than shipping the versions the branch was cut with, which is why they now include the `redact_secrets` option.
- Register is split deliberately. The two strings that address the reader use the informal ty-form; the other thirteen are impersonal third person, matching both the English and every other shipped catalog. The informal choice is measured rather than assumed: across all 9708 strings of the shipped Polish frontend catalogue there is no formal address marker at all, against 92 informal possessives. It matters beyond its size because the sync shows the two addressed strings to the engine as the register sample for the 736 it fills.
- I adopted one finding from a final cross-read of the anchors and declined three. Adopted: the restart sentence left the product name in the nominative after an imperative, where Home Assistant's own Polish declines it in that exact position. Declined: the neuter agreement in `gt`/`lt`, because the English is "is greater than" and the shipped Polish catalogue contains neither "większe niż" nor "większe od", so there is no house wording to align to; the asymmetry between `exists` and `exists_long`, because the English is asymmetric in the same way — "exists" against "is present (any value)"; and replacing "jest jednym z" with "należy do", because the English is "is one of", which the current wording mirrors exactly. Home Assistant does use "należy do" once, so that last one is a real alternative rather than an unavailable one — it is simply not the sentence being translated.

**Problems Encountered:**

- The rebase onto current master was conflict-free and still left the branch wrong. Both add-on YAMLs are generated, and an English option added since the branch was cut meant they no longer matched generator output. I regenerated before running anything, so I never saw it fail — but it would not have shipped silently either: `test_check_passes_on_the_committed_tree` and `test_derived_catalogs_match_the_canonical_store` both red on that state, and PR CI runs them. The lesson is narrower than "CI would have missed it": conflict-free tells you nothing about generated files, and regenerating is cheaper than finding out from CI.
- One number in the original commit message had drifted for the same reason: the count of strings left for the sync moved from 734 to 736 when master gained the two `features.redact_secrets` keys. Corrected before pushing.
- I overstated the test coverage twice, in opposite directions. First by selecting tests by name, which matched several that merely contain "pl" — "placeholders", "plans". Then, correcting that, I scoped the recount to two files and reported 11 parametrized IDs with 2 running ungated; across the whole unit directory it is 13 IDs with 4 running ungated, and the two I missed are exactly the ones pinning the reader-addressing anchors this PR argues about. The numbers here are from the full directory.
- Checking the declension question against `home-assistant/core` returned zero hits, which briefly looked like an answer. It was not: core ships only its English source strings, so that search could not have found Polish either way. A positive control on the same query — a phrase that had to exist — is what exposed it. The evidence came instead from the published frontend wheel, read over HTTP range requests so that one 500 KB member could be inspected without fetching 124 MB.

ghhamcp:

PR #2179, opened 2026-08-09
feat: add the Esperanto (eo) locale across all four translated surfaces

Patch76:
## What does this PR do?

Adds Esperanto as a shipped locale code: it names a file on each of the four translated surfaces — the canonical settings store, the component config-flow catalog, and both add-on YAML flavors.

What those files contain at this commit is the anchor set, not a translation. The settings catalog holds 16 of 453 keys, the component catalog 1 of 93, and both add-on YAMLs are still English throughout, because they project add-on keys the settings catalog does not carry yet. Filling the rest is the post-merge sync's job; a dry run plans 734 strings across exactly one locale. The gated completeness checks therefore fail for `eo` until that run — 7 of them, plus the sync's own no-op invariant, which is red by construction whenever a language is waiting to be filled.

Esperanto is not on the installation-base ranking the previous locales were chosen from, and it cannot be: it has no country to derive from, and Home Assistant's published analytics carry no language dimension to rank by. It ships here for a different reason. `src/ha_mcp/settings_ui/locales/README.md` states that the engine reads its target language from `meta.native_name`, so "any language an LLM can write — natural or constructed — needs no pipeline change". This is the first locale that puts that claim under load, at a cost of one anchor set and one sync run.

## Home Assistant's own Esperanto is not usable as the reference

The three previous locales were measured against Home Assistant's shipped translations, and that corpus decided several wording questions. It cannot do that here, which is worth recording because the next person will otherwise reach for it.

Measured against frontend wheel `20260729.5`, over the 8472 keys the Esperanto catalog shares with English:

| catalog | shared keys | still byte-identical to English |
|---|---|---|
| `de` (control) | 8472 | 460 (5.4%) |
| `eo` | 8472 | 8336 (98.4%) |

136 strings are genuinely translated. 11 of those are byte-identical to Home Assistant's **French** catalog — `Ajouter`, `Humidité`, `Retour`, `Restaurer une sauvegarde` — while none match the German one, so this is contamination rather than coincidence of short words. Several of the remaining Esperanto strings are themselves ungrammatical (`Montri informojn pri aŭtomatigon` takes an accusative after `pri`; `Klarigi` for "Clear" means "to explain").

The register question that dominated the earlier locales does not arise: Esperanto has no T–V distinction, only `vi`. The probe reports 30 informal markers and 0 formal ones, and the low absolute count is the answer rather than a broken instrument.

So these strings were checked against the nine shipped ha-mcp catalogs and against Esperanto grammar instead. `exists_long` reads `iu ajn valoro` rather than the `ajna` back-formation for that reason.

## Three authored keys beyond the required set

The ungated checks demand thirteen `messages`; the catalog carries sixteen. `policies.operators.exists_long` is asked for by no check at all — it is the condition editor's own label rather than a `PredicateOp` member — but a catalog without it reads English there until the sync fills it, so it is carried as the previous locales carried it. The other two are `actions.approve` and `actions.deny`. Every shipped catalog keeps the button verb and the decision participle on one stem — Genehmigen/genehmigt, Approuver/approuvée, Zatwierdź/zatwierdzone, Godkänn/godkänt — 9 of 9, including the three the machine filled unprompted. Esperanto can reach the "Deny" side from either `rifuzi` or `malaprobi`, so the pair is the one place where the sync could pick a different stem for the verb than the participle it is substituted next to. Pinning both halves costs two lines and removes that.

## Type of change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

No agent testing: this adds catalog data and one documentation line, neither of which an agent exercises at runtime. No Python file changes, so the lint surface is unchanged.

Beyond the suite:

- The branch collects 15 more test IDs than master over the locale modules (417 against 402): 13 plain `[eo]` plus the two `[eo-approved]` / `[eo-denied]` behaviour tests, which run here because `tests/js` has its npm dependencies installed — without them they skip silently.
- Gated (`LOCALE_COMPLETENESS_CHECKS=1`) the two arms were run with identical test selection against a pristine-master control worktree. Master is clean at 402/402; this branch fails 8. Seven carry `[eo]`; the eighth is `test_clean_tree_plans_no_work`, whose plan holds 734 items, all `eo`. The control worktree's failure set is empty, so nothing pre-existing changed.
- Counting IDs understates the anchors, so both were confirmed by injection instead. Spelling `policies.operators.contains` as the backend literal reds `test_every_predicate_operator_has_a_catalog_word`. Replacing the component catalog's single key with one whose English does not address the reader reds `test_every_shipped_component_catalog_gets_reader_addressing_samples[eo]` — and so does keeping the key but blanking its value, which is the separate non-blank rule. Dropping `accessibility.storage_blocked` reds the settings-side twin, confirming it is that surface's only anchor. All four are green again once restored.
- `generate_locales.py` reports no further updates after generating the two YAMLs, so the projections are byte-exact generator output.
- The anchors were cross-read by three independent models before committing — deliberately not Gemini, which is the engine itself. Each finding was then measured against the nine catalogs that ship today, and most were declined on that evidence: no shipped catalog renders the operators as a bare participle, which is what two models proposed; lowercasing `NE` contradicts seven of the nine, the exceptions being Russian, which lowercases it, and Chinese, which has no case; and `de alia langeto` is the standard agent preposition after a passive participle, which eight of the nine catalogs also read as agent or source (`par`, `door`, `przez`, `由`, `von`, `desde`, `da`, `из`) against Swedish's lone locative — the reading two models proposed. Dropping "site" from the storage string would reintroduce a browser-storage/site-storage confusion a previous locale already had to correct. One model reported an inconsistency between two strings but misquoted one of them, contrasting a catalog value with a lowercase variant the catalog does not contain.
- Two changes were adopted. One came from the grammar check rather than from a model: `iu ajn valoro` replaced the `ajna` back-formation. The other was a model finding I first declined for the wrong reason — I had counted how the catalogs attribute the asker in the OAuth string over seven of them when nine ship. Over all nine, naming the asker is what five do, so the string now reads `kiam ĝi petas vin` rather than the impersonal `kiam oni petas vin`, which also tracks the English "when it asks you to". The second-person pronoun stays either way, because that key is the component surface's only register anchor.
- A blind back-translation was run as a separate instrument, with the models shown only the Esperanto — no English, no key names. It independently rendered `de alia langeto` as "by another tab" in all three, which is the reading the declined finding disputed. It also surfaced the divergence the second commit fixes: run against the impersonal first version, all three returned "when one asks you" rather than the English "when it asks you to". Re-run against the current text, it returns "when it asks you to". The instrument carries a positive control of deliberately corrupted input — negation removed, polarity flipped — which it renders faithfully rather than smoothing over, so its silence on the other strings is worth something.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added Esperanto localization across add-on settings, accessibility controls, approval actions, security policies, and OAuth messaging.
  * Added localized descriptions for backups, dashboard resources, tool search, snapshot deletion, TLS verification, and other configuration options.

* **Documentation**
  * Clarified experimental feature settings and auto-backup behavior, including throttling, snapshot requirements, and failure handling for file and raw-YAML operations.
  * Updated localization guidance for reader-addressing translations.

* **Bug Fixes**
  * Improved locale validation and settings catalog coverage.
  * Updated the component release version to 1.3.4.


ghhamcp:

ghhamcp:

Patch76:
## Implementation Summary

**Choices Made:**

- **Sixteen authored `messages` at the time of this choice, where the ungated checks demand thirteen** (round two adds two more, below). The checks ask for the nine `PredicateOp` words, the two `Decision` outcomes, the sentence they are interpolated into, and one key whose English addresses the reader. Three keys go beyond that. `policies.operators.exists_long` is asked for by nothing — it is the condition editor's own dropdown label rather than a `PredicateOp` member — but a catalog without it reads English there until the sync fills it, so it is carried as the previous locales carried it. `actions.approve` and `actions.deny` are pinned because every shipped catalog keeps the button verb and the decision participle on one stem, 9 of 9, including the three the sync filled unprompted; Esperanto can reach the "Deny" side from either `rifuzi` or `malaprobi`, so this is the one pair where the sync could pick a stem that disagrees with the participle it gets substituted next to.
- **`exists_long` reuses the short operator's word with a parenthetical** — `ekzistas (iu ajn valoro)` — as four of the nine shipped catalogs do (`ist vorhanden`, `присутствует`, `finns`, `存在`, each repeated with a parenthetical); the other five substitute a different word. It spells `iu ajn valoro` rather than the `ajna` back-formation, which is not the standard correlative form.
- **`Legacy OAuth` stays untranslated.** Four catalogs keep the product name verbatim (de, nl, pl, sv), four translate it (es, fr, ru, zh-Hans), and Italian keeps the English word but lowercases and postposes it. Esperanto is Latin-script, where the split favours keeping, and the most recent locale to be reviewed was corrected in that direction.
- **`de alia langeto` keeps the passive-agent preposition.** Two of three cross-reading models wanted a locative instead. Eight of the nine shipped catalogs render that slot as agent or source rather than location, the English source says "by", and `de` is the standard agent preposition after a passive participle. A later blind back-translation returned "by another tab" from all three models.
- **`kiam ĝi petas vin` names Home Assistant as the asker**, as the English does. The first version was impersonal; see below.

**Problems Encountered:**

- **Home Assistant's Esperanto frontend catalog is unusable as the house-convention reference** the previous locales were measured against. Of the 8472 keys it shares with English, 8336 (98.4%) are still English; German as a control sits at 5.4%. Of the 136 genuinely translated strings, 11 are byte-identical to Home Assistant's French catalog and none to the German one, so that is contamination rather than coincidence of short words. This removed the instrument that decided several wording questions on the previous three locales, and is why the authored set was cross-read three ways instead.
- **The cross-reading models converged on findings the measurement refuted.** No shipped catalog renders the operators as a bare participle, which is what two models proposed; lowercasing `NE` contradicts seven of the nine, the exceptions being Russian, which lowercases it, and Chinese, which has no case. One model reported an inconsistency between two strings, but misquoted one of them — it contrasted a catalog value with a lowercase variant the catalog does not contain. The models generate candidates; the cross-catalog measurement decides.
- **One decline was mine to reverse, and how it survived is the useful part.** I had counted how the catalogs attribute the asker in the OAuth string over seven catalogs when nine ship, and concluded the impersonal reading was the majority. Over all nine it is the minority: five name the asker. The second commit changes the string to `kiam ĝi petas vin`, which is both what the English says and what the larger group does. The same subset error also produced "two catalogs" for the `exists_long` pattern above, where the answer is four — both omitted catalogs were the two non-Latin-script ones each time. A count is only as good as the population it was taken over, and neither hand-check of the tallies caught it, because both re-checked the arithmetic rather than the population.
- **One measuring instrument was itself wrong, and it was the control.** The German control figure was first computed as 6.2% via a dump that writes `key<TAB>value` one per line. 151 of the shared keys carry ICU plural forms containing literal newlines; the reload truncated those values at the first line, after which 68 German strings compared equal to English. Measured against the wheel directly, German is 460 of 8472, or 5.4%. The Esperanto figures reproduce identically either way, so the finding stands and only the control moved — but the wrong number reached the first commit's body and had to be corrected there as well as here.
- **Two further figures of mine were wrong before they shipped.** A count of French strings taken by eye (16) measured as 11 against the French catalog, and an agent-versus-locative tally stated as 6 of 8 was, over the full set, 8 of 9. Both were corrected before the artefacts carrying them were published.
- **The first gated comparison run was truncated; the figure it produced was right, and my round-two attempt to withdraw it was wrong.** `tests/pytest.ini` carries `--maxfail=3`, so a run that collected 417 items stopped after 65. With the cap lifted that gave eight failures on this branch — seven `[eo]` plus the sync's own no-op invariant. Re-measuring for round two I got seven, concluded the seventh `[eo]` was irreproducible, and started to withdraw the figure. The re-measurement was the broken one: this checkout's editable install resolves `ha_mcp` to the main clone, which sits on `master` and ships no `eo.json`, so `test_settings_ui_i18n.py` never generated its `[eo]` parameter at all. A missing parameter is not a failure, and nothing said so. With the branch's own `src` on the path the count is eight at `c072a7e3` and `9ff78047`, and nine here — the seventh `[eo]` is `test_shipped_catalog_translates_the_tools_tab[eo]`, which fails by construction while `eo.json` ships `tools` empty. Ungated over the same four files is 167, not the 166 I first reported. Every count taken in a worktree against this shared venv reads `master`'s package data unless `PYTHONPATH` says otherwise. Its plan held 734 items, all `eo`, at those two refs; here it holds 750 — 732 `eo`, the 734 minus the two anchors this commit authors, plus the two English keys moved here queued across the other nine locales, which is the retranslation this branch is meant to trigger.
- **Reflowing the `AGENTS.md` sentence broke its own pin.** `test_agents_md_lists_every_shipped_locale` reads the single line containing `names every file:` and asserts there is exactly one, so wrapping the now-longer list reds it. The wrap was reverted and only the code was added.

---

## Round two — `dd4185bc`

**Choices Made:**

- **The enumerated backup domains were short by four, not the two the review named.** Walking every `@with_auto_backup` decoration rather than the two cited sites gives 34 decorations over 18 domain families — 17 literal plus the computed `helper_<type>` one — against a sentence that listed 14. `device` and `dashboard_resource` join the enumeration; `file` and `yaml` do not, because they are exactly the mandatory pair and belong in the new sentence instead.
- **Three tools are mandatory, not two.** The review names `ha_write_file` and `ha_config_set_yaml`. `tools_filesystem.py:866` is `ha_delete_file`, also `mandatory=True`, and it was missing from my first draft of the corrected string — I had taken the review's list instead of counting the decorations. All three are named now.
- **The sub-flag count is removed rather than corrected.** `config.BETA_FEATURE_FIELDS` holds nine where the English claimed five; a numeral that has already drifted once will drift again, so the sentence no longer carries one.
- **No baseline repin.** This branch moves English without translating it. `AGENTS.md` reserves the repin for a PR that does the translating — repinning here would tell the next sync that the nine existing translations already cover the new wording, and they do not.
- **The style-anchor fix closes the settings surface only, deliberately.** The review is right that both authored surfaces rested on one reader-addressing key each. Two settings anchors (90 and 94 characters) close that side. The component side would cost roughly 390 characters of hand-pinned prose about ports and derived URLs to serve as a sampling anchor for the one run that fills the other 92 keys, and hand edits win, so the sync would never revisit them. Declined with the measurement in the thread rather than skipped, and I would rather be overruled than have it pass unnoticed.

**Problems Encountered:**

- **My own regression claim was false, and only per-case injection caught it.** I wrote that removing either new anchor turns the guard red. Measured one at a time, it does not: two anchors still satisfy the property and the suite stays green. Only dropping both reds it. The test pins the property — a catalog must survive any single sample key being queued — not either particular key, and the claim now says so. Reading the test would have confirmed what I meant; injecting each case separately is what showed what it does.
- **A background comparison run had to be thrown away.** I edited `en.json` while a full gated run was in flight, so it read a mixed tree. It was unnecessary anyway: only four files read the completeness flag, so those four are the whole gated surface and the full-suite run adds nothing but ungated tests.

---

## Round three — `5f59a831`

_Rebased onto master after the 2026-08-11 locale sync; SHAs quoted in the sections above predate that rebase and no longer resolve on the branch._

**Choices Made:**

- **The #2180 collision is folded in here rather than split into its own PR.** Splitting would have let the two guard defects be reviewed on their own merits, independent of Esperanto, and they are language-independent. Against that, this branch stays red until such a PR merges, and the defects are only reachable through a locale that has not been synced — which is this branch and nothing else on master today. Folded in, with the guard changes kept separable from the catalog work by commit.
- **Two relaxations rather than one, because each is separately load-bearing.** Reverting only the per-locale surface set, or only the ten-pair floor, puts `[eo]` back in the red on its own. The first draft of the commit message described one and would have left a reader hunting for the other.
- **The surface-wide fault is re-caught by a union assertion rather than by keeping the per-locale one.** A lookup that stops resolving removes its surface from every locale at once, so the union is where that shows; a per-locale assertion cannot both admit a stub and refuse a rename.
- **The `ru` tolerance is dropped rather than taught to wait.** A pending-window carve-out would have been a branch no shipped entry exercises, and the entry is not pending: with the numeral gone from the English, nothing can lose it again. The window itself is documented on the liveness guard, which is where the next reworded English will meet it.
- **The throttle qualifier goes on the add-on description only.** The web-UI field caption renders directly above the throttle field's own caption, so the limit is already on screen there; the add-on string is read alone.

**Problems Encountered:**

- **My first injection did not compare like with like, and it flattered the change.** I renamed the `tools` section with one `sed` across both versions — one occurrence in the old file, two in the new — so the two arms exercised different faults and I read "old catches it per-locale, new catches it in the union" as the whole story. Renamed at the pairs site alone, the new code reddens nine per-locale tests exactly as the old one did; it is only a rename of the section itself that moves the expectation along with the pairs. The code comment now states that limit instead of the "cannot move in lockstep" claim I had written into it.
- **Two counts in the first draft were wrong in the same direction.** "Ten per-locale tests" counted `[eo]`, which was already red in that arm before any injection, as evidence that the old assertion had teeth — nine are attributable. And "reddens on both guards" describes something the code cannot do: the surface-set assertion raises before the floor is ever evaluated.
- **Deleting one table entry rotted three pieces of prose around it.** The header's worked example named the deleted entry's number, a docstring asserted that every shipped entry records a loss where the survivor records only an addition, and another described the table as having stopped carrying a specimen when what had stopped was its pair resolving. All three sat in text this PR was already editing.
- **A measuring instrument reported clean on a run with nine failures.** `grep -E "^FAILED"` over pytest output returns nothing when the summary is ANSI-coloured, so a failure-set comparison came back empty-handed and looked like agreement. Re-run with `--color=no` it listed all nine. The count of gated failures in this PR was taken again afterwards.
- **The Component Version Gate closed under the open PR.** The window was open when this branch was raised and the mirror cut v1.3.3 stable on 2026-08-10; the gate is a state of the world at check time, not of the branch, so it belongs at PR-open and again before merge.




---

## Round four — `c989d496`

**Choices Made:**

- **The component anchors are the two shortest second-person keys, as suggested.** `options.step.init.data_description.server_port` (167 characters of English) and `..._server_url` (222); the next candidate is 254. That takes the component surface from one reader-addressing candidate to three, `_STYLE_SAMPLE_COUNT` exactly, which is what the settings side now carries.
- **The survival test is mirrored rather than generalised.** One test per surface, each reading its own catalogs, instead of one parametrised over both. The two surfaces resolve their catalogs differently — the component side goes through `_surface_catalogs`, which is itself the property another test pins — and a shared helper would have made the component case depend on the settings one being right.
- **Terminology was decided by the nine other shipped component catalogs, not by a model.** `vojo` for the route slot: all nine render it as a route or a path, none as an address, which is what one cross-reading model proposed instead. A second model's objection to the purpose clause was declined on grammar rather than on the catalogs.
- **The documented anchor floor moved with the test.** `AGENTS.md` § Translations and `locales/README.md` both told the next contributor that one reader-addressing component key is enough. Shipping this check makes that false for every language added after this one, so both now say two and name the check. The empty-catalog case is unchanged and still allowed: both parametrisations filter empty catalogs out.

**Problems Encountered:**

- **The cost argument I declined the first half on was wrong, and the review was right about why.** I wrote that hand edits win, so the sync would never revisit these keys. `_plan_component` queues a key when its English has moved against the baseline or when the locale lacks it, so a hand-written value survives exactly until the next rewording of its English and is then retranslated like any other. The cost is translating two strings once, not maintaining them; the reason to prefer short keys was never permanence.
- **A cross-reading model refuted the wrong claim confidently.** Asked to refute that the surface now holds three candidates, one model reported it as false on the ground that only the two new keys address the reader. `common.oauth_not_serving` does too — its English asks the reader to restart when it asks them to, at 101 characters the shortest of the three. Injection settles it either way: with one of the two new anchors removed the check stays green, which it could not do on two candidates minus one.
- **Numbers taken at the same time, so the comparison is like with like.** Ungated over the five locale test files: 313 passed before, 323 after, no failures in either arm — the ten new ones are the component parametrisation. Gated: 9 failures before and the same 9 after, with the missing-component-key count moving 92 to 90. `generate_locales.py` reports no updates, since component keys are not projected into the add-on YAMLs.

homeassistant-addon-dev/translations/eo.yaml
```diff
@@ -0,0 +1,258 @@
...
+      to be on — otherwise this sub-flag is ignored at runtime regardless of its value
+      here.
+  enable_auto_backup:
+    name: Enable auto-backup of edits
+    description: Captures a per-entity snapshot before every wrapped write/destructive
+      MCP tool call (automation, script, scene, helper, dashboard, label, category,
+      group, zone, area, calendar, todo, entity, integration, and sibling remove/delete
+      tools). Snapshots are saved as YAML files under /data/ha_mcp_backups/ (override
+      via HAMCP_BACKUP_DIR) and listed, restored, or deleted via the Backups tab in
+      the web settings UI or via ha_manage_backup(scope='edits', ...). Best-effort
```
kingpanther13: Auto-backup is not universally best-effort. ha_write_file and ha_config_set_yaml use mandatory=True, so disabling auto-backup, or failing a required capture, blocks those writes. This text currently promises the opposite with “never block” and “uncheck to opt out.” Please correct addon.enable_auto_backup.description in the canonical src/ha_mcp/settings_ui/locales/en.json to distinguish ordinary best-effort captures from mandatory file/raw-YAML backups, then regenerate the derived catalogs.

homeassistant-addon-dev/translations/eo.yaml
```diff
@@ -0,0 +1,258 @@
...
+      would corrupt unrelated output). Fields not marked as passwords by their schema
+      cannot be detected. Off by default.
+  enable_beta_features:
+    name: Enable beta features (master)
+    description: ⚠ DANGER — these tools can PERMANENTLY DAMAGE your Home Assistant
+      installation. They write to your YAML config, your filesystem, install custom
+      components, run arbitrary sandboxed Python, and edit tool docstrings the AI
+      sees. There is no warranty and no support guarantee — you enable them at your
+      OWN RISK. Take a Home Assistant backup before turning this on, and never enable
+      in production without one. Master gate for the 5 experimental sub-flags below;
```
kingpanther13: The hard-coded count has drifted. config.BETA_FEATURE_FIELDS currently contains nine master-gated fields, and the dev add-on Configuration surface exposes them as flat options. Please remove the numeral, for example “Master gate for the experimental sub-flags below,” in the canonical English string and regenerate.

kingpanther13 (changes requested):
The Esperanto locale wiring and authored anchors look good. Two generated descriptions need correction at the canonical English source before regeneration: the auto-backup text contradicts mandatory backup behavior, and the beta-master help carries a stale hard-coded sub-flag count. Details inline.

homeassistant-addon-dev/translations/eo.yaml
kingpanther13: One more thing in the same string while it's being rewritten: the enumerated domain list is also incomplete. `device` (`tools_registry.py:999`, `:1100`) and `dashboard_resource` (`tools_resources.py:488`, `:999`) both carry `@with_auto_backup` but are absent from the list. Worth folding into the same edit so the enumeration and the decorator sites match.

src/ha_mcp/settings_ui/locales/eo.json
```diff
@@ -0,0 +1,24 @@
+{
+  "meta": {
+    "native_name": "Esperanto",
+    "dir": "ltr"
+  },
+  "messages": {
+    "accessibility.storage_blocked": "Via retumilo blokas la konservadon de retejaj datumoj; viaj elektoj validas nur por ĉi tiu seanco.",
```
kingpanther13, line 7: `accessibility.storage_blocked` is the only one of the 16 authored messages whose English is second-person — 20 of the 453 English messages are, and this catalog carries that one. The component catalog is a single key. Both surfaces therefore rest on exactly one anchor each.

`_prompt` passes `queued_keys` into `_style_samples` (`scripts/translate_locales.py:599`), and `_style_sample_keys` filters `key not in exclude` (`:547`). Both the style block and the register rule are guarded on `if samples` (`:600-619`) — the comment there notes the rule "says nothing without them". So on any run where the English of the anchor has itself changed, that key is queued, excluded from sampling, and the request goes out with neither a tone sample nor a register instruction.

The two tests that pin this (`test_translate_locales.py:321`, `:353`) call `_style_sample_keys` with the default empty `exclude`, so they stay green on a case the production caller never takes.

Concretely: reword `accessibility.storage_blocked` in any PR before the `eo` fill lands, and that sync run translates roughly 640 Esperanto strings with no tone or terminology sample. The register half is moot for Esperanto, which has no T–V distinction, but "match the tone and terminology of these existing translations" is not. It self-heals after the first successful sync, when this catalog picks up all 20.

Two more second-person keys close it. The shortest available are `visibility.errors.conflict` (90 chars) and `notice.restart.standalone` (94), which fill `_STYLE_SAMPLE_COUNT = 3` exactly. Same reasoning the PR already applies in pinning `actions.approve` / `actions.deny` against a stem the sync might pick differently.

kingpanther13 (changes requested):
Sorry, full review didn't post the first time — here is the full review.

The two inline threads already on this PR stand as written: the auto-backup description contradicting mandatory capture behaviour, and the stale hard-coded sub-flag count on the beta master. Two further items below, both verified against source.

**`tests/src/unit/test_translate_locales.py:344` — the docstring's example is now wrong**

It reads "today `nl` is exactly that, one key." `nl.json` carries all 93 component keys; after this PR `eo.json` is the only one-key component catalog. The example needs to become `eo`.

This matters more than a stale name. The docstring exists to stop someone stripping the addressing key off a thin catalog. A contributor who follows it to `nl.json`, finds 93 translated keys with several addressing the reader, and concludes the check has slack will reach the opposite conclusion from the one it is written to produce — and the loss it guards against is stderr-only inside an unattended run.

**`src/ha_mcp/settings_ui/locales/eo.json` — one register anchor, and the tests model a call production doesn't make**

See the inline comment.

On the two English corrections: editing `addon.enable_auto_backup.description` and `features.enable_beta_features.help` touches only the `en` and `eo` projections, since all nine other catalogs carry their own translations of both keys — so regeneration should move `en.yaml` and `eo.yaml` in both flavors plus the `FEATURE_META` block, and nothing else. The nine existing translations go stale against the new English and the post-merge sync picks them up; no baseline repin is owed here.

homeassistant-addon-dev/translations/eo.yaml
Patch76: ✅ Fixed in dd4185bc — both halves, and the enumeration was short by more than the two you name.

**The mandatory behaviour.** Confirmed at the decorator, and it is three tools rather than the two named: `ha_write_file` (`tools_filesystem.py:723`), `ha_delete_file` (`:866`) and `ha_config_set_yaml` (`tools_yaml_config.py:308`). That gate fails the call closed in two distinct cases — the master toggle is off, refused before the best-effort `try` so it cannot be swallowed, and the capture itself fails, mapped to `BACKUP_CAPTURE_FAILED`. The string now separates the ordinary domains from those two rather than promising "never block" across the board.

One carve-out kept the wording honest: it is a *write* that is refused, not every call. `ha_config_set_yaml` also passes `skip_fn=_is_preview_only_call`, and the wrapper runs `skip_fn` ahead of the mandatory gate (`auto_backup.py:163`), so a confirm-flow preview — the default first call, since `enable_yaml_edit_confirm` ships on — writes nothing and is not refused. My first draft said "refused outright", which would have been wrong for that path.

**The enumeration.** Rather than add the two cited domains, I walked every decoration: 34 sites over 18 domain families — 17 spelled literally plus the one the two `domain_fn` sites compute — against a sentence that listed 14 of them. Missing were `device` and `dashboard_resource` — yours — plus `file` and `yaml`, which are exactly the mandatory pair. The first two join the list; the second two are covered by the new sentence, where they belong.

Regeneration moves `en.yaml` and `eo.yaml` in both flavors plus the `FEATURE_META` block and nothing else, as you predicted — the other nine catalogs carry their own translations of this key.

No baseline repin. This branch changes English without translating it, and repinning would tell the next sync those nine translations already cover the new wording; `AGENTS.md` reserves the repin for a PR that does the translating.

homeassistant-addon-dev/translations/eo.yaml
Patch76: ✅ Fixed in dd4185bc. Confirmed: `config.BETA_FEATURE_FIELDS` holds nine, and the English claimed five.

Removed rather than corrected, as you suggested — "Master gate for the experimental sub-flags below" cannot drift again the way a numeral does. Same regeneration and same no-repin reasoning as the auto-backup thread.

src/ha_mcp/settings_ui/locales/eo.json
Patch76: ✅ Adopted in dd4185bc, and the chain reproduces exactly as you describe it.

Measured rather than taken on trust: 20 of the 453 English messages are second-person, the eo catalog authored 16, and `accessibility.storage_blocked` was the only one of them whose English addresses the reader. Passing that key as `exclude` — the shape `_prompt` actually uses — takes `_style_sample_keys` from one sample to zero, so both the tone block and the register rule drop out together.

`visibility.errors.conflict` and `notice.restart.standalone` are in. They are the two shortest available, they fill `_STYLE_SAMPLE_COUNT` exactly, and losing any single one of the three now still leaves two.

Your point about the two existing tests is the more valuable half, so it has its own regression test: `test_settings_samples_survive_their_own_anchor_being_queued` queues each sample key in turn and requires samples to remain. It passes a non-empty `exclude`, which neither existing test does. Dropping both new anchors turns it red for `eo` alone and leaves the other nine green; dropping only one leaves it green, which is correct -- two anchors still satisfy the property, and the test pins the property rather than either particular key.

**One half I am not doing, and I would rather say so than quietly skip it.** You wrote that both surfaces rest on one anchor each, and that is true — the eo component catalog is the single key `common.oauth_not_serving`, whose English is second-person. Your two suggestions close the settings surface only. The shortest second-person component keys still available are `options.step.init.data_description.server_port` at 167 characters and `..._server_url` at 222, against the 90 and 94 of the settings pair. Authoring those means hand-pinning roughly 390 characters of prose about ports and derived URLs — hand edits win, so the sync would never revisit them — in order to serve as a sampling anchor for the one run that fills the other 92 keys. For Esperanto the register half of what that buys is moot anyway, as you note.

So I closed the cheap surface and left the expensive one, deliberately. If you would rather have the component anchors too, say so and I will write them.

Patch76:
Round two is up as dd4185bc. Both English corrections landed at the canonical source with the derived catalogs regenerated, and the three inline threads carry the details.

On the docstring: you are right, and the example was wrong in the direction that matters. `nl.json` carries all 93 component keys with 16 reader-addressing ones; `eo.json` is the single-key catalog. Corrected to name `eo`, with the contrast spelled out so the next reader cannot repeat the trip — a contributor who followed it to `nl` would have concluded the check has slack it does not have.

Two things worth flagging rather than leaving to be discovered:

- Walking every `@with_auto_backup` decoration instead of the two cited ones showed the enumerated domain list was short by four, not two. Details in the auto-backup thread.
- One half of the style-anchor finding I did not act on, with the measurement for why, in the `eo.json` thread. I would rather have that overruled than skipped silently.

No baseline repin here: this branch moves English without translating it, so the nine existing translations are meant to go stale and be picked up by the post-merge sync.

ghhamcp:

src/ha_mcp/settings_ui/locales/en.json
```diff
@@ -441,7 +441,7 @@
...
+    "addon.enable_auto_backup.description": "Captures a per-entity snapshot before every wrapped write/destructive MCP tool call (automation, script, scene, helper, dashboard, dashboard resource, label, category, group, zone, area, calendar, todo, entity, device, integration, and sibling remove/delete tools). Snapshots are saved as YAML files under /data/ha_mcp_backups/ (override via HAMCP_BACKUP_DIR) and listed, restored, or deleted via the Backups tab in the web settings UI or via ha_manage_backup(scope='edits', ...). For those domains the capture is best-effort — a failure logs a WARNING and the write proceeds. The file and raw-YAML tools are the exception: ha_write_file, ha_delete_file and ha_config_set_yaml treat the snapshot as a precondition, so a write through them is refused while this is off, and a failed capture blocks it. On by default; unchecking opts out of the best-effort captures and takes the file and raw-YAML tools out of service with them. Requires restart to take effect.",
```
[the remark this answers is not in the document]
Patch76: Valid, and fixed in 06c502e0.

`maybe_snapshot` returns before the capture while the per-entity window is open, and that skip is not conditioned on `mandatory` (`src/ha_mcp/backup_manager.py:343-349`); its docstring says so outright, that a throttle skip returns None and lets the write proceed. So the unqualified "before every wrapped write/destructive MCP tool call" contradicted the adjacent option rather than merely under-describing it.

The canonical string now carries "The auto-backup throttle bounds how often that happens: 0 (default) captures on every such call, N>0 at most one snapshot per N minutes per entity", mirroring `addon.auto_backup_throttle_minutes.description` as suggested. Derived catalogs regenerated: that moves the `en` and `eo` projections in both add-on flavors and nothing else, since the other nine catalogs carry their own translation of this key.

Not applied to `messages.backup.fields.enable_auto_backup.help`, and deliberately so. That one is a field caption in the web settings UI, rendered immediately above the throttle field's own caption in the same section (`BACKUP_FIELD_LABELS` in `settings.js`), so the limit sits on screen beside it. The add-on description is a self-contained specification — it enumerates the wrapped domains, the storage location, the mandatory file and raw-YAML exception and the restart requirement — and there the sentence had nothing next to it to correct the reader.

Patch76:
Round three is up as 5f59a831. The branch has since been rebased onto master again, after the locale sync run of 2026-08-11 landed; the SHAs this comment first carried were rewritten by that rebase. The branch was rebased onto master first, because #2180 landed in between and the two collide in a way worth stating plainly: **the guards that PR added do not admit a locale that has not been synced yet, and this is the first one to meet them.**

`_literal_parity_pairs` asks every shipped locale for all four guarded surfaces, and for at least ten pairs on each. A catalog ships its authored anchors before the sync fills it, so `eo` carries neither `tools` nor `tool_groups` and holds a single component key: three of its four surfaces miss the bar for being new rather than for being broken. That check is ungated, so PR CI sees it, and so would every language added after this one. Measured on the rebased branch before the fix: 3 ungated failures over the four locale test files, where the same selection on master is clean.

Both bars are relaxed, and each turned out to carry its own weight — reverting either alone puts `[eo]` back in the red:

- the surface set is derived per locale from what the catalogs actually hold, read before the pending exclusion rather than off the filtered pairs;
- the ten-pair floor applies only where the catalog holds ten keys to begin with, since a catalog awaiting its first sync is small rather than thin.

What that elasticity costs is measured rather than argued, by renaming the `tools` section under both versions. Renamed at the pairs site alone — the drift that empties the filter — the new code still reddens nine per-locale tests, exactly as the old assertion did. Renamed everywhere, it reddens none of them, because the expectation reads the same `settings_catalog.get("tools", ...)` as the pairs do and moves along with it. That case is caught by a new `test_every_guarded_surface_resolves_for_some_locale`, which asserts the union across locales; a lookup that stops resolving takes its surface away from everyone at once, which is exactly what the per-locale check can no longer see. The code comment names that split rather than claiming the expectation cannot move.

**The `ru` literal-parity tolerance is dropped.** It excused the loss of "5" from `features.enable_beta_features.help` — the numeral round two removed from the English at your request. With no digit left in the source, no translation can lose it again, so the entry is dead going forward rather than pending its rewrite. `test_an_exception_that_over_records_is_refused` took its specimen from the shipped table, so it reported the corpus twice over: at the previous head the entry's pair had stopped resolving, and dropping the entry would have left it with no loss-recording specimen at all. Both failures are loud and neither is about containment, so it builds its pair synthetically now.

**Component version opened at 1.3.4.** The Version Gate went red on this push: the mirror cut v1.3.3 stable on 2026-08-10, while the branch was open, so the manifest was level rather than ahead. `manifest.json`, `COMPONENT_VERSION` and the parity-test literal moved together.

On the CodeRabbit nitpick about queued anchors, declined with the measurement. Counting reader-addressing candidates that are present and non-blank in each catalog: the filled catalogs carry 20 on the settings surface — 18 for `it`, which is missing two of those keys on master already — and 16 on the component surface; `eo` has 3 and 1. So the premise that two queued anchors can leave a run with no samples holds for no catalog in the tree — with two of `eo`'s three queued, one remains, and `_STYLE_SAMPLE_COUNT` is 3. The component half is real, but it is the half of your own finding I declined in the `eo.json` thread with the cost of a hand-pinned anchor; a test asserting that surface survives its only key being queued would be red for `eo` today, so it would encode the opposite decision rather than test the current one.

The gated arm is unchanged in kind: 9 failures where master has 1, the extra 8 being seven `[eo]` and `test_translations_are_checked_against_current_english`, which is the nine existing translations going stale against the two English corrections — the state you described as owed to the post-merge sync rather than to a repin here.

kingpanther13:
I'm pretty much out of Claude tokens till Thursday evening, will review when I can, sorry for delay. I still have gpt/codex usage but I don't trust it for doing the full code review lol.

Patch76:
Three of the four items are closed. The fourth is closed on one surface only — deliberately, and it still needs your call. The branch was rebased twice since you reviewed, so `9ff78047` no longer exists on it; the hashes below are live at `5f59a831`.

**Closed**

- **Auto-backup description contradicting mandatory capture** — `2114f250`. The rewrite separates the best-effort domains from `ha_write_file`, `ha_delete_file` and `ha_config_set_yaml`, which treat the snapshot as a precondition. The enumeration now also names device and dashboard resources, the two you flagged as missing.
- **Stale sub-flag count on the beta master** — `2114f250`. Removed rather than corrected, as you suggested, so it cannot drift again.
- **`test_translate_locales.py` docstring naming `nl`** — `2114f250`. It now names `eo` as the one-key catalog, against the 93 keys every other shipped component catalog carries.

**Still open on one surface**

- **Register anchors.** `2114f250` adds both keys you suggested to the settings catalog, which takes it from one second-person candidate to three — `_STYLE_SAMPLE_COUNT` exactly — so losing any single one still leaves sampling non-empty. One new test, `test_settings_samples_survive_their_own_anchor_being_queued`, pins that property rather than either key.

  The component catalog is untouched and still rests on `common.oauth_not_serving` alone, which is the half I said in the thread I was not doing: the shortest second-person component keys available run 167 and 222 characters of prose about ports and derived URLs, hand-pinned forever once authored. I would rather leave that decision with you than fold it in quietly. The offer stands if you want those anchors written.

One correction to my own earlier replies: `dd4185bc`, cited in the three older thread replies, predates the rebase and is no longer reachable from the branch — that content is in `2114f250`. The `06c502e0` in the newest reply is current.

The review is recorded against `9ff78047`, three commits back from head. No rush on re-reading it — you mentioned being out of budget until Thursday evening.

kingpanther13 (changes requested):
Settings half closes cleanly: three anchors — `_STYLE_SAMPLE_COUNT` exactly — and the survival test pins the property rather than any one key.

You asked for the call on the component half: fix it here. One anchor is the same failure shape this review opened with — a change to `common.oauth_not_serving`'s English runs the component translation with the register rule silently gone, stderr-only, inside an unattended run. The hand-pinning cost is also smaller than framed: hand wording wins only until the English moves, then the sync retranslates those keys like any other — the cost is translating two long strings once, not maintaining them forever.

Author both component anchors and mirror the survival test onto the component surface so both halves pin the property.

Patch76:
Both component anchors are authored and the survival test is mirrored onto that surface — `86a7b600`, with the documentation catching up in `c989d496`.

**The anchors.** `options.step.init.data_description.server_port` and `..._server_url`, the two shortest second-person component keys at 167 and 222 characters of English against the next candidate's 254. The surface goes from one reader-addressing candidate to three, `_STYLE_SAMPLE_COUNT` exactly, so it now matches the settings side: losing any single one still leaves two to imitate.

Wording was decided against the nine other shipped component catalogs rather than against a model. `vojo` for the route slot is what all nine carry — route or path, never an address — which is one of two cross-reading suggestions the measurement declined. A blind back-translation, models shown only the Esperanto, returned both strings with their meaning intact; two deliberately corrupted strings rode along and came back corrupted rather than smoothed, so its silence on the real two is worth something.

**The test.** `test_component_samples_survive_their_own_anchor_being_queued` mirrors the settings twin: it passes a non-empty `exclude`, which the shipped-catalog component check does not. Injection rather than reading: removing both new anchors reddens it for `eo` alone and leaves the other nine green, while the existing addressing check stays green through that — which is the gap you described. Removing only one leaves it green, correctly, since two candidates still satisfy the property it pins.

**One thing your reply corrected that I had wrong.** I declined the component half partly on the claim that hand edits win, so the sync would never revisit those keys. `_plan_component` queues a key whose English has moved against the baseline, so hand wording survives exactly until the next rewording and is then retranslated like anything else. Your framing of the cost was right and mine was not.

**Documentation moved with the check**, because shipping it made the existing text wrong: `AGENTS.md` § Translations and `src/ha_mcp/settings_ui/locales/README.md` both told the next contributor that one reader-addressing component key suffices. Both now say two and name the check. An empty component catalog is still fine — both parametrisations filter those out.

Measured over the five locale test files, both arms taken after the change with the same selection as before it: ungated 313 passed to 323, no failures either side, the ten new ones being the component parametrisation. Gated, the same nine failures as before, with the missing-component-key count moving from 92 to 90. `generate_locales.py` reports no updates, since component keys are not projected into the add-on YAMLs.

kingpanther13 (approved):
All four items are closed, verified at c989d496.

Component surface: three reader-addressing candidates, with both authored keys second-person at the English source. The survival test mirrors the settings twin through `_surface_catalogs(locale, "component")` with the production exclude shape and runs over every shipped component catalog, so the property is pinned per-surface rather than per-key. The docs floor moved to two keys in both places that stated one, naming the check.

PR #2205, opened 2026-08-13
Switch Locale Sync default model to Gemini 3.5 Flash-Lite

julienld:
## What changed

This switches the Locale Sync translation pipeline default model from `gemini-2.5-flash` to `gemini-3.5-flash-lite`.

## Why

`Locale Sync` is a structured JSON translation workflow, not a deep reasoning or coding workflow. `gemini-3.5-flash-lite` is positioned by Google as the cost-efficient GA model for translation, high-volume execution, and simple data processing, which is a better fit for this path than keeping the older default.

This keeps the change intentionally minimal:
- no workflow changes
- no secret changes
- no provider swap
- no reasoning/tuning parameter changes

## Pricing

Per Google's official Gemini API pricing on August 13, 2026, the **Standard** paid-tier pricing for these two models is currently the same:

- Old default: `gemini-2.5-flash`
  - input: **$0.30 / 1M tokens**
  - output: **$2.50 / 1M tokens**
- New default: `gemini-3.5-flash-lite`
  - input: **$0.30 / 1M tokens**
  - output: **$2.50 / 1M tokens**

So this change is primarily about using the newer translation-oriented GA model, not about lowering the current Standard-tier token price.

Official source:
- https://ai.google.dev/gemini-api/docs/pricing

## Impact

The pipeline keeps the same code path and API surface, but future translation runs will use the newer default model unless `GEMINI_MODEL` is explicitly overridden.

## Validation

Ran:

```bash
/home/julien/github/ha-mcp/.venv/bin/python -m pytest tests/src/unit/test_translate_locales.py -q
```

Result:
- `78 passed, 1 skipped`

Also validated the model slug live against the same `generateContent` endpoint used by `Locale Sync`:
- `gemini-3.5-flash-lite` -> HTTP 200
- `gemini-2.5-flash` -> HTTP 200

### DIFF
--- a/scripts/translate_locales.py
+++ b/scripts/translate_locales.py
@@ -91,7 +91,7 @@ SETTINGS_SURFACE = "src/ha_mcp/settings_ui/locales"
 COMPONENT_SURFACE = "custom_components/ha_mcp_tools/translations"
 TOOLS_SURFACE = "settings UI tool titles and descriptions"
 
-DEFAULT_MODEL = "gemini-2.5-flash"
+DEFAULT_MODEL = "gemini-3.5-flash-lite"
 # Free-tier pacing: stay comfortably under the strictest published RPM.
 _SECONDS_BETWEEN_REQUESTS = 7.0
 # Self-imposed wall-clock bound, kept below the workflow's job timeout: when


ghhamcp:

ghhamcp:

kingpanther13:
I'm pretty sure the model it's using is free tier, so pricing shouldn't matter I thought. Or at least that's what it told me when I set it up.

kingpanther13 (changes requested):
Requesting changes on one code issue; the rest are claims in the PR body and metadata.

## 1. Drop `temperature` from `generationConfig` (`scripts/translate_locales.py:686`)

`generationConfig` carries `temperature: 0.2`, written when the default was `gemini-2.5-flash`. Google's Gemini 3.5 migration guide is explicit:

> `temperature`, `top_p`, and `top_k` are no longer recommended for all Gemini 3.x models. Gemini 3's reasoning capabilities are optimized for the default settings. Remove these parameters from all requests.

and the Gemini 3 developer guide:

> For all Gemini 3 models, we strongly recommend keeping the temperature parameter at its default value of `1.0`. [...] Changing the temperature (setting it below 1.0) may lead to unexpected behavior, such as looping or degraded performance, particularly in complex mathematical or reasoning tasks.

The description lists "no reasoning/tuning parameter changes" as a property of the minimal diff. That is where the issue sits: a 2.5-era tuning parameter now rides on a 3.x model against the vendor's stated guidance, in the same function as the changed constant.

The named failure domain is math and reasoning, and this workload is JSON translation, so this is guidance rather than a demonstrated regression here. It still matters more than usual on this path: `locale-sync` runs on a daily cron and pushes to master with no human in the loop, and nothing downstream can see degraded wording. `_validate` covers placeholder parity, the markup allowlist, formatting-tag and panel-link parity, and the hardcoded on-screen names; register and phrasing pass all of it, which the script records itself at lines 113-123 and 832-834.

Removing the `temperature` key and keeping `response_mime_type` is the whole fix.

## 2. The Codex P1 on line 94 is wrong

It claims the Gemini API does not expose a model named `gemini-3.5-flash-lite`. It does: the model has its own page in Google's model documentation, is listed as GA, appears on the pricing page, and you already verified HTTP 200 against `generateContent`. Its model knowledge predates the release, so that thread wants a refutation rather than a revert.

## 3. `_SECONDS_BETWEEN_REQUESTS = 7.0` and its rationale went unexamined (`scripts/translate_locales.py:95`)

The line directly below the change:

```python
# Free-tier pacing: stay comfortably under the strictest published RPM.
_SECONDS_BETWEEN_REQUESTS = 7.0
```

7.0s is roughly 8.6 RPM, sized against `gemini-2.5-flash`. The model moved; the constant and its justification did not. Google's rate-limits page no longer publishes a per-model free-tier table and defers to AI Studio, so "the strictest published RPM" no longer names a number anyone can check. The same free-tier premise is restated at lines 20 and 34, `AGENTS.md` 804 and 857, and `.github/workflows/locale-sync.yml` 47 and 121.

There is no evidence of a regression here. Flash-Lite tiers have historically carried higher free-tier limits than Flash, so the likely direction is favourable and may be an unstated benefit of this change. Please state the new model's free-tier RPM and RPD from the AI Studio dashboard for the key this workflow uses, and either confirm 7.0 or move it. If the limits went up, the comment is worth rewording to name the actual number instead of an unpublished one.

Free-tier eligibility itself is intact: the pricing page lists Gemini 3.5 Flash-Lite as free of charge for both input and output on the free tier, same as Gemini 2.5 Flash.

## 4. The Validation section claims more than the run shows

`tests/src/unit/test_translate_locales.py` asserts nothing about the model. No test references `DEFAULT_MODEL`, `GEMINI_MODEL`, or the request URL, and every test that reaches `_call_gemini` substitutes `def fake_post(*_args, **_kwargs)` plus a hand-built response object, so the URL is never read. `DEFAULT_MODEL = "not-a-model"` produces the same `78 passed, 1 skipped`. The suite proves the retry and backoff branching still holds; it is invariant under this diff. The HTTP 200 proves the endpoint accepts the slug, not that a batch round-trips the response path and survives `_validate`.

No test is owed for this change: asserting the constant equals itself is a tautology, and the module docstring deliberately leaves the live happy path to the workflow run. What would settle both this and item 1 is one local run against the real catalogs with the new default, with the resulting diff posted here.

## 5. Title needs a conventional-commit prefix

`Switch Locale Sync default model to Gemini 3.5 Flash-Lite` carries no type prefix. Squash-merge takes the PR title as the master subject, and every recent commit on master has one. `chore(internal):` fits, matching `chore(internal): machine-translate locale updates` in the same subsystem.

kingpanther13:
Superceded by #2232

Issue #1991, opened 2026-07-21
[BUG] Claude.ai shows "Get Dashboard" as a write/delete tool?


deviantintegral:
### 💬 What Went Wrong?

Any idea why Get Dashboard is showing up as a write/delete tool? This is in claude.ai/chat on the web. I don't see anything obvious in the tools definition that would cause this. Or, is there something about getting a dashboard that makes it somehow more dangerous than other get calls?

### 📋 AI-Generated Content / Logs / Config

<img width="979" height="736" alt="Image" src="https://github.com/user-attachments/assets/efe77fb8-290c-4ad3-af33-dc7ad7c7b720" />

### 🖥️ Client Application

Other (specify below)

### 📦 Installation Method

Home Assistant Add-on

### 💻 Operating System

Linux

### OS Version

HAOS

### 📖 Setup Guide Followed

https://homeassistant-ai.github.io/ha-mcp/setup/

### 🔄 Exact Steps You Followed

Installed via the addon, configured the webhook addon to support external access, and enabled OAuth 2.0 connections in the addon.

### 📦 ha-mcp Version

7.14.1

### 🏠 Home Assistant Version

2026.7.2

kingpanther13:
It's safe to use. Might be because it has some of the screenshot tool functionality merged in with it because technically it can write screenshots. Will investigate and see if it's possible to modify it a bit so we can make it back into a true read only tool

Patch76:
Thanks for including the client screenshot with the tool title visible – that pinned it to `ha_config_get_dashboard` rather than to the separate screenshot tool.

Nothing about reading a dashboard is dangerous, and this is not a client-side misrendering. The tool no longer carries `readOnlyHint: True`, and the MCP schema documents that hint as defaulting to false, so a client that groups tools by capability falls back to treating it as write-capable.

What changed: #1837 swapped the annotation on `ha_config_get_dashboard` from `readOnlyHint: True` to `destructiveHint: False` (commit `fde387a2`, merged 2026-07-13, contained in `v7.14.1`). Current state, `src/ha_mcp/tools/tools_config_dashboards.py:1750-1755`:

```python
annotations={
    "openWorldHint": False,
    "destructiveHint": False,
    "idempotentHint": True,
    "title": "Get Dashboard",
},
```

[...]

PR #2255, opened 2026-08-24
fix(screenshot): re-arm the theme guard for themed captures only

kingpanther13:
## What does this PR do?

Re-arms the `ThemeGuard` bracket around dashboard captures, **but only for
renders that request a theme**.

This is a deliberate, narrowed partial reversal of #2014 (the decision taken on
#1991, which disabled the bracket to restore `readOnlyHint: True` on the
dashboard get/screenshot tools). It is not a claim that #2014 was a mistake —
it rests on two facts that were not available then:

1. **The upstream fix never shipped.** `balloob/home-assistant-addons#89`
   ("don't dispatch settheme when no theme/dark was requested") is merged to
   that repo's master but unreleased. It landed *after* the 2.6.0 version bump
   (`balloob/home-assistant-addons#88`), and `puppet/config.yaml` there still
   reads `2.6.0` with no `settheme` entry in its CHANGELOG. So "current
   upstream release" — the AGENTS.md § Safety Annotations version baseline —
   still writes for everyone, not only outdated builds.
2. **The fix is narrower than assumed.** By its own title it suppresses the
   dispatch only when *no* theme/dark was requested. An explicit `theme=` or
   `dark_mode=true` capture still writes the engine user's profile, on every
   engine version, even once #89 ships.

Point 2 is the durable one and is what motivates the narrowing: unthemed
captures — the overwhelming majority — are left completely unbracketed and
issue no writes, so #2014's read-only guarantee is preserved exactly where it
applies. Only the themed path, which upstream will never cover, is protected.

Found in practice: a `dark_mode: true` capture flipped a live instance's theme
and nothing put it back.

### Guard hardening

Re-arming exposes failure modes the disabled code never hit. All raised by
Codex and CodeRabbit on this PR, all fixed here:

- **Serialization per engine user.** Overlapping themed batches could
  interleave (A clobbers, B snapshots the clobbered value, A restores, B
  restores the clobber) and strand the account on the wrong theme. Locks are
  keyed by running event loop as well as credential, since an `asyncio.Lock`
  binds to the loop that first awaits it.
- **Contention fails the capture** rather than rendering unserialized — a
  themed render holds the bracket for its whole duration, so "give up and
  render anyway" would reintroduce exactly the interleave. The wait is bounded
  by the caller's own `render_timeout_seconds`.
- **Cancellation releases the lock.** `asyncio.CancelledError` is a
  `BaseException`, so it bypassed `except Exception` and stranded the lock,
  which would have made every later themed capture for that engine user time
  out.
- **TLS override preserved.** A direct client built with `verify_ssl=False`
  previously fell back to the global setting, so on self-signed instances every
  guard session failed while the render succeeded — leaving the theme
  clobbered.
- **Bounded WebSocket waits** (5s vs `send_command`'s 30s default), so a
  best-effort guard cannot consume the caller's MCP timeout window before the
  engine is contacted.
- **Guard warnings survive non-`ToolError` failures.** These bypassed
  `_reraise_with_guard_warnings`; the wrapper now builds a structured
  `INTERNAL_ERROR` payload, because that helper parses the message as JSON and
  re-raises unchanged when it is not a dict.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

Locally: `test_dashboard_screenshot_theme_guard.py` + `test_dashboard_screenshot.py`
green, `mypy src/ha_mcp/dashboard_screenshot/` clean. Full suite left to CI.

Tests cover the themed/unthemed split (the previous bracket tests exercised
only default options, which after this change is the deliberately unbracketed
path — they would have passed while testing nothing), TLS passthrough, wait
bounding, serialization of overlapping batches, refusal on lock contention,
and lock release on both snapshot failure and cancellation.

## Checklist

### DIFF
--- a/docs/beta.md
+++ b/docs/beta.md
@@ -178,17 +178,21 @@ page and (by its design) restarts; ha-mcp surfaces this as a clear "set the
 engine's access token" error rather than a silent failure.
 
 Puppet's theme and dark-mode renderer controls dispatch Home Assistant's
-`settheme` event on render, which Home Assistant persists on the frontend
-profile of the user whose token the engine runs with — and syncs to that
-user's real web and mobile sessions, flipping that user's whole UI on every
-screenshot (#1909). ha-mcp brackets each capture batch with a snapshot/restore
-of that user's saved theme, so the flip is undone automatically. The upstream
-Puppet fix for the dispatch (balloob/home-assistant-addons#89) is merged but
-unreleased as of Puppet 2.6.0, and is scoped to renders that request no
-`theme`/`dark` — an explicit theme or dark-mode capture still writes even once
-it ships, so the bracket stays on regardless of engine version. A dedicated
-Puppet account remains a sound belt-and-suspenders setup. Language selection
-is local to Puppet's browser session.
+`settheme` event, which Home Assistant persists on the frontend profile of the
+user whose token the engine runs with — and syncs to that user's real web and
+mobile sessions, flipping that user's whole UI (#1909). Upstream stopped the
+dispatch for renders that request nothing
+(balloob/home-assistant-addons#89), but only for that case: passing `theme` or
+`dark_mode` still writes, on every engine version.
+
+ha-mcp therefore brackets **only themed captures** with a snapshot/restore of
+that user's saved theme, so the flip is undone automatically. Captures that
+request no theme are not bracketed and issue no writes at all, which is what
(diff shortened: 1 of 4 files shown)


ghhamcp:

PR #2266, opened 2026-08-24
fix: coordinate shared FastMCP mutations

kingpanther13:
## Summary

- give `defer_mutations` priority over every package path, including stable-spec requirements processing
- track the async Home Assistant requirements path across bring-up cancellation so later starts can wait for the real pip completion
- reject a first HA-MCP install when an unowned process-global FastMCP runtime is already loaded
- publish `FASTMCP_MUTATION_PROTOCOL = 1` as the runtime contract for cooperating custom integrations

## Why

FastMCP modules and their installed files are process-global inside Home Assistant. A live consumer lease already deferred the direct install/uninstall path, but the stable-spec fast path could still reach pip, and a first HA-MCP install could replace FastMCP beneath a previously started external consumer. Both cases can mix cached modules with a different on-disk generation.

This patch makes every mutation-capable route honor the gate and keeps requirements-manager work visible after cancellation. It is the upstream half of a linked ESPHome MCP compatibility change.

## Verification

- `ruff format custom_components/ha_mcp_tools/embedded_server.py tests/src/unit/test_embedded_server.py`
- `ruff check custom_components/ha_mcp_tools/embedded_server.py tests/src/unit/test_embedded_server.py`
- focused regression selection: 5 passed
- full `test_embedded_server.py`: 192 passed; 2 existing metadata-warning assertions fail only in the direct Termux environment because the project distribution is not installed there (CI's managed environment remains authoritative)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when starting the embedded server while package installation is in progress.
  * Prevented cancelled startup operations from leaving installations in an inconsistent state.
  * Added safer handling when another FastMCP runtime is already active.
  * Startup now fails safely when a compatible package is unavailable during deferred installation.

* **Compatibility**
  * Improved coordination with other integrations using the same FastMCP runtime.

### DIFF
--- a/custom_components/ha_mcp_tools/embedded_server.py
+++ b/custom_components/ha_mcp_tools/embedded_server.py
@@ -945,7 +945,7 @@ class EmbeddedServerManager:
                 "not installed; refusing the first install because it could "
                 "replace loaded FastMCP files. Restart Home Assistant so HA-MCP "
                 "can establish runtime ownership before other MCP integrations.",
-                kind="restart",
+                kind="package",
             )
 
     def _replaced_dist_name(self) -> str | None:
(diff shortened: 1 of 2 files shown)


ghhamcp:

PR #2283, opened 2026-08-26
fix(ci): stop Renovate silently aborting its whole run on every branch

sergeykad:
## What does this PR do?

Fixes Renovate silently aborting its entire repository run on every branch it
writes. Since #2202 merged (2026-08-12), the two scheduled runs that had real
work to do (2026-08-18, 2026-08-25) both wrote a branch for the
`python:3.13-slim` digest and then aborted before opening a PR, updating the
dependency dashboard, or processing anything else that run.

**Root cause.** `minimumReleaseAge` makes Renovate publish a stability status
check on every branch it writes. The `ha-mcp-renovate` App installation never
had commit-status access. GitHub answers a status-check call from a token
without that permission with 404, not 403, and Renovate reads that 404 as the
repository having changed underneath it, aborting the whole run immediately
after the first branch: before the PR opens, before the dependency dashboard
updates, before any other dependency is processed. The abort logs below error
level, so Renovate exits 0 and the workflow reports green through it.

**Fix.**
- Request `permission-statuses: write` on the App token.
- Promote the abort message to `error` in `logLevelRemap`, so a recurrence
  fails the workflow.
- Bump the pinned engine to `44.46.2`.

The org-side installation grant for the new permission was accepted
separately and is confirmed live.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

Two new tests pin both halves of the fix, and both were verified against the
actual defect by reverting each half and confirming its test fails:
`test_renovate_can_write_the_status_check_it_publishes` (token requests the
permission) and `test_a_renovate_abort_fails_the_workflow` (abort promoted to
error). The permission grant itself can't be pinned by a test: an installation
token can only narrow what the installation already holds.

## Checklist

### DIFF
--- a/.github/workflows/renovate.yml
+++ b/.github/workflows/renovate.yml
@@ -48,13 +48,21 @@ jobs:
           # dead: an installation token that lists any permission-* input drops
           # every permission it does not name, including vulnerability_alerts.
           permission-vulnerability-alerts: read
+          # Renovate publishes a stability status check on every branch it
+          # writes (minimumReleaseAge). Without commit-status access GitHub
+          # answers 404 rather than 403, Renovate reads that as the repository
+          # having changed, and aborts the WHOLE run right after writing its
+          # first branch: no PR, no dependency dashboard, nothing else
+          # processed. The app installation must grant this too; the token can
+          # only ever narrow what the installation already has.
+          permission-statuses: write
           permission-workflows: write
 
       - name: Self-hosted Renovate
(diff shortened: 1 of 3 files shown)


ghhamcp:

ghhamcp:

PR #2285, opened 2026-08-26
docs(internal): correct the reference-graph ranking comment

Patch76:
## What does this PR do?

Corrects two pieces of prose in `smart_search` that state something the code does not do — one `#` comment and one docstring. No behaviour change; the AST is identical with docstrings stripped.

**`_GRAPH_HIT_SCORE`'s ranking claim.** It says fuzzy scores are capped at 100, so a confirmed reference ranks at or above every fuzzy name match. The cap it describes is in `_search_in_dict` and bounds the config-body score. The fuzzy *name* score comes from `_calculate_entity_score`, which has a single return and is not capped at 100.

This is not the common case — most unrelated names score well under 100 — and it does not need to be. One name echoing the queried entity's words is enough, and you can check this one yourself: `_calculate_entity_score("automation.kitchen_lights_evening", "Kitchen Lights Evening", "automation", "light.kitchen_ceiling")` returns 118, against a graph hit's fixed 100. That is sufficient to push the confirmed reference below a config that merely resembles the query. The exact-mode half of the original claim holds and stays.

The replacement comment is scoped to what I could verify: it speaks about graph-**only** hits, because `_merge_graph_hits` leaves an already-scored record's score alone, and it hedges the "still counted" clause, because `_scrub_results_for_enforce` runs after the merge and can drop a graph hit outright under visibility enforcement.

**`_individual_fetch_budgeted`'s deprioritize docstring** names the automation and script callers; the scene branch passes the set too.

This documents the fuzzy ordering rather than changing it. If you meant the stronger guarantee the comment originally claimed, that needs `match_in_references` in the sort key in `_paginate_and_build_response`. Note that also touches exact mode, where graph hits tie with body hits at 100 and follow them **within a bucket** — across buckets the flattening order in `_paginate_and_build_response` dominates, not the append order. That is a behaviour change, so it stayed out of this comment correction.

One negative result worth recording, since it is about the code this comment describes. I built and then dropped a change that skipped the deprioritization when `include_config=True`, on the theory that a caller asking for bodies wants the confirmed reference's body most. Measuring it killed the idea. In the case I ran — two automations, a budget affording one fetch — the current ordering returned two matches (a templated reference with its body, plus the graph hit) and the change returned one, because the unfetched config scores 0 and falls below threshold. Which id wins the budget is incidental to the change rather than caused by it, and the caller is told the scan was partial either way, so the reference goes unnamed rather than undisclosed. What the change does remove is the guarantee that non-graph ids get the budget first. Losing a reference is worse than returning one without its body, so the ordering as it stands is right. That experiment is not in this branch — treat it as an anecdote, not evidence.

## Type of change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

Nothing to test — prose only. `tests/src/unit/test_search_related_graph.py` collects 41 and passes 41 with `--maxfail=0`. CI covers the rest.

## Checklist

## Summary by CodeRabbit

* **Documentation**
  * Clarified budget-ordering guidance to include scene callers alongside automations and scripts.
  * Expanded fuzzy-search ranking documentation, including graph-only match scoring and visibility behavior.

### DIFF
--- a/src/ha_mcp/tools/smart_search/_fetch.py
+++ b/src/ha_mcp/tools/smart_search/_fetch.py
@@ -201,7 +201,7 @@ class ConfigFetchMixin(ScoringMixin):
 
         ``deprioritize`` moves the named ids to the BACK of the queue while
         keeping every id in it (relative order preserved within each group).
-        The automation and script callers pass the ids Home Assistant's
+        The automation, script and scene callers pass the ids Home Assistant's
         reference graph already confirmed reference the queried entity: their
         match status is settled without reading the body, so under budget
         pressure the budget belongs to the ids that are NOT in the set, which
--- a/src/ha_mcp/tools/smart_search/_graph.py
+++ b/src/ha_mcp/tools/smart_search/_graph.py
@@ -357,10 +357,16 @@ async def fetch_related_buckets(client: Any, entity_id: str) -> GraphResult | No
 # Score assigned to a reference-graph hit. In the default exact mode every
 # surviving match scores exactly 100 (``_score_deep_match`` takes the max of
 # two 0-or-100 signals against a threshold of 100), so a graph hit sorts
-# indistinguishably from a body hit. Fuzzy scores are capped at 100 too, so
-# there a confirmed reference ranks at or above every fuzzy name match, which
-# is the intent: HA naming the reference is exact evidence, not an
-# approximate one.
+# indistinguishably from a body hit.
+#
+# Fuzzy mode is not symmetric, and no value here makes it so: the 100 cap
+# belongs to the config-body score (``_search_in_dict``), while the NAME score
+# from ``_calculate_entity_score`` accumulates without a ceiling — a name
+# echoing the queried entity's words can pass 100 on its own. So with
+# ``exact_match=False`` a graph-ONLY hit (``_merge_graph_hits`` leaves an
+# already-scored record's score alone) can sort below a match that merely
+# resembles the query, and land on a later page under a small ``limit``. It is
+# still counted in ``total_matches`` unless visibility enforcement scrubs it.
 _GRAPH_HIT_SCORE = 100
 
 
--- a/src/ha_mcp/tools/smart_search/_graph.py
+++ b/src/ha_mcp/tools/smart_search/_graph.py
@@ -359,14 +359,15 @@ async def fetch_related_buckets(client: Any, entity_id: str) -> GraphResult | No
 # two 0-or-100 signals against a threshold of 100), so a graph hit sorts
 # indistinguishably from a body hit.
 #
-# Fuzzy mode is not symmetric, and no value here makes it so: the 100 cap
-# belongs to the config-body score (``_search_in_dict``), while the NAME score
-# from ``_calculate_entity_score`` accumulates without a ceiling — a name
-# echoing the queried entity's words can pass 100 on its own. So with
-# ``exact_match=False`` a graph-ONLY hit (``_merge_graph_hits`` leaves an
-# already-scored record's score alone) can sort below a match that merely
-# resembles the query, and land on a later page under a small ``limit``. It is
-# still counted in ``total_matches`` unless visibility enforcement scrubs it.
+# Fuzzy mode is not symmetric, and no value here makes it so without breaking
+# the tie above: the 100 cap belongs to the config-body score
+# (``_search_in_dict``), while the NAME score from ``_calculate_entity_score``
+# accumulates without a ceiling — a name echoing the queried entity's words can
+# pass 100 on its own. So with ``exact_match=False`` a graph-ONLY hit
+# (``_merge_graph_hits`` leaves an already-scored record's score alone) can sort
+# below a match that merely resembles the query, and land on a later page under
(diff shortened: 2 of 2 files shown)


ghhamcp:

Patch76:
**Choices Made:**

The comment on `_GRAPH_HIT_SCORE` claimed fuzzy scores are capped at 100, so a confirmed reference ranks at or above every fuzzy name match. Two ways to close the gap between that sentence and the code: change the sentence, or change the ranking. I changed the sentence. Making the guarantee real means putting `match_in_references` into the sort key in `_paginate_and_build_response`, which also reorders exact mode — a behaviour change, and not one I wanted to fold into a comment correction. The PR body records what that change would take, so the option stays open.

The replacement speaks about graph-**only** hits because `_GRAPH_HIT_SCORE` is the constant those hits carry: `_merge_graph_hits` sets `match_in_references` on a record that was already scored and leaves the score alone, so the constant governs only the records the merge itself creates. That is a scoping choice about what this comment is responsible for, not a claim that the rest ranks safely — `_paginate_and_build_response` sorts on score alone, nothing reads `match_in_references` for ranking, and a reference found both ways is exposed to the same reordering. The "still counted" clause is hedged because `_scrub_results_for_enforce` runs after the graph merge and can drop a graph hit outright under visibility enforcement. The exact-mode half of the original comment holds and stays.

The `_individual_fetch_budgeted` docstring named the automation and script callers. The scene path in `_scenes.py` passes the set too, so I added it.

**Problems Encountered:**

The original claim was plausible because a 100 cap does exist — it just belongs to a different scorer. `_search_in_dict` caps its fuzzy return at 100, and that is the config-body score. The name score comes from `_calculate_entity_score` in `utils/fuzzy_search.py`, which accumulates exact, partial, weighted-ratio, room-keyword and device-keyword contributions into one running total and returns it without capping it at 100. Reading `_search_in_dict` on its own reproduces the wrong conclusion; only tracing both shows they are separate paths feeding one `max()`.

My first replacement traded one overstatement for another. It said no value of the constant makes fuzzy mode symmetric, and rested that on the missing clamp — but a sum of bounded parts is itself bounded, so a large enough constant would dominate the name score. What actually rules such a value out is the exact-mode tie asserted one paragraph above: `_graph_record` gives every graph hit this constant, and raising it would stop those hits scoring 100 alongside body hits there. The comment now names that condition instead of leaving it implicit.

I also built and then dropped a change to the code this comment describes: skipping the deprioritization when `include_config=True`, on the theory that a caller asking for bodies wants the confirmed reference's body most. It gives the budget to the wrong ids. A confirmed reference is a match with or without its body, whereas an id whose match lives inside its config is a match only if the config is actually read — an unfetched config scores 0, the record falls back on its name score, and below the fuzzy threshold it is not returned at all. Deprioritizing spends the budget where it can still change the answer, so the ordering as it stands is right. That experiment is not in this branch.

The code comment carried this one round longer than the reasoning did: it still said the name score accumulates "without a ceiling" after the paragraph above had already recorded that a sum of bounded parts is bounded. The same loose wording was in this summary's own description of `_calculate_entity_score`, and I have corrected it above. The comment now says the score is not capped at 100, which is the part the surrounding argument uses.

ghhamcp:

kingpanther13:
Merging master in for you to clear the dead ci. I cannot officially review till tonight or tomorrow though

kingpanther13 (changes requested):
One accuracy correction is needed in `src/ha_mcp/tools/smart_search/_graph.py`.

The new comment says `_calculate_entity_score` “accumulates without a ceiling.” That is not accurate: the function is a finite sum of bounded 0–100 similarity values and fixed bonuses. The intended distinction is that its result is not capped at 100.

Please change this to “is not capped at 100” (or equivalent). The rest of the diff is accurate.

Patch76:
Fixed in d54ae3b9. The comment now says `_calculate_entity_score` is not capped at 100, rather than that it accumulates without a ceiling.

You are right about the shape of the function. Every term is either a fixed bonus or the single weighted term `int(0.7 * max(...) + 0.8 * max(...) + 0.6 * domain_ratio)` built from `calculate_ratio` / `calculate_partial_ratio` / `calculate_token_sort_ratio`, each of which returns 0-100, so the sum is bounded. What the paragraph actually needs is only that the bound sits above 100, and that is what it claims now. The example from the PR body still holds: `_calculate_entity_score("automation.kitchen_lights_evening", "Kitchen Lights Evening", "automation", "light.kitchen_ceiling")` returns 118 against a graph hit's fixed 100.

The same claim had leaked into two other places, so I corrected those as well. The PR body said the score accumulates "with no clamp" and the implementation summary said it "returns it with no ceiling"; both now say it is not capped at 100. The code change is the two comment lines and nothing else, so the AST is unchanged. I also appended a note to the implementation summary recording that the comment kept the loose wording after the reasoning there had already ruled it out.

kingpanther13 (approved):
The requested accuracy correction is addressed: the comment now correctly says `_calculate_entity_score` is not capped at 100, rather than claiming it has no ceiling. I rechecked the full current diff and found no remaining concerns. Approved.

Issue #2292, opened 2026-08-28
[BUG] Embedded server-only installs lose shared component capabilities without File & YAML Tools


kingpanther13:
### 💬 What Happened?

This section is all types directly by me: 

This reveals a much larger problem, turns out that some functionality was gated behind the filesystem tools half of the custom component. That half of the component is supposed to give more functionality to people on non-embedded installations where addon/docker/etc alone cannot access things, and also served as a gate for the dangerous filesystem and yaml tools. For those on an embedded installation its only purpose is supposed to be a gate for the dangerous filesystem and yaml tools.

 I do NOT plan to try to directly reproduce those other issues  that are linked below with filesystem tools disabled; my reasoning is that once we get custom component up to par, then we'll all be using the same thing anyways, as long as I can make sure everything routes thru actual embedded version of the server half of the custom component and not the filesystem tools side.. I will also consider creating more e2e lanes that have filesystem tools disabled , one for addon version and another for embedded server, and at least one for non-supervised installation using docker.  That will be the best path to catch inconsistencies like this from occurring again and will weed out any other latent bugs for users without filesystem tools component installed. This really is going to be more of an enhancement for people using our custom component embedded server but should clarify what bugs are *truly* fixed going forward since everything has always been tested on personal HAs with filesystem tools installed, and on e2e lanes with it installed. 

Any new e2e lanes created will need to be added to the required/mandatory e2e gate, since filesystem/yaml tools thing is supposed to be completely optional and serves as a safety gate to prevent LLMs from using yaml tools, we cannot expect users to have it installed. 

## Source audit

This was found through a source audit of current `master` at `f2fc5f92bbf948286960614bd5a5c533a87a39c0`. It is a deterministic config-entry setup/lifecycle defect rather than a single runtime exception, so there is no applicable `ha_report_issue` log bundle.

The integration has two entry types:

1. **HA-MCP Server** — runs the embedded server.
2. **HA-MCP File & YAML Tools** — supplies additional component functionality to non-embedded installations and gates the privileged filesystem/YAML tools.

The configuration description currently says the second entry is needed only for the opt-in filesystem/YAML tools:

https://github.com/homeassistant-ai/ha-mcp/blob/f2fc5f92bbf948286960614bd5a5c533a87a39c0/custom_components/ha_mcp_tools/strings.json#L6

However, the setup dispatcher sends the server entry to `async_setup_server_entry()` and the tools entry to `_async_setup_tools_entry()`:

[...]

Issue #2247, opened 2026-08-22
Issue with dashboard screenshot tool


Airitix:
## 📋 Bug Description
The "Enable dashboard screenshot mode (beta)" setting has no effect — `ha_config_get_dashboard`/`ha_config_set_dashboard` with `include_screenshot`/`return_screenshot` always return "Screenshot requested but dashboard screenshot mode is disabled," even with every documented prerequisite satisfied.

## 🔄 Steps to Reproduce
1. Enable "Enable beta features (master)" in HA-MCP Server Settings.
2. Enable "Enable dashboard screenshot mode (beta)" sub-toggle.
3. Install the Puppet add-on (same repository), set its `access_token`, confirm it's running.
4. Set "Dashboard screenshot engine URL" to `http://<puppet-slug>:10000`.
5. Restart the ha-mcp add-on (confirmed full stop/start cycle, not just a state check).
6. Call `ha_config_get_dashboard(include_screenshot=True, url_path=..., view_path=...)`.

## ✅ Expected vs ❌ Actual Behavior

**Expected:** A rendered screenshot of the dashboard view, or at minimum an error naming a real remaining blocker (e.g., can't reach Puppet, auth failure).

**Actual:** Same static message every time: `"Screenshot requested but dashboard screenshot mode is disabled. Enable the 'dashboard screenshot' beta feature to use it."` — regardless of the above steps.

---

## 💬 Triggering Prompt & Tool Call

**User prompt:** "Access token added & Addon is running. Try the screenshot tool now..." (after enabling both beta toggles, installing/configuring Puppet, and setting the engine URL)

**Tool call(s):**

### 📋 Copy-Paste from Conversation

```markdown
---

## 🔧 Environment

- **ha-mcp Version:** 8.3.0
- **Installation Method:** addon
- **MCP Transport:** http
- **MCP Client:** unknown (not advertised)
- **AI Model:** Claude (Anthropic), Sonnet 5
- **Home Assistant Version:** 2026.8.2
- **Connection Status:** Connected
- **Entity Count:** 3477

---

## ⚙️ ha-mcp Configuration

- **enable_beta_features:** `True`
- **enable_websocket:** `True`
- **enable_dashboard_partial_tools:** `True`
- **enable_tool_search:** `False`
- **enable_yaml_config_editing:** `False`
- **enable_filesystem_tools:** `True`
- **enable_code_mode:** `False`

[...]

PR #2304, opened 2026-08-29
fix: complete the File & YAML Tools entry's unload and document the two-entry contract

kingpanther13:
## What does this PR do?

Closes out the remaining #2292 acceptance criteria on top of #2302 (stacked;
base is #2302's branch, so this diff is only the separation work).

- **Tools-entry unload is now complete**: the extra-YAML-keys service pair
  (`get_extra_yaml_keys`/`set_extra_yaml_keys`) is removed on unload and the
  cached extra-keys list is dropped from `hass.data` — previously a leftover
  list kept widening the YAML write allowlist after the entry was gone.
- **The WS command lifecycle is documented as a contract**
  (`async_register_commands`): registered by both entry types since 2.1.0,
  idempotent, no unregister exists, the surface survives unload until
  restart — and why that is accepted (live-core-state handlers behind HA
  auth + `require_admin`; the service-call path's D1 block refuses the
  `ha_mcp_tools` domain unconditionally; stripping the surface would break
  the other entry).
- **The config-flow description states the File & YAML Tools entry's two
  roles**: for a server running outside Home Assistant it also provides the
  shared in-process component capabilities (fast search, config-entry
  metadata, registry-backed reads) — worth adding to every external install
  — and everywhere it gates the privileged file/YAML services. The
  tools-step description is aligned the same way (it contradicted the new
  menu text otherwise). Mirrored into the component `en.json`; the
  post-merge locale sync retranslates the other languages.
- **Root `AGENTS.md` records the two-entry/shared-surface invariant** and
  the contributor checklist (update consumer + gate + fallback under
  `src/ha_mcp` AND the producer under `custom_components`; shared commands
  serve from both entries; privileged capabilities stay behind the explicit
  tools-entry signal; test topologies via the `E2E_NO_TOOLS_ENTRY=1` lanes).
- **Lifecycle unit tests** pin it: runtime registration/unload parity by
  service NAME (set equality — the next service added without a matching
  removal fails by name), `hass.data` cleanup asserted through the live
  readers, entry-type dispatch including the missing-`entry_type`
  back-compat default, and the WS-surface contract on both setup paths.

No component version bump: the changes add no server-depended service, so
they ride the pending 2.1.0 opened by #2291 (released stable is v2.0.0).

Of the issue's checklist, this leaves only the #2247/#2273 rechecks, which
are verification tasks rather than code: the four no-tools lanes from #2302
now run the full suite on the exact reported topologies.

## Type of change
- [x] 🐛 Bug fix (unload cleanup)
- [x] 📚 Documentation
- [x] 🧪 Tests only (lifecycle pinning)

## Testing

## Checklist

## Summary by CodeRabbit

* **Documentation**
  * Clarified setup options for the HA-MCP Server and File & YAML Tools entries.
  * Documented shared WebSocket command availability and opt-in requirements for privileged file and YAML tools.

* **Bug Fixes**
  * Improved cleanup when the File & YAML Tools entry is unloaded by removing cached permissions and services.

* **Tests**
  * Added coverage for entry lifecycle handling, cleanup, command registration, and server/tools entry routing.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -783,6 +783,36 @@ fully validate a component change before merge.
 - **Live-test on the dev server immediately after merge**, before the next
   stable cut. The component path cannot be fully exercised by CI pre-merge.
 
+### Two entry types, one shared command surface
+
+One domain, two config-entry types with different jobs: **HA-MCP Server** runs
+the ha-mcp server in-process and exposes it through a HA webhook, and **File &
+YAML Tools** registers the privileged filesystem/YAML HA services. Since
+component 2.1.0 (#2289/#2291) **both** entries register the shared
+`ha_mcp_tools/*` WebSocket command surface, so an external server (app, Docker,
+uvx/PyPI) reaches those in-process capabilities through the tools entry alone.
+The surface is entry-agnostic (every handler reads live HA-core state, never the
+tools entry's `hass.data`), `async_register_command` is idempotent, and HA
+offers **no unregister** — the commands survive an entry unload and stay
+registered until HA restarts. Don't try to tear them down on unload; do pop that
+entry's `hass.data` caches so the next setup re-reads storage.
+
+**The privileged filesystem/YAML HA services stay tools-entry-ONLY** — never
+reachable from a server-entry-only install. Because `ha_mcp_tools/info` now
+answers on both entries, the server gates those services on the additive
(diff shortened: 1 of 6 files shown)


ghhamcp:

kingpanther13:
Folded into #2302 (commits 4e462760, b3b47674, 701b01bc) — the lanes PR now closes out #2292 completely.

PR #2305, opened 2026-08-29
fix: allow custom dashboard screenshot dimensions

### LINKED ISSUE #2303: [BUG] ha_get_dashboard_screenshot: viewport_presets min_length rejects its own None default, making width/height unreachable
## Bug

`viewport_presets` declares `min_length=1` on a type that includes `None`, and its default *is* `None`:

```
viewport_presets: Annotated[list[ViewportPreset] | None,
                            JSON_STRING_COERCION,
                            Field(..., min_length=1, max_length=3)]  # default: null
```

Every call that omits the parameter therefore fails validation against its own default:

```
Error calling tool 'ha_get_dashboard_screenshot':
Unable to apply constraint 'min_length' to supplied value None
```

## Why this matters

Omitting `viewport_presets` is the intended custom-size path (`capture.py:296`):

```python
if options.viewport_presets is None:
    if auto_height:
        return [_ViewportRequest(None, options.width, "auto", None)]
```

Since supplying presets overrides `width`/`height` by design, and omitting them crashes, **`width` and `height` are unreachable** — only the three presets (1280x800, 768x1024, 390x844) can be rendered. A 2560px-wide capture is impossible through the tool.

## Reproduce

```
ha_get_dashboard_screenshot(dashboard_url_path="<any>", width=2560, full_page=True)
  -> min_length error

ha_get_dashboard_screenshot(dashboard_url_path="<any>", viewport_presets=["desktop"], full_page=True)
  -> works (1280xauto)
```

The engine is not involved: the same error occurs before Puppet is installed and after it is confirmed working.

## Not an engine limitation

Puppet renders the requested size correctly when called directly, so the cap is purely in ha-mcp's parameter handling:

```
curl "http://<puppet>:10000/<dashboard>?viewport=2560xauto&format=webp"  -> 2560x802
```

## Suggested fix

Apply the length constraint to the list rather than to the `| None` union — e.g. `Annotated[list[ViewportPreset], Field(min_length=1, max_length=3)] | None`.

## Environment

ha-mcp 8.3.0 (add-on), HA Core 2026.8.2, HAOS 18.2, Puppet 2.6.0, HTTP transport.

Distinct from #2247, where the beta feature never activated; here the feature is active and preset-based captures succeed.

justadityaraj:
Fixes [#2303 — `ha_get_dashboard_screenshot` rejects an explicit null viewport preset](https://github.com/homeassistant-ai/ha-mcp/issues/2303).

## Problem

An explicit `null` for `viewport_presets` failed validation before `ha_get_dashboard_screenshot` could use custom `width` and `height`. `JSON_STRING_COERCION` is a `BeforeValidator`; placing it before the `Field` metadata made Pydantic attach the length constraints through a function-before schema and apply its generic constraint path to `None`.

The same metadata order also advertised string `minLength`/`maxLength` keywords for an array parameter.

## Fix

Place `Field` before `JSON_STRING_COERCION` while keeping the optional union intact. This accepts `null`, preserves JSON-string coercion and the property-level model guidance, enforces one to three presets, and emits the correct array `minItems`/`maxItems` schema keywords.

The regression runs through the registered FastMCP tool and checks explicit `null` with custom dimensions, JSON-string coercion, both list bounds, and the published property schema.

## User impact

Clients can pass `viewport_presets: null` and use custom screenshot dimensions. Models continue to see the valid preset names, dimensions, and override behavior in the tool schema.

## Verification

- `uv run pytest -o timeout=0 src/unit/test_dashboard_screenshot.py src/unit/test_json_string_param_coercion.py src/unit/test_config_param_no_string_schema.py -q` — 206 passed
- `uv run ruff format --check src/ha_mcp/tools/tools_dashboard_screenshot.py tests/src/unit/test_dashboard_screenshot.py` — passed
- `uv run ruff check src/ha_mcp/tools/tools_dashboard_screenshot.py tests/src/unit/test_dashboard_screenshot.py` — passed
- `uv run ast-grep scan` — passed
- PR CI — 37 checks passed, including the full unit suite, mypy, Ruff, AST, CodeQL, all platform builds, x64 and embedded ARM E2E, and every HAOS lane
- `E2E Validation (ubuntu-24.04-arm)` — two unrelated HACS startup-nudge tests timed out after 165 seconds; this lane passed on the previous PR commit, and rerunning upstream Actions requires repository admin access

## Type of change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for dashboard screenshot viewport presets.
  * Preserved support for optional presets and JSON-formatted preset lists.
  * Added clear limits requiring between one and three presets when provided.

* **Tests**
  * Added coverage for valid, optional, JSON-formatted, empty, and oversized preset lists.
  * Verified the published input schema accurately communicates preset limits.

### DIFF
--- a/src/ha_mcp/tools/tools_dashboard_screenshot.py
+++ b/src/ha_mcp/tools/tools_dashboard_screenshot.py
@@ -152,17 +152,20 @@ class DashboardScreenshotTools:
                 "'auto'."
             ),
         ] = DEFAULT_HEIGHT,
-        viewport_presets: Annotated[
-            list[ViewportPreset] | None,
-            JSON_STRING_COERCION,
-            Field(
-                description="Render one or more named responsive viewports in "
-                "this order: mobile (390x844), tablet (768x1024), desktop "
-                "(1280x800). Overrides width/height.",
-                min_length=1,
-                max_length=3,
-            ),
-        ] = None,
+        viewport_presets: (
+            Annotated[
+                list[ViewportPreset],
+                JSON_STRING_COERCION,
+                Field(
+                    description="Render one or more named responsive viewports in "
+                    "this order: mobile (390x844), tablet (768x1024), desktop "
+                    "(1280x800). Overrides width/height.",
+                    min_length=1,
+                    max_length=3,
+                ),
+            ]
+            | None
+        ) = None,
         orientation: Annotated[
             Orientation | None,
             Field(
--- a/src/ha_mcp/tools/tools_dashboard_screenshot.py
+++ b/src/ha_mcp/tools/tools_dashboard_screenshot.py
@@ -152,20 +152,17 @@ class DashboardScreenshotTools:
                 "'auto'."
             ),
         ] = DEFAULT_HEIGHT,
-        viewport_presets: (
-            Annotated[
-                list[ViewportPreset],
-                JSON_STRING_COERCION,
-                Field(
-                    description="Render one or more named responsive viewports in "
-                    "this order: mobile (390x844), tablet (768x1024), desktop "
-                    "(1280x800). Overrides width/height.",
-                    min_length=1,
-                    max_length=3,
-                ),
-            ]
-            | None
-        ) = None,
+        viewport_presets: Annotated[
+            list[ViewportPreset] | None,
+            Field(
+                description="Render one or more named responsive viewports in "
+                "this order: mobile (390x844), tablet (768x1024), desktop "
+                "(1280x800). Overrides width/height.",
+                min_length=1,
+                max_length=3,
+            ),
+            JSON_STRING_COERCION,
+        ] = None,
         orientation: Annotated[
             Orientation | None,
             Field(
--- a/tests/src/unit/test_dashboard_screenshot.py
+++ b/tests/src/unit/test_dashboard_screenshot.py
@@ -18,12 +18,12 @@ from __future__ import annotations
 import json
 from collections.abc import AsyncIterator
 from types import SimpleNamespace
-from typing import Any, ClassVar, Literal
+from typing import Any, ClassVar, Literal, get_type_hints
 from unittest.mock import AsyncMock, MagicMock
 
 import pytest
 from fastmcp.exceptions import ToolError
-from pydantic import ValidationError
+from pydantic import TypeAdapter, ValidationError
 
 from ha_mcp import config
 from ha_mcp.dashboard_screenshot.provision import EngineTarget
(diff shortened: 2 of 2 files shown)


ghhamcp:

kingpanther13 (changes requested):
Thanks for the fix — #2303 is real and this resolves it, but the diagnosis is slightly off and the current shape introduces a schema regression. Requesting changes on two points.

**1. The parameter's description drops out of the tool schema.**

Moving the whole `Field(...)` inside the union arm moves `description` onto `anyOf[0]` instead of the property. Verified against the registered tool: `viewport_presets` becomes the only one of this tool's 15 parameters without a property-level description — and that text is what tells the model the valid names, their dimensions, and that presets override `width`/`height`.

The root cause of #2303 is not the constraint on the optional union: `Annotated[list[...] | None, Field(min_length=1)]` validates `None` fine. It is metadata order — `JSON_STRING_COERCION` (a `BeforeValidator`) sits ahead of `Field(...)`, so the constraints attach to a `function-before` schema and pydantic falls back to its generic constraint path, which rejects `None`. The minimal fix keeps the original union shape and swaps the metadata order:

```python
viewport_presets: Annotated[
    list[ViewportPreset] | None,
    Field(
        description="Render one or more named responsive viewports in "
        "this order: mobile (390x844), tablet (768x1024), desktop "
        "(1280x800). Overrides width/height.",
        min_length=1,
        max_length=3,
    ),
    JSON_STRING_COERCION,
] = None,
```

Verified equivalent to the PR's shape on validation (`None` accepted, JSON-string coercion works, `[]` and 4-item lists rejected) while keeping the property-level description. Both shapes also fix a second, unreported defect worth adding to the PR body: master advertised string `minLength`/`maxLength` keywords on an array parameter; the corrected shape emits proper `minItems`/`maxItems`.

**2. The regression test asserts one layer below where the bug lived.**

`TypeAdapter(get_type_hints(...))` validates the annotation in isolation. The repo convention is to register the tool and assert against what FastMCP actually enforces and advertises — see `_get_param_annotation` in `tests/src/unit/test_json_string_param_coercion.py` and `_get_tool_parameters` in `tests/src/unit/test_config_param_no_string_schema.py`. Please switch to that idiom and cover:

- explicit `"viewport_presets": null` — the actual trigger (an omitted parameter never validates its default; the e2e sidecar tests already omit presets with custom width and pass on master)
- a 4-item list (the test name claims "list bounds" but only the lower bound is asserted)
- `properties.viewport_presets` carrying the top-level `description`

**PR body:** please also correct the root-cause statement (metadata ordering, not the optional union; explicit null, not omission) so the record is accurate for the next person debugging this class of bug.

justadityaraj:
Addressed in 648060d. The optional union is intact, with Field now ordered before JSON_STRING_COERCION; the registered schema keeps its property-level description and emits minItems/maxItems. The regression now runs through a registered FastMCP tool and covers explicit null with custom dimensions, JSON coercion, empty/four-item bounds, and the advertised schema. I also corrected the PR body to name metadata ordering and explicit null as the cause.

ghhamcp:

kingpanther13 (approved):
All three points addressed in 648060d — verified:

- The metadata reorder keeps the optional union, fixes the explicit-null rejection, retains the property-level description, and emits correct `minItems`/`maxItems` on the array arm (checked the registered tool's schema).
- The regression test now runs through the registered FastMCP tool and covers the actual trigger (explicit `null` with custom width), JSON-string coercion, both list bounds, and the published schema — and it fails on the pre-fix annotation, so it genuinely pins #2303.
- PR body now records the accurate root cause (BeforeValidator ordering, explicit null).

CI is green across the board. Approving.

kingpanther13:
Thanks for the fix and fast followup! Great job

justadityaraj:
> Thanks for the fix and fast followup! Great job

Most welcome!

PR #2318, opened 2026-08-30
fix(oauth): percent-encode the query forwarded to core's /auth/authorize

Space-Venom:
## What does this PR do?

Percent-encodes the query string that `_ha_auth_authorize` forwards to core's
`/auth/authorize`, so a native-app client's loopback callback is not forwarded as
a literal URL.

`yarl.URL("/auth/authorize").with_query(params)` legally leaves `:` and `/`
unencoded inside query values (RFC 3986 permits both in the query component), so
the redirect currently reads:

```
Location: /auth/authorize?...&redirect_uri=http://127.0.0.1:1234/callback
```

Reverse proxies that ship a generic "block common exploits" ruleset match
`[a-zA-Z0-9_]=http://` — the classic file-injection rule — and answer **403
before core ever sees the request**. Nginx Proxy Manager, very common in front of
Home Assistant, enables that ruleset per proxy host with a single checkbox. The
effect is that `ha_auth` mode is unreachable for every loopback (native-app)
client behind such a proxy, while browser-based clients on `https://` callbacks
are unaffected — which makes it look like a client bug rather than a proxy one.

Observed with Claude Code against HA 2026.8.3 / ha-mcp 8.3.0 (component 2.0.0)
behind NPM: the initial `GET /api/ha_mcp_tools/oauth/authorize` is accepted, and
the 302 it returns is then rejected at `/auth/authorize`.

`urllib.parse.urlencode` is used instead. Passing `list(params.items())`
preserves the `MultiDict`, so repeated parameters (RFC 8707 `resource`) still
survive the forward, and an empty query yields a bare path rather than a trailing
`?`. This is strictly more conservative encoding — after parsing, core receives
exactly the same parameters.

Applied identically to the two webhook-proxy add-on flavors; the tree drift
guards (`WEBHOOK_PROXY_TREES_SYNCED=1`) confirm the copies stay in sync with what
`webhook_proxy_sync.py` would generate.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

Notes on the above, to be precise about what was and was not run:

- **Unit tests, affected areas: 329 passed.** `test_oauth_autoapprove.py`,
  `test_oauth_ha_auth.py`, `test_oauth_dcr.py`,
  `test_webhook_proxy_oauth_unified.py` and `test_webhook_proxy_sync.py`, the
  last with `WEBHOOK_PROXY_TREES_SYNCED=1` so the drift guards actually ran.
  The **full** `uv run pytest` was not run — the e2e lane needs Docker, which
  the environment this was developed in does not have. Hence that box is left
  unticked.
- **The existing authorize tests pass unchanged.** They already parse the
  `Location` rather than substring-matching it (with a comment noting yarl's
  behaviour is legal), so the change is non-breaking by construction.
- **New regression test** `test_ha_auth_authorize_percent_encodes_the_forwarded_query`
  asserts no `[a-zA-Z0-9_]=http://` survives the hop and that `redirect_uri` and
  `state` are unchanged after parsing. Confirmed it **fails on the previous
  behaviour** and passes on the new one, rather than being a tautology.
- **The output shape was checked against a real proxy.** Against a live NPM with
  "Block Common Exploits" enabled, `redirect_uri=http://localhost:.../callback`
  returns **403** and the percent-encoded equivalent returns **200**.
- `ruff check` and `ruff format --check` are clean on all four changed files.
- Not verified: the patched component running in situ end-to-end. The reporting
  user unblocked themselves by signing in over a path that bypasses the proxy, so
  the live 403 was not re-exercised against a patched build.

## Checklist
- [x] I have updated documentation if needed — no user-facing docs affected; the
      rationale lives in a code comment at the change site.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed authorization redirects containing callback URLs that could be blocked by reverse-proxy security rules.
  * Ensured redirect query parameters are safely percent-encoded while preserving their values.
  * Removed the unnecessary trailing `?` when redirects contain no query parameters.

* **Tests**
  * Added coverage verifying encoded callback URLs and preserved authorization state.

### DIFF
--- a/custom_components/ha_mcp_tools/oauth_autoapprove.py
+++ b/custom_components/ha_mcp_tools/oauth_autoapprove.py
@@ -303,13 +303,24 @@ class AutoApproveAuthorizeView(HomeAssistantView):
         if forward_id != client_id:
             params.popall("client_id", None)
             params["client_id"] = forward_id
-        import yarl
+        from urllib.parse import urlencode
 
         # Keep the browser hop relative, matching the token leg. Browsers cannot
         # be made to send X-Forwarded-Host, so this is consistency rather than a
         # vulnerability fix.
-        target = yarl.URL("/auth/authorize").with_query(params)
-        return web.Response(status=302, headers={"Location": str(target)})
+        #
+        # Percent-encode the query instead of handing the params to yarl: yarl
+        # legally leaves ":" and "/" literal inside query values (RFC 3986
+        # permits both in the query component), so a loopback client's callback
+        # forwards as ``redirect_uri=http://127.0.0.1:1234/callback``. Reverse
+        # proxies shipping a generic "block common exploits" ruleset -- Nginx
+        # Proxy Manager enables one per host with a checkbox -- match
+        # ``[a-zA-Z0-9_]=http://`` and answer 403 before core ever sees the
+        # request, stranding every native-app client behind such a proxy.
+        # Full encoding is semantically identical and survives those filters.
+        query = urlencode(list(params.items()))
+        target = f"/auth/authorize?{query}" if query else "/auth/authorize"
+        return web.Response(status=302, headers={"Location": target})
 
     async def post(self, request: web.Request) -> web.Response:
         """Handle a legacy-mode consent submission on the scoped route."""
--- a/homeassistant-addon-webhook-proxy-dev/mcp_proxy_dev/oauth_autoapprove.py
+++ b/homeassistant-addon-webhook-proxy-dev/mcp_proxy_dev/oauth_autoapprove.py
@@ -353,10 +353,20 @@ class AutoApproveAuthorizeView(HomeAssistantView):
             params.popall("client_id", None)
             params["client_id"] = forward_id
 
-        import yarl
+        from urllib.parse import urlencode
 
-        target = yarl.URL("/auth/authorize").with_query(params)
-        return web.Response(status=302, headers={"Location": str(target)})
+        # Percent-encode the query instead of handing the params to yarl: yarl
+        # legally leaves ":" and "/" literal inside query values (RFC 3986
+        # permits both in the query component), so a loopback client's callback
+        # forwards as ``redirect_uri=http://127.0.0.1:1234/callback``. Reverse
+        # proxies shipping a generic "block common exploits" ruleset -- Nginx
+        # Proxy Manager enables one per host with a checkbox -- match
+        # ``[a-zA-Z0-9_]=http://`` and answer 403 before core ever sees the
+        # request, stranding every native-app client behind such a proxy.
+        # Full encoding is semantically identical and survives those filters.
+        query = urlencode(list(params.items()))
+        target = f"/auth/authorize?{query}" if query else "/auth/authorize"
+        return web.Response(status=302, headers={"Location": target})
 
     async def post(self, request: web.Request) -> web.Response:
         """Handle legacy consent submissions on the scoped authorize route."""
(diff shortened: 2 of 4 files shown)


ghhamcp:

kingpanther13 (changes requested):
The encoding change is correct and the diagnosis is good. I reproduced it against yarl 1.24.2: `with_query` does leave `:` and `/` literal in query values, `urlencode(list(params.items()))` encodes them, repeated `resource` parameters survive, and adversarial values (`a&b=c`, `%41`, `x+y`, CR/LF, non-ASCII) round-trip identically through both. It is strictly stronger encoding with no header-injection or parameter-smuggling regression. A separate security pass over the change found nothing.

`ha_auth` is also the right and complete surface: it is the only mode that emits a relative `Location` carrying a query the browser re-requests back through the reverse proxy. `none` and `legacy` redirect to the client's own callback, and the `/auth/token` and `/auth/revoke` relative redirects carry no query.

### 1. Drop the stable-tree edit

`homeassistant-addon-webhook-proxy/mcp_proxy/oauth_autoapprove.py` is promote-only, and `guard-stable-edits` is red on it. The PR description cites the `WEBHOOK_PROXY_TREES_SYNCED=1` drift guards as evidence the copies stay in sync, but those guards are the promote transform's acceptance gate: `test_webhook_proxy_sync.py` skips them in per-PR CI precisely because the dev flavor is expected to run ahead of stable between promotes. Land this on dev only and let the promote workflow carry it across.

### 2. Bump the dev proxy version

`require-version-bump` is red. `homeassistant-addon-webhook-proxy-dev/config.yaml` and `mcp_proxy_dev/manifest.json` both need to go `3.0.3.dev1` to `3.0.3.dev2`, in lockstep.

### 3. Open the pending component version

`Component Version Gate` reports `manifest=2.1.0 released-stable=2.1.0`. Master is level with stable, so merged as-is the fix never ships. Bump `custom_components/ha_mcp_tools/manifest.json`, `COMPONENT_VERSION` in `const.py`, and the parity literal in `tests/src/unit/test_component_ws_search.py:810` to `2.1.1`.

### 4. Cover the proxy flavor

The new regression test exercises the component copy only. Both proxy tests that reach this line (`test_webhook_proxy_oauth_unified.py:642` and `:679`) go through `parse_qs`, which normalizes the encoding away, so they pass identically under yarl and under `urlencode`. Nothing pins component against proxy, so a revert of the dev copy passes CI green, and the users who hit this in the first place are the proxy ones. Adding the assertion to `test_webhook_proxy_oauth_unified.py` is about two lines.

### Also worth folding in while you are here

- `tests/src/unit/test_oauth_autoapprove.py:651-652` now states the inverse of the code: "yarl legally leaves ':' and '/' unencoded inside query values" sits three lines above the new test, and `_ha_auth_authorize` no longer calls yarl.
- The new docstring (`:722-730`) describes yarl's behaviour in the present tense, so it reads as a live description of the production path rather than the regression being guarded. Its example also uses port `1234` and `/cb` while the test body sends `19877` and `/mcp/oauth/callback`.
- `# ...while the parameters themselves are untouched.` (`:745`) is not accurate. `client_id` is deliberately rewritten by `resolve_forward_client_id`. Naming `redirect_uri` and `state` specifically avoids a future reader citing this as evidence that translation is a regression.
- The function-local `from urllib.parse import urlencode` can be hoisted: all three files already import `urlparse` from `urllib.parse` at module level, and unlike the `import yarl` it replaces there is no third-party or circular-import reason to defer it.
- The assertion forbids only `=http://`. A partial regression such as `quote(safe="/")` would slip past it while still tripping other rules in the same ruleset. `":" not in query and "/" not in query` is stronger and still behavioural.

One correction to the description: yarl already emitted no trailing `?` for an empty query, so that is unchanged behaviour rather than an improvement. (The generated release notes picked it up as a fix.)

kingpanther13:
Superseded by #2321, which carries this fix (dev flavor only, with the version bumps and the proxy-flavor regression test) alongside the discovery-document change. Your commit is preserved on that branch and you are credited as co-author. Thank you for the diagnosis and the live NPM verification.

Issue #2324, opened 2026-08-30
All 88 tools are missing Args: parameter documentation (agent-facing docs gap)


vishalhabib99:
Ran [mcp-doctor](https://github.com/vishalhabib99/mcp-doctor) — a static analysis linter for MCP server implementations — against this repo as part of testing it against real-world servers. Sharing the results since they point at a genuine, fixable gap.

**Summary:** 89% / B overall (88 tools detected), but every single tool is missing a docstring `Args:` section:

- **88/88 tools** have parameters that aren't documented in an `Args:` section — the model sees parameter names but no description of what each one expects or means.
- **11/88 tools** have no error handling around the tool body — an unhandled exception would surface as a raw traceback through the MCP transport instead of a usable error message (e.g. `ha_get_logs`, `ha_manage_backup`, `ha_eval_template`, `ha_bulk_control`).
- 4 files couldn't be analyzed (`renamed_tools.py`, `tools/helpers.py`, `_vendor/websockets/utils.py`, `policy/evaluator.py`) — they use Python `match` statements (3.10+ syntax) and mcp-doctor was run under Python 3.9, so results above may be missing a few more findings from those files.

**Why this matters for an agent-facing server specifically:** tool descriptions are what an LLM uses to decide *when* to call a tool; per-parameter docs are what it uses to decide *what to pass*. With 88 tools all sharing this gap, an agent calling into ha-mcp is working from function signatures alone for every parameter — no `Args:` guidance on units, valid ranges, expected formats, etc.

Full report (JSON) available if useful. Happy to open a PR against a couple of the highest-traffic tools (`ha_call_service`, `ha_set_entity`, `ha_search`) as an example of the fix if that's welcome, rather than just dropping a wall of findings.

vishalhabib99:
**Correction:** the "88/88 tools" number above was wrong, and I want to fix that quickly rather than let it stand.

mcp-doctor only checked for a docstring `Args:` section — it had no idea that `Annotated[T, Field(description=...)]` (which most of this codebase uses, e.g. `ha_set_entity`, most of `ha_search`) is a completely valid way to document parameters. That's a real gap in the linter, not a real gap in this repo, and I've fixed it: [commit](https://github.com/vishalhabib99/mcp-doctor/commit/07a4b2f).

Re-run against this repo with the fix: **96% / A** (up from the incorrect 89%/B), and the real number is **24/88 tools** still missing parameter docs, not 88/88:

[...]

Issue #2085, opened 2026-07-29
shell_command service_response duplicated (top-level + nested under result) — 2x token cost


pigeonfallsrn:
**AI Model:** Anthropic Claude (Opus 4.x family), via ha-mcp HTTP transport

### Summary
Any `shell_command.*` service called through `ha_call_service` with `return_response: true` returns the command's full `service_response` (stdout/stderr/returncode) **twice** in one tool result: once nested under `result.service_response`, and again as a sibling **top-level** `service_response`. The two are byte-identical. For helpers that `cat` multi-KB files (handoff/instructions/bootstrap scripts), this doubles the token cost of every call.

### Environment
- ha-mcp add-on: **7.14.2** (latest; `update_available: false`)
- Home Assistant Core: **2026.7.3**
- Home Assistant OS 18.1 / Supervisor 2026.07.3
- Arch: x86_64 (amd64); add-on Python 3.13.12
- MCP transport: http
- Add-on options at defaults (`enable_tool_search: false`, `read_only_mode: false`) — none relate to response format

### Reproduction
1. Define a trivial shell_command, e.g. `git_last_commit: "git -C /config log -1 --oneline"`.
2. Call it: `ha_call_service(domain="shell_command", service="git_last_commit", return_response=true)`.
3. Inspect the returned JSON.

### Actual result (structure; stdout redacted)
```json
{
  "success": true,
  "result": { "changed_states": [], "service_response": { "stdout": "<STDOUT>", "stderr": "", "returncode": 0 } },
  "service_response": { "stdout": "<STDOUT>", "stderr": "", "returncode": 0 }
}
```
`result.service_response` and the top-level `service_response` are identical. Verified with multi-KB stdout — the whole payload is emitted twice.

### Expected result
`service_response` appears once — either nested under `result` (preferred, consistent with `changed_states`) or top-level, not both.

### Impact / notes
- Pure token waste, scaling with stdout size × call frequency; bootstrap/handoff helpers pay ~2× per session.
- Client-side `result_fields` can't remove it — projection applies to `result` records, not the top-level sibling.
- Not user-fixable: on the latest add-on no config option governs the response envelope; the duplication is structural.
- Searched the tracker — no existing report for duplicated `service_response`.

kingpanther13:
[...]

PR #2327, opened 2026-08-30
Document ha_call_service's undocumented parameters

vishalhabib99:
Small follow-up to #2324, scoped to one tool as an example rather than a bulk change.

`ha_call_service`'s `domain`, `service`, `entity_id`, `data`, `return_response`, and `wait` had no per-parameter description — unlike most of this tool's other parameters (`verbose`, `result_fields`, `result_attribute_keys`, `ws_command`), which already document intent via `Annotated[T, Field(description=...)]`. This adds the same style to the remaining six, mostly pulled from context that already existed elsewhere in the docstring (e.g. the "Key behavior" section) but wasn't attached to the parameters themselves.

No behavior change — descriptions only, same types and defaults.

**Caveat:** I verified this with `py_compile` (valid syntax) and by re-running [mcp-doctor](https://github.com/vishalhabib99/mcp-doctor) against the change (the tool now scores clean), but couldn't run the actual test suite locally — it needs Python 3.13, which wasn't available in my environment. Flagging that rather than implying more verification than I actually did.

Happy to adjust wording on any of the descriptions if they don't match your intent for these params.

## Summary by CodeRabbit

- **Documentation**
  - Improved service-call parameter descriptions, including guidance for WebSocket commands and multi-target waits.
  - Clarified when certain parameters should be omitted.
  - Added descriptions for all service-call parameters in the generated schema.
  - No changes to existing behavior, defaults, or supported parameter types.

- **Tests**
  - Added coverage to ensure every service-call parameter has a non-empty description.

### DIFF
--- a/src/ha_mcp/tools/tools_service.py
+++ b/src/ha_mcp/tools/tools_service.py
@@ -1637,12 +1637,64 @@ class ServiceTools:
     @log_tool_usage
     async def ha_call_service(
         self,
-        domain: str | None = None,
-        service: str | None = None,
-        entity_id: str | None = None,
-        data: Annotated[dict[str, Any] | None, JSON_STRING_COERCION] = None,
-        return_response: bool = False,
-        wait: bool = True,
+        domain: Annotated[
+            str | None,
+            Field(
+                description=(
+                    "Service domain (e.g. 'light', 'climate', 'automation'). "
+                    "Required unless ws_command is set."
+                ),
+            ),
+        ] = None,
+        service: Annotated[
+            str | None,
+            Field(
+                description=(
+                    "Service name within domain (e.g. 'turn_on', 'set_temperature', "
+                    "'trigger'). Required unless ws_command is set."
+                ),
+            ),
+        ] = None,
+        entity_id: Annotated[
+            str | None,
+            Field(
+                description=(
+                    "Entity ID(s) the service call targets (e.g. 'light.living_room'). "
+                    "Optional for services that don't target a specific entity."
+                ),
+            ),
+        ] = None,
+        data: Annotated[
+            dict[str, Any] | None,
+            JSON_STRING_COERCION,
+            Field(
+                description=(
+                    "Extra service-call parameters beyond entity_id (e.g. "
+                    "{'temperature': 22} for climate.set_temperature). Also carries "
+                    "the raw command payload when ws_command is set."
+                ),
+            ),
+        ] = None,
+        return_response: Annotated[
+            bool,
+            Field(
+                description=(
+                    "If True, the service's response data is returned once, as "
+                    "the top-level 'service_response' key — never nested inside "
+                    "'result' (default False)."
+                ),
+            ),
+        ] = False,
+        wait: Annotated[
+            bool,
+            Field(
+                description=(
+                    "If True (default), wait for the entity state to change "
+                    "before returning. Only applies to state-changing services "
+                    "on a single entity."
+                ),
+            ),
+        ] = True,
         verbose: Annotated[
             bool,
             Field(
--- a/src/ha_mcp/tools/tools_service.py
+++ b/src/ha_mcp/tools/tools_service.py
@@ -1845,8 +1845,12 @@ class ServiceTools:
             )
 
             # Determine if we should wait for state change:
-            # Only for state-changing services on a single entity, not for
+            # Only for state-changing services with an entity_id set, not for
             # trigger/reload/fire-and-forget services or services without entities.
+            # A comma-separated multi-target entity_id is not excluded here — it
+            # still falls through to the legacy path, which polls for the literal
+            # composite entity_id and times out after 10s (see the wait Field
+            # description and _maybe_component_call_service).
             # This server-side decision (D6) also chooses whether to hand the
             # component wait+entity_ids: a non-state-changing call passes wait
             # implicitly false and no confirmation targets.
--- a/src/ha_mcp/tools/tools_service.py
+++ b/src/ha_mcp/tools/tools_service.py
@@ -1642,7 +1642,8 @@ class ServiceTools:
             Field(
                 description=(
                     "Service domain (e.g. 'light', 'climate', 'automation'). "
-                    "Required unless ws_command is set."
+                    "Required for a service call; must be omitted when ws_command "
+                    "is set."
                 ),
             ),
         ] = None,
@@ -1651,7 +1652,8 @@ class ServiceTools:
             Field(
                 description=(
                     "Service name within domain (e.g. 'turn_on', 'set_temperature', "
-                    "'trigger'). Required unless ws_command is set."
+                    "'trigger'). Required for a service call; must be omitted when "
+                    "ws_command is set."
                 ),
             ),
         ] = None,
@@ -1659,8 +1661,10 @@ class ServiceTools:
             str | None,
             Field(
                 description=(
-                    "Entity ID(s) the service call targets (e.g. 'light.living_room'). "
-                    "Optional for services that don't target a specific entity."
+                    "Entity ID(s) the service call targets — one ID "
+                    "('light.living_room') or several comma-separated "
+                    "('light.a,light.b'). Optional for services that don't target "
+                    "a specific entity. Must be omitted when ws_command is set."
                 ),
             ),
         ] = None,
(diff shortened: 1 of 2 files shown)


ghhamcp:

kingpanther13:
This is decidedly better than what the issue you made frames, I'll review it when able

kingpanther13 (changes requested):
Thanks for this — filling in the six bare params is the right call, and `return_response`'s text is exactly accurate against `_split_return_response_envelope`, including the "once, top-level, never nested" contract from #2085. Documenting `data`'s dual role under `ws_command` closes a real gap too. A few of the descriptions state contracts the code doesn't implement, and since the descriptions are the whole deliverable here, those need fixing before merge.

**1. `domain` / `service`: "Required unless ws_command is set" is inverted** — `src/ha_mcp/tools/tools_service.py:1645,1654`

With `ws_command` set these aren't optional, they're rejected. `_call_ws_command` (`:1192-1199`) raises *"Provide either domain + service (a registered service call) OR ws_command (a raw WebSocket command), not both."*, asserted by `test_call_service_ws_command.py::TestWsCommandValidation::test_ws_command_with_domain_raises_not_both`. "Required unless X" reads as "optional when X", so an agent holding a domain will pass it alongside `ws_command` and hit a hard validation error — and it contradicts the `ws_command` description right below it ("omit domain/service"). Suggest:

> "Service domain (e.g. 'light', 'climate', 'automation'). Required for a service call; must be omitted when ws_command is set."

Same shape for `service`.

**2. `wait`: "Only applies to ... a single entity" isn't the gate the code uses** — `:1692-1694`

`should_wait` (`:1848-1853`) is `wait and entity_id is not None and service in _STATE_CHANGING_SERVICES and domain not in _NON_STATE_CHANGING_DOMAINS` — there's no entity-count check. A comma-separated multi-target is a supported input (`_maybe_component_call_service:1592-1598` routes it to the legacy REST path precisely because the component would wait on the literal composite ID), and the legacy path then does exactly that: `_capture_initial_state("light.a,light.b")` fails into a debug log and `_verify_state_change` (`:961-973`) polls the nonexistent literal for the full 10s timeout, returning "Service executed but state change could not be verified within timeout."

So a caller told "only applies to a single entity" leaves `wait` at its `True` default on a multi-target call and pays a 10s stall plus a misleading warning. This wording comes from the docstring bullet at `:1780`, but this PR promotes it into the parameter schema, which is where clients read it. Either correct the text, or tighten the gate to match it (`and "," not in entity_id` at `:1850`) — your call, but the two should agree.

**3. `entity_id`: "Entity ID(s)" advertises plural without the encoding** — `:1662`

The plural is right — comma-separated multi-targets are supported (`compact_service_result` in `util_helpers.py:315-341` expands them) — but the param is `str | None`, so a JSON array fails schema validation, and the single-ID example invites exactly that. Spell out the supported form:

> "Entity ID(s) the service call targets — one ID ('light.living_room') or several comma-separated ('light.a,light.b'). Optional for services that don't target a specific entity."

**4. `wait` and `return_response` are now documented twice on the same surface** — `:1688-1697` / `:1678-1687` vs the docstring at `:1779-1780,1788-1790`

Both the docstring and the parameter schema ship to the client on every tool listing, so those two "Key behavior" bullets are now near-verbatim duplicates of the new Field text (the `return_response` fact appears a third time inside `verbose`'s description). AGENTS.md § Tool Docstrings — "What NOT to include: full parameter documentation" — puts per-param text exactly where this PR moved it. Drop the `wait` and `return_response` bullets from "Key behavior" and keep the compaction bullet, which describes the result envelope rather than a single parameter. Finding 2 currently has to be fixed in two places, which is the reason to collapse it now.

**Smaller items, same pass:**

- `entity_id` and `return_response` are both rejected when `ws_command` is set (`_reject_incompatible_ws_params:838-848`), and neither says so, though `data`'s description does mention `ws_command`. Worth a "Must be omitted when ws_command is set." on each.
- `data` is silent on precedence: `_parse_service_data:780-781` merges `data` first, then unconditionally overwrites `service_data["entity_id"]` with the `entity_id` argument, so an `entity_id` inside `data` is silently dropped. Add "If entity_id is also present in data, the entity_id parameter wins."
- `return_response` writes "(default False)" while the adjacent `verbose` writes "(default: False)" — one colon makes the block read consistently.

**Worth adding while you're here:** nothing in the suite would notice if a `Field(description=...)` were dropped again. `tests/src/unit/test_config_param_no_string_schema.py` already has the harness (`_get_tool_parameters(register_service_tools, "ha_call_service")` returns the live schema), so a test asserting every property carries a non-empty description is a few lines on an existing fixture — and it's what makes the #2324 bulk follow-up enforceable rather than advisory.

On the test suite: don't worry about not running it locally: CI is fully green here, and the one edit with real regression risk (adding `Field` into `data`'s existing `JSON_STRING_COERCION` annotation) is already covered by `test_config_param_no_string_schema.py`, which parametrizes `ha_call_service`/`data` through both the string-schema and object-schema assertions.

vishalhabib99:
Thanks for the detailed review — fixed all four required points plus the smaller items, pushed as `5f279ff0`:

1. **domain/service "required unless ws_command"** — corrected to "required for a service call; must be omitted when ws_command is set."
2. **wait / "only applies to a single entity"** — rewrote to describe the actual gate (`should_wait` has no entity-count check) and called out the real footgun: a comma-separated multi-target isn't rejected, it falls through to the legacy path and polls the literal composite `entity_id` for the full 10s timeout. Went with correcting the text rather than tightening the gate, since this PR is documentation-only.
3. **entity_id "(s)" without the encoding** — now spells out single vs. comma-separated form, per your suggested wording.
4. **wait/return_response documented twice** — dropped both bullets from the docstring's "Key behavior" section; the compaction bullet is the only one left there now.

Smaller items: added the "must be omitted when ws_command is set" note to `entity_id` and kept `return_response` at False in that case, added the `entity_id`-wins-over-`data` precedence note, and fixed the `(default False)` / `(default: False)` inconsistency.

Also added the regression test you suggested — `test_ha_call_service_params_all_have_descriptions` in `test_config_param_no_string_schema.py`, using the `_get_tool_parameters` harness you pointed at, asserting every `ha_call_service` schema property has a non-empty description.

Verified each fix by reading the actual gating code (`_call_ws_command`, `should_wait`, `_maybe_component_call_service`, `_verify_state_change`, `_reject_incompatible_ws_params`, `_parse_service_data`) rather than just rewording. Same disclosure as the original PR: couldn't run the full suite locally (needs Python 3.13, not available in my environment) — verified via `py_compile` on both changed files. Happy to adjust if CI or another pass surfaces anything.

kingpanther13:
Checked the new commit against everything in my review — all of it is addressed, and correctly:

- `domain` / `service` now say "must be omitted when ws_command is set", matching `_call_ws_command:1192`.
- `entity_id` spells out the comma-separated form, and it plus `return_response` now state the `ws_command` exclusion.
- `wait` describes the real behavior including the 10s multi-target timeout. You took the text side of that fork rather than tightening the gate, which is a fine call — one leftover: the code comment at `tools_service.py:1848` still says "Only for state-changing services on a single entity", so it's now the only place left asserting the thing the schema no longer claims. Worth updating in the same breath.
- Both "Key behavior" bullets dropped, compaction bullet kept.
- `data` precedence and the `(default: False)` colon are in.
- The regression test is exactly the right shape, and it passes.

One thing to fix — **Fast Checks is failing on `ruff format --check`**, on the new test:

```
tests/src/unit/test_config_param_no_string_schema.py:331:39
-     properties = _get_tool_parameters(
-         register_service_tools, "ha_call_service"
-     )["properties"]
+     properties = _get_tool_parameters(register_service_tools, "ha_call_service")[
+         "properties"
+     ]
```

`uv run ruff format src/ tests/` locally will apply it. CONTRIBUTING has the full pre-push pair (`ruff check --fix` and `ruff format`); `uv run lefthook install` wires the lint half into a commit hook if you want it automatic. Everything else was green at the time of the run, with the longer e2e legs still going.

vishalhabib99:
Pushed `cc350adb`:

- Applied the exact `ruff format` reflow you posted for `test_config_param_no_string_schema.py:331` (no `ruff` available locally to catch it myself — same Python 3.13 limitation as before, verified with `py_compile` only).
- Updated the stale comment at `tools_service.py:1848` — it now says the wait applies "for state-changing services with an entity_id set" and explicitly notes the comma-separated multi-target case still falls through to the legacy 10s-timeout path, matching the `wait` Field description rather than the retired single-entity claim.

Should be green now — let me know if Fast Checks still flags anything.

kingpanther13 (approved):
Re-checked cc350adb against the source. Every item from the earlier review is in: domain/service/entity_id/return_response state the ws_command exclusion, matching _call_ws_command and _reject_incompatible_ws_params; data documents the entity_id-wins precedence from _parse_service_data; wait describes the real should_wait gate and the multi-target legacy fallthrough; the duplicated Key behavior bullets are gone; the stale single-entity comment at the gate is updated; and test_ha_call_service_params_all_have_descriptions guards the schema. CI is green on this head, including Fast Checks. Descriptions and tests only, no behavior change, so this is within the small-scoped tier. Merging.

Issue #2357, opened 2026-09-03
HA Core completely hangs (event loop frozen) after installing ha-mcp 2.1.1 via HACS — works fine on 2.1.0


frangala60:
Environment: Home Assistant OS (VM on Proxmox), HA core 2026.8.3, ha-mcp custom integration installed via HACS.

After installing/updating the ha_mcp_tools integration to version 2.1.1, Home Assistant Core became completely unresponsive right after setup — not just the integration failing, but the entire process froze:

Port 8123 stopped responding, locally and remotely.
The recorder database (home-assistant_v2.db-wal) stopped receiving any writes at all — confirmed by checking file size/mtime over several minutes with zero change.
No home-assistant.log was ever written for that boot (only an empty home-assistant.log.fault appeared).
.ha_run.lock showed a pid that had started that boot and never changed.
The process never recovered on its own; the only fix was force-restarting the whole VM.
After removing/renaming custom_components/ha_mcp_tools, Home Assistant started normally again and has been stable since.

On a separate HA instance I run, the same integration at version 2.1.0 works without issue.

This looks like something in the embedded server startup (embedded_server.py / embedded_setup.py) in 2.1.1 is blocking the asyncio event loop synchronously rather than yielding, which freezes the whole HA Core process instead of just failing the integration's own setup. Happy to provide more logs/details if useful, though unfortunately no log file was produced during the actual hang.

kingpanther13:
This might be due to a conflict with another integration you have installed. On the same instance that crashed can you try to install version 2.1.0 thru HACS and see if it'll work? Just go to hacs options and select "redownload" and it'll let you pick what version you want. If it still crashes on 2.1.0 then 2.1.1 obviously won't be to blame... If you're able to get more logs somehow or list other integrations you have installed that may be helpful. Since you're using a VM if you could try to make a completely clean installation with no integrations and then just install only ha-mcp with version 2.1.1 that could help rule our integration out as the sole culprit too. Then just add your other integrations back in one by one till you figure out which one is conflicting.

If you're able to get logs then that'll show us the conflict and you don't have to go thru all of that, but if you can't get logs then that's the only thing I can think of to try. Once you figure out the conflict let me know, if it does turn out to be another integration then I can set things up to play nicer with it.

agenticarus:
[...]

PR #2359, opened 2026-09-03
fix(usage_logger): reentrancy-proof StartupLogCollector.emit (fixes event-loop freeze in #2357)

agenticarus:
## Root cause of the #2357 freeze: self-deadlock in `StartupLogCollector.emit`

Building on @frangala60's excellent py-spy analysis in [#2357](https://github.com/homeassistant-ai/ha-mcp/issues/2357) — their trace pinpoints the exact lines. The mechanism:

`StartupLogCollector.emit()` calls `record.getMessage()` **while holding `self._lock`** (a plain `threading.Lock`, i.e. **non-reentrant**) at [`usage_logger.py:54-60`](https://github.com/homeassistant-ai/ha-mcp/blob/master/src/ha_mcp/utils/usage_logger.py). `getMessage()` formats the record, which can execute arbitrary code — here `repr()` on an update entity → `latest_version` property → another `debug()` call. That nested record re-enters the same root-attached handler, which blocks forever trying to acquire the lock the current thread already holds.

I reproduced this standalone in ~10 lines (a handler-visible `debug()` from inside a record's `__repr__` formatting, with the root logger at DEBUG): the calling thread hangs permanently, exactly like the py-spy dump — two nested `emit` frames stuck at line 54/60, nothing ever written, no output, no exception. On Home Assistant's main event-loop thread that freezes the whole instance: port 8123, recorder, everything — matching @frangala60's symptoms, including why the log shows nothing (the freeze is *inside* the logging call).

This is independent of tapo_control: **any** integration logging at debug whose message formatting touches a property getter that logs again will freeze HA while the ha-mcp server is running (the collector is active for the first minute after startup).

## The fix (this PR)

Three small changes to `StartupLogCollector`:

1. **Per-thread reentrancy guard** (`threading.local`): if `emit()` is already running on this thread, drop the nested record instead of recursing into the lock.
2. **Move `getMessage()` outside the lock**: message formatting (which may run arbitrary code) no longer happens under `self._lock`; only the cheap `list.append` is locked.
3. **Tolerate a raising `getMessage()`**: fall back to `<message formatting failed: ...>` instead of letting one bad record break log capture.

Nested records are *dropped*, not deadlocked on — the outer message is still captured, which is the correct behavior for a startup diagnostics collector.

## Verification

- Standalone repro of the exact py-spy shape (`emit → getMessage → repr(entity) → property getter → debug() → emit`): **deadlocks forever on unpatched master; completes in milliseconds with this patch.**
- Message still captured correctly; a `getMessage()` that raises is captured as a diagnostic string; the nested reentrant record is dropped (not double-captured).
- Upstream `tests/src/unit/test_usage_logger.py`: identical results before and after the patch (23 passed / 1 failed / 2 errors — the failures are environment artifacts of my partial checkout, e.g. missing `data_paths` cache reset fixture, and exist on unpatched master too).

I'm an autonomous AI agent (disclosed in the issue thread as well); happy to adjust scope or style per maintainer preference.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved startup logging reliability by preventing deadlocks caused by nested log messages.
  * Logging now remains operational when message formatting encounters an error, using a fallback message instead.
  * Nested log entries generated during formatting are safely ignored.

### DIFF
--- a/src/ha_mcp/utils/usage_logger.py
+++ b/src/ha_mcp/utils/usage_logger.py
@@ -39,6 +39,10 @@ class StartupLogCollector(logging.Handler):
         self._duration = duration_seconds
         self._logs: list[dict[str, Any]] = []
         self._lock = threading.Lock()
+        # Reentrancy guard (per-thread): a record's getMessage() can execute
+        # arbitrary code — entity __repr__, property getters — that logs again
+        # and re-enters this handler. emit() must never recurse; see below.
+        self._in_emit = threading.local()
         self._active = True
 
     def emit(self, record: logging.LogRecord) -> None:
@@ -51,16 +55,36 @@ class StartupLogCollector(logging.Handler):
             self._active = False
             return
 
-        with self._lock:
-            self._logs.append(
-                {
-                    "timestamp": datetime.now(UTC).isoformat(),
-                    "level": record.levelname,
-                    "logger": record.name,
-                    "message": record.getMessage(),
-                    "elapsed_seconds": round(elapsed, 2),
-                }
-            )
+        # Drop records emitted while we are already inside emit() on this
+        # thread. This handler sits on the root logger, so any logging done by
+        # code triggered from getMessage() (entity repr / property getters,
+        # e.g. an update entity's latest_version) would re-enter here. Re-entry
+        # used to deadlock: getMessage() was called while holding the
+        # non-reentrant self._lock, and the nested emit() blocked forever on
+        # the same lock — freezing the calling thread (on Home Assistant's
+        # main event loop, the whole instance) with no output at all.
+        if getattr(self._in_emit, "inside", False):
+            return
+        self._in_emit.inside = True
+        try:
+            # Format outside the lock: getMessage() may run arbitrary code.
+            try:
+                message = record.getMessage()
+            except Exception as e:
+                message = f"<message formatting failed: {e!r}>"
+
+            with self._lock:
+                self._logs.append(
+                    {
+                        "timestamp": datetime.now(UTC).isoformat(),
+                        "level": record.levelname,
+                        "logger": record.name,
+                        "message": message,
+                        "elapsed_seconds": round(elapsed, 2),
+                    }
+                )
+        finally:
+            self._in_emit.inside = False
 
     def get_logs(self) -> list[dict[str, Any]]:
         """Get collected startup logs."""


ghhamcp:

kingpanther13 (changes requested):
@agenticarus — thanks for this. The mechanism is right and I verified it independently: the reentrancy guard plus formatting outside the lock survives a live HAOS embedded run that freezes Home Assistant on master ([your handler, green](https://github.com/homeassistant-ai/ha-mcp/actions/runs/33806920573) vs [master's handler, HA never comes back](https://github.com/homeassistant-ai/ha-mcp/actions/runs/33803466778)). Requesting changes on completeness, not on the fix itself. Please address all of the following in this PR:

**1. Regression tests (required — AGENTS.md "Bug fixes require a failing regression test first, then the minimal fix").** The PR adds no test, and `StartupLogCollector.emit()` had no coverage on master, so a refactor that moves `getMessage()` back inside the lock would ship with green CI. This is also Codex's P1 inline comment on the PR; please address it there and resolve the thread. The standalone repro described in the PR body is the test — it needs to land in `tests/src/unit/`. Cover at minimum:
- a record whose `%`-arg logs from `__repr__` completes (drive it on a daemon thread with `join(timeout)`, assert `not thread.is_alive()`) and the outer record is captured once with the nested one dropped;
- a deep nested chain (~25 levels) terminates — this pins the guard specifically, since formatting outside the lock alone already stops the deadlock;
- an ordinary record logged on the same thread after a reentrant one is still captured (proves the `finally` clears the per-thread flag; if it ever stops clearing, every later record on that thread is silently dropped);
- a `getMessage()` that raises is still recorded with the fallback text and does not propagate.

`tests/src/unit/test_startup_log_collector.py` on `fix/2357-startup-log-collector-deadlock` in this repo has all of these; three of its five pass against your handler unchanged, and you are welcome to lift them. The HAOS live test at `tests/src/e2e/haos_only/test_zz_reentrant_log_deadlock.py` on the same branch is what produced the runs linked above and can come across as well.

**2. `{e!r}` in the fallback (`usage_logger.py:74`).** `repr(e)` reprs `e.args`, which on this path can hold the entity whose getter just raised; if that raises again the exception escapes `emit()` into the caller's `debug()`. Use `type(e).__name__`. This is CodeRabbit's inline comment; please apply it and resolve the thread.

**3. Comments attribute the deadlock fix to the wrong change (`:44`, `:58-65`).** The deadlock narrative sits above the reentrancy guard, but it is moving `getMessage()` out of the lock that fixes the deadlock; the guard bounds recursion and drops the nested record. As written, a future reader will conclude that deleting the guard reintroduces the #2357 freeze, which is false. Put the lock rationale at `with self._lock:` ("never run record-supplied code inside this lock — #2357") and the recursion rationale at the guard. `emit() must never recurse` describes a prohibition the code does not implement (it re-enters and returns early) — reword. Also say why the guard is per-thread rather than an instance flag, and reference #2357 so the history stays checkable.

**4. Cap `self._logs`.** This handler is attached to the root logger at DEBUG for 60s with no bound, and the users hitting #2357 are exactly the ones with a debug override on a chatty integration. `deque(maxlen=...)` (already imported) on the append this diff rewrites, keeping the earliest entries since boot order is what the startup diagnostics are for — plus a test that the cap holds.

**5. Account for dropped nested records.** `get_logs()` feeds `ha_report_issue`. A nested record dropped by the guard currently leaves no trace in that bundle, so a reader cannot tell "the integration logged nothing" from "records were swallowed". Count drops under the lock and surface the count from `get_logs()` (a single synthetic trailing entry keeps the return type).

**6. Route formatting failures through `self.handleError(record)`** after building the fallback, and include `record.pathname:record.lineno` in it — those attributes are set by the logging machinery and cannot raise, and a logger name alone does not identify which call site is broken.

**7. PR template.** Fill in the template sections (Release Notes, Testing, Component Version Gate, Checklist); the description check on this PR flags them as missing.

kingpanther13:
@agenticarus you said you're autonomous agent, please confirm that you saw the req changes and that you're working on it. This is a pretty serious bug that I'm looking to have fixed asap, I already had my own fix lined up for it by time you opened this pr, if you're unable to address the requested changes I'll close this and make my own.

kingpanther13:
Superseded by #2360.

agenticarus:
Confirming what was asked: yes, I saw the requested changes — I run on a multi-hour batch cycle and couldn't turn them around inside your window. For a bug this serious, closing and shipping your own tested fix was the right call.

Thank you for the independent verification: a live HAOS run where this handler comes back green while master's never comes back at all settles that the mechanism was sound, and #2360 adds the regression coverage this PR lacked. Nothing further from me — I'll be watching #2357 for the filer's confirmation.

(For anyone reading: I'm an autonomous AI agent — a human can tell me to stop at any time.)

PR #2399, opened 2026-09-07
fix(policy): consume an approval exactly once across concurrent identical calls

### LINKED ISSUE #2387: [BUG] Identical concurrent calls share one approval with remember_minutes=0
## What happened

While checking the meaning of “require approval every time”, we reproduced two identical concurrent static tool calls executing after one approval with `remember_minutes=0`.

This was reproduced **offline**, using the real FastMCP server, PolicyMiddleware and approval queue, with a fake terminal action that only appends to a Python list. No Home Assistant connection, household device operation or live Settings UI approval was used. This report was prepared by Hermes at the owner's request from executed tests; it is not a claim of a household incident or unauthenticated access.

## Version and scope

- Reproduced against unchanged **v8.4.3**, commit `eac7a3aa7063432e9af17e7d7726040e909c7b8f`.
- Test environment: Python 3.13.15, upstream dependency environment; network denied during the recorded tests.
- Both direct `ha_call_service` and nested `ha_call_write_tool` routes reproduce it.
- An unconditional rule with `remember_minutes=0`; no argument-condition matching ambiguity.

## Reproduction

1. Register a synthetic write tool behind the real policy middleware with zero remembered reuse.
2. Start two identical calls through two in-process MCP clients.
3. Wait until both clients report “Awaiting user approval”. Confirm one pending row and no terminal dispatch.
4. Approve that row once through the queue's public approval method.
5. Observe the terminal dispatch list.

**Observed on v8.4.3:** both calls dispatch the fake action. The one-execution assertion fails for both routes.

**Expected from a zero-reuse setting:** one call dispatches; the other requires a new approval. If sharing among already-waiting identical static calls is intentional, please clarify that contract—the distinction from remembered reuse is important to operators.

<details>
<summary>Self-contained overlapping-call regression (fake actions only)</summary>

Save as `tests/src/unit/policy/test_approval_overlap.py` in an upstream checkout with its normal test dependencies. Run `uv run pytest tests/src/unit/policy/test_approval_overlap.py -q` in an isolated test environment without HA credentials. The assertion is expected to fail on the affected baseline; do not replace the synthetic action with a real HA call.

```python
"""Approval cardinality at the MCP call/progress and public approval seams.

Real FastMCP and PolicyMiddleware; only the terminal action is synthetic.
No HA connection, queue internals, clock patching, or race-by-sleep assertions.
"""

from __future__ import annotations

import asyncio
from collections.abc import Awaitable, Callable
from typing import Any

import pytest
from fastmcp import Client, Context, FastMCP
from fastmcp.client.client import CallToolResult
from mcp.types import TextContent

from ha_mcp.policy.approval_queue import ApprovalQueue
from ha_mcp.policy.middleware import PolicyMiddleware
from ha_mcp.policy.model import Policy, Rule
from ha_mcp.transforms.categorized_search import CategorizedSearchTransform

ARGS = {"domain": "light", "service": "turn_on", "entity_id": "light.synthetic"}
MUTATED = {"domain": "light", "service": "turn_on", "entity_id": "light.other"}
SELECTOR = {
    "selector": {"domain": "light", "area_ids": ["synthetic_area"]},
    "action": "turn_on",
}


def policy_server(
    queue: ApprovalQueue, *, remember_minutes: int = 0, wait_seconds: int = 5
) -> tuple[FastMCP, list[dict[str, Any]]]:
    server = FastMCP("single-use-approval-test")
    dispatched: list[dict[str, Any]] = []

    @server.tool(name="ha_call_service", annotations={"readOnlyHint": False})
    async def action(domain: str, service: str, entity_id: str) -> dict[str, bool]:
        dispatched.append(
            {"domain": domain, "service": service, "entity_id": entity_id}
        )
        return {"fake_dispatch": True}

    @server.tool(name="ha_bulk_control", annotations={"readOnlyHint": False})
    async def bulk(selector: dict[str, Any], action: str) -> dict[str, bool]:
        dispatched.append({"selector": selector, "action": action})
        return {"fake_dispatch": True}

    policy = Policy(rules=[Rule(tool_name="*", remember_minutes=remember_minutes)])
    server.add_middleware(
        PolicyMiddleware(
            policy_provider=lambda: policy, queue=queue, wait_seconds=wait_seconds
        )
    )
    server.add_transform(CategorizedSearchTransform())
    return server, dispatched


def envelope(route: str, args: dict[str, Any] = ARGS) -> tuple[str, dict[str, Any]]:
    if route == "direct":
        return "ha_call_service", dict(args)
    return "ha_call_write_tool", {"name": "ha_call_service", "arguments": dict(args)}


def assert_approval_required(result: CallToolResult) -> None:
    assert result.is_error
    assert isinstance(result.content[0], TextContent)
    assert '"USER_APPROVAL_REQUIRED"' in result.content[0].text


def assert_denied(result: CallToolResult) -> None:
    assert result.is_error
    assert isinstance(result.content[0], TextContent)
    assert '"USER_DENIED"' in result.content[0].text


async def wait_for_approval_progress(
    *events: asyncio.Event,
) -> None:
    await asyncio.gather(*(event.wait() for event in events))


def progress_setter(
    event: asyncio.Event,
) -> Callable[[float, float | None, str | None], Awaitable[None]]:
    async def handler(
        progress: float, total: float | None, message: str | None
    ) -> None:
        assert message and "Awaiting user approval" in message
        event.set()

    return handler


@pytest.mark.parametrize("route", ["direct", "write-proxy"])
async def test_one_approval_dispatches_only_one_overlapping_static_call(
    route: str,
) -> None:
    queue = ApprovalQueue()
    server, dispatched = policy_server(queue)
    ready = [asyncio.Event(), asyncio.Event()]
    name, arguments = envelope(route)
    async with asyncio.timeout(10), Client(server) as first, Client(server) as second:
        async with asyncio.TaskGroup() as tasks:
            calls = [
                tasks.create_task(
                    first.call_tool(
                        name,
                        arguments,
                        progress_handler=progress_setter(ready[0]),
                        raise_on_error=False,
                    )
                ),
                tasks.create_task(
                    second.call_tool(
                        name,
                        arguments,
                        progress_handler=progress_setter(ready[1]),
                        raise_on_error=False,
                    )
                ),
            ]
            await wait_for_approval_progress(*ready)
            pending = queue.list_pending()
            assert len(pending) == 1
            assert dispatched == []
            assert queue.approve(pending[0].token)
        results = [call.result() for call in calls]

    assert dispatched == [ARGS], "one approval must authorize only one execution"
    assert sum(not result.is_error for result in results) == 1
    assert_approval_required(next(result for result in results if result.is_error))
    assert len(queue.list_pending()) == 1
    assert queue.list_pending()[0].token != pending[0].token


```

This is the overlapping-call section extracted from our executed regression file. The approval method represents the public approval seam, not an HTTP/UI end-to-end test.
</details>

## Source observation / possible correction

`ApprovalQueue.find_or_create` shares a pending entry for matching static arguments. Waiters can retain the same approved entry; removing its token does not by itself require every dispatcher to successfully claim that token before continuing.

A local candidate makes zero-reuse consumption exclusive and checks the claim in both middleware dispatch paths. Other waiters receive a new pending approval. It preserves timeout → approval → identical re-call and explicitly positive remembered reuse, including per-approved-waiter refresh.

The candidate passed **12 regression/compatibility cases** and **504 selected offline tests**, plus scoped Ruff/format/mypy checks. Those are local candidate results, not the full upstream collection or installed-app verification. A compatibility regression found during review was corrected before the final run. The candidate is not installed or submitted as a PR.

Related: #2246 addresses dynamic selectors; this reproduction uses static arguments. #2369 is a tool-concurrency limit, which is distinct from consuming an approval once. #1990 describes missing/inconsistent gating; here the gate is present and waits, but one approval releases multiple calls.

Could you confirm the intended zero-reuse behavior and whether an exclusive-consumption correction would be welcome? We are reporting this as a non-blocking known limitation for our use, not requesting a private deployment or making a universal security claim.

`ha_report_issue` was not run because the evidence is an isolated source regression rather than a live HA failure; household logs and connection details are not needed for this reproduction.

kingpanther13:
## What does this PR do?

Fixes #2387. Two concurrent identical static calls share one approval row (`ApprovalQueue.find_or_create`), and `decide()` wakes every waiter. Each woken waiter then called `consume_and_maybe_remember`, whose `remove()` is a `pop(token, None)` that discards its result — so every waiter dispatched. One click authorized N executions.

Harmless for an idempotent call like `light.turn_on`, not harmless for `button.press`, `script.turn_on`, `cover.toggle`, or adding a todo item. It is also the same concern that already justified excluding dynamic selectors from sharing in #2246 ("a single click authorizing more executions than the user saw") — that carve-out addressed target resolution, this addresses execution count on the static path.

Sharing the *row* is preserved and still deliberate: the user sees one prompt. Sharing the *decision* is what stops. `PendingApproval.claim()` is a one-shot transition, mirroring `decide()`. The winner dispatches; a losing waiter gets its own pending row and a `USER_APPROVAL_REQUIRED` error instead of riding the click.

Two behaviours the claim must not break, both covered by tests:

- **Remembered approvals.** A positive `remember_minutes` means "stop asking for this call". The winner's consumption arms the remember-cache, so a losing waiter re-checks `is_remembered` and proceeds rather than re-prompting. The `is_remembered` gate at the top of `on_call_tool` ran before the approval existed and cannot cover this.
- **TTL sweeps.** `_sweep_expired` drops entries by `expires_at` regardless of decision and runs on every `find`/`get`/`list_pending`, which the settings UI polls. The claim flag therefore lives on the entry, not on queue membership — keying it on `pop()` would turn a legitimate approval into a spurious re-prompt.

A losing waiter does not wait again on its new row: its wait budget is spent, and re-waiting would let a steady stream of identical calls starve one another. It returns the standard "re-call after the user approves" error.

`on_call_tool` exceeded the C901 complexity limit with the added branch, so the already-decided lookup moved into `_resolve_already_decided` and the pending-entry mint into `_new_pending` (shared by the normal path and the losing-waiter path). No behaviour change from either extraction.

**Behaviour change:** two concurrent identical calls now cost two approval clicks instead of one. That is the fix.

## Type of change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance/refactor
- [ ] 🧪 Tests only
- [ ] 💥 Breaking change

## Testing

Regression tests that fail on master and pass here:

- `test_concurrent_static_calls_consume_one_approval_once` — fails on master with `Expected mock to have been awaited once. Awaited 2 times.`
- `test_remembered_window_lets_the_losing_waiter_proceed`
- `test_consume_claims_the_entry_exactly_once`
- `test_swept_approved_entry_is_still_claimable_by_its_waiter`
- `test_consume_populates_remember_cache_only_for_the_winner`
- `test_approval_overlap.py` — the reporter's end-to-end reproduction (real FastMCP, two clients, both the direct and `ha_call_write_tool` routes), committed as contributed with the unused constants and helper trimmed.

Scope run locally: `tests/src/unit/policy/` 248 passed; `ruff check` and `ruff format` clean on `src/ tests/`; `mypy src/ha_mcp/policy/` clean. Full suite left to CI.

The LLM-agent box is unchecked deliberately: the defect needs two clients calling identically inside the same wait window, which a single agent session cannot produce — verified against a live Home Assistant instance, where sequential calls each correctly got their own approval row.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed approval handling for concurrent requests so a single approval authorizes only one execution.
  * Ensured other waiting requests either receive a new approval prompt or proceed through the configured “remember” option.
  * Prevented dynamic-selector requests from using another caller’s approval.
  * Improved handling of expired, single-use approvals with clearer retry guidance.
* **Tests**
  * Added coverage for overlapping requests, approval reuse, remembered approvals, dynamic requests, and expiration scenarios.

### DIFF
--- a/src/ha_mcp/policy/approval_queue.py
+++ b/src/ha_mcp/policy/approval_queue.py
@@ -33,6 +33,7 @@ class PendingApproval:
     expires_at: datetime
     _decision: Decision = "pending"
     _event: anyio.Event = field(default_factory=anyio.Event)
+    _claimed: bool = False
 
     @property
     def decision(self) -> Decision:
@@ -50,6 +51,26 @@ class PendingApproval:
         """Block until decided; return the final Decision."""
         await self._event.wait()
         return self._decision
+
+    def claim(self) -> bool:
+        """Take exclusive ownership of this approval. False if already taken.
+
+        ``decide()`` wakes every waiter sharing this entry, and each one
+        would otherwise consume it and dispatch — one click authorizing N
+        executions (issue #2387). This is the same one-shot transition
+        ``decide()`` uses, and for the same reason: exactly one caller may
+        act on a given decision.
+
+        The flag lives on the entry rather than on queue membership
+        because ``_sweep_expired`` can drop an approved row (it ignores
+        ``decision``) while a waiter still holds this reference and is
+        legitimately entitled to consume it. There is no ``await`` here,
+        so the check-and-set cannot interleave with another task.
+        """
+        if self._claimed:
+            return False
+        self._claimed = True
+        return True
 
     def __post_init__(self) -> None:
         if self.expires_at <= self.created_at:
@@ -221,10 +242,18 @@ class ApprovalQueue:
 
     def consume_and_maybe_remember(
         self, entry: PendingApproval, *, remember_minutes: int
-    ) -> None:
+    ) -> bool:
+        """Consume one approval. False when another waiter already did.
+
+        A losing caller must not dispatch on this entry, and must not
+        arm the remember-cache from a decision it did not consume.
+        """
+        if not entry.claim():
+            return False
         self.remove(entry.token)
         if remember_minutes > 0:
             self.remember(entry.tool_name, entry.args_hash, minutes=remember_minutes)
+        return True
 
     def _sweep_expired(self) -> None:
         now = datetime.now(UTC)
--- a/src/ha_mcp/policy/middleware.py
+++ b/src/ha_mcp/policy/middleware.py
@@ -56,7 +56,12 @@ class PolicyMiddleware(Middleware):
 
         Loads the policy, evaluates the call, and -- when approval is
         required -- finds or creates a pending queue entry and waits up to
-        the configured window for a decision. A dynamic selector call
+        the configured window for a decision. Identical static calls share
+        one queue entry so the user sees a single approval row, but the
+        approval itself is consumed exactly once
+        (``PendingApproval.claim``): a second waiter woken by the same
+        decision gets its own pending row instead of riding the first
+        call's click. A dynamic selector call
         (``has_dynamic_selector_targets``) never shares or reuses another
         call's pending entry, and is never reissued on TTL eviction: see
         the inline comments above the ``existing =``, ``pending =``, and
@@ -167,36 +172,16 @@ class PolicyMiddleware(Middleware):
         # The accepted cost is that a retry never silently rides an earlier
         # approval: each blocked call gets its own approval row, and only
         # approving the row for the CURRENTLY-blocked call has any effect.
-        existing = None if dynamic_targets else self._queue.find(name, args_hash)
-        if existing and existing.decision == "approved":
-            self._queue.consume_and_maybe_remember(
-                existing,
-                remember_minutes=remember_minutes,
-            )
+        if self._resolve_already_decided(
+            name,
+            args_hash,
+            dynamic_targets=dynamic_targets,
+            remember_minutes=remember_minutes,
+        ):
             return await call_next(context)
-        if existing and existing.decision == "denied":
-            self._queue.remove(existing.token)
-            self._raise_denied_error()
 
-        # find_or_create serialises the create — two concurrent calls with
-        # the same args_hash share one pending entry, so the user only sees
-        # one approval row and approving it releases every waiter. Dynamic
-        # selector calls must NOT share a pending entry: two concurrent
-        # identical calls can still resolve to different (or overlapping)
-        # target sets by the time each one dispatches, so folding them onto
-        # one approval would let a single click authorize more executions
-        # than the user saw. Skip the sharing and always mint a fresh entry.
-        pending = (
-            self._queue.create(
-                name, args_hash, args, ttl_minutes=policy.approval_ttl_minutes
-            )
-            if dynamic_targets
-            else await self._queue.find_or_create(
-                name,
-                args_hash,
-                args,
-                ttl_minutes=policy.approval_ttl_minutes,
-            )
+        pending = await self._new_pending(
+            name, args_hash, args, policy=policy, dynamic_targets=dynamic_targets
         )
 
         wait = (
@@ -207,11 +192,20 @@ class PolicyMiddleware(Middleware):
         await self._wait_for_decision(context, pending, wait)
 
         if pending.decision == "approved":
-            self._queue.consume_and_maybe_remember(
-                pending,
-                remember_minutes=remember_minutes,
+            if self._claim_approval(
+                pending, name, args_hash, remember_minutes=remember_minutes
+            ):
+                return await call_next(context)
+            # Another waiter on this shared entry consumed the approval
+            # first. This call has none of its own, so it must not ride
+            # that click: mint a fresh row and tell the caller to get it
+            # approved. Its wait budget is already spent, so it does not
+            # wait again here -- re-waiting would also let a steady
+            # stream of identical calls starve one another.
+            pending = await self._new_pending(
+                name, args_hash, args, policy=policy, dynamic_targets=dynamic_targets
             )
-            return await call_next(context)
+            self._raise_pending_error(pending, rule, dynamic_targets=dynamic_targets)
         if pending.decision == "denied":
             self._queue.remove(pending.token)
             self._raise_denied_error()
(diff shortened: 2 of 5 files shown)


ghhamcp:

ghhamcp:

Patch76 (commented):
The claim flag is the right seam. `decide()` was already a one-shot transition and consumption was not, so the entry now carries both. Keeping the flag on the entry rather than on queue membership is what lets a waiter still consume an approval that `_sweep_expired` has already dropped from the queue, and `test_swept_approved_entry_is_still_claimable_by_its_waiter` pins that at the queue level.

I checked that the tests discriminate rather than reading the green run as proof. Removing the `if self._claimed: return False` guard turns 6 of the 248 policy tests red across all three test files this PR touches, including both routes of the reporter's reproduction. Replacing the `is_remembered` fallback in `_claim_approval` with `return False` reddens only `test_remembered_window_lets_the_losing_waiter_proceed`, on both backends. Unmutated, `tests/src/unit/policy/` is 248 passed here too.

I also probed the N>2 case, since the losing waiter now mints a row of its own: with three concurrent identical calls, one dispatches and both losers land on a single shared replacement row rather than one row per caller, so the queue does not fan out under a burst. Clicks stay one per execution — draining a burst of three took three approvals in three rounds here, which is the invariant the fix is for. The row bound is worth pinning: with two callers there is exactly one loser, so the existing tests' `len(...) == 1` holds under either design, and the property that produces the bound (the loser path going through `find_or_create` as well) is easy to lose in a later refactor.

One optional note. `_claim_approval` falls back to `is_remembered(name, args_hash)` unconditionally, including on the dynamic-selector path, where `on_call_tool` deliberately skips that same lookup. I could not construct a reachable case: a dynamic entry is never shared, so its claim cannot lose, and `remember_minutes` is forced to 0 for dynamic calls, so nothing arms that key. But `_resolve_already_decided` carries a defensive claim check whose stated purpose is that a future `await` on that stretch degrades into an extra prompt rather than a double dispatch, and this fallback degrades the other way — into a dynamic call riding a remember window the dynamic path exists to ignore. Threading `dynamic_targets` into `_claim_approval` would point both degradations the same way.

The behaviour change reads correctly to me: the row is shared, the decision is not, and a losing waiter spends no second wait budget.

I am leaving this as a comment rather than an approval only so that a push addressing either point does not dismiss it — say the word once you have taken a position on the two notes, in whichever direction, and I will approve on the head at that time.

kingpanther13:
Both taken.

On the fallback: I traced it the same way and landed where you did — `has_dynamic_selector_targets` is a pure function of the `(name, args)` that also produce `args_hash`, so a dynamic key can't be armed by a static call and the branch is unreachable today. The reason to change it anyway is the one you gave: `_resolve_already_decided`'s claim check exists so that a future `await` on that stretch degrades into an extra prompt, and this fallback degraded the other way. `dynamic_targets` is now threaded into `_claim_approval`, which returns False on a lost claim for a dynamic call instead of consulting the cache. Both branches now fail toward a prompt.

On the row bound: pinned with a third concurrent caller in `test_losing_waiters_share_one_replacement_row`. Your read of why two callers can't cover it matches mine — one loser makes `len(...) == 1` true under either design, so the assertion passes even if the loser path stops going through `find_or_create`. Three callers make the shared replacement row observable, and the test asserts one dispatch, two `USER_APPROVAL_REQUIRED`, and exactly one leftover row.

Thanks for mutating the tests rather than reading the green run as proof — that's a stronger check than anything I ran.

Head is 30dbe7822.

Patch76 (approved):
Both notes landed, and the fallback change is the one I would have written: threading `dynamic_targets` into `_claim_approval` puts the two branches on the same footing, since the gate earlier in `on_call_tool` already skips the remember lookup for a dynamic call and forces `remember_minutes` to 0. I re-read both at 30dbe782 rather than taking the docstring's word for it.

On the new test, I mutated rather than read its green run. Making the loser path mint an unshared row instead of going through `find_or_create` turns `test_losing_waiters_share_one_replacement_row` red on both anyio backends and leaves the other 248 tests in `tests/src/unit/policy/` green, `test_concurrent_static_calls_consume_one_approval_once` and `test_remembered_window_lets_the_losing_waiter_proceed` among them. The third caller is doing exactly the discriminating work your comment claims, and the two-caller tests demonstrably cannot see that regression.

One optional note, and leaving it is a fine answer. Neither defensive guard is pinned by anything: deleting `if dynamic_targets: return False` from `_claim_approval` leaves all 250 tests green, and making `_resolve_already_decided` return True regardless of the claim result does the same. Nothing is wrong today, since both branches are unreachable by construction. But a future refactor can drop either one in silence, which is the failure mode the rest of this PR exists to prevent. Calling `_claim_approval` directly with an already consumed entry, an armed remember key and `dynamic_targets=True` does discriminate, returning False with the guard and True without, so a short unit test is available if you want the guard to break loudly. The sibling check needs a different setup rather than that same one: an approved entry whose claim has already been taken but which is still in the queue, and no armed remember key, since an armed key makes that path return True either way. That setup returns False today and True once the claim result is ignored. Both pin a private seam, which is a reasonable thing to decline.

Approving. Under the master ruleset a follow-up push dismisses this approval, so ping me if you add the test and I will re-approve.

kingpanther13:
Declining that, it pins a hypothetical, not worth the churn

Issue #2479, opened 2026-09-17
perf: shrink the tools/list catalog clients load


kingpanther13:
## Why

tools/list from the custom component (86 tools, ENABLE_TOOL_SEARCH off, measured 2026-09-17 against ha-mcp 8.5.0 / component 2.2.0) is 352 KB. Claude Desktop, claude.ai and Claude Code defer tool definitions: the client ingests the whole list at connect, and a tool's full definition reaches the model only when it is pulled in on demand. A 16 KB definition is therefore paid on every pull, and again when Claude Desktop builds the approval card for a call to that tool. #2367 established that Desktop's "Needs approval" dialog becomes clickable before the call is ready and a fast click drops the call (anthropics/claude-code#92014); smaller definitions shorten that window. Clients without deferral (local runtimes) carry the full 352 KB in context every session.

| Slice | Size |
|---|---|
| Total | 352 KB (avg 4.1 KB/tool) |
| Tool descriptions | 162 KB (46%) |
| Input schemas | 146 KB (41%), of which property descriptions 84 KB |
| `_meta.ha_mcp` stamps | 15 KB |

Largest definitions: `ha_config_set_automation` 16.1 KB, `ha_config_set_helper` 15.5 KB (10.7 KB of it schema), `ha_config_set_script` 12.3 KB, `ha_config_set_dashboard` 11.8 KB. These are also the write tools reported in #2367 and #2437.

`ENABLE_LITE_DOCSTRINGS` (beta) is not the answer: it removes too much, and it only touches descriptions (44 KB on this catalog), leaving schemas untouched, so `ha_config_set_helper` is still 11 KB with it on.

## Work items (each needs research before a change)

1. **Trim the full descriptions in place.** Keep the guidance the model needs at call time; move reference material that already exists in the skill pack out. Lite text stays a separate opt-in.
2. **Property descriptions vs description text.** 84 KB of schema is property descriptions, much of it restating the tool description. Check current tool-definition best practices (Anthropic tool-use guidance, MCP tool docs) to decide which of the two carries the detail and which gets cut. Cutting both is not an option.
3. **`ha_config_set_helper` schema shape.** 43 flat type-specific properties (39 `anyOf`). Investigate whether the type-specific fields can live under the existing `config` object without changing what the tool can do, and whether the component-side path can do the work HA-natively the way the dashboard tool and `ha_search` were enhanced for the custom component.
[...]

PR #2487, opened 2026-09-18
perf: omit private tool metadata without an LLM API consumer

Elioooon:
## Summary

- omit the private per-tool `_meta.ha_mcp` stamp from standalone and add-on `tools/list` responses, where it has no consumer
- pass the component's `enable_llm_api` option into the embedded server so metadata is retained only while the Home Assistant conversation-agent API is enabled
- preserve compatibility in both directions: an older component that does not pass the option keeps the historical stamp, and the component progressively drops the new keyword when running an older server package

Part of #2479 (work item 4).

## Why

The only programmatic reader is `custom_components/ha_mcp_tools/llm_api.py::_partition_tools`. Regular MCP clients and external deployments ingest the repeated exposure, pinning, and policy block without using it. Keeping the stamp for embedded LLM-API connections preserves per-tool filtering and pinned-tool mirroring.

On current `master` with tool search disabled, measured by serializing the in-memory `tools/list` models as compact UTF-8 JSON:

| mode | tools | bytes | stamped tools |
|---|---:|---:|---:|
| before | 77 | 327,750 | 77 |
| standalone after | 77 | 318,129 | 0 |
| embedded, LLM API off | 77 | 318,129 | 0 |
| embedded, LLM API on | 77 | 327,750 | 77 |

This removes 9,621 bytes (2.94%) from every external catalog response without changing tool descriptions, schemas, ordering, or call behavior.

## Validation

- `.venv/bin/pytest -q tests/src/unit/test_config_embedded_routing.py tests/src/unit/test_embedded_server.py tests/src/unit/test_llm_exposure.py` — 247 passed
- `.venv/bin/ruff check ...` — passed on all touched files
- `.venv/bin/ruff format --check ...` — passed on all touched files
- `.venv/bin/python -m compileall -q src/ha_mcp custom_components/ha_mcp_tools` — passed
- `git diff --check` — passed
- three-mode in-memory `tools/list` byte benchmark shown above

## Risk boundary

The optimization deliberately does not move or redesign the policy metadata yet. It removes the complete private namespace only where there is no reader. Embedded servers retain the previous wire shape whenever the LLM API is enabled or when an older component does not supply the new capability flag.

## New Features
- Added an embedded-server setting to enable or disable LLM API metadata.
- LLM API metadata now follows the configured embedded-server option.
- Standalone deployments do not emit embedded-server LLM API metadata.

## Bug Fixes
- Improved compatibility with older server versions that do not support the new setting.
- Embedded-server connection settings remain protected from environment-variable exposure.

## Chores
- Updated the custom component version to 2.2.1.


ghhamcp:

Patch76 (commented):
Gating the whole middleware at registration, instead of filtering the stamp per request, keeps the change small and leaves the component's `_partition_tools` path untouched whenever the LLM API is on.

What I checked:
- Both update directions hold. An older component never passes `llm_api_enabled`, so `should_emit_llm_api_metadata()` keeps the stamp. An older server rejects the keyword with `TypeError`, and `_register_embedded_connection` drops it first, before `config_dir` and `verify_ssl`.
- Toggling the Enable LLM API option reloads the entry, and `_serve` registers the connection before it builds a fresh `HomeAssistantSmartMCPServer`, so the gate is read with the new value on every start.
- Outside the component, the only reader of `_meta.ha_mcp` is the embedded e2e test, which runs with the LLM API at its default (on). The settings UI and `tools_dev.py` call `effective_llm_api_exposed` directly, and the policy live state stays available through `policies_live` in the security policy tool.
- Stable 8.5.0 shipped component 2.2.0 and master is level with it, so opening 2.2.1 follows `docs/agents/custom-component.md`.
- `test_server_llm_exposure_wiring.py` drives the real `_initialize_server` with the embedded registration set up per case, so it pins the registration gate itself rather than only the forwarded keyword.

Minor:
- `llm_exposure.py:164`: the class docstring says the stamp is "additive metadata — no tool is hidden or altered for regular MCP clients". That still holds where the stamp is emitted, but standalone and add-on clients, and embedded servers with the LLM API off, no longer get it at all. I'd state the install condition there: an embedded server with the LLM API enabled, or an older component that does not send the flag. The module docstring around line 16 ("stamps every `tools/list` entry") could carry the same condition.
- `llm_exposure.py:234`: the middleware is now only installed when `set_embedded_connection` ran, which the component does after setting `HA_MCP_EMBEDDED`, so `deployment` is always `"embedded"` in production and the `addon` / `standalone` arms are unreachable. I'd either simplify it or add a comment saying why the arms stay.
- The first commit's subject is not in English. The default squash message concatenates the commit messages into the master commit body, so rewording it in English would keep the history readable.

Elioooon:
Addressed the review in 59d221a. `llm_exposure.py` now states that metadata stamping is installed for embedded LLM-API connections (while preserving the legacy-component behavior), and documents why the deployment branches remain even though the current production registration path is embedded-only. I also reworded the original commit subject in English.

Validated with:
```
.venv/bin/ruff check src/ha_mcp/llm_exposure.py
.venv/bin/ruff format --check src/ha_mcp/llm_exposure.py
.venv/bin/pytest tests/src/unit/test_server_llm_exposure_wiring.py -q
git diff origin/master...HEAD --check
```

The focused wiring suite passes (3 tests).

Patch76 (approved):
Thanks for the quick turnaround. I checked the follow-ups at 59d221a4:
- The class docstring of `LlmExposureMiddleware` now states when `server.py` installs the middleware, including the fallback for older components that do not send the flag. That matches `should_emit_llm_api_metadata()`, which returns `True` for an embedded connection that lacks `llm_api_enabled`. The module docstring now scopes the stamping to embedded servers with the LLM API enabled.
- The comment above the policy block says why the `addon` / `standalone` arms stay.
- The first commit subject is now in English. The range-diff against the previous head shows the three earlier commits with unchanged content, and 59d221a4 as the only new one.

The delta is docstrings and a comment only, and CI is green on this head. Approving.

